PR #21, opened 2026-03-24
feat: Add MemWal memory plugin for OpenClaw

hungtranphamminh:
### Why
We need a reference implementation showing how MemWal integrates with an AI agent framework. OpenClaw is the target — its plugin system supports memory replacement, making it a good fit to demonstrate MemWal's encrypted, user-owned memory in a real agent workflow.

### What
Reference implementation of a MemWal-backed memory plugin for OpenClaw at `packages/openclaw-memory-memwal/`. Demonstrates auto-recall, auto-capture, namespace-based agent isolation, and prompt injection protection — all backed by MemWal + Walrus.

### Solution
- **Auto-recall** (`before_prompt_build` hook) — searches MemWal for relevant memories and injects into LLM context
- **Auto-capture** (`agent_end` hook) — extracts facts from conversations via server-side LLM (`analyze()`)
- **Agent tools** — `memory_search` and `memory_store` (opt-in, requires `tools.allow`)
- **CLI** — `openclaw memwal search` and `openclaw memwal stats`
- **Namespace-based agent isolation** — each agent gets its own memory scope derived from session key
- **Prompt injection protection** — injection detection + HTML escaping on both read and write paths
- **Feedback loop prevention** — memory tags stripped during capture to avoid re-storing recalled memories

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Documentation (non-breaking change that doesn't change code behavior, can skip testing)
- [x] Security awareness (changes that affect permission scope, security scenarios)

## Testing

- [ ] I have added/updated unit tests

## Checklist

- [x] My change requires a change to the documentation
- [x] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed


cmdossprime:
Overall, I like the docs structure but there are some nits we need to fix. Will comment directly on each relevant part.

docs/openclaw/overview.md
```diff
@@ -0,0 +1,53 @@
+---
+title: "OpenClaw Plugin"
+description: "Give your OpenClaw AI agents persistent, encrypted long-term memory powered by MemWal."
+---
+
+The MemWal memory plugin replaces OpenClaw's default file-based memory with a **cloud-based, encrypted memory backend**. After setup, it runs silently — your agent remembers things from past conversations and learns new facts automatically.
```
cmdossprime: It's not a complete replacement of OpenClaw's default memory behavior.

docs/openclaw/overview.md
```diff
@@ -0,0 +1,53 @@
...
+  </Card>
+  <Card title="Multi-Agent Isolation" icon="users">
+    Each agent gets its own memory space via namespaces — no cross-contamination
+  </Card>
+  <Card title="Injection Protected" icon="shield">
+    Prompt injection detection and HTML escaping on both read and write paths
+  </Card>
+</CardGroup>
+
+## What it does
```
cmdossprime: I think this section more belongs to "how it works" at a high level which we can move at the top of "How it works" page. We can transform this section into "Features" to briefly introduce features and what comes with this plugin by bullet points, i.e Automatically capture memories, inject memories, CLI, etc

docs/openclaw/quick-start.md
```diff
@@ -0,0 +1,185 @@
...
+title: "Quick Start"
+description: "Install the MemWal memory plugin for OpenClaw and verify it works."
+---
+
+Get the plugin running and test the memory loop in a few minutes.
+
+## Prerequisites
+
+- [OpenClaw](https://openclaw.ai) `>=2026.3.11` installed and running
+- [bun](https://bun.sh) as the package manager
```
cmdossprime: Do we only support `bun`? or we also support other package managers?

docs/openclaw/quick-start.md
```diff
@@ -0,0 +1,185 @@
...
+- A MemWal account with **Ed25519 key pair**, **MemWalAccount ID**, and a **server URL**
```
cmdossprime: 1. I don't think we need a whole keypair?
2. Can we use a more friendly wording like "delegate key"? Likely that OpenClaw users won't even know the web3 behind the scenes
3. I think it could be better to show them how to create the key, and the simplest way is to point them to our website dashboard

docs/openclaw/reference.md
cmdossprime, line 1: This page is more like a "Reference". I think we can find the right title for it as "API Reference"

cmdossprime (dismissed):

ducnmm:
Can Openclaw be deployed on the railway @hungtranphamminh, I think it will be easier to test and evaluate for that team. If possible, deploy one to the dev environment to try

Aaron1924:
Please update the docs to use the current relayer URLs. The old endpoint is stale and breaks setup. Supported URLs should be: 
- production = https://relayer.memwal.ai
- staging/dev = https://relayer.dev.memwal.ai.

Aaron1924:
Auto-capture currently includes assistant-generated messages as input to analyze(). That means summaries or hallucinated assistant text can end up being persisted as memory, not just user-provided facts. but again, this could take as an advantage for user, like planning, etc.

Config validation looks too permissive at the moment. Invalid private keys, account IDs, or server URLs still get through setup and only fail later at runtime with unclear errors, which makes onboarding/debugging much harder than it needs to be.

Aaron1924 (approved):
Auto-capture currently includes assistant-generated messages as input to analyze(). That means summaries or hallucinated assistant text can end up being persisted as memory, not just user-provided facts. but again, this could take as an advantage for user, like planning, etc.

Config validation looks too permissive at the moment. Invalid private keys, account IDs, or server URLs still get through setup and only fail later at runtime with unclear errors, which makes onboarding/debugging much harder than it needs to be.

PR #50, opened 2026-03-26
fix: Anthropic compatibility for memory injection middleware

tamashi095:
## Summary

- **Anthropic compatibility**: `injectMemoryContext` inserted memories as a separate `{ role: "system" }` message. When the consumer already passes a system prompt to `streamText`, this creates multiple system messages separated by user/assistant messages, which Anthropic rejects with `AI_UnsupportedFunctionalityError`. Fix: append memory context to the existing system message instead of inserting a new one.
- **Publish safety**: Added `prepublishOnly` script to ensure `tsc` runs before `npm publish`, preventing packages from being published without `dist/`.
- **Tests**: Added unit tests for `findLastUserMessage`, `injectMemoryContext`, and `formatMemories` using `bun:test`.

## Context

Discovered while integrating MemWal with [Sona](https://github.com/user/sona)'s Miso chat assistant, which uses `@ai-sdk/anthropic` + `streamText` with a system prompt. The middleware worked with OpenAI (which tolerates multiple system messages) but broke on Anthropic.

## Test plan

- [x] `bun test packages/sdk/src/ai/middleware.test.ts` — 13 tests pass
- [x] `bun run build` in `packages/sdk/` — clean compile
- [x] Verified end-to-end with `@ai-sdk/anthropic` + `streamText` + system prompt

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -26,7 +26,8 @@
     "scripts": {
         "build": "tsc",
         "dev": "tsc --watch",
-        "typecheck": "tsc --noEmit"
+        "typecheck": "tsc --noEmit",
+        "prepublishOnly": "tsc"
     },
     "dependencies": {
         "@noble/ed25519": "^2.3.0",
--- a/packages/sdk/src/ai/middleware.test.ts
+++ b/packages/sdk/src/ai/middleware.test.ts
@@ -1,0 +1,143 @@
+import { describe, expect, test } from "bun:test";
+import { findLastUserMessage, injectMemoryContext, formatMemories } from "./middleware.js";
+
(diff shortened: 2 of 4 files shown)


ducnmm:
Hi @tamashi095 Thanks for the fix! Before approving, could you share a screenshot or recording of the e2e test passing with `@ai-sdk/anthropic` + `streamText` + system prompt? The unit tests look good, but I'd like to see proof the Anthropic error is actually resolved.

Could you include:
- A log/screenshot showing the `AI_UnsupportedFunctionalityError` before the fix
- A log/screenshot showing it working correctly after the fix

This helps verify the root cause and confirms the fix resolves it. Thanks!

PR #54, opened 2026-03-27
feat(sidecar): sponsor certify + metadata tx via Enoki

ducnmm:
Handle feat(sidecar): sponsor certify + metadata tx via Enoki

- Add optional allowedAddresses param to executeWithEnokiSponsor
- Use Enoki sponsorship for blob metadata set + transfer (was direct sign)


Aaron1924 (approved):
lgtm

PR #64, opened 2026-03-30
fix(server): use UTF-8 safe truncation in log preview strings

Ashwin-3cS:
## Summary

When testing the app by adding some notes with emojis came across this issue.

Byte-level slicing via `&text[..text.len().min(50)]` in the 4 main API handlers (`remember`, `recall`, `analyze`, `ask`) panics when byte 50 lands inside a multi-byte UTF-8 character. This crashes the server for all connected users.

Any non-ASCII input longer than ~50 bytes triggers it — emoji (4 bytes each), CJK characters (3 bytes), accented Latin (2 bytes) etc.

---

## What this PR does

Adds a `truncate_str(s, max_bytes)` helper that backs up to the nearest `char_boundary` when the limit falls inside a multi-byte sequence. Replaces all 4 instances.

Only the log preview is truncated — the full user text is passed through untouched to the LLM, SEAL encrypt, Walrus upload, and DB insert pipeline.

---

## Affected endpoints

All 4 log-preview sites in `routes.rs`:

* `remember` (line 135)
* `recall` (line 194)
* `analyze` (line 384)
* `ask` (line 606)

All apps in the repo (noter, chatbot, researcher, playground) hit these endpoints — so all benefit from this fix.

---

## Test plan

* Sent multi-byte UTF-8 input via noter frontend — server panicked without fix
* Applied fix, same input — server truncated cleanly at char boundary, full flow completed (extract, encrypt, upload, store)
* Verified no other byte-slicing patterns in `db.rs`, `seal.rs`, `walrus.rs`, `auth.rs`, `sui.rs`
* Confirmed existing e2e tests (`tests/e2e_test.py`) still pass — they use ASCII-only input so were unaffected

Screenshot 2026-03-30 183353

### DIFF
--- a/services/server/src/routes.rs
+++ b/services/server/src/routes.rs
@@ -9,6 +9,19 @@ use crate::walrus;
 use crate::types::*;
 use crate::db::VectorDb;
 
+/// Truncate a string to at most `max_bytes` bytes without splitting a UTF-8
+/// character.  Falls back to the nearest char boundary when `max_bytes` lands
+/// inside a multi-byte sequence (e.g. emoji).
+fn truncate_str(s: &str, max_bytes: usize) -> &str {
+    if s.len() <= max_bytes {
+        return s;
+    }
+    let mut end = max_bytes;
+    while !s.is_char_boundary(end) {
+        end -= 1;
+    }
+    &s[..end]
+}
 
 // ============================================================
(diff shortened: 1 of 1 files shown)


ducnmm:
LGTM!

ducnmm (approved):

PR #72, opened 2026-04-01
fix(server): restore quota accounting and enoki fallback behavior

ducnmm:


Aaron1924:
### Code review

No issues found. Checked for bugs, git history context, CLAUDE.md compliance, and code-comment consistency.

**What the PR does** (verified correct):
- Restores `blob_size` tracking in the `restore` handler (`routes.rs`) — previously passed `0`, bypassing storage quota accounting introduced in `be4e439` (Mar 31). Fix pre-computes `blob_sizes` from downloaded encrypted blobs and passes the real size to `insert_vector`.
- Adds `ENOKI_FALLBACK_TO_DIRECT_SIGN` env var (default `true`) to `executeWithEnokiSponsor` (`sidecar-server.ts`) — falls back to direct signing on Enoki sponsor failure, restoring behavior that existed before `41c295c0` (Mar 31).
- Adds clarifying comment on middleware ordering (`main.rs`) — accurately documents that auth runs before rate limiting.

**Historical context** — both bugs were introduced in back-to-back commits by the same author on Mar 31: `blob_size=0` was a gap when quota tracking was added; the `throw err` on Enoki failure was a later addition that removed a prior fallback. The fix correctly addresses both without introducing regressions.



<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Aaron1924 (approved):
LGTM — both fixes are correct and well-targeted. No issues found.

PR #76, opened 2026-04-02
feat(server): add /api/embed route to expose embedding generation

Ashwin-3cS:
## Summary

Adds a missing `/api/embed` endpoint to support the SDK’s `embed()` method, which is already exposed and documented.

This enables users—especially in manual flows (`rememberManual` / `recallManual`)—to generate embeddings consistent with the server’s model via the API.

---

## Description

The SDK exports an `embed()` method:

```js
await memwal.embed("some text") // → { vector: [...] }
```

However, the corresponding `/api/embed` route was not implemented on the server, resulting in 404 responses.

This PR adds the missing route and wires it to the existing internal `generate_embedding()` function.

---

## What this PR does

* Adds `routes::embed()` handler
* Uses existing `generate_embedding()` logic
* Defines request/response types in `types.rs`
* Registers `/api/embed` route in `main.rs`

---

### Request

```json
{ "text": "I prefer dark mode" }
```

### Response

```json
{ "vector": [0.059, -0.043, ...] }
```

* Returns 400 on empty input
* Requires authentication (same as other `/api/*` routes)

---

## Verification

* HTTP 200 with correct vector dimensions on valid input
* HTTP 400 on empty input

### DIFF
--- a/services/server/src/main.rs
+++ b/services/server/src/main.rs
@@ -128,7 +128,7 @@ async fn main() {
         .route("/api/recall", post(routes::recall))
         .route("/api/remember/manual", post(routes::remember_manual))
         .route("/api/recall/manual", post(routes::recall_manual))
-
+        .route("/api/embed", post(routes::embed))
         .route("/api/analyze", post(routes::analyze))
         .route("/api/ask", post(routes::ask))
         .route("/api/restore", post(routes::restore))
--- a/services/server/src/routes.rs
+++ b/services/server/src/routes.rs
@@ -599,6 +599,24 @@ async fn extract_facts_llm(
     Ok(facts)
 }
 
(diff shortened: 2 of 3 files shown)


ducnmm:
Thanks for the catch! Instead of adding the server route, we'll remove embed() from the SDK — exposing it would let users call OpenRouter embeddings at our cost with no business value. I'll handle the removal on our side. Closing this PR.

Ashwin-3cS:
> Thanks for the catch! Instead of adding the server route, we'll remove embed() from the SDK — exposing it would let users call OpenRouter embeddings at our cost with no business value. I'll handle the removal on our side. Closing this PR.

thanks for the clarification!!

PR #83, opened 2026-04-08
fix(server): guard patchGasCoinIntents behind enokiApiKey check

Ashwin-3cS:
## Summary
Small fix: guard patchGasCoinIntents() behind enokiApiKey check in sidecar walrus/upload handler.

In direct-sign mode (no ENOKI_API_KEY), the patch breaks gas selection with 'No valid gas coins found for the transaction'. Only apply when Enoki is enabled.

## Testing
E2E tested on nautilus-memwal-relayer self-hosted template (AWS Nitro enclave)
- Remember: blob stored with direct signing
- Recall: semantic search works

### DIFF
--- a/services/server/scripts/sidecar-server.ts
+++ b/services/server/scripts/sidecar-server.ts
@@ -220,11 +220,12 @@ async function callEnoki<T>(path: string, payload: unknown): Promise<T> {
     return parsed.data;
 }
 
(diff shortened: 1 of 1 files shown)


ducnmm:
Hi @Ashwin-3cS we will see the lightning later within the team

PR #96, opened 2026-04-13
fix(security): MEM-17 cap recall limit and bound restore concurrency

ducnmm:
Resolves MED-3, MED-6, MED-13 (Phase 3 Medium severity).


harrymove-ctrl:
Hi  I have merged #87 into sec/security_fix and  rebased onto sec/security_fix (conflict was caused by #87 landing)

Verified fixes:
  - MED-13: items cap 50 ✅ (kept tighter than the 100 already in base)
  - MED-3: limit.min(100) on recall + recall_manual ✅
  - MED-6: buffer_unordered(10) on restore download ✅

LGTM.
cc: @ducnmm

PR #97, opened 2026-04-13
fix(security): MEM-18 rate limiter fail-closed on Redis error

ducnmm:
Resolves MED-19, MED-20, MED-21 (Phase 3 Medium severity).


harrymove-ctrl:
Merged

PR #104, opened 2026-04-14
fix(sec): rate limiting + input validation on /sponsor endpoints   (MEM-HIGH-4)

harrymove-ctrl:
## What

Closes MEM-HIGH-4: The /sponsor and /sponsor/execute endpoints  were fully public with no rate limiting, no input validation, and permissive CORS.

## Changes

  **Rate limiting (phase 01)**
  - Atomic sliding-window via single Lua script (checks minute + hour in one
  Redis call, no race)
  - Two axes: per-IP (XFF last-entry, IPv6-normalized) + per-sender (Sui address
   from body)
  - Default: 10 req/min, 30 req/hr — configurable via
  SPONSOR_RATE_LIMIT_PER_MINUTE/HOUR
  - Fail-open on Redis down (rate limit is abuse prevention, not hard gate)
  - /sponsor and /sponsor/execute share the same bucket (both consume Enoki gas)

  **Input validation + error masking (phase 02)**
  - Body size caps enforced before JSON parse (10 KB / 4 KB)
  - sender: must match ^0x[0-9a-fA-F]{64}$
  - transactionBlockKindBytes: valid base64, decoded size 10–8192 bytes
  - signature: valid base64, decoded 65 or 97 bytes
  - Enoki error responses masked — generic messages only, full error logged
  server-side
  - /sponsor/execute now accepts optional sender for per-sender rate limit
  parity

  **CORS**
  - Replaced CorsLayer::permissive() with env-driven allowlist (ALLOWED_ORIGINS)
  - Default is deny-all when env var unset — fail-safe

  ## Testing
  - 39 unit tests passing (cargo test)

### DIFF
--- a/services/server/.env.example
+++ b/services/server/.env.example
@@ -41,3 +41,19 @@ SEAL_KEY_SERVERS=0x...,0x...
 # Enoki Sponsored Transactions (sidecar)
 ENOKI_API_KEY=
 ENOKI_NETWORK=mainnet
+
+# Sponsor endpoint rate limiting (per IP + per sender, sliding window)
+SPONSOR_RATE_LIMIT_PER_MINUTE=10
+SPONSOR_RATE_LIMIT_PER_HOUR=30
+
+# CORS — comma-separated list of allowed browser origins
+# If unset, CORS is deny-all (browsers blocked) — always set this in production
+#
+# Production:
+# ALLOWED_ORIGINS=https://memwal.ai,https://noter.demo.memwal.ai,https://chatbot.demo.memwal.ai,https://researcher.demo.memwal.ai
+#
+# Staging:
+# ALLOWED_ORIGINS=https://staging.memwal.ai,https://noter.demo.staging.memwal.ai,https://chatbot.demo.staging.memwal.ai,https://researcher.demo.staging.memwal.ai
+#
+# Local dev:
+ALLOWED_ORIGINS=http://localhost:3000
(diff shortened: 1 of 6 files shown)


harrymove-ctrl:
Linear: https://linear.app/commandoss/issue/MEM-6/sec-add-auth-and-rate-limiting-to-sponsor-and-sponsorexecute

cc: @ducnmm

ducnmm (changes requested):
Hi @harrymove-ctrl, thanks for the PR! The input validation, error masking, and rate-limiting look great.

However, I found two critical regressions in `services/server/src/main.rs` where some previously implemented fixes were accidentally removed:

1. **Missing HTTP Timeout (DoS Risk):** You reverted the `http_client` initialization to use `reqwest::Client::new()`. This drops the 30-second timeout, making the server vulnerable to hanging requests (e.g., from LLM or Walrus APIs) which was previously addressed (LOW-9). Please restore:
   ```rust
   let http_client = reqwest::Client::builder()
       .timeout(std::time::Duration::from_secs(30))
       .build()
       .expect("Failed to build HTTP client");
   ```

2. **Missing Background Cache Eviction (Storage Leak):** The background task (`tokio::spawn`) that periodically calls `evict_expired_delegate_keys()` has been completely deleted. Without this, the `delegate_key_cache` table will grow infinitely. Please restore the background eviction loop.

Please address these regressions before we can merge. Thanks!

PR #121, opened 2026-04-23
ENG-1406 + ENG-1408: Async remember pipeline and bulk remember

ducnmm:
## Summary
- **ENG-1406**: Convert `/api/remember` to async — returns 202 Accepted with `job_id`, poll via `GET /api/remember/:job_id`
- **ENG-1408**: Add `/api/remember/bulk` — batch up to 20 items in one request, returns array of `job_id`s
- Add per-wallet pinned Apalis workers (`WalletJob`) to avoid wrong-signer races
- Add `POST /api/remember/bulk/status` for batch status polling
- SDK: `remember()` now polls automatically, add `rememberBulk()`

## Files changed (13 files, +2334/-149)
- `services/server/migrations/005_remember_jobs.sql` — job tracking table
- `services/server/migrations/006_bulk_remember.sql` — bulk job support
- `services/server/src/jobs.rs` — WalletJob, BulkRememberJob workers
- `services/server/src/routes.rs` — async remember, bulk endpoints
- `services/server/src/main.rs` — Apalis workers, per-wallet queues
- `services/server/src/types.rs` — async/bulk request/response types
- `packages/sdk/src/memwal.ts` — polling + rememberBulk
- `packages/sdk/src/types.ts` — SDK types

## Test plan
- [ ] `cargo check` passes
- [ ] Server starts and all Apalis workers spawn
- [ ] `POST /api/remember` returns 202 + job_id
- [ ] `GET /api/remember/:job_id` returns job status progression
- [ ] `POST /api/remember/bulk` accepts up to 20 items
- [ ] `POST /api/remember/bulk/status` returns batch statuses

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ const memwal = MemWal.create({
   namespace: "demo",
 });
 
-await memwal.remember("User prefers dark mode and uses TypeScript.");
+const job = await memwal.remember("User prefers dark mode and uses TypeScript.");
+await memwal.waitForRememberJob(job.job_id);
 const memories = await memwal.recall("What are the user's preferences?");
 await memwal.restore("demo");
 ```
--- a/SKILL.md
+++ b/SKILL.md
@@ -95,14 +95,16 @@ const memwal = MemWal.create({
 
 ```ts
 // Store a memory
-await memwal.remember("User prefers dark mode and works in TypeScript.");
+const job = await memwal.remember("User prefers dark mode and works in TypeScript.");
+await memwal.waitForRememberJob(job.job_id);
 
 // Recall by meaning
 const result = await memwal.recall("What are the user's preferences?");
 console.log(result.results);
 
 // Extract and store facts from text
-await memwal.analyze("I live in Hanoi and prefer dark mode.");
+const analyzed = await memwal.analyze("I live in Hanoi and prefer dark mode.");
+await memwal.waitForRememberJobs(analyzed.job_ids);
 
 // Check relayer health
 await memwal.health();
@@ -127,9 +129,11 @@ await memwal.health();
 
 | Method | Description | Returns |
 |---|---|---|
-| `remember(text, namespace?)` | Store one memory (relayer embeds, encrypts, uploads) | `{ id, blob_id, owner, namespace }` |
+| `remember(text, namespace?)` | Accept one memory job immediately | `{ job_id, status }` |
+| `rememberAndWait(text, namespace?, opts?)` | Store one memory and wait for completion | `{ id, job_id, blob_id, owner, namespace }` |
 | `recall(query, limit?, namespace?)` | Semantic search for memories | `{ results: [{ blob_id, text, distance }], total }` |
-| `analyze(text, namespace?)` | Extract facts via LLM, store each as a memory | `{ facts: [{ text, id, blob_id }], total, owner }` |
+| `analyze(text, namespace?)` | Extract facts and accept one memory job per fact | `{ job_ids, facts, fact_count, status, owner }` |
+| `analyzeAndWait(text, namespace?, opts?)` | Extract facts and wait for all fact jobs to complete | `{ results, facts, total, succeeded, failed, owner }` |
 | `restore(namespace, limit?)` | Rebuild missing index entries from Walrus | `{ restored, skipped, total, namespace, owner }` |
 | `health()` | Check relayer health | `{ status, version }` |
 | `getPublicKeyHex()` | Get hex-encoded public key | `string` |
--- a/apps/app/src/pages/Dashboard.tsx
+++ b/apps/app/src/pages/Dashboard.tsx
@@ -243,7 +243,8 @@ const memwal = MemWal.create({
 })
 
 // Remember something
-await memwal.remember("I'm allergic to peanuts")
+const job = await memwal.remember("I'm allergic to peanuts")
+await memwal.waitForRememberJob(job.job_id)
 
 // Recall memories
 const result = await memwal.recall("food allergies")
--- a/apps/app/src/pages/Playground.tsx
+++ b/apps/app/src/pages/Playground.tsx
@@ -515,15 +515,15 @@ const data = await memwal.health()
                 <DemoStep
                     number={2}
                     title="remember"
-                    description="store a memory → embed → encrypt → Walrus"
+                    description="accept a memory job → embed → encrypt → Walrus"
                     code={`const result = await memwal.remember(
   "${rememberText.slice(0, 60)}..."
 )
 // namespace: "${namespace || 'default'}"
-// → { id, blob_id, owner, namespace }`}
+// → { job_id, status }`}
                     onRun={runRemember}
                     result={rememberResult}
-                    resultLabel="stored on Walrus (encrypted)"
+                    resultLabel="memory job accepted"
                     error={rememberError}
                     loading={rememberLoading}
                 >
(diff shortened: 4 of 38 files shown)


ducnmm:
## Benchmark: Before vs After (commit 8e4796d)

Tested locally against mainnet (real Sui RPC, Walrus, SEAL).

### Endpoints affected
Only **2 existing endpoints** changed behavior:
- `POST /api/remember` — sync 200 → async 202
- `POST /api/analyze` — sync 200 → async 202

3 new endpoints added: `GET /api/remember/:job_id`, `POST /api/remember/bulk`, `POST /api/remember/bulk/status`

All other endpoints (`/api/recall`, `/api/recall/manual`, `/api/remember/manual`, `/api/ask`, `/api/restore`) are unchanged.

### Results

| Endpoint | Before (sync, HTTP 200) | After (async, HTTP 202) | Improvement |
|----------|------------------------|------------------------|-------------|
| `POST /api/remember` | 13.9s – 19.3s | **542ms – 1.05s** | **~20-30x** |
| `POST /api/analyze` | 37.2s | **3.26s – 3.50s** | **~10x** |

### What changed

**Before:** Client blocks until the full pipeline completes — embed → SEAL encrypt → Walrus upload → Sui metadata tx → DB vector insert — all synchronous in the request handler.

**After:** Server returns 202 immediately after embed + encrypt (~300ms–1s). Walrus upload + Sui metadata transfer run in background Apalis workers. Client polls `GET /api/remember/:job_id` for final result.

> Total server-side work is the same; the improvement is in **user-perceived latency** (time-to-first-response).

harrymove-ctrl:
Hey @ducnmm found 5 issues. Benchmark numbers look great for the happy path but the issues below all live on failure paths the benchmark doesn't exercise:

### `execute_upload_and_transfer` swallows transfer failure → vector silently lost

  **`services/server/src/jobs.rs:537-554`**

  ```rust
  tracing::warn!("[wallet-job:upload] set-metadata failed: {} blob_id={}", e, blob_id);
  return Ok(());  // ← Apalis: "great, done!" — vector never inserted.
  ```

  When `set-metadata-and-transfer` fails permanently the row is marked `failed` and the function returns `Ok(())` **before** `insert_vector`. Apalis treats the job as successful (no  retry, no dead-letter). End state: blob is paid-for and certified on Walrus, the wallet pool key still owns the Sui object, **but the embedding is never inserted, so the memory
  is forever unrecallable**. The user sees `failed` but their data is gone.

  **Suggested fix**: either return `Err(WalletJobError::Internal(...))` so Apalis retries via the queue, or restructure so `insert_vector` runs **before** the transfer (so a stuck  transfer still leaves a recallable row that a sweeper can finish later).


##  Auth body cap (1 MiB) < bulk route cap (2 MiB) → max-size bulk requests undeliverable

**`services/server/src/auth.rs:170` vs `services/server/src/main.rs:314`**

  The auth middleware caps request body at `1024 * 1024` bytes:
  ```rust
  // auth.rs:170-172
  let body_bytes = to_bytes(body, 1024 * 1024).await
      .map_err(|_| StatusCode::BAD_REQUEST)?;
  ```
  But `/api/remember/bulk` is configured for 2 MiB. A 20-item bulk at the documented `MAX_REMEMBER_TEXT_BYTES = 64 KiB` per item is 1.28 MiB plaintext alone — auth rejects with `400  BAD_REQUEST` **before** the route ever sees it.

  **Suggested fix**: raise the auth cap to 2 MiB (and audit implications), or lower the bulk route cap to 1 MiB and reduce `MAX_BULK_ITEMS` to match. Add a regression test that   submits a max-size bulk payload.


## Bulk path re-introduces the wrong-signer race that per-wallet pinning was supposed to fix

  **`services/server/src/jobs.rs:949-1014` + `services/server/src/main.rs:133-134, 244`**

  Per-wallet `wallet_storages[i]` queues correctly serialize `WalletJob::UploadAndTransfer` per wallet — that's the design. But `bulk_job_storage` is a **single global queue**  drained by 2 concurrent workers, and inside one `BulkRememberJob` we run `BULK_UPLOAD_CONCURRENCY=3` concurrent uploads. Items are pre-stamped with round-robin `wallet_index`  (`routes.rs:423-426`), so a 20-item bulk on a 3-key pool produces ~7 items per wallet, **3 of which try to upload concurrently using the same Sui key** — exactly the  coin-object-lock conflict the per-wallet pinning was justified by.

  It gets worse with 2 bulk workers: two simultaneous `BulkRememberJob`s from any callers can both target wallet 0 in parallel.

  **Suggested fix**: route bulk uploads through the per-wallet `wallet_storages[i]` queues as N × `WalletJob::UploadAndTransfer` items — same pattern `analyze` already uses  (`routes.rs:968-983`). The bulk worker becomes a fan-out planner, not an uploader.



## No sweeper / TTL on `remember_jobs` → zombie `running` rows on worker crash

  **`services/server/migrations/005_remember_jobs.sql` + `services/server/src/main.rs:282-291`**

  `005` has no expiration column, no scheduled cleanup. `main.rs:282-291` already has an hourly evict task — but only for `delegate_key_cache`. Nothing equivalent for `remember_jobs`.

  If a worker panics, gets SIGKILL'd, OOMs, or the pod is rolled mid-job, the corresponding `remember_jobs` row is stuck at `running` forever. The SDK polls indefinitely (no  terminal state). PR test plan explicitly listed this and it wasn't addressed.

  **Suggested fix**: add a periodic sweeper that transitions `running` rows older than e.g. 10 minutes to `failed` with reason "stale/orphaned", **or** reconcile against `apalis_jobs.status='Dead'` from the apalis-sql tables. The hourly tick in `main.rs:282-291` is the natural place.

  Bonus: workers that panic stop their queue silently — `Monitor::new()...run().await.unwrap_or_else(|e| tracing::error!(...))` (`main.rs:202-208`) logs and exits the spawned task,  never restarts. Wrap each spawn in a respawn loop with backoff so a bad job doesn't kill the whole queue.


SDK polling: `not_found` JSON branch is dead code; real 404 throws an opaque error mid-poll
  **`packages/sdk/src/memwal.ts:170-198`**

  The SDK polls `GET /api/remember/:job_id` and checks `status.status === "not_found"` (line 193-198). But the server returns **HTTP 404** for "no such job / not yours"
  (`routes.rs:309-314`, `types.rs:634`), never a JSON body with `status: "not_found"`. Because the GET call doesn't pass `acceptedStatuses=[200, 404]`, the 404 throws inside
  `signedRequest` with a sanitized generic error — the rich `{status, jobId}` info the polling loop is supposed to handle never reaches the catch.

  Combined with no transient-error tolerance (any 401/429/5xx/network drop kills the whole poll loop, unlike `waitForRememberJobs` which retries on `429 || >=500 || 0` at lines
  326-334), users hit `remember()` failures during ordinary infra blips even when their job ultimately succeeds.

  **Suggested fix**:
  1. Add `acceptedStatuses: [200, 404]` to the GET in `waitForRememberJob` and dispatch on `res.status` directly.
  2. Mirror the bulk path's transient-retry logic for single-job polls.
  3. Add jittered backoff (current 1500 ms constant → thundering herd at scale).

  While you're in `memwal.ts`: `RememberResult.id` silently changed meaning from "memory id" to "job id" (`memwal.ts:181`). Callers persisting `result.id` will break at runtime once   the row gets reaped. Either rename the field or include the on-chain memory id in the status response.

ducnmm:
Hi @harrymove-ctrl please fix it and test it carefully in local

hien-p:
LGTM

harrymove-ctrl (approved):

PR #124, opened 2026-04-28
Optimization/v1

Olympusxvn:
Pushed 807eef4 to optimization/v1 , removed npm install -g npm@latest from both release workflows, updated CHANGELOG (entry #15) and lessons.md (entry #9).


ducnmm:
Hi @Olympusxvn there are quite a lot of changes at the moment, I think PR is no longer suitable

PR #141, opened 2026-05-11
feat(mcp): @memwal/mcp stdio bridge + browser login 

harrymove-ctrl:
#  Summary
Adds end-to-end MCP (Model Context Protocol) support to MemWal so any MCP-aware editor — Cursor, Claude Desktop, Claude Code, Antigravity  can read and write to a user's encrypted Walrus memory via 4 tools (memwal_remember, memwal_recall, memwal_analyze, memwal_restore).

# Test plan

  - pnpm -F @memwal/mcp build clean
  - cargo build -p memwal-server clean
  - First-run login: spawn bridge with --local, browser opens, complete consent, creds land in ~/.memwal/credentials.json
  - All 4 tools end-to-end against --local relayer (encrypt → SEAL → Walrus upload → blob id returned)
  - Recall round-trip: previously-saved fact retrievable by query
  - Bridge auto-reconnect: kill sidecar mid-session, bridge reconnects and serves next call
  - Auth-required fallback: delete creds, restart bridge from an editor — friendly error appears inline
  - Env preset switch: change --local → --dev in .mcp.json, creds get relayer override patched, calls hit dev relayer
  - Dashboard /connect/mcp page renders config snippets for Cursor / Claude Desktop / Claude Code / Antigravity with copy buttons

image

image


ducnmm (changes requested):
Found blockers before this should merge.

1. `services/server/Dockerfile:49` only copies `scripts/*.ts` into the runtime image, but `sidecar-server.ts` now imports `./mcp/index.js` and that directory is not copied. Local source runs can pass, but the deployed Docker image will fail at sidecar startup with module-not-found for `scripts/mcp/*`. Please copy the MCP directory into the image, e.g. `COPY scripts/ ./scripts/` or an explicit `COPY scripts/mcp/ ./scripts/mcp/`.

2. The public MCP routes are long-lived and not rate-limited, while `resolveAuth()` only validates bearer/account shapes and derives a pubkey before a session is allocated. A client can open many `/api/mcp/sse` or `/api/mcp` sessions with random 64-hex bearer values and valid-looking account ids, causing the relayer to hold 24h proxy streams and the sidecar to allocate transports/heartbeats without any on-chain auth ever happening. Please add an IP/session rate limit or cap before session creation, and preferably force a real credential validation before opening long-lived transports.

3. Streamable HTTP sessions are not bound to the authenticated caller on subsequent requests. `handleStreamableHttp()` resolves auth for every request, but when `streamableSessions.get(sessionId)` succeeds it directly calls `conn.transport.handleRequest(req, res)` without checking `conn.sessionKey === auth.sessionKey`. Please reject mismatched session/auth pairs with 403 so a leaked/session-id cannot be reused under another bearer/account context.

Verification I ran:
- `cargo check` in `services/server` passed
- `pnpm --filter @memwal/app build` passed
- `pnpm --filter @memwal/mcp build` could not be run locally because this checkout has not installed the new package dependencies, while CI's Compile & CLI Smoke is passing

harrymove-ctrl:
HI @ducnmm  Thanks for the review addressed all three blockers in .

ducnmm:
One gap to cover before merging: auth should be discoverable from MCP clients, not require a separate manual `memwal-mcp login` command. This affects Codex and Claude Code similarly for the stdio package, since the server is spawned non-interactively and cannot safely block startup on a browser login.

Could we add either a `memwal_login` auth-required tool for stdio clients, or implement native remote MCP OAuth for `/api/mcp` so clients like Claude Code/Codex can trigger the browser flow from their MCP auth UI?

harrymove-ctrl:
Yes, i think  we should aim to ship both eventually; first can add  a `memwal_login` tool that invokes the existing loginFlow() in-process so users sign in from inside their MCP client without dropping to a terminal. 


However Native MCP OAuth  is our long term answer but need more time to implement it (auth code + PKCE). I can create tickets to implement it 

@ducnmm

ducnmm (approved):

PR #142, opened 2026-05-11
Fix/playground poll job status

harrymove-ctrl:
Task(fix) https://linear.app/commandoss/issue/MEM-36/playground-remember-step-stuck-at-status-running

### DIFF
--- a/apps/app/src/pages/Playground.tsx
+++ b/apps/app/src/pages/Playground.tsx
@@ -21,6 +21,7 @@ import {
 } from '@mysten/dapp-kit'
 import { useSponsoredTransaction } from '../hooks/useSponsoredTransaction'
 import { MemWal } from '@mysten-incubation/memwal'
+import type { RememberJobStatus } from '@mysten-incubation/memwal'
 import { MemWalManual } from '@mysten-incubation/memwal/manual'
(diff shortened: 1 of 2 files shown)


ducnmm (changes requested):
Found one behavior issue that should be fixed before approval.

`getRememberStatus()` treats HTTP 404 as an accepted status and returns the raw JSON body as `RememberJobStatus`, but the server's `GET /api/remember/:job_id` returns `{ \"error\": ... }` for not-found/unauthorized jobs, not `{ \"job_id\": ..., \"status\": \"not_found\" }`. That violates the new method's public contract and makes the playground polling loop see `current.status === undefined`, append an undefined transition, and continue until timeout instead of surfacing `not_found` immediately.

Suggested fix: normalize 404 inside `getRememberStatus()` to `{ job_id: jobId, status: \"not_found\", error: ... }`, or stop accepting 404 and handle it explicitly in the caller.

Verification I ran:
- `pnpm --filter @mysten-incubation/memwal build`
- `pnpm --filter @memwal/app build` after SDK build

ducnmm (approved):
Fix verified locally and pushed. Approving after normalizing 404 remember status responses to the documented not_found state.

PR #176, opened 2026-05-20
feat(seal): default testnet SEAL to Mysten committee aggregator

harrymove-ctrl:
Switch the built-in SEAL server config on testnet to Mysten's initial committee aggregator so fresh relayer/sidecar deployments use the decentralized committee path by default where the official committee is available.

What changed:
- Testnet built-in sidecar/server default is now one committee entry: objectId `0xb012...e1e98`, weight `1`, aggregatorUrl `https://seal-aggregator-testnet.mystenlabs.com`.
- Mainnet built-in default stays on the existing independent key server pair until Mysten publishes an official mainnet committee aggregator.
- `SEAL_SERVER_CONFIGS` still wins first and remains the override path for any custom committee by providing `objectId`, `weight`, and `aggregatorUrl`.
- `SEAL_KEY_SERVERS` still wins second and preserves the legacy independent-key-server path.
- Docs and env examples now explain committee defaults, custom committee overrides, legacy independent key servers, and the migration warning for existing testnet data encrypted with the previous independent defaults.

SDK note:
- This PR does not change the published TypeScript SDK defaults, so it does not require an SDK release for the relayer/sidecar default behavior. The SDK already supports committee aggregators through explicit `sealServerConfigs`.

Threshold note:
- The new built-in testnet sidecar default has one decentralized committee config, so MemWal's outer default threshold resolves to `1`. The Mysten aggregator fronts the Seal committee threshold internally.
- Legacy 2-of-2 independent deployments are still supported by pinning `SEAL_KEY_SERVERS=0x73d0...,0xf5d1...`; the new test covers that this path still resolves to threshold `2`.

Validation:
- `npm --prefix services/server/scripts test` passes: 29/29.
- `pnpm -C docs exec mintlify validate` passes.
- GitHub checks pass; Mintlify Deployment is skipped by workflow.

### DIFF
--- a/packages/sdk/src/manual.ts
+++ b/packages/sdk/src/manual.ts
@@ -45,9 +45,9 @@ import { sha256hex, hexToBytes, bytesToHex, normalizeServerUrl, sanitizeServerEr
 type ResolvedSealServerConfig = Omit<SealServerConfig, "weight"> & { weight: number };
 
 // Default SEAL server configs per network.
-// Testnet uses Mysten's initial committee aggregator by default. Mainnet
-// remains on the legacy independent servers until an official committee
-// aggregator is available.
+// Keep testnet on the legacy independent servers so Manual mode can decrypt
+// data written by the hosted testnet relayer. Committee aggregators remain
+// supported through explicit sealServerConfigs.
 const DEFAULT_SEAL_SERVER_CONFIGS: Record<string, ResolvedSealServerConfig[]> = {
     mainnet: [
         {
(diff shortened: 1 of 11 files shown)


packages/sdk/src/manual.ts
```diff
@@ -61,12 +61,9 @@ const DEFAULT_SEAL_SERVER_CONFIGS: Record<string, ResolvedSealServerConfig[]> =
     ],
     testnet: [
         {
-            objectId: "0x73d05d62c18d9374e3ea529e8e0ed6161da1a141a94d3f76ae3fe4e99356db75",
-            weight: 1,
-        },
-        {
-            objectId: "0xf5d14a81a982144ae441cd7d64b09027f116a468bd36e7eca494f750591623c8",
+            objectId: "0xb012378c9f3799fb5b1a7083da74a4069e3c3f1c93de0b27212a5799ce1e1e98",
```
ducnmm: Why do we delete the key, if this is the case, does the system run well to ensure 2/2 threadhold, have you tested it? @harrymove-ctrl

ducnmm (approved):

PR #180, opened 2026-05-21
fix(server): mark single remember jobs failed on recovery enqueue errors

harrymove-ctrl:
## Summary

This PR fixes a stuck-state bug in the async single-remember pipeline.

When a remember upload succeeds but the follow-up recovery/finalization wallet job cannot be enqueued, the corresponding `remember_jobs` row could remain in a non-terminal state such as `running` or `uploaded`, even though no downstream worker would ever resume it.

This change makes those enqueue-handoff failures terminal.

## Problem

The async remember pipeline has a recovery path for partial success:

- upload succeeds
- metadata/transfer fails
- server tries to enqueue a follow-up wallet job to finish recovery

Before this PR, if that follow-up enqueue failed, the `remember_jobs` row might already have been updated to `uploaded`, but nothing else would process it after that. Clients polling `GET /api/remember/:job_id` could hang until timeout because the row still looked active.

## What changed

In `services/server/src/jobs.rs`:

- added a focused helper to mark `remember_jobs` rows as `failed` with an `error_msg`
- used that helper for enqueue-handoff failures that would otherwise orphan a single remember job
- return a terminal/permanent error for those handoff failures instead of leaving the row in a live-looking state

Covered paths:

1. `WalletOperation::UploadAndTransfer`
   - upload succeeded
   - metadata/transfer recovery enqueue fails
   - now marks the remember job as `failed`

2. `WalletOperation::SetMetadataAndTransfer`
   - metadata/transfer succeeded
   - uploaded-blob finalization enqueue fails
   - now marks the remember job as `failed`

3. legacy `execute_remember`
   - upload succeeded
   - metadata/transfer recovery enqueue fails
   - now marks the remember job as `failed`

## Why this is correct

Polling reads directly from `remember_jobs`.

If no downstream worker exists anymore, the row must be terminal. Otherwise the system reports `running` / `uploaded` even though the job is effectively dead.

This PR enforces the state-machine rule:

- if work is still actively recoverable by an enqueued worker, the row may remain non-terminal
- if the handoff to that worker fails, the row must transition to `failed`

## User impact

- no more permanently stuck single remember jobs for these enqueue failure paths
- `wait_for_remember_job()` will observe a terminal failure instead of timing out indefinitely
- operators get a concrete `error_msg` indicating which enqueue handoff failed

## Validation

Ran:

- `cargo test --manifest-path services/server/Cargo.toml jobs::tests -- --nocapture`

Result:

- `7 passed`

## Scope note

This PR is scoped to single remember job lifecycle handling only.

It does not address the separate bulk remember partial-fanout issue.

### DIFF
--- a/services/server/src/jobs.rs
+++ b/services/server/src/jobs.rs
@@ -181,6 +181,20 @@ async fn update_remember_job_after_wallet_error(
     .await;
 }
 
+async fn mark_remember_job_failed(state: &AppState, remember_job_id: Option<&str>, msg: &str) {
+    let Some(jid) = remember_job_id else {
+        return;
+    };
+
+    let _ = sqlx::query(
+        "UPDATE remember_jobs SET status = 'failed', error_msg = $1, updated_at = NOW() WHERE id = $2",
+    )
+    .bind(msg)
+    .bind(jid)
+    .execute(state.db.pool())
+    .await;
+}
+
 /// A wallet job.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct WalletJob {
@@ -379,6 +393,7 @@ pub async fn execute_wallet_job(job: WalletJob, ctx: Data<Arc<AppState>>) -> Res
                         )
                         .await
                         {
+                            let finalize_remember_job_id = remember_job_id.clone();
                             if let Err(enqueue_err) = enqueue_finalize_uploaded_blob(
                                 state,
                                 enqueued_wallet_index,
@@ -392,7 +407,19 @@ pub async fn execute_wallet_job(job: WalletJob, ctx: Data<Arc<AppState>>) -> Res
                             )
                             .await
                             {
-                                return Err(enqueue_err.into_apalis_error());
+                                let msg = enqueue_err.to_string();
+                                mark_remember_job_failed(
+                                    state,
+                                    finalize_remember_job_id.as_deref(),
+                                    &msg,
+                                )
+                                .await;
+                                tracing::error!(
+                                    "[wallet-job:set-metadata] job_id={} {}",
+                                    finalize_remember_job_id.as_deref().unwrap_or("-"),
+                                    msg,
+                                );
+                                return Err(WalletJobError::Permanent(msg).into_apalis_error());
                             }
                             tracing::warn!(
                                     "[wallet-job:set-metadata] finalization failed after transfer; enqueued index-only retry: {}",
@@ -722,22 +749,9 @@ async fn execute_upload_and_transfer(
                 .await
             {
                 let msg = format!("failed to enqueue metadata/transfer recovery job: {}", e);
-                let classified = WalletJobError::classify_sidecar_error(&msg);
-                update_remember_job_after_wallet_error(
-                    state,
-                    recovery_remember_job_id.as_deref(),
-                    &classified,
-                    &msg,
-                )
-                .await;
-                tracing::error!(
-                    "[wallet-job:upload] job_id={} {} classification={} retryable={}",
-                    job_id_for_log,
-                    msg,
-                    classified.kind(),
-                    !classified.is_permanent()
-                );
-                return Err(classified);
+                mark_remember_job_failed(state, recovery_remember_job_id.as_deref(), &msg).await;
+                tracing::error!("[wallet-job:upload] job_id={} {}", job_id_for_log, msg,);
+                return Err(WalletJobError::Permanent(msg));
             }
 
             tracing::info!(
@@ -1022,7 +1036,7 @@ pub async fn execute_remember(
             .await;
 
             let mut storage = state.wallet_storage.clone();
-            storage
+            if let Err(e) = storage
                 .push(WalletJob {
                     wallet_index: key_index,
                     operation: WalletOperation::SetMetadataAndTransfer {
@@ -1044,12 +1058,12 @@ pub async fn execute_remember(
                     },
                 })
                 .await
-                .map_err(|e| {
-                    RememberJobError::Internal(format!(
-                        "failed to enqueue metadata/transfer recovery job: {}",
-                        e
-                    ))
-                })?;
+            {
+                let msg = format!("failed to enqueue metadata/transfer recovery job: {}", e);
+                tracing::error!("[remember-job] {} job_id={}", msg, job.job_id);
+                mark_remember_job_failed(state, Some(&job.job_id), &msg).await;
+                return Err(RememberJobError::Internal(msg));
+            }
 
             tracing::info!(
                 "[remember-job] job_id={} enqueued metadata/transfer recovery for blob_id={} key={}",
--- a/services/server/src/jobs.rs
+++ b/services/server/src/jobs.rs
@@ -181,18 +181,67 @@ async fn update_remember_job_after_wallet_error(
     .await;
 }
 
-async fn mark_remember_job_failed(state: &AppState, remember_job_id: Option<&str>, msg: &str) {
+async fn mark_remember_job_failed(
+    pool: &sqlx::PgPool,
+    remember_job_id: Option<&str>,
+    msg: &str,
+) -> Result<(), sqlx::Error> {
     let Some(jid) = remember_job_id else {
-        return;
+        return Ok(());
     };
 
-    let _ = sqlx::query(
+    let result = sqlx::query(
         "UPDATE remember_jobs SET status = 'failed', error_msg = $1, updated_at = NOW() WHERE id = $2",
     )
     .bind(msg)
     .bind(jid)
-    .execute(state.db.pool())
-    .await;
+    .execute(pool)
+    .await?;
+
+    if result.rows_affected() == 0 {
+        return Err(sqlx::Error::RowNotFound);
+    }
+
+    Ok(())
+}
+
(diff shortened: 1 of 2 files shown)


harrymove-ctrl:
Live validation result with real account credentials against `https://relayer.dev.memwal.ai`:

```text
CONFIG {"base_url": "https://relayer.dev.memwal.ai", "account_id": "0x40d66224593e30103c71e415004d2d85a4c91a4457e800d162c0bf6ef4d27448", "namespace": "remember-terminal-check-475b9ff6"}
ACCEPTED {"job_id": "8f86ea77-bb3a-437f-b041-14d639f1bd8c", "status": "running"}
STATUS {"job_id": "8f86ea77-bb3a-437f-b041-14d639f1bd8c", "status": "running", "owner": "0xe5c91145cb92ac82df7285022ee4f73c40be5c995d632c74700b1fc04e5f4994", "namespace": "remember-terminal-check-475b9ff6"}
STATUS {"job_id": "8f86ea77-bb3a-437f-b041-14d639f1bd8c", "status": "done", "owner": "0xe5c91145cb92ac82df7285022ee4f73c40be5c995d632c74700b1fc04e5f4994", "namespace": "remember-terminal-check-475b9ff6", "blob_id": "IXgOzidrrPDw1SAL-sJLLLlldTSczNNiEia3vuDhIAE"}
TERMINAL {"result": "done", "statuses": ["running", "done"]}
RESULT {"ok": true, "job_id": "8f86ea77-bb3a-437f-b041-14d639f1bd8c", "blob_id": "IXgOzidrrPDw1SAL-sJLLLlldTSczNNiEia3vuDhIAE", "note": "Happy path completed. If PR #180's recovery enqueue failure occurs, this same poll path should return status=failed with error instead of hanging."}
```

ducnmm:
Hi @harrymove-ctrl The direction makes sense: recovery enqueue failures should surface a terminal `remember_jobs` failure instead of leaving clients polling `uploaded`/`running`. I do not think this breaks the normal retry paths, since ordinary upload/sidecar transient errors still return `Transient`.

Two things I would fix before merge:
- `mark_remember_job_failed` currently ignores the UPDATE result. If enqueue failed because Postgres/storage is unavailable, the status update may fail too; returning `Permanent` afterward can still leave the row non-terminal. Please return/handle that error and keep the handler retryable if the failure cannot be persisted.
- The new live script only exercises the happy path. Please add a deterministic test that simulates recovery enqueue failure and asserts `status=failed` plus a useful `error_msg`.

jasong-03 (changes requested):
I independently verified both of @ducnmm's concerns by reading `jobs.rs` and running `cargo test`. Both are real and blocking. Adding a third finding plus a concrete patch.

### 1. `mark_remember_job_failed` silently swallows the UPDATE error (CONFIRMED)

```rust
async fn mark_remember_job_failed(state: &AppState, remember_job_id: Option<&str>, msg: &str) {
    let Some(jid) = remember_job_id else { return };
    let _ = sqlx::query("UPDATE remember_jobs SET status = 'failed' ...")
        .bind(msg).bind(jid)
        .execute(state.db.pool())
        .await;          // ← Result discarded
}
```

The most likely cause of an enqueue-handoff failure is **infrastructure trouble** (Redis/Postgres degraded, connection pool exhausted). In exactly that scenario the follow-up `UPDATE remember_jobs` is also likely to fail — and `let _ =` makes it silent. All three callers then return `Permanent`/`Internal` regardless, leaving the row in `running` / `uploaded`. **The very bug this PR claims to fix is not fixed for the most realistic trigger condition.**

**Suggested signature**:

```rust
async fn mark_remember_job_failed(
    state: &AppState,
    remember_job_id: Option<&str>,
    msg: &str,
) -> Result<(), sqlx::Error> {
    let Some(jid) = remember_job_id else { return Ok(()) };
    sqlx::query("UPDATE remember_jobs SET status = 'failed', error_msg = $1, updated_at = NOW() WHERE id = $2")
        .bind(msg)
        .bind(jid)
        .execute(state.db.pool())
        .await
        .map(|_| ())
}
```

And at each call site, if the UPDATE fails, return `Transient` so Apalis retries — the row is then better off `running` (still retryable) than `failed-but-not-actually-marked`:

```rust
if let Err(db_err) = mark_remember_job_failed(state, jid, &msg).await {
    tracing::error!("[wallet-job] could not persist failed status: db={} orig={}", db_err, msg);
    return Err(WalletJobError::Transient(format!(
        "{} (also failed to persist terminal state: {})", msg, db_err
    )));
}
return Err(WalletJobError::Permanent(msg));
```

### 2. No deterministic test for the new behavior (CONFIRMED)

The 7 `jobs::tests` that pass (verified locally — `cargo test jobs::tests --bin memwal-server` → `7 passed`) are all **pre-existing** error-classification tests:

- `classify_object_lock_as_transient`
- `classify_move_abort_as_permanent`
- `classify_balance_split_move_abort_as_transient`
- `classify_network_errors_as_transient`
- `display_includes_classification_tag`
- `permanent_errors_abort_apalis_retries`
- `transient_errors_remain_retryable`

None exercise `mark_remember_job_failed` nor the new enqueue-failure paths. The only new artifact is `live_remember_terminal_state_check.py`, which (per its own docstring) just polls a real relayer on the happy path.

Suggested deterministic coverage — for each of the 3 covered paths:

1. seed `remember_jobs` row in `running` status
2. invoke the function with `wallet_storage.push` mocked to return `Err`
3. assert row transitions to `status = 'failed'` with non-empty `error_msg`
4. **and** a second variant: DB UPDATE also fails → assert `Transient` returned (no orphan)

A `#[tokio::test]` + sqlx test pool would be enough; pattern already exists in `routes/remember.rs:1079` and `types.rs:1165`.

### 3. Behavior regression in `execute_upload_and_transfer` (NEW finding)

The old code at the upload-and-transfer path classified the push error and returned `Transient` for transient-looking strings (e.g., `503`, `ECONNRESET`, `sidecar timeout`). The new code unconditionally returns `Permanent` for any push failure:

```diff
-let classified = WalletJobError::classify_sidecar_error(&msg);
-update_remember_job_after_wallet_error(state, ..., &classified, &msg).await;
-return Err(classified);
+mark_remember_job_failed(state, recovery_remember_job_id.as_deref(), &msg).await;
+return Err(WalletJobError::Permanent(msg));
```

This intentionally trades retry capability for orphan-prevention, which is defensible — but a transient Redis blip during `wallet_storage.push` will now permanently fail a remember whose upload already succeeded, instead of letting Apalis retry the enqueue. At minimum, add a comment explaining the deliberate trade-off, and consider whether `classify_sidecar_error(&msg)` still applies here for known-transient push errors.

### Verdict

Direction is correct, scope is right (single remember only). Blocking on #1 + #2 before merge. #3 is a discussion point but worth at least a comment in the code documenting the chosen semantics.

Will re-review once #1 + #2 are addressed.

jasong-03 (approved):
Re-reviewed after the two new commits (`10b7fc4`, `cda4cea`). All 3 of my prior concerns are fully addressed.

### ✅ Concern #1 — `mark_remember_job_failed` swallowed UPDATE error

Now returns `Result<(), sqlx::Error>` and additionally treats `rows_affected() == 0` as `Err(RowNotFound)`. New `classify_wallet_remember_handoff_failure` wraps it and falls back to `WalletJobError::Transient` when persistence fails, with a clear diagnostic suffix `"; failed to persist remember_jobs failed status: {err}"`. Exactly the semantics needed to avoid orphan rows when DB is unhealthy.

### ✅ Concern #2 — deterministic test for enqueue failure

Five new `#[tokio::test]` tests added, covering every branch I asked for plus a couple I did not think of:

- `recovery_enqueue_failure_marks_remember_job_failed` — happy DB → row terminates `failed` + `error_msg` set
- `terminal_recovery_handoff_failure_overrides_transient_looking_queue_error` — even a `503` message still returns `Permanent` (locks in concern #3 semantics by test)
- `failed_status_persistence_keeps_wallet_handoff_retryable` — closed pool simulates DB down → returns `Transient` with diagnostic
- `mark_remember_job_failed_returns_real_update_error` — helper surfaces real sqlx error
- `missing_remember_job_keeps_handoff_retryable` — `rows_affected = 0` path → `Transient`

Local run (`DATABASE_URL=postgresql://memwal:memwal_secret@localhost:5433/memwal cargo test jobs::tests --bin memwal-server -- --test-threads=1`) → **12 passed; 0 failed**.

Also nice touch in `cda4cea` — switching tests from full `VectorDb::new(...)` setup to a focused `test_pool()` that only loads `005_remember_jobs.sql`, with a `OnceLock<Mutex<()>>` to serialize migration application. Keeps these tests independent of unrelated migration churn.

### ✅ Concern #3 — Permanent-for-all-push-errors regression

The new docstring on `classify_wallet_remember_handoff_failure` explicitly documents the deliberate trade-off:

> Recovery handoff failures happen after an external side effect already succeeded: a Walrus upload and sometimes an on-chain transfer. Once the polling row is durably terminal, abort retries so clients see `failed` instead of polling `uploaded` / `running` forever. If that terminal state cannot be persisted, keep the wallet handler retryable so the row is not orphaned.

And `terminal_recovery_handoff_failure_overrides_transient_looking_queue_error` enforces it by test. Good — the intent is now both written and locked.

### CI

All 7 checks green on `cda4cea` (Mintlify skipped — no docs touched).

LGTM, approving.

ducnmm (approved):

PR #194, opened 2026-05-26
Feat: Filter recall results by Walrus blob lease epoch

hungtranphamminh:
### Why

Walrus blobs don't live forever. Each blob has a **storage lease** purchased at upload time, and once the lease ends Walrus is free to garbage-collect the blob — at which point any attempt to fetch it returns 404.

Auto-recall didn't know about this. When a user asked a question, recall would do its vector search, return the top-K blob IDs, and try to hydrate them. Two things would go wrong with blobs whose lease had expired:

**1. 404 on hydration.** The most visible symptom. Recall returns a blob_id, the engine goes to Walrus, Walrus says "gone", the memory is dropped and `dropped_count` ticks up. User sees fewer results than they should, and a dead memory burned a top-K slot that could have gone to a live one further down the list.

**2. Worse — the Redis blob cache silently masking expiry.** This is the part that's easy to miss. The ciphertext cache has a 14-day TTL, but on testnet Walrus leases run only ~5 days. So a blob that expired five days ago can still be sitting in the cache, getting served to recall as if it were alive, for **another nine days past its real lease end**. The user gets back what looks like a valid memory, decrypts it, uses it — but the underlying blob has been GC'd. Any operation that has to round-trip Walrus (e.g. a cache miss) breaks; the system has been lying about what's available.

Why can't we fix this with a timestamp heuristic? Walrus leases are denominated in **epochs**, not wall-clock seconds. Epochs advance via on-chain events (a validator quorum), not a fixed clock — so the duration drifts. A formula like `created_at + epochs_bought × nominal_duration` drifts in both directions: too late and we serve dead blobs (the bug we're fixing); too early and we wrongly hide memories that are still alive. The only correct way to ask "is this blob expired?" is to read the **actual** end-of-lease epoch off the on-chain Blob object and compare it to the **actual** current Walrus epoch.

So that's what this change does — and crucially, it does the comparison **at the SQL level, before the cache gets involved**, so the cache can't paper over expiry.

### What

- New migration `010` adds two nullable columns to `vector_entries`: `end_epoch` (the blob's on-chain lease end) and `object_id` (the Sui object id the lease lives on). NULL means "no lease state tracked" — handled as always-served, so adding the columns changes no behavior until a row actually has values.
- New sidecar endpoint `GET /walrus/current-epoch` returns the live Walrus epoch (from the SDK's `stakingState()`). Rust caches the result in Redis with a 1h positive TTL and a 30s negative-cache on failure, so this costs effectively one chain read per recall window per replica.
- Sidecar `/walrus/upload` and `/walrus/query-blobs` now include `endEpoch` per blob, read off the on-chain object that's already being fetched — zero extra chain calls in either path.
- `search_similar` gains an expiry filter: `AND (end_epoch IS NULL OR end_epoch > $current_epoch)`. Applied to `/api/recall`, `/api/recall/manual`, `/api/ask`, and the analyze pre-extraction context search. Expired rows never become a blob_id we'd try to hydrate.
- Write path threads `end_epoch` + `object_id` from the upload result into `insert_vector`, so every memory written after deploy is correctly marked. All five existing call sites updated (including the deferred-transfer recovery paths and the restore flow).
- Migration runner now serializes concurrent app boots under a session `pg_advisory_lock` so a rolling deploy can't crash-loop a replica that loses a `CREATE INDEX IF NOT EXISTS` race.
- `scripts/backfill-end-epoch.ts` (new, manual-run): fills `end_epoch` for legacy rows by reading each blob's `storage.end_epoch` straight from chain. Dry-run by default; `--apply` writes; `--limit` / `--owner` scope it.

### Solution

A few design decisions worth calling out, because they're load-bearing:

**Filter in the SQL `WHERE`, not at hydration.** This is the one that makes the cache bug above unfixable any other way. The blob cache reads happen *after* `search_similar` returns; if we filter at hydration, we've already retrieved (and potentially served from cache) a row whose lease ended weeks ago. Putting the filter in the SQL `WHERE` means expired rows never become a blob_id passed to `cache_get`. As a side benefit it also fixes top-K quality — `LIMIT $K` now returns up to K **live** rows, instead of letting expired rows squat in slots that should have gone to the next-best memory.

**NULL means "always-served" everywhere.** Three concrete categories of row write NULL: benchmark rows (no Walrus blob), recovery paths that lack a fresh upload result, and any row not yet backfilled. All are handled identically — the filter passes them. That makes the entire change *safe* by construction: it's structurally impossible to wrongly hide a memory unless a real `end_epoch` value was deliberately written. The migration is purely additive, and behavior is identical to today's for every row until a real value lands.

**Fail-open on the current-epoch lookup.** If Redis is unreachable, the sidecar 502s, the chain RPC times out, or anything else goes wrong — `current_epoch_cached` returns `None` and the SQL binds `i64::MIN` as the threshold, which every real `end_epoch` is greater than. Effect: the filter is silently disabled and recall serves everything. Pre-existing 404 cleanup still catches truly-dead blobs on the hydration path. Recall stays available no matter what's broken upstream; safety doesn't depend on the sidecar being healthy.

**The backfill is the only thing that can wrongly hide a memory, so it's the most conservative piece of code in the change.** Writing `end_epoch=0` ("blob is gone, filter forever") is the one irreversible operation in this whole PR — the script's own re-run guard (`WHERE end_epoch IS NULL`) will never revisit a row that has that value. So we classify "gone" only on the SDK's structured `ObjectError.code` (`notExists` / `deleted`). Anything else — transport 429, transient RPC error, unknown SDK error, anything message-based — leaves the row NULL for the next pass. We never write 0 on a guess.

**Migration is non-blocking, runs at boot, doesn't rewrite the table.** Two `ADD COLUMN IF NOT EXISTS` (catalog-only changes; no row rewrite) plus a `DROP INDEX IF EXISTS` to clean up a partial index from an earlier draft. The boot-time advisory lock prevents concurrent rolling-deploy replicas from racing on the `IF NOT EXISTS` checks, which would otherwise let one replica crash-loop on "already exists".

**Operator safety on the backfill script.** Three startup guards because this is the one piece of code that writes to a real DB: `SUI_NETWORK` is cross-checked against the sidecar's reported network (otherwise a misconfigured run silently miscovers every owner); the schema is checked for the new columns (friendly error if migration hasn't run yet); `SIDECAR_AUTH_TOKEN` is required up front (a missing token would otherwise 401 every owner and produce a clean-looking no-op).

## Types of Changes

- [ ] Security awareness (changes that affect permission scope, security scenarios)

## Testing

Validated end-to-end on a complete clone of the Neon dev DB (785 production rows / 18 distinct owners), pulled via `pg_dump` v18 and restored into a local pg17 container.

**Migration 010 on the clone:** both columns added (nullable), `DROP INDEX IF EXISTS` no-ops cleanly on fresh DBs, re-running the whole migration is a clean no-op (every statement guarded with `IF NOT EXISTS`).

**SQL filter behaviour with seeded `end_epoch` values** (current_epoch=100):
- `end_epoch=0` (chain-confirmed gone) → filtered
- `end_epoch=99` and `=100` (boundary) → filtered (correct: `>` matches the "expired = `end_epoch ≤ current`" definition)
- `end_epoch=101` and `=999999` → served
- `end_epoch IS NULL` → served
- Fail-open sentinel (`i64::MIN`) → all rows served

**Live sidecar/chain integration:** booted the new sidecar locally, `GET /walrus/current-epoch` returned `{"epoch":409,"suiNetwork":"testnet"}` against testnet; 401 without token.

**Full-clone backfill:** `--apply` converged at `set=665 markedGone=0 leftNull=120` in one pass; subsequent passes mopped up stragglers (the `WHERE end_epoch IS NULL` re-run guard verified idempotent — never clobbers a value written between passes). At the live epoch 409 the post-backfill filter split the 785 rows as 291 served / 494 filtered / 17 untracked NULL. An earlier full run produced 438 alive / 346 expired / 1 unresolvable, which confirms the precise-classification path was necessary — a "mark legacy expired" shortcut would have wrongly hidden 438 live memories.

**Operator-error guards on the backfill script:**
- `SUI_NETWORK=mainnet` against a testnet sidecar → FATAL before any DB work.
- Missing migration 010 → FATAL with a friendly schema-check message (URL password masked).
- Missing `SIDECAR_AUTH_TOKEN` → FATAL up front.
- `--limit` with no value → FATAL (no silent cap-at-1).

**Build/type:** `cargo check` clean. Both new/touched TS files (`backfill-end-epoch.ts`, `sidecar-server.ts`) type-check clean. (3 pre-existing errors in `mcp/tools/*.ts` are unrelated — already present on `origin/dev`.)

## Checklist

- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

## Additional Notes

**Merge sequence (manual, post-merge):**

1. Merge this PR into `dev`.
2. Wait for the dev deploy. Migration 010 auto-runs at server boot.
3. Verify the new columns exist on the dev DB (`\d vector_entries` shows `end_epoch` + `object_id`).
4. Verify the write path is recording lease state — a fresh `/api/remember` on dev should land a row with a non-NULL `end_epoch` and `object_id`.
5. **Back up the dev DB** — `pg_dump` v18 → restore into a local throwaway in the Docker container. (Mainnet backup location is a separate, infra-owned decision; out of scope here.)
6. Run `scripts/backfill-end-epoch.ts` — dry-run first (no `--apply`), inspect the `WOULD SET` lines, then `--apply`. Expect testnet 429s; the script is re-runnable so each pass mops up the previous one's stragglers.

**Boundary asymmetry worth knowing:** filtering at `end_epoch ≤ current` may briefly hide a blob whose lease has technically ended but whose ciphertext Walrus hasn't GC'd yet. The opposite direction (we serve, Walrus has already GC'd) is covered by the existing 404 cleanup backstop. Both are acceptable trade-offs.

**Performance note:** intentionally no dedicated `end_epoch` index. EXPLAIN confirms the recall filter rides the existing `(owner, namespace)` index from migration 002; `end_epoch` is a cheap recheck on that already-narrow per-owner row set. A partial `WHERE end_epoch IS NOT NULL` index can't satisfy the `OR end_epoch IS NULL` branch and would only add write-maintenance cost — see the inline comment in 010 for the full reasoning. If a prune is ever added (`DELETE WHERE end_epoch <= current`), a tuned index can come with it.

**Backfill script is committed for review + reproducibility + audit trail** — but is manual-run only (no scheduler, no CI hook, `--apply` defaults off). The operator invokes it deliberately post-deploy.

**Mainnet pre-rollout work (out of scope for this PR):**
- Decide on a dedicated DB-backup location (the local-Docker pattern used here is testnet-only).
- Add a `--confirm` prompt on `--apply` to catch a same-shape wrong-DB paste.
- Re-measure the per-row / convergence timing on mainnet's actual owner count + RPC limits.

### DIFF
--- a/services/server/migrations/010_blob_end_epoch.sql
+++ b/services/server/migrations/010_blob_end_epoch.sql
@@ -1,0 +1,46 @@
+-- Blob-expiry tracking on `vector_entries`. Walrus storage leases are
+-- denominated in EPOCHS (integers), not wall-clock time, and epoch duration
+-- drifts (advances via on-chain events, not a fixed clock). Any
+-- created_at-based heuristic drifts in both directions — too-late serves a
+-- dead blob (404), too-early prunes a still-alive memory (data loss). Working
+-- in exact epoch numbers removes the drift entirely: each blob is sealed
+-- until a fixed `end_epoch`, and "is it expired?" is the integer test
+-- `end_epoch <= current_epoch`.
+--
+-- `end_epoch`  — the blob's on-chain `storage.end_epoch` (a u32 on chain;
+--   stored BIGINT for headroom). Recorded at write time from the upload
+--   result (no extra chain call) and filled in for legacy rows by a separate
+--   re-runnable backfill script. NULL means "no expiry tracked" — benchmark
+--   rows and any row not yet backfilled. The recall filter treats NULL as
+--   ALWAYS-SERVED, so this migration is purely additive: behaviour doesn't
+--   change for any row until a real `end_epoch` is written.
+--
+-- `object_id` — the blob's Sui object id. Already returned by the upload but
+--   previously discarded. Persisting it lets the backfill (and any future
+--   on-chain re-read) skip the expensive blob_id → object_id resolution
+--   scan, leaving that cost only for the pre-existing legacy backlog.
+--
+-- Both columns are NULLABLE with no default and no data backfill here: the
+-- migration is pure DDL and must stay cheap (it runs at every server boot).
+-- The data step is the standalone backfill script, deliberately separate so
+-- a chain-read failure can't block startup.
+
+ALTER TABLE vector_entries
+    ADD COLUMN IF NOT EXISTS end_epoch BIGINT NULL;
+
+ALTER TABLE vector_entries
+    ADD COLUMN IF NOT EXISTS object_id TEXT NULL;
+
+-- No dedicated end_epoch index. The recall filter
+--   WHERE owner=$ AND namespace=$ AND (end_epoch IS NULL OR end_epoch > $)
+-- is driven by the existing (owner, namespace) index from migration 002;
+-- end_epoch is a cheap recheck on the already-narrow per-owner row set.
+-- A standalone partial index on `(end_epoch) WHERE end_epoch IS NOT NULL`
+-- can't serve this query because the `OR end_epoch IS NULL` branch
+-- contradicts the index predicate — confirmed via EXPLAIN; the planner
+-- rechecks regardless. It would just add write-maintenance cost. If a
+-- prune (`DELETE WHERE end_epoch <= current`) is ever added, a tuned
+-- index can come with it.
+--
+-- DROP an index left over from an earlier draft of this migration.
+DROP INDEX IF EXISTS idx_vector_entries_end_epoch;
--- a/services/server/scripts/backfill-end-epoch.ts
+++ b/services/server/scripts/backfill-end-epoch.ts
@@ -1,0 +1,398 @@
+/**
+ * Backfill `vector_entries.end_epoch` for legacy rows.
+ *
+ * Rows written before the expiry feature shipped have no `end_epoch`, so the
+ * recall filter falls back to "always serve" for them — safe, but means the
+ * fix is incomplete until this script runs and fills in real lease state.
+ *
+ * Strategy: read each blob's `storage.end_epoch` straight from chain via
+ * `walrusClient.getBlobObject(objectId)`. That's exact (no timestamp→epoch
+ * estimation, no assumption about how many epochs were originally bought) and
+ * works for our `deletable: true` blobs (`getVerifiedBlobStatus` doesn't —
+ * it omits `endEpoch` for the deletable variant).
+ *
+ * `getBlobObject` needs the Sui object id, but legacy rows store only
+ * `blob_id`. We resolve in this order:
+ *   1. Row already has `object_id` (written by post-feature paths) — use it.
+ *   2. Otherwise, the sidecar's `/walrus/query-blobs` does a per-owner wallet
+ *      scan that returns `blobId` already converted to the base64url form
+ *      stored in `vector_entries.blob_id` — so map lookups match the DB
+ *      directly with no u256↔base64url encoding step on this side.
+ *
+ * Safety properties:
+ *   - DRY-RUN BY DEFAULT. `--apply` writes; `--limit N` caps rows; `--owner
+ *     0x…` scopes to one wallet.
+ *   - Re-runnable: the UPDATE is guarded `WHERE end_epoch IS NULL`, so it
+ *     can't overwrite a value the live server has written.
+ *   - Benchmark rows (`plaintext IS NOT NULL`) are skipped — they have no
+ *     Walrus blob.
+ *   - We only write `end_epoch = 0` (= "filtered/dead") when the chain itself
+ *     confirms the blob is gone (ObjectError.code = notExists/deleted). Any
+ *     other failure — transient RPC, unresolved object id, partial scan — is
+ *     left NULL for the next run. We never guess in the dead direction.
+ *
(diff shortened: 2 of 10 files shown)


ducnmm:
Thanks for digging into this. We are not going to take this PR as-is.

For the current scope, this is better handled as a cache-policy/config fix rather than adding new lease-tracking state, a migration, and a backfill path. The Redis blob cache TTL should be aligned with the lease duration we purchase when uploading blobs. Mainnet currently uploads with 3 Walrus epochs, which is roughly 42 days, so I have fixed the Railway configuration by setting:

```
BLOB_CACHE_TTL_SECS=3628800
```

That keeps the cache lifetime aligned with the current mainnet upload policy without adding an `end_epoch` column or a legacy-row backfill job. If we decide later that we need stricter correctness than TTL-based cache expiry, we should revisit it in a smaller design; even then, using the upload timestamp plus the configured storage duration may be sufficient for our needs, instead of threading a new `end_epoch` field through the whole system.

Closing this PR since the immediate issue has been addressed in Railway and this implementation is larger than we want to carry right now.

PR #199, opened 2026-05-27
fix(MEM-34): handle Walrus package version mismatch without redeploy

jasong-03:
Linear: [MEM-34](https://linear.app/commandoss/issue/MEM-34/handle-walrus-package-version-mismatch-without-redeploy)

## Issue

When the Walrus on-chain package is upgraded, the sidecar's cached `@mysten/walrus` client keeps the stale package metadata and every upload fails with `MoveAbort EWrongVersion` from `walrus::system::inner_mut` until the relayer is redeployed.

## What was done

- **Detect EWrongVersion** in the sidecar upload catch block and call the existing `refreshWalrusClient()` so the next retry uses fresh package metadata
- **Fix classifier**: `EWrongVersion` was being marked Permanent → Apalis would not retry. Carved it out as Transient
- **Slack alert** via the existing `AlertManager` (`WalrusPackageUpgradeDetectedAlert`) so the team knows a Walrus upgrade just hit the upload path
- **Log on-chain version** before + after refresh in the sidecar warn line (via `walrusClient.systemObject().version`)
- **Bump deps**: `@mysten/walrus` 1.0.3 → 1.1.7, `@mysten/sui` 2.5.0 → 2.17.0

## Tests

- Rust: 249/249 pass (4 new — classifier carve-out, detector, alert payload, truncation)
- Sidecar TS: 41/41 pass (11 new — real JSON-RPC + gRPC formats, false-positive guards)


ducnmm:
@jasong-03 quick review:

- Main path looks good, but legacy `RememberJob` still marks `UploadBlobError::App` as failed immediately on `EWrongVersion` instead of retrying after the sidecar refreshes the Walrus client. Since that worker still drains old rows, can we carve out `is_walrus_package_version_mismatch` there too or route it through the same transient classifier?
- The package-upgrade Slack alert is described as one-shot, but it currently fires for every matching upload failure. During an upgrade, concurrent queued jobs could spam Slack. A small per-network/version time-window dedupe would help.
- Minor: the new TS test has `@ts-expect-error` directives that become unused if the test file is typechecked with the repo’s current flags. Runtime tests pass, but `undefined as unknown as string` / `null as unknown as string` would avoid future typecheck breakage.

jasong-03:
@ducnmm thanks for the review — addressed all three points:

1. **Legacy `RememberJob` path** → carved out `is_walrus_package_version_mismatch` ahead of `fail!()` in the `UploadBlobError::App` branch (`services/server/src/jobs.rs:1293`). The transient case now fires the same informational alert and returns `Err` without writing `status='failed'`, so Apalis retries against the refreshed Walrus client. Commit `95b8059`.

2. **Slack alert spam during upgrade burst** → added an in-memory dedup gate inside `AlertManager` keyed by `(sui_network, sidecar_walrus_dep_version)` — same upgrade event suppresses follow-up alerts within the window; different networks / dep bumps stay independent. Default 10 min, env-overridable via `WALRUS_PACKAGE_UPGRADE_ALERT_DEDUP_SECS`. 4 new unit tests cover first-fire / burst-suppress / network+dep independence / re-fire-after-expiry. Commit `6d9e7c7`.

3. **`@ts-expect-error` directives** → replaced with `undefined as unknown as string` / `null as unknown as string` casts in the nullish-input test so it stays clean under any tsconfig strictness. Commit `c4f2eed`.

Tests: 253/253 Rust + 41/41 sidecar TS green.

ducnmm (approved):
Reviewed locally against WALM-33/MEM-34 scope. No blocking findings; tests/checks are green.

PR #202, opened 2026-05-28
Feat: WALM-55 — extract.v6 timestamp injection for temporal recall

hungtranphamminh:
### Why

Temporal recall is the weakest category on both LOCOMO and LongMemEval — by a wide margin. Diagnosis of failing temporal questions on the prior `extract.v5` baseline showed a specific, mechanical cause: the overwhelming majority of failing temporal answers hinge on a date, and almost all of those dates are entirely absent from the memories retrieved at recall time. The dates exist in the source conversations (and in our benchmark dataset's session metadata) but get silently dropped before reaching the extractor — so the LLM has no anchor to write a dated fact, and vector search cannot match temporal queries against undated stored memories.

This is a write-side input problem, not a retrieval or reasoning problem. No amount of better recall, reranking, or query expansion can find a date that was never stored in the first place.

### What

This PR introduces an optional `occurred_at` field on `AnalyzeRequest` that carries an RFC 3339 UTC timestamp from caller to extractor. When present, the server renders it as a `<context occurred_at="..."/>` tag and supplies it to the LLM as a dedicated user message alongside the existing `<related_memories>` block. A new prompt iteration (`extract.v6`) instructs the LLM how to use this anchor — resolving in-turn relative-time references (`"last Friday"`, `"yesterday"`, `"two weeks ago"`) into absolute dates and attaching them to the resulting fact text in a verbose `Weekday, D Month YYYY (YYYY-MM-DD)` shape.

This PR also includes a set of transient-failure fixes that the bench investigation uncovered along the way. These convert several silently-dropped upstream-flake failure modes (HTTP 200 wrapping a 504-gateway-timeout error envelope; mid-body connection drops; null-content completions) from HTTP 500 (not retried) to HTTP 503 (in the SDK + benchmark harness retry set), so they recover via the existing retry policy instead of dropping turns.

The harness side wires both LOCOMO and LongMemEval to normalize their dataset-specific timestamp formats to RFC 3339 UTC and pass them as `occurred_at` on every ingestion call.

### Solution

**Theory — how the resolved date ends up improving temporal retrieval.**

Three independent mechanisms compound when the absolute date enters the fact text before embedding:

1. **Date tokens become matchable signal in vector search.** Modern dense embedding models capture not just semantic similarity but also lexical co-occurrence on distinctive tokens. Rare, information-dense tokens like `"2023"`, `"May"`, or `"2023-05-19"` give a query like *"What did the user do in May 2023?"* something concrete to match against. Without dates in the stored fact text, this signal has nothing to land on; with them, the same semantic-only vector search has the date token in its searchable surface.

2. **The answer LLM can reason about time only when it can see time.** A query like *"How long ago did X happen?"* requires the model to read a date and do arithmetic. If retrieval surfaces a memory that says *"User ran a charity race"* with no date, the model has nothing to work with even if it found the right memory. With the absolute date embedded in the fact, time-arithmetic questions become directly answerable.

3. **Relative-time resolution at extraction is strictly cheaper and more reliable than at retrieval.** The extractor has the absolute anchor (the caller-supplied `occurred_at`) at the moment it sees the input. It can resolve *"last Friday"* exactly once, at write time, with full context. The answer model at read time wouldn't have that anchor, so doing the same resolution there is either impossible or error-prone.

**Architectural constraint — Architecture A.**

The resolved date lives ONLY inside the extracted fact text. That text is SEAL-encrypted before storage on Walrus, and the same text feeds the embedding. **There is no new metadata column on `vector_entries`.** This is a deliberate, permanent trade: the server can never filter, range-scan, or rank by event time (only by ingestion time, which it already has via `created_at`). What we gain is the temporal-retrieval lift without weakening the privacy floor — date information remains inside the encrypted-storage boundary, not in a queryable plaintext column.

This trade was the load-bearing design decision; the rest of the implementation is just the cleanest way to realize it.

**Implementation summary.**

Server side:
- New optional `AnalyzeRequest.occurred_at: Option<chrono::DateTime<chrono::Utc>>` field (RFC 3339 via serde).
- `Extractor::extract_with_context` trait signature extended with an `occurred_at` argument; default impl ignores it (backwards-compatible).
- `LlmExtractor` builds 1–3 user messages depending on which contexts are present (related-memories, occurred-at, and the actual input text).
- New `extract.v6` prompt: temporal-anchor rules, defensive opt-out ("when uncertain, prefer no date"), worked examples covering current-event resolved, recounted-past-event-keeps-own-date, vague-future-month-anchored, stable-fact-no-date.
- Three transient-failure fixes (separate but related):
  - New `AppError::UpstreamUnavailable` variant → HTTP 503.
  - New shared `parse_openrouter_error_envelope` helper to detect the `{"error":{"code":NNN}}`-wrapped-in-200 pattern, called from both extractor and embedder.
  - `ChatMessageResp.content` widened from `String` to `Option<String>` to gracefully handle null-content completions; the Option ripples through three consumer sites (extractor, ask, summariser).

Harness side:
- Two new normalizers (one per benchmark) that convert dataset-specific date formats to RFC 3339 UTC at parse time.
- `build_ingest_text_*` return type widened from `(label, text)` to `(label, text, occurred_at)`.
- `run.py` unpacks the new 3-tuple and threads `occurred_at` through to `client.analyze()`.

**Validation.** Both benchmarks (LOCOMO and LongMemEval) were re-run on this branch end-to-end against the prior `extract.v5` baseline, with the primary target category (temporal) moving cleanly past the judge-noise floor in the predicted direction on both benchmarks. Other categories held within noise — no measurable regression. Bench completion rate also improved from the v5 baseline thanks to the transient-failure fixes. Detailed numbers are captured in the internal benchmark archive.

16 new unit tests pin the load-bearing behaviour: envelope detection across 5 distinct body shapes (including the choices+error edge case that the guard explicitly handles); occurred_at-block rendering in two timezones; default-impl pass-through for occurred_at; the v6 prompt's anti-leak rules; the 503-mapping and observability-label contracts; and the timestamp-normalization helpers in both benchmark adapters.

## Types of Changes

- [ ] Security awareness (changes that affect permission scope, security scenarios)

The "Bug fix" check covers the transient-failure handling (500 → 503 mapping + null-content type widening + envelope detection) that landed alongside the main feature work. They share the same code paths and were diagnosed during the same investigation, so they ship in one PR.

## Testing

- [x] I have added/updated integration tests

End-to-end testing against the LOCOMO and LongMemEval benchmark harnesses (full ingest + eval) was the primary validation path: both benchmarks completed against the patched server with positive results vs the prior baseline, exercising the entire flow from harness → `AnalyzeRequest.occurred_at` → server → extractor → LLM → embedded fact → recall → answer-model evaluation. Full numeric results are in the internal benchmark archive.

Unit tests (267 Rust, 42 Python) pass on the rebased branch. The 16 new tests are scoped to the load-bearing pieces called out in the Solution section.

## Checklist

The change does not require user-facing documentation updates yet: the new `occurred_at` field is server-side only and not yet exposed in the public SDK clients. SDK exposure is a separate follow-up.

## Screenshots/Recordings

Not applicable — server-side and benchmark-harness changes only.

## Additional Notes

- **No new database migration.** All changes go through existing schema; `occurred_at` lives only in the encrypted fact text + embedding, never as a metadata column on `vector_entries`.
- **No SDK changes in this PR.** The TS + Python SDK clients do not yet expose `occurredAt` / `occurred_at`. Until that follows up, production callers still get the baseline behaviour; the bench harness is the only consumer exercising the new field right now.
- **Observability:** one new `AppError::kind()` label (`upstream_unavailable`). Any Grafana/Loki dashboards that filter on error-kind will see a new category; existing `internal` label volume should drop proportionally for upstream transient failures.
- **Operational impact:** production callers will see slightly more HTTP 503 responses where they previously saw 500, with corresponding fewer 500s. Net 5xx volume drops because the harness/SDK retries the 503s. Worth a heads-up in the release notes if there's automated alerting that doesn't distinguish 500 vs 503.

### DIFF
--- a/.changeset/walm-55-occurred-at-openclaw.md
+++ b/.changeset/walm-55-occurred-at-openclaw.md
@@ -1,0 +1,14 @@
+---
+"@mysten-incubation/oc-memwal": patch
+---
+
+Wire temporal anchoring through the agent-side memory tools.
+
+### Added
+
+- `memory_store` tool now accepts an optional `occurredAt` argument (RFC-3339 / ISO-8601 string) so agents can anchor recounted past events to the date they actually occurred. Description tells the LLM to omit it when unknown rather than guess.
+
+### Changed
+
+- Auto-capture hook (`agent_end`) now passes `new Date()` as `occurredAt` to `analyze()`. Every captured conversation now gets temporal anchoring automatically — the server extractor resolves in-turn relative references ("yesterday", "last Friday") into absolute dates inside the stored fact text. Facts captured by this version now carry resolved dates.
+- SDK dependency bumped from published `^0.0.2` to `workspace:*` to consume the new `AnalyzeOptions` signature.
--- a/.changeset/walm-55-occurred-at-sdk.md
+++ b/.changeset/walm-55-occurred-at-sdk.md
@@ -1,0 +1,11 @@
+---
+"@mysten-incubation/memwal": patch
+---
+
+Add optional `occurredAt` to `analyze()` and `analyzeAndWait()` for temporal anchoring of extracted facts.
+
+- New `AnalyzeOptions` overload: `analyze(text, { namespace, occurredAt })` accepts `Date` or RFC-3339 string. The legacy `analyze(text, namespace?)` signature still works unchanged.
+- When `occurredAt` is supplied, the server resolves in-turn relative references ("last Friday", "yesterday") into absolute dates inside the extracted fact text before embedding and encryption.
+- Wire format is RFC-3339 UTC with millisecond precision (e.g. `"2023-05-25T17:50:00.000Z"`).
+- Invalid `Date` instances (constructed from malformed input) now throw a diagnostic `TypeError` from the SDK rather than an opaque `RangeError` from `toISOString()`.
+- Field is omitted from the request body when not supplied — existing callers see byte-identical wire payloads.
--- a/packages/openclaw-memory-memwal/package.json
+++ b/packages/openclaw-memory-memwal/package.json
@@ -40,7 +40,7 @@
     "node": ">=20.0.0"
   },
   "dependencies": {
-    "@mysten-incubation/memwal": "^0.0.2",
+    "@mysten-incubation/memwal": "workspace:*",
     "@sinclair/typebox": "0.34.48",
     "zod": "^3.23.0"
   },
--- a/packages/openclaw-memory-memwal/src/hooks/capture.ts
+++ b/packages/openclaw-memory-memwal/src/hooks/capture.ts
@@ -47,8 +47,28 @@ export function registerCaptureHook(api: any, client: MemWal, config: PluginConf
         .join("\n\n");
 
       // analyze() calls the server LLM for fact extraction — retry once
-      // since transient failures are common with remote LLM calls
-      const result = await withRetry(() => client.analyze(conversation, namespace));
+      // since transient failures are common with remote LLM calls.
+      //
+      // Pass `occurredAt: new Date()` so the server extractor can
+      // resolve in-turn relative references ("yesterday", "last
+      // Friday") into absolute dates inside the fact text before
+      // encryption.
+      //
+      // Caveat: this is hook-fire time, not strictly per-message
+      // time. The hook fires on agent_end — *after* the LLM finishes
+      // responding — and `event.messages` may carry the last N turns
+      // (captureMaxMessages) spanning a longer window. All extracted
+      // facts share this one anchor. For coarse relative references
+      // ("yesterday", "last Friday") this is accurate enough; for
+      // fine-grained ones ("an hour ago", "this morning") the anchor
+      // can be off by minutes-to-hours. Acceptable for an opt-in
+      // auto-capture path where the alternative is no anchor at all.
+      // (The "no silent now() fallback" rule is about the server
+      // defaulting for callers that passed nothing; here the caller
+      // IS passing, with real-ish knowledge of when the event happened.)
+      const result = await withRetry(() =>
+        client.analyze(conversation, { namespace, occurredAt: new Date() }),
+      );
 
       if (result.facts?.length) {
         api.logger.info(
--- a/packages/openclaw-memory-memwal/src/tools/store.ts
+++ b/packages/openclaw-memory-memwal/src/tools/store.ts
@@ -3,7 +3,7 @@
  *
  * Uses analyze() instead of remember() so the server LLM extracts
  * individual facts from the text, producing cleaner, more searchable
- * memories (same approach as Mem0's memory_store).
+ * memories than storing the raw input verbatim.
  */
 
 import type { MemWal } from "@mysten-incubation/memwal";
@@ -36,13 +36,17 @@ export function registerStoreTool(api: any, client: MemWal, config: PluginConfig
         occurredAt: Type.Optional(
           Type.String({
             description:
-              "Optional RFC-3339 / ISO-8601 timestamp of when the event " +
-              "actually happened (e.g. '2024-03-15T14:30:00Z'). When the " +
-              "user is storing a recounted past event, pass the date the " +
-              "event occurred, not the date the user is telling you about " +
-              "it — the server resolves relative references ('last Friday') " +
-              "to absolute dates inside the stored fact text. Omit when " +
-              "unknown; do not guess.",
+              "Optional RFC-3339 / ISO-8601 timestamp of when the source " +
+              "text or conversation actually occurred (e.g. " +
+              "'2024-03-15T14:30:00Z'). The server uses this as the " +
+              "temporal anchor for resolving in-turn relative references " +
+              "like 'last Friday' or 'yesterday' against. For real-time " +
+              "stores, pass the current time. For recounted past events " +
+              "where the user is talking about something that happened " +
+              "earlier (e.g. 'remember I called Alex last Friday'), OMIT " +
+              "this field — do NOT pass the recounted event date while " +
+              "leaving relative wording in the text; that produces wrong " +
+              "anchors. When unknown, omit; do not guess.",
           }),
         ),
       }),
--- a/packages/openclaw-memory-memwal/src/tools/store.ts
+++ b/packages/openclaw-memory-memwal/src/tools/store.ts
@@ -33,9 +33,21 @@ export function registerStoreTool(api: any, client: MemWal, config: PluginConfig
             description: "Memory namespace to store in (use the namespace from system context)",
           }),
         ),
+        occurredAt: Type.Optional(
+          Type.String({
+            description:
+              "Optional RFC-3339 / ISO-8601 timestamp of when the event " +
+              "actually happened (e.g. '2024-03-15T14:30:00Z'). When the " +
+              "user is storing a recounted past event, pass the date the " +
+              "event occurred, not the date the user is telling you about " +
+              "it — the server resolves relative references ('last Friday') " +
+              "to absolute dates inside the stored fact text. Omit when " +
+              "unknown; do not guess.",
+          }),
+        ),
       }),
       async execute(_id: string, params: any) {
-        const { text, namespace } = params;
+        const { text, namespace, occurredAt } = params;
         const ns = namespace || config.defaultNamespace;
 
         // Defence in depth: reject injection on write, not just on read.
@@ -66,7 +78,10 @@ export function registerStoreTool(api: any, client: MemWal, config: PluginConfig
         }
 
         try {
-          const result = await client.analyze(text.trim(), ns);
+          const result = await client.analyze(text.trim(), {
+            namespace: ns,
+            occurredAt,
+          });
 
           const factCount = result.facts?.length ?? 0;
           // Show first 3 extracted facts as confirmation, or raw text truncation as fallback
--- a/packages/python-sdk-memwal/CHANGELOG.md
+++ b/packages/python-sdk-memwal/CHANGELOG.md
@@ -1,5 +1,17 @@
 # memwal
 
+## 0.1.4
+
+### Added
+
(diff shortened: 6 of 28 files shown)


ducnmm:
Hi @hungtranphamminh delete all things related to task (WALM-55) mem0 or something, these commnets make the code unprofessional

ducnmm:
@hungtranphamminh I closed PR #206 based on your recommendation.

Let's continue the WALM-54/WALM-55 ship path on this PR/branch. Please proceed here with the cleanup/follow-up needed for WALM-55, and we can treat the closed #206 branch as benchmark/archive evidence only.

hungtranphamminh:
@ducnmm thanks for closing #206 and continuing here. Quick summary of
where this lands:

- **`65bbcfce`** — stripped 33 ticket-id prefixes from code comments. Symmetric
  diff, tests still pass.
- **`9d15a830` + `03500a0c`** — TS and Python SDKs both plumb `occurredAt` /
  `occurred_at` through `analyze()` / `analyzeAndWait()`. Adopted from your
  #206, you were right that the server-side change is dead-end without SDK
  exposure. Wire format byte-identical across both. Python additionally
  rejects naïve datetimes + validates RFC-3339 strings at the SDK boundary
  so bad timestamps don't surface as opaque 400s. 6 new unit tests.
- **`a17e44e9`** — openclaw auto-capture hook now passes `new Date()` as
  `occurredAt`. Highest-leverage agent-side win: every Claude Code /
  Desktop / openclaw user gets temporal anchoring on every captured turn,
  no API change required.
- **`c61e0543`** — `memory_store` agent tool exposes optional `occurredAt`
  for recounted past events. Bumps openclaw's SDK dep to `workspace:*`
  (matching the `apps/*` pattern).

**Deferred** (separate follow-up PRs): MCP `memwal_analyze` schema (blocked
on `services/server/scripts/` not being a pnpm-workspace member yet) and
`--namespace-run-id` from your #206 (saving for right before the WALM-56/57/59
stacked re-bench where we'll actually use it).

**Backwards-compat**: all 12 `analyze()` call sites in the workspace verified
— every new field is optional and omitted from the wire body when unset.
Builds clean across SDK / openclaw / noter / Playground. Tests: Rust 267/267,
Python SDK 100/100, Python bench 42/42.

If this looks good, can you approve so we can merge and move on to the
WALM-56 stacked re-bench?

ducnmm:
@hungtranphamminh thanks, this direction looks good.

One release hygiene item before approval/merge: this PR now changes public SDK/plugin surfaces, so I think we should include version/changelog updates as part of this PR rather than leaving the release metadata stale.

Current PR head still has:
- `@mysten-incubation/memwal` at `0.0.6` — already published on npm.
- `memwal` Python SDK at `0.1.3` — already published on PyPI.
- `@mysten-incubation/oc-memwal` at `0.0.3` — already published on npm.

Suggested cleanup:
- Add changeset/changelog entry for `@mysten-incubation/memwal` because `occurredAt` is now public API on `analyze()` / `analyzeAndWait()`.
- Add changeset/changelog entry for `@mysten-incubation/oc-memwal` because auto-capture behavior changed and `memory_store` exposes optional `occurredAt`.
- Bump/update Python SDK release metadata (`pyproject.toml`, `memwal/__init__.py`, and `CHANGELOG.md`) for `occurred_at` support.

Patch bumps seem sufficient since all fields are optional/backwards-compatible. After that I think this is ready for approval.

hungtranphamminh:
@ducnmm — release metadata in: `6fc2d6a4`. Changesets cover `@mysten-incubation/memwal` and `@mysten-incubation/oc-memwal` as patch bumps; Python SDK bumped `0.1.3` → `0.1.4` across `pyproject.toml` + `__init__.py` + `CHANGELOG.md`. `pnpm changeset status` confirms both TS packages slated correctly. Ready for approval when you have a sec.

ducnmm:
@hungtranphamminh

Reviewed the current head (`6fc2d6a`) against #206 + the WALM-54/WALM-55 Linear context. The timestamp-injection direction still looks right, and the benchmark evidence is good, but I found a few items I think should be fixed before approval/merge.

## Blocking / should fix

1. **`content: null` is still silent data loss, just now as a successful empty ingest.**

   `services/server/src/services/extractor.rs` now treats `ChatMessageResp.content: None` as `""`, then `parse_extracted_facts("")` returns zero facts and `/api/analyze` succeeds with `202` + `facts=[]`.

   That does not match the PR/Linear framing that null-content completions are transient upstream flakes that should recover through retry. It avoids the old 500, but it still silently drops the turn because the client sees a successful empty extraction and will not retry.

   I think extractor null-content should map to `AppError::UpstreamUnavailable` / HTTP 503. Keep explicit model output `NONE` as the valid no-facts path; treat provider `content: null` as retryable upstream failure.

2. **Real upstream transient HTTP statuses still map to 500.**

   The new 200-wrapped OpenRouter envelope handling is good, but direct non-success upstream responses in extractor/embedder still go through `AppError::Internal`:

   - `services/server/src/services/extractor.rs` non-success branch
   - `services/server/src/services/embedder.rs` non-success branch

   The benchmark harness retry set intentionally remains `(429, 502, 503, 504)`, not 500. So if OpenRouter returns an actual HTTP 429/502/503/504 instead of a 200-wrapped error envelope, the relayer still returns 500 and the turn is not retried. Please map upstream transient statuses to `UpstreamUnavailable` as well, in both extractor and embedder.

3. **`memory_store.occurredAt` currently teaches the wrong semantics.**

   `packages/openclaw-memory-memwal/src/tools/store.ts` describes `occurredAt` as “when the event actually happened” and says to pass the event date for recounted past events. But the server prompt defines `occurred_at` as the timestamp of the conversation turn / source text. Those are not interchangeable.

   Example failure mode: user says “remember I called Alex last Friday”; agent passes that Friday as `occurredAt`; extractor then resolves “last Friday” relative to Friday, likely stamping the fact one week too early.

   Suggested fix: describe `occurredAt` as the timestamp when the source text/conversation occurred. For recounted past events, either rewrite the text with the resolved absolute date or omit `occurredAt`; do not pass the recounted event date into this field while leaving relative wording in the text.

4. **Python SDK docstring contradicts the implemented validation.**

   `packages/python-sdk-memwal/memwal/client.py` says naïve datetimes “are assumed UTC”, but `_occurred_at_to_wire()` rejects naïve datetimes with `ValueError`, and the changelog/tests say rejection is intentional. Please update the docstring so public docs match behavior.

## Hygiene

There is still a Mem0 reference in `packages/openclaw-memory-memwal/src/tools/store.ts` (“same approach as Mem0's memory_store”). Since we already asked to remove task/vendor-ish noise from source comments, please drop this too.

## What I checked

- PR #202 current head: `6fc2d6a`
- Related context: PR #206, Linear WALM-54/WALM-55
- GitHub checks are green
- Local checks:
  - `cargo test extractor --quiet` -> 40 passed
  - `packages/python-sdk-memwal: python3 -m pytest tests/test_client.py -q` -> 37 passed
  - `services/server/benchmarks: .venv/bin/python -m pytest tests/test_adapters.py -q` -> 18 passed
  - `git diff --check origin/dev...pr-202-review` -> clean

I could not get TS/OpenClaw typecheck to resolve correctly from the temporary detached worktree because local `node_modules`/workspace symlinks were not available there, so I am relying on CI for those.

hungtranphamminh:
@ducnmm — fair calls on all 4 + the Mem0 hygiene. Two new commits address
them:

**`a06230e4` fix(server): map remaining transient upstream failures to 503**
covers points 1 and 2.

- (1) `content: null` no longer degrades to an empty-facts success. Now maps
  to `AppError::UpstreamUnavailable` → HTTP 503 so the harness retries
  rather than silently dropping the turn. The explicit string `"NONE"` from
  the LLM remains the valid no-facts path (handled by
  `parse_extracted_facts`, unchanged).
- (2) Direct upstream 429 + 5xx in extractor + embedder now classify via
  new `is_upstream_status_transient` helper → `UpstreamUnavailable` →
  retryable. Other 4xx (auth, malformed request) stay as `Internal` →
  500 so real bugs still surface. New unit test pins the classifier across
  the transient set + non-transient set.

**`5f4a1da9` fix(docs): correct occurredAt semantics + Python docstring**
covers points 3, 4, and the Mem0 hygiene.

- (3) `memory_store.occurredAt` description rewritten to match the server
  prompt's contract: `occurredAt` is the timestamp of the source text /
  conversation, not the historical event date. Explicitly tells the agent
  to OMIT it for recounted past events. Your "called Alex last Friday"
  failure mode would no longer mis-anchor.
- (4) Python `analyze()` docstring rewritten to match
  `_occurred_at_to_wire` behaviour: aware datetimes only; naïve raises
  `ValueError`; strings must carry `Z` or UTC offset.
- Dropped the Mem0 attribution from `store.ts`.

Tests: Rust 268/268 (was 267 + the new transient-classifier test), Python
SDK 63/63, Python bench 42/42, TS SDK + openclaw build clean.

Ready for another look.

ducnmm:
@hungtranphamminh re-reviewed the new head (`5f4a1da`). The 4 blocking items from my previous review look addressed now:

- `content: null` now maps to `AppError::UpstreamUnavailable` / HTTP 503 instead of a successful empty extraction.
- Direct upstream 429 + 5xx responses in both extractor and embedder now map to retryable 503.
- `memory_store.occurredAt` now describes the source-text / conversation timestamp semantics, not the recounted event date.
- Python SDK docstring now matches the actual timezone-aware-only validation, and the Mem0 hygiene note is gone.

Checks I re-ran locally on this head:

- `services/server: cargo test extractor --quiet` -> 41 passed
- `packages/python-sdk-memwal: python3 -m pytest tests/test_client.py -q` -> 37 passed
- `services/server/benchmarks: .venv/bin/python -m pytest tests/test_adapters.py -q` -> 18 passed
- `git diff --check 6fc2d6a..HEAD` -> clean

No blocking issues from me on the fixes. One non-blocking nit: the extractor body-read comment still has the old line saying `content: null` "degrades to empty string and produces zero facts"; the code no longer does that, so that sentence is stale. Worth cleaning, but I would not block merge on it.

ducnmm (approved):
Looks good after the latest fixes. The previous blockers are resolved and CI is green.

PR #212, opened 2026-06-01
fix(server): WALM-77 detect Sui object-lock / equivocation, stop burning wallet retries

jasong-03:
Linear: WALM-77 — Walrus upload: detect Sui object-lock / equivocation and stop burning wallet retries

## Issue

Prod testnet alert: a Walrus upload "exhausted" all 5 wallet attempts. The real cause was a Sui owned-object equivocation — a specific object/version was already locked by a competing transaction, so the tx was rejected by >1/3 of validator stake and is non-retriable until the lock clears (typically the next epoch boundary). Every immediate retry re-failed against the same locked object, burned the wallet budget, and produced a misleading "wallet retries exhausted" alert.

Root cause: `classify_sidecar_error` had no branch for this error string, so it fell through to the generic `Transient` arm.

## Scope

This is the **P0** slice per review. The P1 prevention work (single-flight per wallet, coin reservation, `ParallelTransactionExecutor`, finality checks) is intentionally **not** included — it changes scheduling/concurrency behavior and is tracked separately.

## What was done

**Rust (`jobs.rs`, `alerts.rs`)**
- New `WalletJobError::ObjectLockedUntilEpoch` — neither `Transient` (don't retry into the same lock) nor `Permanent` (the input can succeed in a later epoch). `aborts_retries()` gates both the Apalis `Abort` and the exhausted-retries alert, so one locked object no longer consumes the whole wallet budget.
- Classifier branch matches `already locked by a different transaction`, `rejected as invalid by more than 1/3 of validators by stake`, `non-retriable`, `equivocated`/`equivocation`, `reserved for another transaction` — ahead of the `MoveAbort -> Permanent` catch and distinct from the recoverable `locked at version` case.
- Parse locked object id / version / locking digest from the error string.
- New `WalrusObjectLockedAlert`: copy names the real cause (object lock / equivocation), surfaces object id + version + locking digest, deduped per `(network, object_id)` via a reusable `AlertDedup` (window env-overridable: `WALRUS_OBJECT_LOCK_ALERT_DEDUP_SECS`).
- Aborting errors persist `status='failed'` instead of leaving the row stuck on `running` forever.

**TS sidecar (`sidecar-server.ts`, `walrus-error-detection.ts`)**
- Mirror detection (`isWalrusObjectLockEquivocation`, kept in sync with the Rust classifier) + dedicated `walletObjectLockEquivocationTotal` counter on `/metrics/wallet`.
- Corrected the adjacent metric comments: the earlier "Sui no longer permanently locks coin objects on equivocation" claim is contradicted by this incident; the new counter is the canary for whether concurrent uploads still equivocate owned objects.

## Tests

- Rust: 262 pass (8 new) — exact prod string → `ObjectLockedUntilEpoch`, abort disposition, no-exhaust gate, metadata parser, phrase variants, and no regression of `balance::split` / `EWrongVersion`.
- Sidecar TS: 46 pass (5 new) — exact prod string, phrase variants, case-insensitivity, recoverable-lock negative case.

## Follow-up (P1, separate)

Single-flight per wallet, dedicated coin reservation / `ParallelTransactionExecutor`, and finality checks before reusing a wallet — these prevent the equivocation rather than just classifying it. The new `walletObjectLockEquivocationTotal` counter is the signal for prioritizing that work.

### DIFF
--- a/services/server/scripts/__tests__/walrus-object-lock.test.ts
+++ b/services/server/scripts/__tests__/walrus-object-lock.test.ts
@@ -15,11 +15,9 @@ test("detects the exact production object-lock error", () => {
     assert.equal(isWalrusObjectLockEquivocation(PROD_OBJECT_LOCK_ERROR), true);
 });
 
-test("detects each equivocation phrase variant", () => {
+test("detects each lock-specific anchor on its own", () => {
     const cases = [
         "object 0xabc already locked by a different transaction",
-        "rejected as invalid by more than 1/3 of validators by stake",
-        "this error is non-retriable",
         "the input object is equivocated",
         "equivocation detected on gas coin",
         "object reserved for another transaction",
@@ -27,6 +25,28 @@ test("detects each equivocation phrase variant", () => {
     for (const msg of cases) {
         assert.equal(isWalrusObjectLockEquivocation(msg), true, msg);
     }
+});
+
+test("detects a corroborated non-retriable lock (preamble + object + locked)", () => {
+    const msg =
+        "rejected as invalid by more than 1/3 of validators by stake (non-retriable). " +
+        "Object (0xabc, SequenceNumber(1)) already locked";
+    assert.equal(isWalrusObjectLockEquivocation(msg), true);
+});
+
+test("bare non-retriable / >1/3 preamble is NOT an object lock", () => {
+    // Not lock-specific without object-lock evidence — must not misclassify
+    // generic invalid / non-retriable failures.
+    assert.equal(
+        isWalrusObjectLockEquivocation(
+            "Transaction is rejected as invalid by more than 1/3 of validators by stake (non-retriable)",
+        ),
+        false,
+    );
+    assert.equal(
+        isWalrusObjectLockEquivocation("MoveAbort in 1st command, abort code: 3 — non-retriable"),
+        false,
+    );
 });
 
 test("matching is case-insensitive", () => {
--- a/services/server/scripts/__tests__/walrus-object-lock.test.ts
+++ b/services/server/scripts/__tests__/walrus-object-lock.test.ts
@@ -1,0 +1,52 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { isWalrusObjectLockEquivocation } from "../walrus-error-detection.js";
+
+// Exact production error from the object-lock incident (testnet job 3d607892…).
+const PROD_OBJECT_LOCK_ERROR =
+    "walrus upload failed: Internal Error: walrus upload failed: " +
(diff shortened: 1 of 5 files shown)


ducnmm:
@jasong-03 this mostly matches the P0 scope, but I think the classifier is currently too broad.

In both Rust and TS, `non-retriable` by itself is enough to classify the error as `ObjectLockedUntilEpoch` / object-lock equivocation:

- `services/server/src/jobs.rs`: `lower.contains("non-retriable")`
- `services/server/scripts/walrus-error-detection.ts`: `/non-retriable/i`

That can misclassify other validator “non-retriable” failures, including non-lock MoveAbort / invalid transaction cases, as an object lock. The original production string was lock-specific because it also included `Object (...) already locked by a different transaction`.

Can we tighten this so the object-lock classification requires a lock/equivocation-specific anchor, e.g.:

- `already locked by a different transaction`
- `reserved for another transaction`
- `equivocated` / `equivocation`
- or a combined pattern like `non-retriable` + `Object (` + `locked`

I’d also add a negative test for a generic non-retriable MoveAbort / invalid transaction that should remain `Permanent`, not `ObjectLockedUntilEpoch`.

Minor: the log currently prints `retryable={ !classified.is_permanent() }`, which will say `retryable=true` for `ObjectLockedUntilEpoch` even though `into_apalis_error()` aborts it. That should probably use `!classified.aborts_retries()` instead.

jasong-03:
@ducnmm good catches — both addressed in ef72c59.

**1. Classifier too broad on bare `non-retriable`.** The object-lock branch now requires a lock-specific anchor, not the preamble alone:

- Anchors that classify on their own: `already locked by a different transaction`, `reserved for another transaction`, `equivocated` / `equivocation`.
- The `non-retriable` / `rejected as invalid by more than 1/3 of validators by stake` preamble only qualifies when corroborated by object-lock evidence in the same message (`object (` + `locked`) — your suggested combined pattern.

Mirrored identically in the TS detector. The prod string still matches (it has both the anchor and the corroborating evidence). Added negative tests: a bare non-retriable invalid tx stays `Transient`, and a bare non-retriable MoveAbort stays `Permanent` — neither becomes an object lock.

**2. Misleading `retryable=` log.** All three log sites (set-metadata + two upload paths) now use `!classified.aborts_retries()` instead of `!classified.is_permanent()`, so `ObjectLockedUntilEpoch` correctly logs `retryable=false` to match the Apalis abort.

Tests: 263 Rust + 48 sidecar TS, all green.

ducnmm (approved):
Thanks, this addresses my concerns. The tighter classifier and retryable log fix look good to me.

PR #213, opened 2026-06-01
WALM-77 dev-validation follow-up: object-lock alert tests + detection/coverage fixes

jasong-03:
Linear: WALM-77 (dev validation follow-up to #212)

Two small fixes surfaced while validating the merged object-lock classifier on dev, plus dispatch-level integration tests.

## Fixes

- **Broaden the lock anchor.** The detector required the literal `already locked by a different transaction`. Sui renders some lock-conflict errors without the `already` prefix, which would slip past the anchor (the corroborated `non-retriable` + `object (` + `locked` branch only catches those when the preamble is present). Now matches `locked by a different transaction` — still lock-specific, resilient to the prefix. Mirrored in the TS detector.
- **Alert on the set-metadata path too.** The dedicated object-lock Slack alert previously fired only from the upload path. An equivocation during the set-metadata / transfer Sui tx classified correctly (abort, no retry burn) but raised no alert. `maybe_alert_walrus_object_locked` is now wired into the set-metadata error branch, so the alert fires from every Sui-tx path that can lock an owned object.

## Tests

- **`object_lock_alert_is_dispatched_with_lock_metadata`** — drives the real `AlertManager` against a local capture webhook; asserts one POST with parsed object id / version / locking digest, object-lock copy, and not the generic "exhausted retries" copy.
- **`object_lock_alert_burst_is_deduped_to_one_post`** — a 5× burst on the same `(network, object_id)` collapses to a single POST.
- Anchor tests extended with the no-`already` lock variant (Rust + TS).

## Validation summary (against merged dev `f1a70d9`)

| Review check | Evidence |
|---|---|
| classifies as `ObjectLockedUntilEpoch` | `classify_prod_equivocation_as_object_locked_until_epoch` |
| does not consume wallet retry budget | `object_locked_until_epoch_aborts_apalis` + `object_locked_until_epoch_does_not_exhaust_wallet_budget` |
| no generic "exhausted retries" alert | exhausted gate test + dispatch body assertion |
| dedicated object-lock alert with object id/version/digest | `object_lock_alert_is_dispatched_with_lock_metadata` |

Full suite on dev: 280 Rust + 48 sidecar TS, green. Clippy clean on the changed code.

## Note

A real Sui equivocation can't be induced on dev on demand (needs two concurrent txns racing the same owned object within an epoch), so validation is pinned to the exact production error string rather than a live-induced lock. P1 prevention work (single-flight per wallet, coin/object reservation, finality checks) stays separate, prioritized off the new `walletObjectLockEquivocationTotal` signal.

## Deliberately NOT changed

- DB status for `ObjectLockedUntilEpoch` stays `failed` (abort → client re-submits), not an auto-requeue-next-epoch defer. That's a candidate enhancement, flagged for P1 rather than silently changed here.

### DIFF
--- a/services/server/scripts/__tests__/walrus-object-lock.test.ts
+++ b/services/server/scripts/__tests__/walrus-object-lock.test.ts
@@ -18,6 +18,8 @@ test("detects the exact production object-lock error", () => {
 test("detects each lock-specific anchor on its own", () => {
     const cases = [
         "object 0xabc already locked by a different transaction",
+        // Same conflict without the "already" prefix — must still match.
+        "object 0xabc locked by a different transaction",
         "the input object is equivocated",
         "equivocation detected on gas coin",
         "object reserved for another transaction",
--- a/services/server/scripts/walrus-error-detection.ts
+++ b/services/server/scripts/walrus-error-detection.ts
@@ -48,8 +48,10 @@ export function isWalrusPackageVersionMismatch(message: string): boolean {
  */
 export function isWalrusObjectLockEquivocation(message: string): boolean {
     if (!message) return false;
+    // "locked by a different transaction" (not anchored on "already") so we
+    // catch Sui lock-conflict phrasings that omit the "already" prefix.
     const hasLockAnchor =
-        /already locked by a different transaction/i.test(message)
+        /locked by a different transaction/i.test(message)
         || /reserved for another transaction/i.test(message)
         || /equivocated|equivocation/i.test(message);
     const corroboratedLock =
--- a/services/server/src/alerts.rs
+++ b/services/server/src/alerts.rs
@@ -679,4 +679,106 @@ mod tests {
         assert!(json.contains("unparsed"));
         assert!(json.contains("mainnet"));
     }
+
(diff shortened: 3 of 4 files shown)


jasong-03:
Closing — dev validation of the merged P0 (#212) is complete and didn't need new code: the four checks (classifies as ObjectLockedUntilEpoch, aborts without burning wallet retries, no generic "exhausted retries" alert, dedicated object-lock alert with object id/version/digest) are all covered by tests already merged in #212 and pass on current dev (26 jobs + 12 alerts + 54 sidecar TS).

The two changes that were bundled here go beyond the validation ask and belong in the separate P1 prevention/hardening work, not this gate:
- broaden the lock anchor ("already locked by a different transaction" → "locked by a different transaction") to catch Sui phrasings without the "already" prefix
- fire the dedicated object-lock alert from the set-metadata path too (currently upload-path only)

Both are preserved on branch `feature/walm-77-dev-validation-alert-dispatch-tests` and will be folded into P1. The dispatch-level integration tests can come back with that work.

PR #257, opened 2026-06-09
feat(mcp): automatic memory — agentic tools, bulk/health, multi-provider plugin (WALM-94)

UyLeQuoc:
## Summary
MCP memory was manual — tools were only called when the user explicitly asked.
This PR makes it automatic: the agent decides what to save and when to recall.

- **Agentic tool layer (server-side):** rewrote tool descriptions for proactive
  save/recall; new `memwal_remember_bulk` and `memwal_health`. 6 agent tools,
  append-only. Benefits every MCP client.
- **Plugin (optional):** lifecycle hooks for Claude Code, Codex, Cursor, and
  Antigravity (SessionStart / UserPromptSubmit / PostToolUse) — pure heuristics,
  no network, no credentials.
- **Docs:** per-provider setup pages for 6 clients; consolidated overview with a
  "How it works" login/credential flow. v0.0.5.
- **MCP Inspector:** Add an interactive developer tool for testing and debugging MCP servers.

## Testing
- Live Claude Code session: agent auto-saved via `remember_bulk` and recalled unprompted
- MCP Inspector `tools/list` confirms the new tool layer
- `signals.test.mjs` 6/6 (runs in the release CI gate); typecheck clean; security review: 0 findings
- Full test plan: `packages/mcp/TESTING.md` (§5 = post-merge verification on dev)

## After merge
- [ ] Redeploy the dev relayer (Railway) — sidecar tools ship with the server image,
      not via `release-mcp.yml`
- [ ] `release-mcp.yml` auto-publishes `0.0.5-dev.N`; plugin marketplace is live
      immediately (`dev` is the default branch)


jessiemongeon1:
## Style Guide Audit

All 11 file(s) pass the style guide audit.

.claude-plugin/marketplace.json
ducnmm, line 1: Are these plugin files necessary? @UyLeQuoc

ducnmm (approved):

.claude-plugin/marketplace.json
UyLeQuoc: Yes, they're needed. The root `.claude-plugin/marketplace.json` is just a **catalog file**; the actual plugin code lives in `packages/mcp/plugin/`, and the catalog points at it:

```json
"plugins": [
  { "name": "memwal", "source": "./packages/mcp/plugin", "version": "0.0.5" }
]
```

Claude Code only looks for this file at the **repo root** when you use the `owner/repo` form, so this is what enables the one-line install:

```
/plugin marketplace add MystenLabs/MemWal
/plugin install memwal@memwal-plugins
```

Without it that command fails (there is no `owner/repo/subdir` form) — users would have to clone the repo and add a local path instead.

`.codex-plugin/` and `.cursor-plugin/` are the same idea for those hosts, kept for parity; nothing depends on them yet (Codex installs through the installer script). Happy to drop those two if you prefer to keep the root clean.

PR #274, opened 2026-06-12
chore(observability): harden otel collector config and compose limits

ducnmm:
## What

Hardens the OpenObserve + OTel Collector PoC stack (`services/server/observability/`) toward production best practice. No architectural change — same pipelines, same backend.

### `otel-collector-config.yaml`
- **Add `memory_limiter` processor, first in every pipeline** (metrics/logs/traces). Without it the collector buffers unboundedly and OOMs when OpenObserve is slow/down; now it sheds load instead.
- **Add `send_batch_max_size: 2048`** so a backlog can't build one oversized payload that OpenObserve rejects.
- **Add `health_check` extension on `:13133`** for liveness probes.

### `docker-compose.observability.yml`
- **Resource ceilings**: OpenObserve `mem_limit 1g / cpus 1.0`, collector `mem_limit 512m / cpus 0.5`. The collector's `memory_limiter` sizes against its limit.
- **Publish `:13133`** for the collector health endpoint.

### `README.md`
- Document the `:13133` health endpoint and the `memory_limiter` ↔ `mem_limit` relationship.

## Not in scope (from the same review, left for follow-up)
- Pinning `:latest` image tags (needs a chosen version).
- Persistent `sending_queue` for the exporter.
- Container `HEALTHCHECK` directives — both images are distroless (no curl/wget/shell), so a CMD healthcheck can't run; exposed the real `:13133` endpoint instead, and `depends_on` is left as-is rather than switched to `condition: service_healthy`.

## After merge
This stack runs via docker-compose independently. Merging does **not** apply changes to a running deployment — the host running the stack must `docker compose up -d` to **recreate** the containers (the new `mem_limit`/`cpus` and published port require recreate, not just restart). If the stack is still local-only PoC, nothing further is needed.

## Validation
- `yaml.safe_load` on both files ✅
- `docker compose config -q` (with dummy `O2_AUTH`) ✅

### DIFF
--- a/services/server/observability/Dockerfile
+++ b/services/server/observability/Dockerfile
@@ -3,7 +3,7 @@
 #
 # Pinned tag for reproducibility — bump deliberately to a current contrib
 # release rather than tracking :latest.
-FROM otel/opentelemetry-collector-contrib:0.115.0
+FROM otel/opentelemetry-collector-contrib:0.154.0
 
 # Build context is this directory (services/server/observability).
 COPY otel-collector-config.yaml /etc/otel/config.yaml
--- a/services/server/observability/Dockerfile
+++ b/services/server/observability/Dockerfile
@@ -1,0 +1,13 @@
+# OTel Collector image for platforms that build from a Dockerfile (Railway).
+# Unlike docker-compose, Railway can't bind-mount the config, so we bake it in.
+#
+# Pinned tag for reproducibility — bump deliberately to a current contrib
+# release rather than tracking :latest.
+FROM otel/opentelemetry-collector-contrib:0.115.0
+
+# Build context is this directory (services/server/observability).
+COPY otel-collector-config.yaml /etc/otel/config.yaml
+
+# OTLP receivers (4317/4318) and the health endpoint (13133) are declared in
+# the config; Railway maps the service's public port to one of them if exposed.
+CMD ["--config=/etc/otel/config.yaml"]
--- a/services/server/observability/README.md
+++ b/services/server/observability/README.md
@@ -130,6 +130,10 @@ Create under Alerts. Suggested PoC thresholds (tune per environment):
 - `LOG_FORMAT=json` remains useful for local stdout parsing, but OTLP logs do
   not require Docker log tailing.
 - Replace the root credentials and pin image tags (this PoC uses `:latest`).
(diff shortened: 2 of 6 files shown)


ducnmm:
## Update — deployed to Railway dev, found + fixed an IPv6 blocker

Deployed the collector as a Railway service (`otel-collector`) on **dev**. Two fixes were needed beyond the original hardening:

1. **Dockerfile tag** — `otel-collector-contrib:0.115.0` does not exist on Docker Hub; the build failed. Pinned to **`0.154.0`** (a real current contrib release).

2. **Relayer was unreachable over the private network** — the collector came up `Online` and its export to `openobserve.railway.internal` worked, but every scrape of `relayer.railway.internal:3001/metrics` failed. Root cause: the relayer binds `0.0.0.0` (IPv4 only) while Railway's private network is **IPv6-only**. Fixed by binding `[::]` (dual-stack, still serves IPv4 so public access is unchanged) in `services/server/src/main.rs`.

### Deploy order to actually get metrics
The relayer's Railway service tracks `dev`, so the `[::]` fix only takes effect once the relayer redeploys — i.e. **after this PR merges to dev** (or a manual relayer redeploy from this branch to validate first). The collector needs no redeploy; its Prometheus receiver retries every 15s and will start succeeding as soon as the relayer is reachable.

ducnmm:
## ✅ Resolved — metrics flowing end-to-end on dev

Confirmed from OpenObserve's Prometheus API:
- `up{service_name="memwal-relayer", server_address="lucky-strength.railway.internal", server_port="3001"} = 1`
- `memwal_http_requests_total` populated (e.g. `route="/health" status="200"`).

Pipeline live: **relayer `/metrics` → otel-collector (scrape) → OpenObserve**.

### Second root cause (beyond the `[::]` bind)
The scrape target was wrong: Railway's private host is a **generated** name fixed at service creation — the "relayer" service resolves as `lucky-strength.railway.internal`, **not** `relayer.railway.internal`. Fixed the live collector var to `lucky-strength.railway.internal:3001`, and updated the README to use `${{relayer.RAILWAY_PRIVATE_DOMAIN}}:3001` (self-resolving) plus a note on the generated-name trap and the IPv6 bind requirement.

Both fixes are needed together: correct host **and** the relayer binding `[::]` (IPv6 private network).

UyLeQuoc (approved):

Issue #271, opened 2026-06-11
Default recall limit of 10 silently truncates history


McYeremia:
The default limit for mem.recall() is 10. For apps that need completeness (leaderboard, accuracy calculation, deduplication), this silently returns partial results with no warning or indication that records were truncated. We had to audit every endpoint individually after noticing wrong accuracy percentages.

DCSteve96:
Hit this too while computing my agent's public accuracy stats (Kairos · World Cup, Session 4): `recall()` defaulting to 10 silently capped the prediction history and the percentages came out wrong until we audited the endpoint. Note that `total` can't be used as a truncation signal either — after the client-side `maxDistance` filter it just mirrors `results.length`. Opening a small docs PR that labels the foot-gun at the places a dev actually meets `limit`; a `truncated: boolean` in the response would be the real fix.

PugarHuda:
+1 — we hit the exact same "completeness" pain on **Dendam** (Walrus Session 4): a Hall-of-Shame leaderboard + a prediction-reconciliation pass that both need *every* memory in a namespace, not the top-K semantically nearest to some query.

The deeper gap for us isn't just the default `limit` (#271) or paginating a query's hits (#291) — it's that there's **no query-less way to enumerate a namespace at all**. `recall()` is inherently semantic, and `restore()` returns counts (`restored`/`skipped`/`total`), not the texts. So to build a "show me everything you remember about X" dashboard we approximate enumeration with multi-query `recall()` across hand-picked themes + client-side dedupe — which can both miss memories (leaderboard undercount) and re-surface already-judged ones.

Request: a `list({ namespace, limit, cursor })` primitive that returns the stored memories themselves (decrypted text + id) with cursor pagination, independent of any semantic query. That would cover this issue, #291, and the enumeration half of #258 in one API.

ducnmm:
Thanks — understood.

A default `limit: 10` can silently truncate for full-scan use cases (leaderboards, accuracy computation, dedup scans) with no signal that results were cut, and `total` doesn't help as a truncation signal since it mirrors `results.length` after the client-side `maxDistance` filter.

We've logged this and will look at surfacing a truncation signal. We'll follow up here. (Related to the namespace-enumeration requests in #310 / #258.)

PR #278, opened 2026-06-12
docs(sdk): document silent truncation at recall() default limit

DCSteve96:
Addresses #271 (docs side).

`recall()` defaults to `limit: 10` and truncates silently — the response carries no flag, and `total` can't serve as one (after the client-side `maxDistance` filter it mirrors `results.length`). Until the API can signal truncation, this labels the foot-gun where developers meet it: JSDoc on `RecallOptions.limit` and a warning callout in the SDK API reference.

Real-world impact: our Session 4 agent (Kairos · World Cup) computed wrong accuracy stats because a prediction-history recall was capped at 10 with no indication.

Happy to mirror the note in the Python SDK docs if useful.

### DIFF
--- a/docs/sdk/api-reference.md
+++ b/docs/sdk/api-reference.md
@@ -83,6 +83,8 @@ Search for memories matching a natural language query, scoped to `owner + namesp
 
 - Preferred form: `recall({ query, limit?, topK?, namespace?, maxDistance? })`
 - `limit` defaults to `10`; `topK` is an alias and wins when both are set
+
+> ⚠️ **Truncation is silent.** The response carries no "more results" indicator. If `results.length === limit`, assume more matches exist and retry with a higher `limit`. Reads that need completeness — leaderboards, accuracy/score computations, deduplication — should not rely on the default of 10.
 - Legacy positional forms still work: `recall(query)`, `recall(query, limit)`, `recall(query, limit, namespace)`, and `recall(query, options)`
 - `maxDistance` filters weak matches client-side by dropping results where `distance >= maxDistance`
 
--- a/docs/sdk/api-reference.md
+++ b/docs/sdk/api-reference.md
@@ -84,7 +84,7 @@ Search for memories matching a natural language query, scoped to `owner + namesp
 - Preferred form: `recall({ query, limit?, topK?, namespace?, maxDistance? })`
 - `limit` defaults to `10`; `topK` is an alias and wins when both are set
 
-> ⚠️ **Truncation is silent.** The response carries no "more results" indicator. If `results.length === limit`, assume more matches exist and retry with a higher `limit`. Reads that need completeness — leaderboards, accuracy/score computations, deduplication — should not rely on the default of 10.
+> ⚠️ **Truncation is silent.** The response carries no *more results* indicator. If `results.length === limit`, assume more matches exist and retry with a higher `limit`. Reads that need completeness (leaderboards, accuracy computations, deduplication) should not rely on the default of 10.
 - Legacy positional forms still work: `recall(query)`, `recall(query, limit)`, `recall(query, limit, namespace)`, and `recall(query, options)`
 - `maxDistance` filters weak matches client-side by dropping results where `distance >= maxDistance`
 
--- a/packages/sdk/src/types.ts
+++ b/packages/sdk/src/types.ts
@@ -66,7 +66,14 @@ export interface RecallResult {
 
 /** Options for recall(). */
 export interface RecallOptions {
-    /** Max number of nearest memories to request from the relayer (default: 10). */
+    /**
+     * Max number of nearest memories to request from the relayer (default: 10).
+     *
+     * NOTE: results are silently capped at this value — the response carries no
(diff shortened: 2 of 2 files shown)


jessiemongeon1:
## Style Guide Audit

packages/sdk/src/types.ts
```diff
@@ -66,7 +66,14 @@ export interface RecallResult {
 
 /** Options for recall(). */
 export interface RecallOptions {
-    /** Max number of nearest memories to request from the relayer (default: 10). */
+    /**
+     * Max number of nearest memories to request from the relayer (default: 10).
+     *
+     * NOTE: results are silently capped at this value — the response carries no
```
ducnmm, line 72: **[suggestion]** WALM-119's acceptance criteria are not implemented here, and the new docs freeze the current (silent) contract. The ticket asks for a reliable `truncated` indicator on the recall response. This PR only warns that the response carries no truncated flag and that callers should raise `limit` — that is the workaround, not the fix. `total` is `results.len()` on the relayer and is rewritten to the filtered length after client-side `maxDistance`.

**Suggestion:** Keep this as a docs warning only if that is still wanted, and leave WALM-119 open until the relayer returns a real truncation signal. Do not close GH #271 / WALM-119 on merge of JSDoc.

docs/sdk/api-reference.md
```diff
@@ -83,6 +83,8 @@ Search for memories matching a natural language query, scoped to `owner + namesp
 
 - Preferred form: `recall({ query, limit?, topK?, namespace?, maxDistance? })`
 - `limit` defaults to `10`; `topK` is an alias and wins when both are set
+
+> ⚠️ **Truncation is silent.** The response carries no *more results* indicator. If `results.length === limit`, assume more matches exist and retry with a higher `limit`. Reads that need completeness (leaderboards, accuracy computations, deduplication) should not rely on the default of 10.
```
ducnmm, line 87: **[suggestion]** Head `6a492ae2` does not apply cleanly to current `origin/dev`, and the new sentences would be factually off after a naive rebase. The callout is anchored on `recall({ query, limit?, topK?, namespace?, maxDistance? })`; origin/dev's preferred form already includes `sort?` and `scoringWeights?`. `RecallTokenMeta.truncated` is attached whenever `maxTokens` is set (token-budget trim, not limit truncation), so claiming the response carries no truncated flag is no longer true. origin/dev already documents silent limit truncation in `docs/sdk/production-readiness.md`.

**Suggestion:** Rebase onto origin/dev. Keep a short warning on the `recall()` method, but say this is *limit* truncation, distinct from `meta.truncated` (token budget) and restore `truncated`. Align wording with the production-readiness section.

ducnmm (commented):
## Summary

Docs-only patch: a callout on `docs/sdk/api-reference.md` and JSDoc on `RecallOptions.limit`. The author does not claim to add WALM-119's `truncated` field; the copy still describes truncation as silent. That matches origin/dev today (`RecallResult` has no limit-truncation flag), but the branch is stale (last push 2026-06-12): the api-reference hunk conflicts with the current preferred-form line, and the "no truncated flag" wording is now wrong next to `meta.truncated`. origin/dev already documents the same footgun in `docs/sdk/production-readiness.md`. Do not merge as-is, and do not close WALM-119 with this PR.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 1

docs/sdk/api-reference.md
```diff
@@ -83,6 +83,8 @@ Search for memories matching a natural language query, scoped to `owner + namesp
 
 - Preferred form: `recall({ query, limit?, topK?, namespace?, maxDistance? })`
 - `limit` defaults to `10`; `topK` is an alias and wins when both are set
+
```
ducnmm, line 86: **[suggestion]** The recommended detector (`results.length === limit`) is the same unreliable signal the PR body says `total` cannot be. After `maxDistance`, the SDK filters `results` and sets `total` to the filtered length, so the length check is a false-negative whenever weak hits were dropped. The relayer also clamps `limit` at 100, so retrying with 200 still returns at most 100 rows and looks complete. "Assume more matches exist" is also a false-positive when there are exactly `limit` hits.

**Suggestion:** After rebase, describe the length check as a heuristic. Mention the server cap of 100 and that `maxDistance` / `dropped_count` / `maxTokens` make `results.length === limit` miss truncation. That is why WALM-119 wants a server-side `truncated` instead of documenting a workaround.

packages/sdk/src/types.ts
```diff
@@ -66,7 +66,14 @@ export interface RecallResult {
 
 /** Options for recall(). */
 export interface RecallOptions {
-    /** Max number of nearest memories to request from the relayer (default: 10). */
+    /**
+     * Max number of nearest memories to request from the relayer (default: 10).
```
ducnmm, line 70: **[nit]** Follow-up is titled style-guide compliance (no em dashes, no quotation marks) but only rewrote the markdown callout. The JSDoc still uses an em dash and quotation marks around `"truncated"`.

**Suggestion:** If this JSDoc is kept, match the markdown pass.

PR #286, opened 2026-06-17
Feat: memory deletion — clearNamespace / forget / list (soft-delete) [WALM-115]

### LINKED ISSUE #258: No way to delete memories or clear a namespace — stale memories pollute recall after agent resets
## Summary

The SDK exposes `remember()`, `recall()`, `restore()` and `health()`, but no way to
delete a memory or clear a namespace. For agents that are developed iteratively
(i.e. every agent in this hackathon), this means any reset of the agent's local
state leaves orphaned memories behind that keep surfacing in `recall()` forever.

## Environment

- `@mysten-incubation/memwal` ^0.0.7, managed relayer (`https://relayer.memory.walrus.xyz`)
- Runtime: Cloudflare Workers
- Project: Kairos · World Cup (Walrus Session 4) — a self-critical football pundit that
  stores `[PREDICTION]` / `[OUTCOME]` / `[BIAS]` memories and recalls its own biases
  before each new prediction.

## What happened (real incident, twice in one day)

Before the tournament started we reset the agent's prediction log and re-issued its
predictions (we had just integrated betting-market priors). The old `[PREDICTION]`
memories were still recallable, so on "day one" the agent recalled **six stale
memories, including three near-duplicate predictions of the same match** from
discarded runs. Two concrete harms:

1. **Prompt pollution / self-anchoring**: the LLM anchored on its own stale outputs —
   it literally reasoned *"my past predictions on this exact match settled at 50-52%
   confidence, and I see no new reason to deviate"* — defeating the purpose of the reset.
2. **Demo integrity**: our "day-one, no memories yet" before/after moment (the core
   judging signal of this hackathon) was no longer reproducible.

## Workaround we used

Rotate the namespace string (`worldcup` → `worldcup-v2`) and abandon the old one.
It works, but: the orphaned memories remain attached to the MemWalAccount, there is
no way to enumerate existing namespaces or inspect what's in them, and repeated
iterations accumulate junk forever.

## Proposed API (any subset would help)

- `forget(memoryId)` — delete by the id returned from the remember job
- `clearNamespace(namespace)` — bulk reset for iteration/dev loops
- `list({ namespace, limit })` — enumerate memories so a client can at least audit
- If hard deletion is off the table because Walrus blobs persist until expiry, a
  **soft-delete tombstone filtered out at recall time** (relayer-side vector index)
  would solve 100% of our problem.

Happy to provide logs/repro. Overall MemWal was pleasant to integrate — recall
quality on bias-style queries is genuinely good; this is the one sharp edge we hit.

hungtranphamminh:
### Why
The SDK exposed `remember`/`recall`/`restore` but **no way to delete a memory or clear a namespace**. After an agent reset, stale memories keep surfacing in `recall()` — reported by two Walrus-Sessions-4 integrators (#258, +1 in-thread): a reset agent self-anchored on its own discarded outputs, and "day-one vs day-four" demo integrity broke on the same wallet. The only workaround (rotate the namespace string) orphans data forever.

### What
Adds owner-scoped memory deletion across the stack:

- **`clearNamespace(ns)`** — soft-delete every memory in a namespace (the reset primitive).
- **`forget(id)`** — soft-delete a single memory by its per-row id.
- **`list(ns, { limit, cursor })`** — enumerate a namespace (metadata only — id, blob_id, created_at, importance; no decrypted text), cursor-paginated via `has_more` / `next_cursor`.

Delivered in: the relayer (`/api/clear-namespace`, `/api/list`, `/api/memories/forget`), the **TS SDK** (`0.0.8`), and the **Python SDK** (`0.1.5`). MCP sidecar tools are included but **dormant** — see Related.

### Solution
- **Soft-delete via a `deleted_at` tombstone** (migration 010): deletion stops a memory surfacing in `recall()` (and the analyze pre-extraction context) while retaining the row. The underlying Walrus blob is user-owned and persists — "delete" means **un-recallable, not cryptographically erased** (carried honestly in all docstrings). Fits the privacy floor: no new server-readable plaintext.
- **The existing hard `/api/forget` is left untouched** for benchmark-harness cleanup; `clearNamespace` is a separate soft path.
- **`list()` keyset pagination** on `(created_at, id)` (stable tie-break, no skip/dup under concurrent deletes); metadata-only so it's decrypt-free.
- **Owner-scoped everywhere** — every query binds the auth-derived owner; `forget(id)` of an id you don't own is a clean no-op (no cross-owner existence oracle).

## Types of Changes

- [x] New feature (non-breaking)

## Benchmark impact

- [x] No benchmark-affecting change (write/admin path; recall read path only gains an additive `deleted_at IS NULL` filter, no scoring change)

## Testing

- [x] `cargo test` passes — **308 tests, 0 failures**
- [ ] e2e suite run against a live stack 

## Checklist

- [x] Docs updated if needed 

## Related

- Closes #258 (delete / clear namespace). Acknowledged in-thread by both reporting integrators.
- Linear: WALM-115.
- **MCP is deferred to a separate go-live PR.** The sidecar tools (`memwal_clear_namespace` / `memwal_list` / `memwal_forget`) are added but dormant: the sidecar's Docker build installs a pinned published SDK that predates these methods, so the tools resolve them only once the sidecar is built against the new SDK. No MCP version bump in this PR.
- Follow-ups (separate tickets): MCP go-live (sidecar build → pnpm/workspace); generation/epoch reset to make `clearNamespace` ordered against in-flight async writes; hard-purge/erasure job (must filter `deleted_at IS NULL` in restore's blob-discovery); public SDK docs.

### Reviewer note
`clearNamespace` clears what exists at call time; a `remember()` still in flight when it runs can land after the sweep and survive (standard "delete what exists now" semantics + async write path). Documented in the SDK; the durable fix (generation/epoch) is a tracked follow-up, intentionally out of this PR's scope.


hungtranphamminh:
Closing — drafting the PR description locally for review first; will reopen/recreate when ready.

PR #328, opened 2026-06-24
docs: Walrus Memory troubleshooting and FAQ page (BEDU-612)

reemsabawi-mystenlabs:
## Description
Adds a Walrus Memory troubleshooting and FAQ page so readers can self-serve on common support issues instead of filing tickets. Seeded from a Discord support thread (BEDU-612).

**New page** (`docs/troubleshooting/overview.md`)
- Structured symptom, cause, fix format, reusable as new questions come in.
- 401 `AUTH_REJECTED` and the delegate-key registration flow: a likelihood-ordered cause/check/fix table.
- MCP connection issues: only `memwal_login` showing, and tools not appearing after a mid-session connect.
- `memwal_analyze` timeouts: explains the async background-job model (a client timeout is not a failure), with verify-with-recall guidance.
- Quick-reference table and a product-page FAQ section.
- Registered in `docs/docs.json` under a new Troubleshooting tab.

**Sourcing**
- Grounded in existing docs: contract/delegate-key-management, relayer/api-reference, sdk/api-reference, mcp/reference.
- Written to the Sui documentation style guide.

> Note: the original 401 reporter self-resolved without recording the exact fix, so the AUTH_REJECTED entry documents all causes from the auth model rather than a single confirmed one.

## Test plan
- Mintlify build renders the page; admonitions and tables display correctly.
- The page appears in the Troubleshooting tab and internal cross-links resolve.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -206,6 +206,17 @@
         ]
       },
       {
+        "tab": "Troubleshooting",
+        "groups": [
+          {
+            "group": "Troubleshooting",
+            "pages": [
+              "troubleshooting/overview"
+            ]
+          }
+        ]
+      },
+      {
         "tab": "Contributing",
         "groups": [
           {
--- a/docs/troubleshooting/overview.md
+++ b/docs/troubleshooting/overview.md
@@ -64,7 +64,7 @@ This section covers problems that appear before the memory tools work.
 
 **Symptom:** After you connect the MCP server, the agent sees only `memwal_login`, or the memory tools return an authentication-required error.
 
-**Cause:** No credentials file exists at `~/.memwal/credentials.json`. When an MCP host launches the package without credentials, it starts in an authentication-required mode. The tool list still advertises all the `memwal_*` tools, but only `memwal_login` runs until you authenticate; the other tools return an authentication-required error when called, which is why it can look like `memwal_login` is the only tool available.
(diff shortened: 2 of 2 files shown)


jessiemongeon1:
## Style Guide Audit

ducnmm:
Thanks for this @reemsabawi-mystenlabs — really clear page, and most of the technical claims (auth model, URLs, `AUTH_REJECTED`/401, 5-min window + nonce, max-20 delegate keys, SDK method names, cross-links) check out against the code. Two small accuracy fixes before merge:

1. **`debug` is not a core SDK option.** The intro ("enable `debug` in the SDK configuration") and fix step 5 ("`debug: true` in the SDK") don't apply to the core SDK — `MemWalConfig` (`packages/sdk/src/types.ts`) has no `debug` field. It only exists on `WithMemWalOptions` for the `withMemWal()` AI middleware (`packages/sdk/src/ai/middleware.ts`). Passing `debug: true` to `MemWal.create()` has no effect. Suggest scoping that guidance to the `withMemWal()` wrapper, or pointing direct-SDK users at the right logging path.

2. **"Only `memwal_login` appears" — cause is slightly off.** The auth-required server actually lists *all* the tools (`packages/mcp/src/auth-required.ts`); only execution of the non-login tools is gated (they return an auth-required error on call). So "exposes `memwal_login` first" isn't quite what happens. The second clause and the fix are correct — suggest rewording the cause to "all tools appear, but only `memwal_login` runs until you're authenticated."

jessiemongeon1 (approved):

ducnmm (approved):
Fact-checked every technical claim against the codebase (URLs, paths, env vars, the 20-key cap, 5-min window, SDK versions, tool/method names, cross-links, docs.json schema) — all accurate. Pushed two small wording corrections (auth-required tool list scope; analyze sync vs background steps). LGTM.

PR #352, opened 2026-07-02
fix(relayer): recover Walrus register destroy_zero from stale WAL price (#351)

### LINKED ISSUE #351: Mainnet writes fail on both prod relayers: Enoki dry_run_failed (MoveAbort balance::destroy_zero in sponsored upload)
## Summary

All mainnet memory writes through the MCP server currently fail at the Walrus upload step with an Enoki sponsorship error. Reads (`memwal_recall`) and auth work fine. Reproduced consistently on **both** production relayers.

## Error (verbatim)

```
Tool error: remember job failed: walrus upload failed: Internal Error: walrus upload failed:
Enoki API error (400): {"errors":[{"code":"dry_run_failed","message":"Dry run failed, could
not automatically determine a budget: MoveAbort(MoveLocation { module: ModuleId { address:
0000000000000000000000000000000000000000000000000000000000000002, name: Identifier(\"balance\") },
function: 9, instruction: 8, function_name: Some(\"destroy_zero\") }, 0) in command 9"}]}
```

## Repro

1. `@mysten-incubation/memwal-mcp` **v0.0.5**, fresh login via `memwal_login` (delegate key registered on-chain, credentials saved).
2. Call `memwal_remember` with any short text and namespace, e.g. `{"text": "[FACT] smoketest :: ...", "namespace": "gardener-smoketest"}`.
3. Fails identically every time. Tried 2026-07-02 between ~11:25 and ~11:40 UTC:
   - `https://relayer.memory.walrus.xyz`: fails
   - `https://relayer.memwal.ai`: fails with the same error
4. `memwal_recall` on the same account/namespace works (returns "No matching memories found." cleanly), so relayer auth and the read path are healthy.

Account (mainnet): `0x057359bbb9bc3b6eed2b1c04d51bcbee421b9c802e90543b0ca65844c006d7cb`. The object is active with 3 registered delegate keys, verified via `sui_getObject`.

## Analysis

The abort is in the Sui framework's `balance::destroy_zero` (abort code 0 = `ENonZero`) at **command 9** of the sponsored PTB that Enoki dry-runs. That means the transaction being constructed destroys a coin/balance that still has a non-zero value, i.e. the WAL payment split inside the sponsored upload flow leaves a non-zero remainder where the code expects exact change. Nothing client-side affects this (payload size, namespace, relayer choice, fresh login), which points at the sponsor-side coin selection / storage-cost calculation, possibly triggered by a recent storage price or epoch change.

## Impact

Every MCP user is currently unable to write memories to mainnet (Session 5 proof runs included). Reads unaffected.

## Environment

- `@mysten-incubation/memwal-mcp` 0.0.5 (stdio), `--prod`
- Node.js 24.14.0, Windows 11

ducnmm:
## Summary

Fixes #351 — mainnet writes were failing 100% at the `register_sponsor` phase with an Enoki `dry_run_failed` → `MoveAbort` in `0x2::coin::destroy_zero` (ENonZero).

**Not an Enoki fault.** `@mysten/walrus`'s `#withWal` helper pre-funds an *exact* WAL payment (`storageUnits × price × epochs`) from the client's **cached** `systemState` price, then asserts the coin is empty via `coin::destroy_zero`. When mainnet storage/write price drifts **down** between the cached read and execution, the contract deducts less WAL than we split off and the leftover trips `destroy_zero`. Enoki only surfaces it during its budget dry-run.

## Evidence (prod, 2026-07-02)

- On-chain `storage_price_per_unit_size` moved `71464 → 70922` within ~15 min.
- After a relayer restart, **6 writes succeeded** then the **same** sidecar client flipped to **100% `destroy_zero`** at the next price move — a time/price transition, not per-user/blob/wallet.
- `classification=permanent retryable=false`, so every write was Dead-marked with no retry.

## Why it didn't self-heal (two gaps)

1. The sidecar auto-refresh only fired on `isMoveAbortBalanceSplit` (matches `balance`+`split`); this message says `destroy_zero`, so `refreshWalrusClient()` never ran.
2. The Rust worker swept it into the `MoveAbort → Permanent` catch → Apalis never retried.

A newer SDK does **not** fix it: `@mysten/walrus` 1.2.x has no cost/`#withWal`/`destroy_zero` change (installed 1.1.7, latest 1.2.3).

## Fix

- **`enoki.ts`** — add `isMoveAbortWalDestroyZero` detector.
- **`walrus-upload.ts`** — `refreshWalrusClient()` on the abort so the retry rebuilds against the live price.
- **`jobs.rs`** — classify the register `destroy_zero` abort as **Transient** (retryable), disjoint from the `balance::split` gas-budget path.
- **`config.ts`** — drop `WALRUS_CLIENT_MAX_AGE_MS` default `30m → 60s` so the cached price tracks a live-drifting mainnet price (still env-overridable).

## Testing

- New TS detector suite (7 cases) + Rust classification test using the verbatim prod error.
- Full suites green: **85 TS**, **40 jobs** Rust tests.

## Deployment

Already hot-deployed to the **prod relayer** (Railway, deployment `26742fd0`). Verified live: clean sidecar boot, writes succeeding, and `refreshed reason=max_age` lines confirm the new 60s max-age is active. This PR lands the same change on `dev`.

> Follow-up worth filing upstream to Mysten: the exact-change `#withWal` + `coin::destroy_zero` pattern is inherently racy against a moving on-chain price — it should tolerate leftover (return change) or add a small buffer.

### DIFF
--- a/services/server/scripts/__tests__/walrus-wal-destroy-zero.test.ts
+++ b/services/server/scripts/__tests__/walrus-wal-destroy-zero.test.ts
@@ -1,0 +1,62 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { isMoveAbortBalanceSplit, isMoveAbortWalDestroyZero } from "../sidecar/enoki.js";
+
+// Production format reference (issue #351): the Walrus register PTB pre-funds an
+// exact WAL payment from the client's cached storage price, then asserts the
+// coin is empty via `0x2::coin::destroy_zero`. When the on-chain price drops
+// between the cached read and execution, the contract deducts less WAL and the
+// leftover trips `destroy_zero` with ENonZero (abort code 0). Enoki surfaces it
+// during its budget dry-run as a 400 dry_run_failed.
+
+const PROD_DESTROY_ZERO_ERROR =
+    'Enoki API error (400): {"errors":[{"code":"dry_run_failed","message":"Dry run failed, ' +
+    "could not automatically determine a budget: MoveAbort(MoveLocation { module: ModuleId { " +
+    "address: 0000000000000000000000000000000000000000000000000000000000000002, name: " +
+    'Identifier(\\"balance\\") }, function: 9, instruction: 8, function_name: ' +
+    'Some(\\"destroy_zero\\") }, 0) in command 2"}]}';
+
+test("matches the verbatim prod destroy_zero abort", () => {
+    assert.equal(isMoveAbortWalDestroyZero(PROD_DESTROY_ZERO_ERROR), true);
+});
+
+test("matches the compact MoveLocation shape", () => {
+    assert.equal(
+        isMoveAbortWalDestroyZero(
+            "MoveAbort(MoveLocation { module: coin, function_name: Some(\"destroy_zero\") }, 0) in command 9",
+        ),
+        true,
+    );
+});
+
+test("is case-insensitive on both anchors", () => {
+    assert.equal(isMoveAbortWalDestroyZero("moveabort ... destroy_zero"), true);
+});
+
+test("bare destroy_zero without MoveAbort context is rejected", () => {
+    // Guards against unrelated log lines that merely mention the function.
+    assert.equal(isMoveAbortWalDestroyZero("calling coin::destroy_zero"), false);
+});
+
+test("balance::split abort does not match this detector", () => {
+    // The stale-price destroy_zero path must stay disjoint from the gas-budget
+    // balance::split path so the two handlers never double-fire.
+    const balanceSplit =
+        "MoveAbort(MoveLocation { module: balance, function_name: Some(\"split\") }, 2) in command 1";
+    assert.equal(isMoveAbortWalDestroyZero(balanceSplit), false);
+    assert.equal(isMoveAbortBalanceSplit(balanceSplit), true);
+});
+
+test("the destroy_zero abort is NOT matched by the balance-split detector", () => {
+    // The prod message contains the `balance` module name but no `split`, so the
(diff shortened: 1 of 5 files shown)


DrVelvetFog:
Reviewed this against the SDK internals — diagnosis and fix both look right. The two-gaps writeup nails why it didn't self-heal; I'd independently landed on the same `destroy_zero` ≠ `balance::split` distinction, and keeping the two detectors disjoint is the load-bearing call — a `split` ENotEnough is genuinely ambiguous with gas-pool exhaustion, so you don't want a starved gas wallet reclassified as recoverable stale-price. Register `destroy_zero` → Transient plus the `30m → 60s` max-age both read as correct for tracking a live-drifting price.

On the upstream follow-up you flagged — it's already in flight, and it's exactly what you described:

- I filed the race as MystenLabs/ts-sdks#1127 (exact-change `#withWal` + `destroy_zero`, same pattern across `createStorage` / `registerBlob` / `extendBlob`).
- hayes-mysten shipped the fix in ts-sdks#1128: a payment cost buffer, default sized from real `PricesUpdated` history, that over-funds slightly and returns the remainder as an address balance instead of asserting zero — the "tolerate leftover / add a small buffer" you called for, both at once. In review now.

So once that lands and you bump the SDK, the recovery here becomes the defense-in-depth layer: the buffer means the down-drift `destroy_zero` mostly stops firing in the first place, and your refresh-and-retry still catches anything that drifts past it. Good to have both sides covered.

LGTM.

harrymove-ctrl (approved):

ducnmm:
Superseded by #389 — both commits of this branch are included there (the deletion feature was stacked on this fix).

PR #361, opened 2026-07-06
fix(relayer): resolve client IP via trusted proxy hops to stop X-Forwarded-For spoofing (#360)

### LINKED ISSUE #360: bug(relayer): X-Forwarded-For spoofing bypasses per-IP rate limiting on the unauthenticated MCP proxy routes
The MCP SSE/messages proxy derives the client IP for rate limiting by taking the **first** value from an inbound `X-Forwarded-For` header, without validating how many trusted proxy hops sit in front of the relayer. These routes intentionally skip signed-request auth **and** per-key rate limiting (documented in `mcp_proxy.rs`), so the per-IP session cap is the *only* per-source defense against an unauthenticated session flood — and it is trivially spoofable.

The Rust relayer appends its own observed peer IP to any client-supplied `X-Forwarded-For`, leaving the attacker-controlled value at the front — `services/server/src/mcp_proxy.rs` (`inject_forwarded_for`):

```rust
// appends the real peer IP but never strips/validates a client-supplied
// value already present in the header:
let value = match inbound.get("x-forwarded-for")... {
    Some(existing) => format!("{}, {}", existing, peer_ip),
    None => peer_ip,
};
```

The Node sidecar then trusts the first (leftmost / attacker-controlled) hop — `services/server/scripts/mcp/rateLimit.ts` (`clientIpFromRequest`):

```ts
const first = raw.split(",")[0]?.trim();
return first;
```

The same first-hop pattern also exists in `services/server/src/rate_limit.rs` (`sponsor_rate_limit_middleware`):

```rust
.and_then(|s| s.split(',').next())   // first hop = attacker-controlled
```

No `trust proxy` / hop-count configuration exists anywhere in `services/server` to bound how many hops are trusted.

## Steps to reproduce

```bash
# Open many MCP SSE sessions from one host; a fresh fake header each time
# lands in a new per-IP bucket, so maxSessionsPerIp / maxNewSessionsPerIpPerMin
# never trip:
for i in $(seq 1 100); do
  curl -N https://relayer.memory.walrus.xyz/api/mcp/sse \
    -H "X-Forwarded-For: 203.0.113.$((RANDOM % 255))" \
    -H "Authorization: Bearer <any-shape-valid-value>" &
done
```

Each connection is bucketed under a different attacker-chosen IP, so the per-IP caps in `McpRateLimiter.acquire(ip)` are never reached.

## Impact

- **MCP `/api/mcp/sse` + `/api/mcp/messages`** (primary): per `mcp_proxy.rs`, these routes deliberately bypass the relayer's Ed25519 signed-request auth **and** its per-key rate limiting; the sidecar's per-IP session caps (`maxSessionsPerIp`, `maxNewSessionsPerIpPerMin`) are the sole per-source defense. `rateLimit.ts` is keyed purely on IP (`acquire(ip)`), with no per-key/per-sender axis. Spoofing the header therefore fully bypasses per-source limiting and lets a single host open long-lived SSE/streamable sessions well past the intended cap — the exact unauthenticated session-flood DoS the limiter's own comments say it exists to prevent. The global `maxTotalSessions` cap still bounds absolute worst case, but per-IP fairness is defeated.

- **`/sponsor` + `/sponsor/execute`** (secondary, weaker): the same first-hop pattern in `sponsor_rate_limit_middleware` is spoofable, but here it is **not** the only gate. `routes/sponsor.rs` adds a per-sender rate limit (`check_sender_rate_limit`, keyed on the request's Sui `sender`) explicitly described in-code as "a second axis that a distributed IP attack cannot bypass." So spoofing the IP alone does **not** grant extra Enoki gas-sponsorship throughput — an attacker capped at `per_minute`/`per_hour` per sender would additionally have to rotate `sender` addresses to abuse it. Worth fixing for consistency and defense-in-depth, but the per-sender backstop materially limits real-world impact on this route.

## Expected behavior

1. Configure an explicit trusted-hop count (e.g. `axum::extract::ConnectInfo` plus a known number of proxies in front of the relayer — 1 for Railway's edge) and take the IP that many hops from the **right** end of the `X-Forwarded-For` chain, never the left/first client-supplied value.
2. Alternatively, if Railway's edge already guarantees a sanitized header (e.g. `X-Real-IP` or a platform-specific trusted header), prefer that over raw `X-Forwarded-For`.
3. Apply the fix consistently to both the MCP proxy path (Rust → Node sidecar) and the `/sponsor` limiter (Rust).

**Environment:** Deployed on Railway (`railway.json`, axum server + Node sidecar per `Dockerfile`). No `trust proxy` / hop-count config found in `services/server`. Verified against `dev` branch; MCP rate limiter in `services/server/scripts/mcp/rateLimit.ts` is IP-keyed only.

ducnmm:
## Summary

Fixes #360.

The MCP proxy routes (`/api/mcp/sse`, `/api/mcp/messages`, `/api/mcp`) intentionally skip signed-request auth **and** per-key rate limiting, so the per-IP session cap in the Node sidecar is the **only** per-source defense against an unauthenticated session flood. Both that limiter and the `/sponsor` IP limiter derived the client IP from the **first (leftmost)** value of `X-Forwarded-For` — which is fully client-controlled. A single host could send a fresh fake `X-Forwarded-For` on each request, land in a brand-new per-IP bucket every time, and sail past `maxSessionsPerIp` / `maxNewSessionsPerIpPerMin`.

This PR resolves the real client IP by counting a configured number of trusted reverse proxies **in from the socket end** (express-style `trust proxy` semantics) and never trusting the leftmost value.

## Changes

- **`client_ip::resolve_client_ip(xff, peer, trusted_hops)`** — new shared helper. Builds the address chain `[socket_peer, XFF reversed…]`, skips `trusted_hops` trusted proxies, and returns the first address the client cannot forge past. Clamps a too-large hop count instead of underflowing. Unit-tested (spoof-defeat, 0/1/2 hops, IPv6, misconfig).
- **`TRUSTED_PROXY_HOPS`** — new config (default `1` = Railway's single edge proxy; `0` = relayer directly exposed → ignore `X-Forwarded-For`). Documented in `.env.example`.
- **`mcp_proxy::inject_forwarded_for`** — now forwards **one** sanitized client IP to the sidecar instead of appending the peer to a client-controlled chain. The sidecar is loopback-only from the relayer, so that single value is authoritative.
- **`sponsor_rate_limit_middleware`** — uses the same resolver (secondary route; per-sender limit already backstops it, but fixed for consistency / defense-in-depth).
- **Sidecar `clientIpFromRequest`** — takes the rightmost (nearest-relayer) hop as defense-in-depth and documents the trust boundary.

⚠️ **Deploy note:** default is `TRUSTED_PROXY_HOPS=1`, correct for the current Railway topology (one edge proxy). Any environment that exposes the relayer without a proxy in front must set `TRUSTED_PROXY_HOPS=0`; a value larger than the real hop count re-opens the spoofing hole.

## Testing

- `cargo test` — `client_ip` 10/10, `mcp_proxy` inject 5/5, `rate_limit` 26/26 pass.
- Node sidecar — 17/17 rate-limit tests pass.
- `cargo clippy` — no new warnings.

Scope is entirely relayer + sidecar; the Move contract is untouched.

### DIFF
--- a/services/server/.env.example
+++ b/services/server/.env.example
@@ -111,6 +111,14 @@ ENOKI_INVALIDATED_MAX_ATTEMPTS=4
 ENOKI_INVALIDATED_BASE_DELAY_MS=1000
 ENOKI_INVALIDATED_MAX_DELAY_MS=8000
 
+# Number of trusted reverse proxies in front of the relayer. Controls how the
+# real client IP is resolved from X-Forwarded-For for per-IP rate limiting
+# (MCP proxy + /sponsor). MUST equal the actual hop count — a value larger
+# than reality lets clients spoof their rate-limit bucket (issue #360).
+#   1 = one edge proxy (Railway) — correct in production (default if unset)
+#   0 = relayer directly exposed; ignore X-Forwarded-For, use the socket peer
+TRUSTED_PROXY_HOPS=1
+
 # Sponsor endpoint rate limiting (per IP + per sender, sliding window)
 SPONSOR_RATE_LIMIT_PER_MINUTE=10
 SPONSOR_RATE_LIMIT_PER_HOUR=30
--- a/services/server/scripts/mcp/__tests__/rateLimit.test.ts
+++ b/services/server/scripts/mcp/__tests__/rateLimit.test.ts
@@ -141,16 +141,27 @@ test("stats snapshot reflects acquired/released slots", () => {
     // current behavior produces (ip 1.1.1.1 still has an `opens` entry).
 });
 
-test("clientIpFromRequest prefers first hop in x-forwarded-for", () => {
+test("clientIpFromRequest uses the relayer's single sanitized value", () => {
+    // Normal case: the relayer forwards exactly one resolved client IP.
     assert.equal(
-        clientIpFromRequest(fakeReq({ xff: "203.0.113.5, 10.0.0.1, 127.0.0.1" })),
+        clientIpFromRequest(fakeReq({ xff: "203.0.113.5" })),
         "203.0.113.5"
     );
 });
 
-test("clientIpFromRequest trims whitespace around the first hop", () => {
+test("clientIpFromRequest takes the rightmost hop, never the client-forgeable leftmost", () => {
+    // Defense-in-depth: if a multi-value chain ever reaches the sidecar, the
+    // leftmost value is attacker-controlled — we must take the rightmost
+    // (nearest-to-relayer) entry instead. See issue #360.
     assert.equal(
-        clientIpFromRequest(fakeReq({ xff: "  203.0.113.5  , 10.0.0.1" })),
+        clientIpFromRequest(fakeReq({ xff: "1.2.3.4, 10.0.0.1, 203.0.113.5" })),
+        "203.0.113.5"
+    );
+});
+
+test("clientIpFromRequest trims whitespace around the value", () => {
+    assert.equal(
+        clientIpFromRequest(fakeReq({ xff: "  203.0.113.5  " })),
         "203.0.113.5"
     );
 });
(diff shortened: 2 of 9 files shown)


harrymove-ctrl (approved):

ducnmm:
Closing without rebasing. This fix already landed through another path.

#360 was closed on 2026-08-01 as fixed in #513 (`main <- staging`), and that merge is on both `dev` and `main`. Current `dev` resolves the client IP from the trusted side of `X-Forwarded-For` via `canonical_client_ip` (`services/server/src/client_ip.rs`), `TRUSTED_PROXY_HOPS` (default 0), and the sidecar only accepts the single IP the relayer already sanitized (`clientIpFromRequest`).

This branch is the July version of the same fix (`resolve_client_ip`) and now conflicts with that code. Rebasing it would re-implement a change that is already shipped.

ducnmm:
Reopening. Closing this was wrong — the conflict should be rebased, not discarded.

PR #370, opened 2026-07-08
fix(mcp): bind /connect/mcp consent to the local bridge (#368)

### LINKED ISSUE #368: [Security / Medium] /connect/mcp consent screen enables delegate-access phishing (attacker-controlled label + delegate address, no source binding)
**Component:** `apps/app/src/pages/ConnectMcp.tsx` (browser consent flow for the `@mysten-incubation/memwal-mcp` bridge)
**Type:** Security — consent/authorization phishing (UX + missing request binding)
**Severity:** Medium (requires social engineering + victim wallet approval)

## Summary

The `/connect/mcp` page registers a delegate key on the connected wallet's Walrus Memory account based **entirely on query-string parameters**, and its consent screen presents an **attacker-controlled application name** (`label`) plus a delegate address the victim has no way to validate. Because the page can be opened by anyone with any parameters (there is no binding to the local MCP bridge that is supposed to originate the request), an attacker can send a victim a link on the trusted `memory.walrus.xyz` domain that reads *"Cursor wants access to your Walrus Memory"* and, on approval, permanently grants the **attacker's own delegate key** read/decrypt/write access to all of the victim's memories until manually revoked.

This is an app-level consent-phishing surface, not a wallet-level issue: the branded, first-party consent screen actively lends credibility to a request that is fully attacker-defined.

## Verified facts (code)

`apps/app/src/pages/ConnectMcp.tsx`:

- `publicKey`, `delegateAddress`, `label`, `relayer`, `port`, `connectState` are all read directly from `useSearchParams()`.
- `paramsValid` validates **format only** (hex length / port range); it does **not** verify any relationship between the parameters, nor the origin of the request.
- On approve, it builds and signs `account::add_delegate_key(accountId, publicKey, delegateAddress, label, clock)` with the raw params, signed by the **connected wallet (the victim, who is the account owner)** — so the on-chain owner check passes.
- The consent screen (`ConsentCard`) shows the attacker-controlled `label` as the requesting app, a **truncated** `delegateAddress` (`slice(0,16)…slice(-6)`), and the `relayer`. The `publicKey` is not shown, and there is no warning that the link source must be trusted.
- The only cross-origin protection (`connectState`) guards the **localhost callback**, not the on-chain `add_delegate_key`. There is no referrer/opener/bridge-origin check.

I confirmed `paramsValid` returns `true` for entirely attacker-chosen values (random `publicKey`, an attacker-owned `delegateAddress`, arbitrary `port`, a self-generated 64-hex `connectState`).

## Attack flow (no local bridge required on the victim side)

1. Attacker generates their own delegate keypair `(pk, addr)` (a normal, self-consistent keypair they control).
2. Attacker crafts:
   `https://memory.walrus.xyz/connect/mcp?port=20000&publicKey=<pk>&delegateAddress=<addr>&label=Cursor&relayer=<relayer>&connectState=<random 64-hex>`
3. Victim opens the link → sees *"Cursor wants access to your Walrus Memory"* on the real domain → connects wallet → approves.
4. `add_delegate_key` succeeds (victim is the owner). The attacker's key is now a registered delegate on the victim's account.
5. The localhost callback fails (no bridge), showing only a benign "couldn't reach local listener" message — but the **on-chain delegate registration already succeeded**.
6. The attacker now uses `(pk, addr)` to: decrypt the victim's memories via SEAL (`seal_approve` delegate branch matches `addr`) and read/write via the relayer (`auth.rs` matches `pk` in `delegate_keys`). Access persists until the victim manually revokes it.

## Important: this is NOT the `public_key` / `sui_address` desync

A related report framed the root cause as the contract not verifying that `sui_address == blake2b256(0x00 ‖ public_key)` in `add_delegate_key`, and proposed enforcing that derivation. **That fix would not prevent this attack**, and the desync is not the enabling factor:

- Relayer API access checks the **public key** (`verify_delegate_key_onchain`, needs the matching private key); SEAL decryption checks the **address** (`is_delegate_address`).
- An attacker's optimal choice is therefore a **fully consistent keypair they own** (`addr = derive(pk)`), which grants **both** relayer and SEAL access and passes any derivation check.
- A mismatched pair grants strictly less (SEAL only), so an attacker gains nothing from the desync.

The vulnerability is the consent/phishing surface — who can trigger the request and how it is presented — not key derivation.

## Mitigating factors (honest severity)

- Requires social engineering: the victim must open the link, connect their wallet, and approve the transaction.
- The wallet popup does display an `add_delegate_key` move call, so a vigilant user could refuse. In practice the branded first-party consent screen suppresses suspicion, which is exactly what makes consent phishing effective.

## Recommended improvements

1. **Bind the request to the local bridge.** Require proof the page was opened by a locally-running bridge for `connectState` (e.g., only proceed if the localhost `/callback` preflight/handshake on `port` succeeds before signing), so a bare emailed link cannot drive an on-chain add.
2. **Make consent honest and scary enough.** Explicitly state that approving grants a third party persistent read + decrypt + write access to *all* memories across *all* namespaces until revoked; show the full `delegateAddress` and `publicKey`; and warn the user to only continue if they personally started this from their own machine.
3. **Do not present `label` as a verified identity.** Render it as untrusted free text (e.g., "This app calls itself: …") rather than an authoritative "<label> wants access".
4. **Optional relayer allowlist** for the `relayer` param, or a prominent warning when it is non-default.

## Separate low / defense-in-depth note

Independently of the above, `add_delegate_key` never verifies `sui_address` is derived from `public_key`, and dedups only on `public_key` (not `sui_address`). Worth adding as defense-in-depth (and to keep the on-chain `public_key` field meaningful), but — as shown above — it is not what enables the phishing attack and should not be presented as the fix for it.

ducnmm:
## Summary

Fixes the consent-phishing surface reported in #368. The `/connect/mcp` page registered a delegate key on the connected wallet's Walrus Memory account based **entirely on query-string parameters**, with no proof the request came from the local MCP bridge it claims to serve. An attacker could send a victim a link on the trusted `memory.walrus.xyz` domain — *"Cursor wants access to your Walrus Memory"* — and, on approval, permanently grant the **attacker's own delegate key** read / decrypt / write access to all the victim's memories until manually revoked.

This is an app-level consent-phishing issue, not a wallet- or contract-level one: the branded first-party consent screen lent credibility to a fully attacker-defined request. The Move contract is **unchanged** — the owner check already passes for a victim-signed tx, and (per the report) the `public_key` / `sui_address` desync is not what enables the attack.

## What changed

**1. Bind the on-chain write to the local bridge (core fix).** Reuses the existing single-use `connectState` token, lifting it from guarding only the callback to also gating the signature:

- `packages/mcp/src/login.ts`: add `GET /handshake?state=` that returns `{ ok: true }` only when the state matches the token the bridge minted (constant-time compare, behind the same Host / Origin / PNA guards as the callback). Allow `GET` in CORS.
- `apps/app/src/pages/ConnectMcp.tsx`: probe the bridge on load and re-verify **immediately before signing**; refuse to sign `add_delegate_key` unless it confirms. A phishing link reaches no local bridge — or one whose token differs — and fails the handshake, so the page never signs.

**2. Make the consent screen honest (report recommendations 2–4).**

- Show the **full** delegate address and public key (were truncated / hidden).
- Render the client `label` as **untrusted free text**, not a verified "*&lt;label&gt; wants access*" identity.
- State plainly that approving grants persistent **read + decrypt + write** to all memories across all namespaces until revoked; warn on a **non-default relayer**.
- Don't auto-skip the consent screen when a wallet is already connected.

## Testing

- `packages/mcp` — new `test/login-handshake.test.mjs` drives the real localhost listener: correct token → `200 {ok:true}`, wrong / missing token → `403 {ok:false}`, and a valid callback still resolves credentials. Full suite: **9/9 pass**.
- `apps/app` — `tsc -b` clean, `eslint` clean.

## Deploy ordering (fail-closed — please read)

The app now rejects an older bridge that lacks `/handshake`. This is intentional and matches the existing "force a bridge upgrade" behaviour of the `connectState` param. **Publish the updated `@mysten-incubation/memwal-mcp` bridge before deploying the app** so legitimate users aren't blocked. The error screen guides users to update the client (`npx @mysten-incubation/memwal-mcp@latest login`).

Fixes #368.

### DIFF
--- a/apps/app/src/pages/ConnectMcp.tsx
+++ b/apps/app/src/pages/ConnectMcp.tsx
@@ -60,6 +60,41 @@ function hexToBytes(hex: string): number[] {
     return out
 }
 
+/**
+ * Prove the request came from a Walrus Memory MCP bridge running on THIS
+ * machine before we register a delegate key on-chain (issue #368).
+ *
+ * The bridge mints a single-use `connectState` token, keeps it in memory, and
+ * only puts it in the `/connect/mcp` URL it opens itself. We hand that token
+ * back to the bridge's localhost `/handshake`; it answers `{ ok: true }` only
+ * when the token matches the one it minted. A phishing link opened from an
+ * email carries an attacker-chosen token and reaches no local bridge (or a
+ * bridge whose token differs), so this returns false and we refuse to sign.
+ *
+ * Any failure — no listener, wrong token, CORS/PNA block, non-bridge process
+ * on the port — is treated as "not verified". False negatives are safe: they
+ * only ask a legitimate user to restart the MCP login command.
+ */
+/** Compare two relayer URLs ignoring trailing slashes and case. */
+function sameRelayer(a: string, b: string): boolean {
+    const norm = (s: string) => s.replace(/\/+$/, '').toLowerCase()
+    return norm(a) === norm(b)
+}
+
+async function verifyLocalBridge(port: string, state: string): Promise<boolean> {
+    try {
+        const res = await fetch(
+            `http://127.0.0.1:${port}/handshake?state=${encodeURIComponent(state)}`,
+            { method: 'GET' },
+        )
+        if (!res.ok) return false
+        const data = (await res.json().catch(() => null)) as { ok?: unknown } | null
+        return data?.ok === true
+    } catch {
+        return false
+    }
+}
+
 async function resolveAccountId(
     suiClient: ReturnType<typeof useSuiClient>,
     ownerAddress: string,
@@ -129,7 +164,17 @@ export default function ConnectMcp() {
     const [walletPickerOpen, setWalletPickerOpen] = useState(false)
     const [callbackPayload, setCallbackPayload] = useState<McpCallbackPayload | null>(null)
     const [callbackDelivered, setCallbackDelivered] = useState<boolean | null>(null)
+    // Bridge binding (issue #368): null = still checking, true = a local bridge
+    // confirmed it issued this request, false = could not confirm (a phishing
+    // link opened from elsewhere, or the local bridge has exited). We refuse to
+    // sign `add_delegate_key` unless this is true.
+    const [bridgeVerified, setBridgeVerified] = useState<boolean | null>(null)
     const invalidRequestTrackedRef = useRef(false)
+    // True once the user actively started the flow (clicked the consent button).
+    // Gates the wallet-picker auto-proceed effect so a pre-connected wallet does
+    // NOT skip straight past the consent screen on mount — the user must read
+    // and click through the disclosure first (issue #368).
+    const initiatedRef = useRef(false)
 
     // Validate query string up-front.
     const paramsValid = useMemo(() => {
@@ -166,6 +211,9 @@ export default function ConnectMcp() {
     )
 
     const handleConnect = useCallback(async () => {
+        // Mark the flow as user-initiated so the auto-proceed effect below is
+        // allowed to run after the wallet picker closes (issue #368).
+        initiatedRef.current = true
         if (!paramsValid) {
             trackEvent('mcp_connect_failed', { error_type: 'invalid_request' })
             setErrorMsg('Invalid query parameters from MCP client.')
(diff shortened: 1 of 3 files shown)


harrymove-ctrl (approved):

ducnmm:
Closing without rebasing. This fix already landed through another path.

#368 was closed on 2026-08-01 as fixed in #513 (`main <- staging`), and that merge is on both `dev` and `main`. Current `dev` binds `/connect/mcp` to the local bridge with `POST /preflight` (`packages/mcp/src/login.ts`, `apps/app/src/pages/ConnectMcp.tsx`): the page will not offer approval unless this listener confirms `state`, `publicKey`, and `relayer`.

This branch is the July version of the same fix (`GET /handshake`) and now conflicts with that code. Rebasing it would re-implement a change that is already shipped.

ducnmm:
Reopening. Closing this was wrong — the conflict should be rebased, not discarded.

PR #376, opened 2026-07-09
docs: remove Seal encryption claims from Walrus Memory overview

reemsabawi-mystenlabs:
Removes the end-to-end encryption card from the Walrus Memory overview, plus encryption mentions in the storage card, the relayer description, and the default SDK path.
Changed: docs/getting-started/what-is-memwal.md, docs/getting-started/choose-your-path.md. Two files.
Deliberately out of scope (no changes here, flagging so it isn't assumed complete):
Ten reference files still document Seal-related API surface (x-seal-session, x-delegate-key, SERVER_SUI_PRIVATE_KEY) across the relayer, contract, environment variables, and SDK API reference. Removing those would leave shipped auth headers undocumented, so they need engineering review first. sdk/changelog.mdx is a historical record and shouldn't be rewritten. getting-started/quick-start.md references @mysten/seal only as a peer-dependency install for the manual client flow.
fundamentals/architecture/data-flow-security-model.md and choose-your-path.md section 3 both state the relayer never sees plaintext in the manual flow. Unchanged pending confirmation of current behavior.

### DIFF
--- a/docs/getting-started/choose-your-path.md
+++ b/docs/getting-started/choose-your-path.md
@@ -12,7 +12,7 @@ These paths aren't mutually exclusive. You can combine them - for example, use t
 
 Use `@mysten-incubation/memwal` when you want the fastest working integration.
 
-- relayer handles embedding, encryption, retrieval, and restore
+- relayer handles embedding, retrieval, and restore
 - best starting point for most teams
 
 Go to: [SDK Overview](/sdk/overview)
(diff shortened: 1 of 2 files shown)


jessiemongeon1 (approved):
approved but PR description seems outdated/incorrect - mentions 10 files and other changes not in the PR

reemsabawi-mystenlabs:
Updated description.

PR #378, opened 2026-07-10
feat(WALM-264): permanent V1 memory deletion — flag-gated UI + backend

harrymove-ctrl:
# WALM-264 — let users permanently delete their old V1 memories

**Everything is OFF by default** behind two flags (`ENABLE_MEMORY_DELETION` on the relayer, `VITE_ENABLE_MEMORY_DELETION` in the app). With the flags off, users see nothing and the new endpoints answer 404. We flip them only when rollout is agreed.

## What users get

- A **"Delete old memories" card** on the Dashboard: it counts the old memory blobs the wallet owns and shows **one button — "Delete all N forever"**. No per-memory picking (per Henry's review: don't show users a 1000-row list; only the number).
- A **dismissible banner** ("You have N old memories — delete them") on every page, pointing to that card. Dismissed per wallet (localStorage).
- A **user guide**: `docs/fundamentals/deleting-old-memories.md` (Sui docs style-guide compliant).
- Very explicit copy everywhere: deletion is **permanent** — never recoverable, never migrated to V2.

## How it works (simple version)

1. The app scans the wallet's on-chain Walrus `Blob` objects (gRPC or JSON-RPC transport), skips **expired** blobs (their storage is already gone, and one expired blob would abort a whole delete transaction), and keeps **only the ones our relayer DB knows about** (new `POST /api/memory-blob-ids`). Wallets without a delegate-key session see a setup prompt instead of a count, so we never show an unscoped number. The scan is cached and shared between the banner and the card.
2. Click **Delete all** → confirm dialog → the app builds the `system::delete_blob` transactions itself, **40 deletes per batch** (the binding limit is the relayer's `/sponsor` request caps — 7 KB of transaction bytes — not Sui's 1024-command cap), returning reclaimed `Storage` objects to the wallet.
3. Enoki sponsors the gas (`POST /sponsor`) — **signing is the only thing the user does**.
4. The signature goes to the new `POST /api/delete-memories`. The backend submits the transaction, **verifies it actually succeeded on-chain (effects status, not just Enoki's 200)**, checks the **executed sender matches the authenticated owner**, and only then deletes the matching DB rows in one bulk query — one call owns both side effects, so chain and index can't drift.

Failure handling: finished batches stay deleted; on any mid-run failure the card re-scans the chain so a retry only sees blobs that still exist. A double-click can't start a second concurrent run. If the row-delete ever fails after a landed transaction, the existing `cleanup_expired_blob` path self-heals on the next recall.

## Verified

- Full flow tested E2E on **mainnet** with a real wallet: scan → sponsor → sign → backend submit → DB rows deleted.
- Adversarial code review (correctness/security pass) driven into the design: batch size vs sponsor byte caps, expired-blob batch poisoning, effects-status verification, sender↔owner binding, double-click guard, post-success resume, scoped banner counts.
- App: `tsc`, `vite build`, asset verification pass. Server: `cargo test` — 298 pass, 5 pre-existing failures (`jobs::tests`, unrelated, also fail on clean dev). Sidecar module imports verified.

## Known limitations (documented, non-blocking)

- `/sponsor` is rate-limited per sender (10/min, 30/hr), so wallets with thousands of memories need multiple sessions — the guide says so. Raising the limit (or a bigger per-batch byte allowance) is a possible follow-up.
- The blob-id list sent to `/api/delete-memories` is client-supplied and not cross-checked against the transaction's deleted objects; a dishonest caller can only desync **their own** index rows (same blast radius as `/api/forget`), and recall self-heals. Documented in the route docs.

## Ops notes (before enabling in prod)

- **Enoki allowlist**: sponsoring `delete_blob` requires the move-call target `<current walrus package>::system::delete_blob` on the Enoki app allowlist (already added for mainnet: `0x98da433a…::system::delete_blob`). ⚠️ This id is the *latest* Walrus package version — it must be re-added whenever Walrus upgrades their package.
- Enable = set both flags above; nothing else changes.

## Also included (found while testing on mainnet)

- **fix: JSON-RPC nested UID shape in registry lookup** — `fetchAccountIdForOwner` read the Table id as a string; JSON-RPC returns `{ id: { id } }`, so every lookup on the JSON-RPC path failed and SetupWizard tried `create_account` for wallets that already have accounts. Regression from the #355 compat layer, hidden until now because testnet runs gRPC. Matters for any mainnet deploy.
- **fix: apiCall nonce-based auth** — `utils/api.ts` still signed the legacy 4-part message with no `x-nonce`, so auth middleware rejected it with 426; now signs the current 6-part message like SDK v0.4+.
- **fix(relayer): honor `SUI_RPC_URL`**, **recover Walrus register from stale WAL price** — small relayer fixes hit while testing.
- **fix: accept 42-char base58 digests** — ~0.1% of real 32-byte digests encode shorter than 43 chars and were rejected as invalid.


ducnmm (approved):

jessiemongeon1:
## Style Guide Audit

ducnmm:
Superseded by #389, which contains this feature plus the current dev merged in (conflicts with #353/#355 already resolved), the /sponsor-delete rate-limit fix, and the suiscan JSON-RPC defaults — matching what production runs.

PR #451, opened 2026-07-14
allow deletion

hungtranphamminh:
## Summary

Adds the **security-delete** subsystem: an authenticated, backend-orchestrated flow that lets a user permanently delete memory blobs written by the previous deployment.

The server resolves which blobs a wallet owns, builds and sponsors the delete transaction, and the user signs it. The blob object is destroyed on chain and its data is no longer retrievable from Walrus. **Users pay no gas.**

- [Public API contract](docs/api/security-delete.md)
- [Environment variables](docs/reference/environment-variables.md)

## How it works

```text
old-V1 vector insert ── trigger/backfill ──> legacy deletion tracking + claim lineage
                                                    │
wallet auth ──> list ──> prepare/claim ──> sign server-built PTB ──> submit to Sui
                                                    │                    │
                                     resolver + chain recheck     reconciler/replay
                                                    └── terminal progress state ──┘
```

1. **Tracking is isolated to the legacy database.** Legacy-only migrations add tracking, batches, append-preserving claim lineage, and job checkpoints beside `vector_entries`. A hook-first migration captures new writes; an idempotent seed backfills existing rows. Caller-supplied IDs are never enrolled, `vector_entries` is not mutated, and the current database is never touched.

2. **Two authentication steps.** A single-use wallet challenge produces a short-lived bearer token for listing and batch management — the address is derived from the signature, never taken from the request. Every destructive batch additionally requires the wallet to sign the exact `TransactionData`. **The token alone cannot delete anything.**

3. **Claim before building.** `prepare` atomically claims only currently-deletable rows, subject to a per-owner active-batch cap. It refreshes Sui object references, drops missing/transferred/expired objects, and builds a bounded Walrus `delete_blob` PTB. The backend is the only builder and sponsor, so the service cannot be used as an arbitrary gas sponsor.

4. **Chain truth and workflow truth stay separate.** Sui is authoritative for object existence and ownership; the legacy database is authoritative for progress. Every transition is compare-and-swap and scoped to the expected batch. The member ledger preserves each claim outcome after the live `batch_id` pointer is cleared, which prevents overlap/ABA races and keeps terminal state auditable.

5. **Recovery, not guessing.** Submit verifies the owner signature, durably records the execution envelope, adds the sponsor signature, executes, and transactionally finalizes. The reconciler replays exact signed transactions after crashes, resolves ambiguous digests, rolls back abandoned prepares, and detects out-of-band deletes/transfers. A separate resolver maps Walrus blob ids to Sui object ids through owner scans.

6. **Bounded resource pressure.** Routes and jobs have independent flags and fail-closed boot guards. The deletion-only Sui client has rolling-window admission, foreground/background priority, bounded retries, `Retry-After` cooldown handling, and an execution concurrency gate. Metrics expose RPC saturation, provider limits, cooldowns, claims, reconciliation, and terminal outcomes.

7. **Preview plaintext stays in the browser.** The dashboard lists a wallet's remaining old memories with delete-selected / delete-all flows and a Seal preview. Encrypted bytes are fetched and blob-id-bound before in-browser decryption. **Plaintext never traverses the server.**

## Two chain-facing constraints

Both are invisible on a freshly-published local network and destroy data on a live one. The types exist so that neither can be expressed.

**`end_epoch` is a Walrus epoch, not a Sui one.** `Blob.storage.end_epoch` must never be compared against the Sui ledger epoch — the two clocks are independent and diverge without limit. Comparing them marks live blobs `expired`, which is terminal: the blob is never re-offered, so a user's data would remain on Walrus with no way left to remove it.

**The Walrus package id must be read from the System object at runtime.** Walrus upgrades its package; a pinned id makes every `delete_blob` abort on chain. The configured value is a fallback and a drift signal only — and it cannot simply be repointed at the new package, because the same value also types the owned-object filter, and Move types keep naming the package they were published in.

`SuiEpoch`, `WalrusEpoch` and `WalrusCallPackage` are therefore distinct types rather than bare primitives.

## Implementation entry points

| Area | Delivered | Entry points |
|---|---|---|
| Legacy persistence | Dedicated legacy-DB runner; hook-first capture/backfill; tracking, batches, durable submission fields, append-preserving member lineage, job checkpoints | [`migrations_legacy/`](services/server/migrations_legacy/), [`security_delete_store.rs`](services/server/src/storage/security_delete_store.rs) |
| API / auth | Wallet challenge + bearer auth; list, prepare, submit, status, cancel; closed machine-readable error contract; backend-built gas-sponsored PTBs | [`routes/security_delete.rs`](services/server/src/routes/security_delete.rs), [`docs/api/security-delete.md`](docs/api/security-delete.md) |
| Sui execution / recovery | gRPC client, quota/retries/cooldowns, input-freshness checks, PTB construction, signature/effects verification, crash-safe replay, resolver and external-state sweeper | [`services/server/src/sui/`](services/server/src/sui/), [`jobs_security_delete.rs`](services/server/src/jobs_security_delete.rs) |
| Dashboard | Stored/progress tabs, keyset paging, cross-page selection, delete-all batching, recovery UX, client-side Seal preview | [`SecurityDeleteSection.tsx`](apps/app/src/components/SecurityDeleteSection.tsx), [`SecurityDeleteTable.tsx`](apps/app/src/components/SecurityDeleteTable.tsx), [`BlobPreviewModal.tsx`](apps/app/src/components/BlobPreviewModal.tsx) |
| Cutover support | Legacy package/registry/key-server overrides so old-V1 data can still be previewed while the app uses new V1 | [`apps/app/src/config.ts`](apps/app/src/config.ts) |
| Operational controls | Independent route/resolver/reconciler flags, fail-closed boot guards, non-secret runtime introspection, deletion-specific metrics | [`types.rs`](services/server/src/types.rs), [`observability.rs`](services/server/src/observability.rs) |

## Where to focus review

**Only 11 files touch pre-existing code.** Everything else is new files that land without interacting with anything. The 11 are where the risk is:

`types.rs` · `main.rs` · `auth.rs` · `storage/sui.rs` · `routes/{admin,remember,sponsor}.rs` · `Cargo.toml` · `App.tsx` · `Dashboard.tsx` · `config.ts`

## Behaviour changes on existing paths

The subsystem talks to Sui over **gRPC**, scoped to its own client. **The app-wide Sui client stays on JSON-RPC**, and on mainnet the existing server paths are unchanged. No new environment variable is required for an existing mainnet deployment.

Three changes ride along that are **not** part of the delete feature and do affect existing paths:

- **`jobs.rs`** — a Walrus `coin::destroy_zero` abort is reclassified from Permanent to **Transient**, so the job retries instead of being marked dead. Paired with a client refresh in the sidecar.
- **`sidecar/config.ts`** — `WALRUS_CLIENT_MAX_AGE_MS` default drops from **30 min to 60 s**, so the Walrus client (and its `systemState` read) is recreated more often on the write path.
- **`routes/sponsor.rs`** — `validate_digest` accepts **32–44** characters instead of exactly 43–44, to admit base58 digests with leading zeros.

## Verification

410 tests (381 unit, 29 Postgres-backed). Verified end to end against live Sui and Walrus testnet: prepare, sponsor, sign, execute — and the blob object absent from chain afterwards.


jessiemongeon1:
## Style Guide Audit

**8** violation(s) found. All must be fixed before merge.

### `docs/api/security-delete.md` (8 violation(s))

_8 violation(s) (8 regex, 0 claude)_

- **Line 1** — H1 only in frontmatter
  - Current: `# Security Delete API`
  - Fix: `Use ## or lower for section headings`
- **Line 37** — Use "might" not "may"
  - Current: `may`
  - Fix: `might`
- **Line 38** — Use "might" not "may"
  - Current: `may`
  - Fix: `might`
- **Line 91** — "on-chain" is one word → "onchain"
  - Current: `on-chain`
  - Fix: `onchain`
- **Line 186** — No em dashes in prose
  - Current: `—`
- **Line 244** — "on-chain" is one word → "onchain"
  - Current: `on-chain`
  - Fix: `onchain`
- **Line 246** — Use present tense, not future
  - Current: `will`
- **Line 340** — Use "might" not "may"
  - Current: `may`
  - Fix: `might`

---

daniellam258:
### [P1] The global gRPC client breaks existing JSON-RPC call sites

`App.tsx` now injects a `SuiGrpcClient` into `SuiClientProvider` for every real network, but existing consumers such as `SetupWizard.tsx` and `ConnectMcp.tsx` still use JSON-RPC request shapes:

- `getObject({ id, options: { showContent: true } })` instead of the gRPC `getObject({ objectId, include: { json: true } })`
- `getDynamicFieldObject(...)`, which the gRPC client does not provide

The cast to `SuiJsonRpcClient` in `App.tsx` hides the incompatibility from TypeScript. The browser E2E also uses the explicit local JSON-RPC escape hatch, so CI does not exercise the failing real-network path.

On testnet/mainnet, existing-account onboarding can incorrectly resolve no account and attempt account creation, while the MCP connection flow reports no account. The reference PR already converted `SetupWizard` to the `suiClientCompat` helpers, but that conversion was lost in this port; `ConnectMcp` needs the same treatment.

Please either convert every provider consumer to the compatibility helpers or keep the app-wide client on JSON-RPC and scope gRPC to the deletion subsystem.

ducnmm (approved):

PR #453, opened 2026-07-14
fix(app): bring prod app hotfixes (registry-table-id) onto dev

UyLeQuoc:
## What
Merge the production **app** hotfix branch `fix/registry-table-id-invalid-params` into `dev`.

Production's `app` service deploys from this branch (not `main`). It carries two fixes that `dev` is missing:
- fix(app,relayer): default mainnet to the dedicated suiscan JSON-RPC endpoint
- fix(app): accept JSON-RPC's nested UID shape in registry table lookup

## Why
The dev Railway environment currently runs `app` from `feat/security-delete`, which is broken. Production runs the branch merged here and works. This lands the working code on `dev`.

## Notes
- Merges cleanly into `dev` (no conflicts). Reconcile branch used so the prod-live fix branch is never touched by PR pushes.
- Deploy: the dev env `app` service is still pinned to `feat/security-delete`; it must be repointed to `dev` to pick this up (handled separately in Railway).


UyLeQuoc:
Superseded by the combined PR https://github.com/MystenLabs/MemWal/pull/455.

PR #454, opened 2026-07-14
fix(relayer): bring prod relayer hotfixes (walrus destroy_zero) onto dev

UyLeQuoc:
## What
Merge the production **relayer** hotfix branch `fix/walrus-register-destroy-zero-stale-price` into `dev`.

Production's `relayer` service deploys from this branch (not `main`). It carries two fixes that `dev` is missing:
- fix(relayer): honor SUI_RPC_URL env to move off degraded public fullnode
- fix(relayer): recover Walrus register destroy_zero from stale WAL price (#351)

## Why
The dev Railway environment currently runs `relayer` from `feat/security-delete`, which is broken. Production runs the branch merged here and works. This lands the working code on `dev`.

## Conflict resolution
`services/server/scripts/sidecar/config.ts` conflicted and was resolved as a **union**: kept dev's `SUI_GRPC_URL` (gRPC migration) and added the fix branch's `SUI_RPC_URL` env override. No behavior lost from either side.

## Notes
- Reconcile branch used so the prod-live fix branch is never touched by PR pushes.
- Deploy: the dev env `relayer` service is still pinned to `feat/security-delete`; it must be repointed to `dev` to pick this up (handled separately in Railway).

### DIFF
--- a/services/server/scripts/__tests__/walrus-wal-destroy-zero.test.ts
+++ b/services/server/scripts/__tests__/walrus-wal-destroy-zero.test.ts
@@ -1,0 +1,62 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { isMoveAbortBalanceSplit, isMoveAbortWalDestroyZero } from "../sidecar/enoki.js";
+
+// Production format reference (issue #351): the Walrus register PTB pre-funds an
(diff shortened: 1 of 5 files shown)


UyLeQuoc:
Superseded by the combined PR https://github.com/MystenLabs/MemWal/pull/455.

PR #481, opened 2026-07-20
docs(relayer): document SUI_GRPC_URL gRPC write-path option

reemsabawi-mystenlabs:
## Description

Documents the optional `SUI_GRPC_URL` environment variable added in #355, which opts the relayer write path (Walrus register and certify, SEAL, and Enoki build) into gRPC instead of JSON-RPC ahead of the Sui JSON-RPC sunset on 2026-07-31. The option is off by default, so leaving it empty keeps the existing JSON-RPC behavior. The blob query and restore path continues to use JSON-RPC index methods and is not affected.

Covers the four operator-facing pages so the behavior is consistent across the docs:

- **Environment Variables reference** (`reference/environment-variables.md`) — new `SUI_GRPC_URL` row in the self-hosted relayer table plus a note (BEDU-816).
- **Configuration reference** (`reference/configuration.md`) — a "Relayer configuration" pointer to the write-path transport option, since that page otherwise covers SDK config shapes (BEDU-813).
- **Relayer Overview** (`relayer/overview.md`) — a "Sui RPC transport" subsection (BEDU-814).
- **Self-Hosting** (`relayer/self-hosting.md`) — a "Sui RPC Transport (gRPC write path)" subsection with an example and the JSON-RPC caveat (BEDU-815).

## Test plan

- Sourced every claim from #355 and the merged code (`services/server/scripts/sidecar/config.ts`, `clients.ts`, `walrus-query.ts`, and `services/server/src/main.rs`): the opt-in default, which write-path operations move to gRPC, the startup validation of the URL, and the fact that the blob query and restore path stays on JSON-RPC index methods.
- Documentation-only change; no runtime code affected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017oS4PdhkXAzbfhJx2ZhRvh

---
_Generated by [Claude Code](https://claude.ai/code/session_017oS4PdhkXAzbfhJx2ZhRvh)_

PR #483, opened 2026-07-21
chore: align dev with current staging (gRPC migration + docs merged into staging's security-delete work)

harrymove-ctrl:
> **Status: Draft — review only, do not merge yet.** Per team guidance, this should be reviewed now but held until the ongoing V1→V2 migration work lands.

## Summary

This PR is the new dev baseline built on staging as the standard: **base is `staging`**, plus the valid `dev`-only parts. It introduces no new gRPC — the gRPC in this diff is what `staging` already ships (`690888b9`) reconciled with `dev`'s #355, and wherever the two branches disagreed, conflicts were resolved in `staging`'s favor.

Context: `dev` and `staging` had diverged substantially (112 files, +18.7k/−1k lines) — the gRPC migration from #355 landed on `dev` but never reached `staging`, while `staging` independently rebuilt the delete-old-memories feature into the `security_delete` subsystem and hardened the relayer.

## What dev adds on top of staging (19 files)

**Documentation (10):**

| File | Change |
|---|---|
| `docs/guides/delete-memories-programmatically.md` | New guide — deleting memories via API/SDK |
| `docs/sdk/headless-setup.md` | New guide (#482 agent-memory set) |
| `docs/sdk/codebase-memory.md` | New guide (#482 agent-memory set) |
| `docs/fundamentals/concepts/verifiable-memory.md` | New concept doc (#482) |
| `docs/relayer/overview.md`, `docs/relayer/self-hosting.md` | `SUI_GRPC_URL` documentation (#481) |
| `docs/reference/environment-variables.md`, `docs/reference/configuration.md` | Env/config reference updates (#481) |
| `docs/docs.json` | Navigation entries |
| `docs/guides/delete-old-memories.md` | Minor fix |

**Smart contract (2):**

- `services/contract/sources/account.move` + `tests/account_tests.move` — the `seal_approve` authorization-scoping fix with its regression test (`test_seal_approve_delegate_rejects_unrelated_id`).

**Supporting changes (7):**

- *Comment-only (2):* `services/server/src/types.rs` (documents that `sui_rpc::Client` must not be rebuilt per request), `apps/app/src/utils/suiClientCompat.ts` (architecture note updated).
- *Merge glue (3):* `ConnectMcp.tsx` and `SetupWizard.tsx` now use the shared `suiClientCompat` helpers instead of the deleted `suiFields.ts` — same behavior (JSON-RPC), but the pages no longer assume one transport shape.
- *Fixes worth keeping (2):* `sidecar/retry/rpc.ts` adds gRPC-aware retryable status codes (`UNAVAILABLE`, `RESOURCE_EXHAUSTED`, `DEADLINE_EXCEEDED`, …) — without it, the gRPC write path staging already uses silently stops retrying; `useSponsoredTransaction.ts` adds retry/backoff on the `/sponsor` call and settles on the always-sponsored flow.

Note: the GitHub "Files changed" tab diffs against the base (`dev`), so it shows a different, larger set — the list above is the delta against `staging`, which is the meaningful one under the "staging is standard" model.

## Conflict resolutions

| Area | Decision | Rationale |
|---|---|---|
| Frontend transport | Kept `staging`'s architecture | Shared `SuiClientProvider` stays JSON-RPC; gRPC remains scoped to the security-delete subsystem's own client. `dev`'s unconditional `createClientForNetwork` swap was removed. |
| Backend validation | Kept `staging`'s hardening | `validate_memwal_account_type` rejects objects that aren't actually a `MemWalAccount` before trusting their fields. |
| Sidecar testnet policy | Kept `staging`'s stricter stance | gRPC mandatory on testnet: boot-time `validateSuiTransportPolicy` guard, `suiJsonRpcClient` forced to `null`. |
| `Cargo.toml` / `Cargo.lock` | Kept `staging`'s exact pins | Required by the native `sui/` client module. |

The full file-by-file resolution list is in the merge commit message (`03d1f72`).

## Verification

**Deployed environment** — the Railway `dev` environment runs this exact branch on both `app` (dev.memwal.ai) and `relayer` (relayer.dev.memwal.ai):

- Relayer logs show `delegate key verified onchain (gRPC)` on `/api/remember`; no failed phases
- Delegate account flow works end-to-end (dashboard, account resolution, delegate keys)

**Test suites**

- Server: `cargo check` clean; `cargo test` 273/278 (the 5 failures are pre-existing local DB-pool timeouts, unrelated)
- Sidecar: `npm test` 89/89
- App: `tsc -b`, `vite build` + verify-dist-assets, vitest 66/66
- Move contract: 48/48


hungtranphamminh:
### Code review

Reviewed the sync for functional correctness: does it bring the security-delete + relayer changes across intact, and does it preserve the existing #355 gRPC migration rather than falling back to a JSON-RPC path against the now-deprecated public Sui fullnodes (`-32601 "JSON-RPC deprecated"`)? Checked the merge seams, DB/migrations, SDK, CI, and docs. Functionally the merge is clean and regression-free; a few minor items below.

**Verification**

| Check | Result |
|---|---|
| #355 gRPC migration preserved, not reverted | Both #355 commits are ancestors of this branch; it's a superset, not behind. |
| No blockchain interaction regressed gRPC → JSON-RPC | No transport path removed; `storage/walrus.rs` byte-identical. |
| Testnet fails closed without gRPC (no dead JSON-RPC calls) | `validate_security_delete_config` and the sidecar's `validateSuiTransportPolicy` both refuse to boot on testnet without `SUI_GRPC_URL`; the raw `suiRpc()` helper self-guards. |
| Legacy account resolution safe on testnet | `resolve_account` fails closed on testnet with an explicit error instead of a silent JSON-RPC dead-call. |
| Merge seams intact | Clients constructed once and shared; boot guards run before any config unwrap; background jobs registered exactly once; delegate-key `expected_package_id` threaded consistently across call sites. |
| DB migrations safe | New schema targets a separate legacy DB behind default-off flags — no numbering collision, no destructive statement, no unsafe backfill against live `vector_entries`. |
| SDK backward-compatible | `ServerConfig` change is internal; the older-relayer `/config` fallback still resolves (either `suiGrpcUrl` or `suiRpcUrl` satisfies it). |
| Extra hardening present | `validate_memwal_account_type` (object type-check) and the `validate_digest` 32–44 base58 widening are correct. |

**Heads-up: `remember`/`analyze` are behind a 503 stub on this branch**

`/api/remember`, `/api/remember/manual`, `/api/remember/bulk`, and `/api/analyze` route to the `uploads_paused` 503 ("paused while we conduct a security upgrade") rather than the real handlers. It isn't mentioned in the description, so it can read as a merge break when testing the branch — worth a line in the PR body to preempt that.

**Minor — make the testnet example config self-sufficient**

`App.tsx` keeps the shared `SuiClientProvider` on JSON-RPC and scopes gRPC to the deletion subsystem; `jsonRpcUrlFor('testnet')` uses `VITE_SUI_RPC_URL` when set:

https://github.com/MystenLabs/MemWal/blob/24609e92b1cd58e397a21475604f84f0bb19c6d3/apps/app/src/App.tsx#L50-L61

That's fine when the deployment sets `VITE_SUI_RPC_URL`, but the checked-in example leaves it commented out in the active testnet block:

https://github.com/MystenLabs/MemWal/blob/24609e92b1cd58e397a21475604f84f0bb19c6d3/apps/app/.env.example#L58-L61

so a deploy built purely from the example falls back to the now-deprecated public testnet fullnode. Suggest uncommenting `VITE_SUI_RPC_URL` in that block. (Related: the added `deploy-app-walrus.yml` comment "The app's Sui client is gRPC" now contradicts the JSON-RPC-only `App.tsx`.)

**Minor — dangling workspace glob**

`pnpm-workspace.yaml` adds a `tests/*` member glob, but there's no `tests/` directory in the tree, so it matches nothing. Worth removing as leftover config.

**Optional doc follow-up**

`docs/relayer/overview.md` still describes the gRPC switch as "opt-in and off by default … a single reversible switch," which no longer holds on testnet now that the sidecar hard-requires gRPC there (`sui-transport-policy.ts`). A follow-up would keep operators from skipping `SUI_GRPC_URL` and hitting a boot failure.

hungtranphamminh (approved):
Approving. Verified the sync functionally, and also ran the branch locally end-to-end to confirm it fixes the delegate-key flow that's currently broken on `dev`.

**Ran it (local, real staging testnet config):** the branch builds and boots clean — server + sidecar, both DBs with security-delete migrations applied, dual gRPC/JSON-RPC transport, Apalis workers, Redis. Drove the delegate-key/sponsor flow in a browser against the local relayer:

- `POST /sponsor` → 200, `POST /sponsor/execute` → 200 (the calls that CORS-block / `net::ERR_FAILED` on `dev`)
- `delegate key verified onchain (gRPC)` — the migrated transport path works
- delegate key created flawlessly; `/api/remember` → 503 only *after* auth succeeds, i.e. the intended upload-pause gate behind working auth (also confirmed from the Playground)

**Sync checks (no regressions):** #355 gRPC migration preserved (superset of `dev`, not behind); no transport path reverted to JSON-RPC; testnet fails closed without gRPC; `resolve_account` now fails closed on testnet instead of a silent dead JSON-RPC call; merge seams (wiring, jobs, config guards, delegate-key `expected_package_id` threading) intact; migrations land in a separate legacy DB behind default-off flags with no collision/destructive statement.

Non-blocking follow-ups from the earlier comment still stand (self-sufficient testnet `.env.example`, dangling `tests/*` workspace glob, `overview.md` "opt-in" doc wording). None gate this.

LGTM.

jessiemongeon1:
## Style Guide Audit

**10** violation(s) found. All must be fixed before merge.

### `docs/api/security-delete.md` (10 violation(s))

_10 violation(s) (0 regex, 10 claude)_

- **Line 4** — passive voice
  - Current: `were tracked in MemWal's old-V1 database`
  - Fix: `that MemWal tracked in its old-V1 database`
- **Line 6** — passive voice
  - Current: `signs the returned `TransactionData``
  - Fix: `signs the `TransactionData` that the server returns`
- **Line 52** — passive voice
  - Current: `Addresses are canonicalized to lowercase`
  - Fix: `The server canonicalizes addresses to lowercase`
- **Line 52** — passive voice
  - Current: `before they are stored or compared`
  - Fix: `before storing or comparing them`
- **Line 215** — passive voice
  - Current: `are removed from the batch and reported as`
  - Fix: `the server removes from the batch and reports as`
- **Line 217** — passive voice
  - Current: `Its input ordering is persisted`
  - Fix: `The server persists its input ordering`
- **Line 245** — passive voice
  - Current: `the returned digest`
  - Fix: `the digest the server returns`
- **Line 311** — passive voice
  - Current: `Client-supplied values are not echoed in error messages`
  - Fix: `Error messages do not echo client-supplied values`
- **Line 398** — passive voice
  - Current: `Transaction input-to-blob ordering is persisted before signing`
  - Fix: `The server persists transaction input-to-blob ordering before signing`
- **Line 399** — passive voice
  - Current: `Sponsor signatures are generated only at submit time and are never returned or stored`
  - Fix: `The server generates sponsor signatures only at submit time and never returns or stores them`

---

PR #488, opened 2026-07-22
BEDU-913: Walrus Memory trust boundary interactive diagram

reemsabawi-mystenlabs:
## BEDU-913: trust boundary interactive diagram

Adds the interactive trust-boundary diagram to `docs/fundamentals/architecture/data-flow-security-model.md` and rebases this PR onto the latest `dev` (post-#487), superseding the July 22 draft commit on this branch.

### Diagram

- Asset: `docs/diagrams/interactive_trust-boundary_v1.html` (self-contained HTML bundle, same format as the three BEDU-912 diagrams). The artwork is the corrected per-mode export from the security review: plaintext enters managed and self-hosted relayers with encryption inside that boundary, TEE shows plaintext only inside the attested enclave, and only the manual client flow sends ciphertext to the relayer.
- Four deployment models (manual, self-hosted, TEE, managed relayer) auto-cycle. No tabs, no player controls.
- Theme-aware with light-mode contrast.
- Orthogonal arrow routing.
- Embedded in a `<figure>` with an agent-readable `figcaption` so Kapa and other agents can ingest the diagram's claims as text; prose follows the Sui documentation style guide.

**Known limitation (mobile):** the current diagram exports are desktop-first. The iframe is fluid-width, but the drawing inside renders at a fixed size with no media queries or resize handling, so small screens see a clipped or inner-scrolling view at the fixed 440px height. This matches the interactive diagrams already merged in #487; a responsive v2 export is planned as part of the diagram design-system work, and the figcaption carries the full flow as text for readers the diagram doesn't serve.

### Embed path

The iframe src uses `https://docs.wal.app/diagrams/interactive_trust-boundary_v1.html`, matching the convention the three sibling diagrams merged in #487 use. Mintlify does not serve repo `.html` files at relative `/diagrams/` paths, so the docs.wal.app statics deployment serves the bundle. The URL resolves once this file reaches that deployment through the same pipeline as the BEDU-912 diagrams.

### Style pass

The mechanical style audit gates pushes on the whole touched file, so this PR also carries a Sui style-guide pass on `data-flow-security-model.md`: unquoted frontmatter plus a `keywords` field, `Trust and Security Model` title (no ampersand), em dashes replaced in prose, bold table headers, `Seal` casing, `basic` for `simple`, `might` for `may`, active-voice rewrites, and unwrapped prose lines. Mermaid code blocks are untouched.

> [!IMPORTANT]
[...]

PR #490, opened 2026-07-23
docs(relayer): complete the API reference against the live route table

reemsabawi-mystenlabs:
## Description

Closes BEDU-654 (DOC-02 · Walrus Memory UI: relayer API contracts & endpoints). Audits `docs/relayer/api-reference.md` against the actual route table (`services/server/src/main.rs`) and request/response structs (`services/server/src/types.rs`), and closes the drift so the Memory UI team can integrate against accurate contracts. Two files changed (the API reference, plus a review-requested drift fix in `self-hosting.md`); no new page.

## What was missing or wrong

- **Four undocumented routes:** `GET /config` (public deployment parameters the SDK reads to build a Seal SessionKey, including `rateLimitDisabled`), `GET /metrics` (Prometheus, cross-linked to Observability), `POST /api/forget` (deletes the vector index rows for a namespace; Walrus blobs persist so `restore` can re-index), and `POST /api/stats` (memory count + stored bytes). The last two are owner-scoped.
- **Composite scoring:** the optional `scoring_weights` object (accepted by `recall`, `recall/manual`, and `ask`) was entirely undocumented, as were the `score` field on results and `dropped_count` on the recall response. `score` only appears when `recency` or `importance` is nonzero, because the ranker short-circuits to plain cosine order otherwise.
- **`recall/manual` response fields:** `created_at` and `importance` were missing from the documented shape.
- **`analyze`:** the optional `occurred_at` temporal anchor was undocumented, including its privacy property (the date lives only inside the fact text and embedding; the server keeps no event-time metadata).
- **Smaller gaps:** `ask` limit default (5) and cap (100), remember job status values, `404` anti-enumeration on single-job polling, `not_found` items on bulk status, the 2 MB bulk body limit, and the signed-header requirement incorrectly claiming to cover all `/api/*` routes (the MCP transports use bearer auth).
- **New section:** MCP transports (`/api/mcp`, `/api/mcp/sse`, `/api/mcp/messages`) with a cross-link to the MCP reference instead of duplicating it.
- **`self-hosting.md` drift fix (review request):** the `SUI_GRPC_URL` note claimed startup validation rejects non-gRPC endpoints; startup only rejects a malformed URL, and the client connects lazily, so a well-formed URL to a non-gRPC host fails on the first call. The page was also brought to docs style compliance since the fix put it in scope.

## Verification

- Every route in the reference now maps 1:1 to a `.route(...)` registration in `main.rs` (18 documented endpoints plus the 3 MCP transports = all 21 registrations); every documented field maps to a serde struct field in `types.rs`; defaults and caps quoted from source (`default_limit`, `default_semantic_weight`, `unwrap_or(5).min(100)`, `MAX_BULK_ITEMS`).
- Pages brought to style-guide compliance: unquoted frontmatter with keywords, sentence-case headings, bold table headers, body text under every section heading, unwrapped prose. Mechanical style audit passes.

https://claude.ai/code/session_01WoBWxqLzd9hAxrhoqv3iSv


jessiemongeon1:
## Style Guide Audit

docs/relayer/self-hosting.md
```diff
@@ -169,42 +177,42 @@ SEAL_SERVER_CONFIGS=[{"objectId":"0xb012378c9f3799fb5b1a7083da74a4069e3c3f1c93de
...
+- Without `OPENAI_API_KEY`, the server uses deterministic mock embeddings (hash-based), useful for local testing but not production.
+- Use `LOG_FORMAT=json` in production and see [Observability](/relayer/observability) for dashboards and alerts.
 
 ## Docker
 
 - `services/server/Dockerfile` for the relayer
 - `services/indexer/Dockerfile` for the indexer
 
-## Read Next
+## Read next
```
jessiemongeon1: These sections are not necessary

jessiemongeon1 (changes requested):

jessiemongeon1 (changes requested):
1. docs/relayer/self-hosting.md:112 still says Walrus storage epochs default to 50 on testnet and 2 on mainnet. Current Rust + sidecar config default to 5 on testnet and 3 on mainnet, capped at 15, so this changed line is still stale.

2. docs/relayer/api-reference.md:231 and the manual recall section around :288 document limit but omit the server cap. Both /api/recall and /api/recall/manual clamp limit to 100; /api/ask documents that cap, so the recall endpoints should too.

Nit: docs/relayer/api-reference.md:463 lists GET, POST, DELETE for /api/mcp, but the route table also mounts OPTIONS /api/mcp. Only worth fixing if the page is meant to be exhaustive by method.

reemsabawi-mystenlabs:
All three addressed in 0bd7e3e, each verified against source first:

1. **Storage epochs** — `self-hosting.md` now says `5` on testnet / `3` on mainnet (`types.rs` `DEFAULT_TESTNET_WALRUS_STORAGE_EPOCHS` and `default_walrus_storage_epochs_for_network`; sidecar `config.ts` mirrors it), and documents the `15` cap: both the request `epochs` value and the `WALRUS_STORAGE_EPOCHS` override clamp there, with over-cap overrides falling back to the network default.
2. **Recall limit cap** — both `POST /api/recall` and `POST /api/recall/manual` now document the `100` cap (`recall.rs` clamps with `body.limit.min(100)` in both handlers), matching how `/api/ask` was already documented.
3. **OPTIONS on `/api/mcp`** — added, since the page does aim to be exhaustive against the route table (`main.rs` mounts `.options(mcp_proxy::streamable_proxy)`).

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/relayer/api-reference.md
```diff
@@ -25,13 +25,13 @@ goal:
       label: Needs answer summary for AI citation
 questions:
   - "What are the Walrus Memory relayer API endpoints?"
-  - "How do I authenticate requests to the MemWal relayer?"
+  - "How do you authenticate requests to the MemWal relayer?"
```
jessiemongeon1: should still be "How do I...". This is an instance where you'll need to fix this instance that goes against the general style guide rule when you manually review the PR after the agent generates it

jessiemongeon1 (changes requested):

reemsabawi-mystenlabs:
Fixed in 259b385 — the three `questions:` entries across `api-reference.md` and `self-hosting.md` are back to "How do I...". Also encoded the exception in the automated style gate so future passes stop "correcting" first person inside frontmatter `questions:` blocks: the second-person rule now applies to page prose only, and question entries keep the user's search voice.

Heads-up: the same over-correction went into the questions blocks on #488 (one entry) and #498 (three entries) before your approvals — happy to fix those in place (resets your approvals) or in a small follow-up after they merge, whichever you prefer.

---
_Generated by [Claude Code](https://claude.ai/code)_

jessiemongeon1 (approved):

PR #522, opened 2026-08-03
Remove noisy Walrus upgrade alert from security delete

ducnmm:
## Summary

- Remove the Slack alert path from `security-delete` when the configured Walrus package differs from the live system package.
- Keep the live on-chain package selection for delete PTBs unchanged.
- Keep upload-path package mismatch alerting and self-recovery unchanged.

## Why

The security-delete mismatch is expected after a Walrus upgrade because `WALRUS_PACKAGE_ID` remains the immutable Blob type-origin package. The existing alert required no operator action and re-fired on later deletes, creating noise.

## Validation

- `cargo test --manifest-path services/server/Cargo.toml security_delete --lib`
- `cargo fmt --manifest-path services/server/Cargo.toml --check`


harrymove-ctrl (changes requested):
LGTM

harrymove-ctrl (approved):

PR #524, opened 2026-08-03
Fix: remove orphaned custom zkLogin login path in noter (closes #518, #519)

### LINKED ISSUE #518: security(noter): cached zkLogin proof bypasses JWT signature verification and permits existing-account takeover [Security / High]
`completeLogin` trusts identity claims decoded with `jwt-decode`, which does not verify signatures. A fresh session still reaches Mysten's prover, and that prover rejects a forged signature. The bypass appears after the session already contains a cached `zkProof`: `completeLogin` reuses the proof, never calls the prover, and accepts the claims from a second caller-supplied JWT.

In the current implementation, salt derivation is also local and deterministic. A forged JWT carrying a victim's `iss` / `sub` / `aud` therefore produces the same `suiAddress` as the victim. `upsertZkLoginUser` selects by that address and takes its update branch, after which the attacker's session points to the pre-existing victim user.

## Affected code

`apps/noter/package/feature/auth/domain/zklogin.ts:66-83` only decodes claims and checks `exp`, `iss`, `sub`, and `aud`:

```ts
export function decodeAndValidateJwt(token: string): JwtClaims {
  try {
    const claims = jwtDecode<JwtClaims>(token);

    const now = Math.floor(Date.now() / 1000);
    if (claims.exp && claims.exp < now) {
      throw new Error(AUTH_ERRORS.INVALID_JWT);
    }

    if (!claims.iss || !claims.sub || !claims.aud) {
      throw new Error(AUTH_ERRORS.INVALID_JWT);
    }

    return claims;
```

`apps/noter/package/feature/auth/api/input.ts:38-41` accepts any non-empty string as the token:

```ts
export const completeLoginInput = z.object({
  jwt: z.string().min(1),
  sessionId: uuidv7Schema,
});
```

`apps/noter/package/feature/auth/api/route.ts:117-186` verifies the nonce, then skips proof generation whenever the session already has one:

```ts
const jwtClaims = decodeAndValidateJwt(jwt);

if (!verifyNonce(jwtClaims, session.nonce)) {
  throw new TRPCError({
    code: "UNAUTHORIZED",
    message: AUTH_ERRORS.INVALID_JWT,
  });
}

const salt = await fetchUserSalt(jwt);
const suiAddress = await deriveAddress(jwt, salt);

let zkProof;
if (session.zkProof) {
  zkProof = session.zkProof;
} else {
  zkProof = await generateZkProof({
    jwt,
    ephemeralPublicKey: session.ephemeralPublicKey,
    maxEpoch: session.maxEpoch,
    randomness: session.randomness,
    salt,
  });
}

const profile = extractUserProfile(jwtClaims);
const user = await authService.upsertZkLoginUser(ctx.db, {
  suiAddress,
  provider: profile.provider,
  providerSub: profile.providerSub,
  name: profile.name,
  email: profile.email,
  avatar: profile.avatar,
});
```

The project documentation states the opposite at `apps/noter/doc/zklogin-integration.md:158-161`: "JWT signature verification via OAuth provider."

## Why it is exploitable in practice

The attacker needs one session on which a genuine login previously produced a cached `zkProof`. The attacker then submits a second JWT with:

- the session's real nonce;
- a future `exp`;
- the victim's `iss`, `sub`, and `aud`;
- arbitrary profile fields;
- any signature bytes, including a garbage signature.

Nonce validation succeeds because the nonce belongs to the attacker's own session. The cached-proof branch prevents the forged JWT from reaching the only component that would verify its signature. Current deterministic salt derivation makes the forged victim claims resolve to the victim's existing address, so the user upsert returns the victim record rather than inserting a new one.

### Attack path

1. Complete one real zkLogin flow on an attacker-controlled session so `session.zkProof` is populated.
2. Learn the victim's provider subject for the relevant OAuth client.
3. Build a JWT-shaped string with the attacker's session nonce and the victim's `iss` / `sub` / `aud`; use a garbage signature segment.
4. Call `completeLogin` again with that JWT and the attacker's session id.
5. `decodeAndValidateJwt` accepts the claims.
6. `completeLogin` reuses `session.zkProof`; `generateZkProof` and the Mysten prover are not called.
7. The derived address matches the victim's address; `upsertZkLoginUser` returns the victim's existing `user.id`.
8. The attacker's session can call protected routes as that user.

### Cached-proof acceptance against the real auth router

```bash
cd apps/noter
PGHOST=/tmp/noter_pg \
PGPORT=5432 \
PGUSER=postgres \
DATABASE_URL=postgres:///postgres \
NEXT_PUBLIC_SUI_NETWORK=mainnet \
pnpm exec tsx poc_zklogin_cached_proof_reuse.ts
```

Observed output:

```text
[1] Creating session row (bypassing initiateLogin's getCurrentEpoch call)...
    sessionId: 019fae0b-5203-7584-b7f0-fb20118eb8a6 nonce: o6hhUVD5paqcfGJuZV26oxNbSic
[2] Seeding session.zkProof directly (simulating one prior real login)...
[3] completeLogin #2 with a forged, unsigned JWT for a DIFFERENT identity...
    returned suiAddress: 0x06ca70f5bb9f1b1d1fbbea07f2937dc15264abd188f53fcc47c966ffae9a307d
    persisted user row email/providerSub: attacker@evil.example attacker-google-sub-222222222222222222222
    session.zkProof after call is still the seeded placeholder (i.e. generateZkProof/real prover was never invoked): true
```

The token declared `alg: RS256` and carried a deliberately invalid signature segment. The identity was still persisted.

### Existing-user takeover and protected-route access

A second PoC drove the real `/api/trpc` endpoint over HTTP. It seeded a pre-existing victim with one private note, then used an attacker session containing the cached-proof precondition:

```text
victimSuiAddress: 0xefcf770593545ed98905e566e38ffa90e330b5303c7f7b44b950074ef048c9bc
victimId: 019fbd94-ba3d-7614-aa41-d17e5cc82cb3
victimNoteId: 019fbd94-ba7a-7ed6-97d3-50134d924976
attacker sessionId: 019fbd94-ba7f-7485-a8c2-54fa0033a95d

completeLogin with forged victim claims:
HTTP status: 200
returned user.id === pre-existing victim.id: true
session.zkProof untouched (real prover never called): true
session.userId now points at victim: true

protected note.list with the attacker's session:
HTTP status: 200
notes returned count: 1
includes victim's seeded note: true
```

## Negative control and exact scope

The same forged token was sent through `generateZkProof` on a fresh session, using a real Google JWKS key id but a garbage signature. Mysten's live prover rejected it:

```text
Prover HTTP 500: JWSSignatureVerificationFailed: signature verification failed
```

A fresh-session account creation bypass is therefore not present in the tested architecture. The exploitable path is the cached-proof branch, which does not reach the prover. The first PoC seeds the cached proof as a stand-in for the required prior genuine login; it does not claim that a forged token can create that first proof.

## Impact

A caller satisfying the cached-proof precondition can bind their session to an existing victim account and read data behind authenticated application routes. The PoC confirmed access to the victim's private note. The attack does not forge a Sui transaction and does not bypass the prover on a fresh session; it compromises the application authentication layer after proof caching.

## Suggested fix

Verify the ID token locally before trusting any claim, including on cached-proof reuse. Use an explicit issuer allow-list and each provider's live JWKS; validate issuer and audience and reject unsupported issuers fail-closed:

```diff
 import { jwtDecode } from "jwt-decode";
+import { jwtVerify, createRemoteJWKSet } from "jose";

+const ISSUER_JWKS: Record<string, ReturnType<typeof createRemoteJWKSet>> = {
+  "https://accounts.google.com": createRemoteJWKSet(
+    new URL("https://www.googleapis.com/oauth2/v3/certs")
+  ),
+};
+
-export function decodeAndValidateJwt(token: string): JwtClaims {
+export async function decodeAndValidateJwt(token: string): Promise<JwtClaims> {
   const claims = jwtDecode<JwtClaims>(token);
   ...
+  const jwks = ISSUER_JWKS[claims.iss];
+  if (!jwks) {
+    throw new Error(AUTH_ERRORS.INVALID_JWT);
+  }
+  await jwtVerify(token, jwks, {
+    issuer: claims.iss,
+    audience: claims.aud,
+  });
   return claims;
 }
```

The three call sites must await the now-async function. Other configured providers need equivalent JWKS entries before being accepted.

## Regression test

The existing-user takeover PoC was rerun against the patch:

```text
completeLogin with forged victim claims:
HTTP status: 500
message: Invalid or expired JWT token
returned user.id === pre-existing victim.id: false
session.userId now points at victim: false

protected note.list with the attacker's session:
HTTP status: 401
notes returned count: 0
includes victim's seeded note: false
```

**Environment:** `MystenLabs/MemWal`, PoC baseline `dev` at `a67e9f4a633a9431352aaeb7eb606b30c95c70ef`; revalidated unchanged on current `dev` `e7fd5c5a0d1849a65a93786f05baa98aed34542a`.

### LINKED ISSUE #519: security(noter): deterministic zkLogin salt makes OAuth identity-to-address mapping independently computable [Security / Low]
`fetchUserSalt` does not call the configured Mysten salt service. It returns a local SHA-256 derivation over `iss`, `sub`, and `aud`, truncated to 16 bytes. There is no server secret, random per-user input, or HMAC key. Anyone who obtains the victim's OAuth subject for the relevant client can independently derive the same salt and therefore the same zkLogin Sui address.

## Affected code

`apps/noter/package/feature/auth/lib/zklogin-client.ts:98-112`:

```ts
async function generateLocalSalt(jwt: string): Promise<string> {
  const claims = decodeAndValidateJwt(jwt);
  const saltString = `${claims.iss}::${claims.sub}::${claims.aud}`;
  const encoder = new TextEncoder();
  const data = encoder.encode(saltString);
  const hashBuffer = await crypto.subtle.digest('SHA-256', data);
  const hashArray = Array.from(new Uint8Array(hashBuffer).slice(0, 16));
  const hashHex = hashArray.map(b => b.toString(16).padStart(2, '0')).join('');
  const salt = BigInt('0x' + hashHex).toString();
  return salt;
}

export async function fetchUserSalt(jwt: string): Promise<string> {
  return generateLocalSalt(jwt);
}
```

`apps/noter/package/feature/auth/constant.ts:46-54` already configures the intended service:

```ts
export const ZKLOGIN_CONFIG = {
  network: (process.env.NEXT_PUBLIC_SUI_NETWORK || "mainnet") as "testnet" | "mainnet",
  proverUrl: process.env.NEXT_PUBLIC_ZK_PROVER_URL || "https://prover-dev.mystenlabs.com/v1",
  saltServiceUrl: process.env.SALT_SERVICE_URL || "https://salt.api.mystenlabs.com/get_salt",
};
```

A source search for `saltServiceUrl` finds only this definition. The integration documentation says the login flow "fetches user salt from Mysten Labs service," but the implementation never does so.

## Why it matters

`iss` and `aud` are fixed for a deployment and OAuth registration. The remaining input is the user's provider subject. If that subject is disclosed through logs, analytics, another relying party using the same OAuth client, or another application-level exposure, an observer can map the OAuth identity to the user's public Sui address without querying MemWal or possessing the JWT.

This is a privacy/linkability issue, not a fund-custody or authentication bypass by itself. The attacker must know the victim's provider subject.

## Reproduction

```bash
node evidence/poc_zklogin_deterministic_salt.mjs
```

Observed output:

```text
App-computed salt:       27529140644209710438497641292389146440
Attacker-computed salt:  27529140644209710438497641292389146440
Identical (zero secret entropy): true
```

A second run with an unrelated `sub` produced a different numeric salt, while the application-side and independent attacker-side computations still matched. This rules out a constant-output test artifact and confirms that the value varies by subject while containing no secret entropy.

The configured service was also checked directly with a placeholder token:

```text
POST https://salt.api.mystenlabs.com/get_salt
HTTP 400 {"error":"Invalid token"}
```

The service exists and requires a genuine signed token; the application simply does not call it.

## Impact

A party that learns the victim's provider subject can independently derive the victim's zkLogin address and correlate that address's public on-chain activity with the OAuth identity. No access to the MemWal application, database, session, or secret is required after the subject is known.

## Suggested fix

Use the configured salt service by default. Keep local deterministic derivation, if needed, behind an explicit development-only opt-in:

```diff
+async function fetchUserSaltFromService(jwt: string): Promise<string> {
+  const response = await fetch(ZKLOGIN_CONFIG.saltServiceUrl, {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({ token: jwt }),
+  });
+  if (!response.ok) {
+    throw new Error(`Salt service request failed: ${response.status}`);
+  }
+  const data = await response.json();
+  if (typeof data.salt !== "string" || !data.salt) {
+    throw new Error("Salt service returned an invalid response");
+  }
+  return data.salt;
+}
+
 export async function fetchUserSalt(jwt: string): Promise<string> {
-  return generateLocalSalt(jwt);
+  if (process.env.ZKLOGIN_ALLOW_LOCAL_SALT === "true") {
+    return generateLocalSalt(jwt);
+  }
+  return fetchUserSaltFromService(jwt);
 }
```

## Regression tests

- Type-check the async call path with `tsc --noEmit`.
- Verify that production/default configuration calls `fetchUserSaltFromService`.
- Verify that local deterministic derivation is reachable only with explicit `ZKLOGIN_ALLOW_LOCAL_SALT=true`.
- With a genuine signed OAuth token, verify that the service response is used and that independent `SHA-256(iss::sub::aud)` no longer reproduces the application salt.

The first three checks were completed without new type errors. The final live service round trip was not run because no genuine OAuth credential was available; the placeholder request above confirms service liveness and authentication enforcement, not a successful salt response.

**Environment:** `MystenLabs/MemWal`, PoC baseline `dev` at `a67e9f4a633a9431352aaeb7eb606b30c95c70ef`; revalidated unchanged on current `dev` `e7fd5c5a0d1849a65a93786f05baa98aed34542a`.

---

hungtranphamminh:
## Summary

This PR started as the removal of an orphaned, insecure custom zkLogin login path
in the `noter` example app (closing #518 and #519), and — after review — was
extended to harden the Enoki login path that survives.

### Part 1 — Remove the orphaned custom zkLogin path (#518, #519)

The app shipped a **custom** OAuth→JWT→prover→salt zkLogin flow (`initiateLogin` /
`completeLogin`) alongside the Enoki flow the UI actually uses. It was never wired
into the UI (no callers, no `/auth/callback` page) but stayed live over HTTP with
two reported defects:

- **#518 (High) — account takeover.** `completeLogin` decoded the JWT with
  `jwt-decode` (no signature verification) and its cached-proof branch skipped the
  prover — the only step that would reject a forged token. A second forged JWT
  (victim's `iss`/`sub`/`aud` + the attacker's own session nonce + a garbage
  signature) rebound the attacker's session to the victim's account.
- **#519 (Low) — deterministic salt.** `fetchUserSalt` returned
  `SHA-256(iss::sub::aud)` truncated to 16 bytes, computed locally with no server
  secret. Since `iss`/`aud` are public per-deployment, anyone who learned a
  victim's OAuth `sub` could recompute the victim's exact zkLogin address — the
  precondition that made #518 resolve to the victim's exact address.

Removed the path in full (deleted `lib/zklogin-client.ts` and `domain/zklogin.ts`,
the two procedures, the now-dead service functions, dead config/schemas, and stale
docs). The endpoint no longer exists, so both #518 and #519 are closed
structurally rather than patched.

### Part 2 — Harden the surviving Enoki login path (review follow-up)

Review found that the Enoki path itself authenticated on trust: `connectEnoki`
took a client-supplied `suiAddress` with no proof of control, and returned the full
user row — including `delegatePrivateKey`, which is the relayer signing secret used
to sign every memory operation. Fixed, mirroring the pattern already used in the
`researcher` app:

- **Proof of ownership.** A server-issued single-use challenge (`issueEnokiChallenge`,
  backed by Redis: `SET NX` to issue, atomic `GETDEL` to consume). `connectEnoki`
  verifies the signed challenge **before any lookup**, for both phases, and fails
  closed if the challenge store is unavailable. The signature is verified against
  the claimed address and the recovered address must match.
- **No secret leakage.** Every auth path (`connectEnoki`, `connectWallet`,
  `connectDelegateKey`, `getSession`) returns a sanitized user DTO that never
  includes the delegate private key or PII. The key is exposed only through a
  dedicated, challenge-gated `exportDelegateKey` procedure.
- **No credential overwrite.** A guard prevents registration from clobbering an
  existing account's stored delegate credentials.
- **Legacy sessions cut off in code.** The app no longer reads or trusts the legacy
  `zklogin_sessions` table for authentication — sessions resolve only from
  `wallet_sessions` (which require proof of ownership to create). A leftover or
  re-introduced `zklogin_sessions` row can no longer authenticate a request. Logout
  still deletes any leftover row, and a one-time cleanup script
  (`pnpm db:purge-legacy-sessions`) is provided for data hygiene.
- Input addresses are validated with a `0x[64hex]` regex at the boundary.

### New requirement

Enoki sign-in now needs a **`REDIS_URL`** (the challenge store) — the same
dependency the `researcher` app already uses. Sign-in fails closed (503) if it is
unset or unreachable. Documented in the README.

## Types of Changes

- [ ] New feature (non-breaking)
- [x] Bug fix (non-breaking)
- [x] Refactor (no behavior change) 
- [ ] Dependency update
- [x] Documentation
- [ ] Test

## Benchmark impact

Example app only; does not touch recall / answer / extraction paths.

## Testing

- [x] Tested locally — `npx tsc --noEmit` in `apps/noter` is clean of new errors
  (only pre-existing, unrelated `@mysten-incubation/memwal` module-not-found and a
  `pdw-client.ts` implicit-any remain, confirmed identical on `dev`). CI green.
- [ ] Added/updated unit tests — N/A. noter has no test runner. The removed
  procedures are deleted (no branch to revert), and the Enoki hardening was verified
  by adversarial source review reproducing each attack against the fixed code.
- [ ] `cargo test` / harness tests pass — N/A (JS app)

## Checklist

- [x] Docs updated if needed
- [x] Respects the privacy floor (no server-readable plaintext index) — N/A, example app

## Related

- Closes #518
- Closes #519
- Linear: WALM-310, WALM-311

## Additional Notes

- The `zkLoginSessions` table/schema are retained (no migration); the table is
  simply no longer trusted for authentication.
- `connectDelegateKey` proves control via key possession (the address is derived
  from the supplied private key), so it does not use the challenge flow — by design.


ducnmm (changes requested):
Thanks for removing the orphaned custom zkLogin path. The specific forged-JWT path in #518 and the deterministic-salt path in #519 are removed, but I do not think this is ready to merge yet.

Blocking findings:

- **Critical: `auth.connectEnoki` still allows account takeover using only a public Sui address.** The public procedure accepts a caller-controlled `suiAddress` without an Enoki token, ownership signature, nonce, or other server-verifiable proof. For an existing Enoki user, the returning-user branch looks up the user by address and immediately creates an authenticated session. The response also returns the full user record, which includes the stored delegate credentials.
- **Critical: the registration branch can overwrite credentials for an existing address.** Supplying an arbitrary `privateKey` and `accountId` causes `upsertEnokiUser` to update the matching user and replace `delegatePrivateKey` / `delegateAccountId`, again without proving ownership of the Enoki address or validating the delegate relationship.
- **Medium: potentially compromised legacy sessions are retained.** The PR removes the writer for the vulnerable custom flow, but `zkLoginSessions` remains trusted by the tRPC context and memory routes. Any session rebound through #518 before deployment remains authorized until expiry. Please invalidate existing legacy sessions during deployment or enforce a cutoff.
- **Documentation/configuration needs another pass.** The README still references the removed `/auth/callback` flow and a nonexistent `LoginButton`, while the Enoki configuration omits required variables such as `NEXT_PUBLIC_ENOKI_API_KEY`, `NEXT_PUBLIC_GOOGLE_CLIENT_ID`, `NEXT_PUBLIC_MEMWAL_PACKAGE_ID`, and `NEXT_PUBLIC_MEMWAL_REGISTRY_ID`.

Before merging, please require server-verifiable proof of control for both Enoki phases, return a sanitized user DTO that never includes delegate secrets, verify the delegate/account binding before updating a user, and revoke or cut off legacy zkLogin sessions.

Validation performed: `git diff --check`, SDK build, Noter TypeScript check, and Noter production build all pass. There are currently no automated authentication tests covering these flows.

ducnmm:
**Re-review of the latest revision (`bff6a7b`)**

The original unauthenticated `connectEnoki` account-takeover path and legacy `zklogin_sessions` trust issue are resolved. The sanitized session DTO work also looks correct. I still do not think this is ready to approve because of the following remaining issues:

1. **Blocking: `exportDelegateKey` has purpose confusion and no session binding.** `issueEnokiChallenge` always asks the user to sign `Sign in to Walrus Memory Noter`, and that same proof is accepted by both `connectEnoki` and the public `exportDelegateKey` procedure. A signature obtained under the appearance of a normal sign-in can therefore be exchanged for the long-lived delegate private key. Please use a purpose-specific export challenge whose signed message clearly names the export action, make `exportDelegateKey` an authenticated/protected procedure, and require the current session user/address to match the requested address.

2. **The credential-overwrite concern is not fully resolved.** `upsertEnokiUser` still permits existing Enoki/wallet rows to have `delegatePrivateKey` and `delegateAccountId` overwritten, while the route accepts those values without verifying their on-chain relationship. Please make initial registration insert-only once credentials exist, or implement an explicit authenticated rotation flow that verifies the delegate key/account binding on-chain before updating the row.

Please add regression coverage for cross-purpose challenge rejection, unauthenticated/mismatched-session export, replay prevention, and attempts to overwrite existing delegate credentials. Keeping my existing Changes Requested review in place pending these fixes.

hungtranphamminh:
Thanks for the careful re-review — both points were fair. Addressed in `ad6178d0` and `2d6807ae`; CI is green including a new Noter unit-test job.

## 1. Export challenge purpose + session binding

The delegate-key export is no longer reachable with a sign-in signature, and is now bound to the caller's own session:

- **Purpose-scoped challenges.** `issueEnokiChallenge` now takes a purpose (`signin` | `export`). The purpose is baked into both the human-readable message the user signs (`"Sign in to…"` vs `"Export your … delegate key"`) and the stored challenge record; `verifyAndConsumeEnokiChallenge` requires an exact purpose match. A signature obtained under a sign-in prompt can no longer be replayed against the export path, and vice-versa.
- **Dedicated, own-address-only export challenge.** `issueExportChallenge` is a protected procedure that mints an `export`-purpose challenge **only for the caller's own session address** — a caller can't request an export challenge for someone else's address.
- **`exportDelegateKey` is now protected + session-bound.** It requires a valid session (`protectedProcedure`), an `export`-purpose challenge signature, **and** that the session user's address equals the requested address. So a caller can only ever export their own key.

The `user-float` "reveal key" UI was updated to use the export challenge flow.

## 2. Credential overwrite

`upsertEnokiUser` is now **insert-only** for credentials: once a row has a `delegatePrivateKey`, the connect/register path refuses to overwrite it (throws a conflict error) rather than replacing it with caller-supplied values. This closes the case you flagged where an existing enoki/wallet row could still be overwritten. A key rotation, if ever needed, would be a separate authenticated, on-chain-verified flow — out of scope here.

## Regression coverage (the scenarios you asked for)

Stood up vitest in noter (new `Noter / Unit tests` CI job). Coverage maps to your list:

| You asked for | Test |
|---|---|
| cross-purpose challenge rejection | sign-in challenge rejected by an export verify, and vice-versa |
| replay prevention | same challenge consumed once (GETDEL) — a second verify fails |
| unauthenticated export | `exportDelegateKey` with no session → `UNAUTHORIZED` |
| mismatched-session export | session address A requesting address B → `FORBIDDEN` |
| overwrite of existing delegate credentials | overwrite attempt on a row with an existing key → refused, stored key untouched |

Each guard was mutation-checked — removing the purpose check, the session-address binding, or the overwrite guard each turns the corresponding test red — so the tests fail if a future change weakens them.

Let me know if there's anything else you'd like tightened.

ducnmm (approved):
Re-review of the latest revision (`2d6807ae`): the previously blocking findings are resolved.

- Delegate-key export now uses a purpose-specific challenge and is protected/session-bound.
- Sign-in and export challenges are cross-purpose isolated and single-use.
- Existing delegate credentials can no longer be overwritten through the registration path.
- Requested regression coverage is present and exercises unauthenticated/mismatched-session export, purpose isolation, replay prevention, and credential overwrite attempts.

Validation: `git diff --check`, Noter unit tests (14/14), SDK build, and Noter `tsc --noEmit` all pass. Approved.

PR #527, opened 2026-08-04
fix(server): use address balances for Walrus uploads

ducnmm:
## Summary

- require SUI and WAL address balances across Walrus register, certify, and metadata-transfer transactions
- fail closed before the Sui SDK can fall back to owned coin objects, including retry and Enoki direct-sign fallback paths
- preserve legitimate async Walrus Blob inputs across repeated builds while rejecting resolver-added coin objects
- preserve the existing legacy fallback policy while extending address-balance enforcement to durable and standalone upload flows

## Verification

- `pnpm test` in `services/server/scripts`: 175/175 passed
- TypeScript no-emit check for `sidecar-server.ts` and `walrus-upload.ts`: passed
- `git diff --check`: passed
- independent agent review after PR creation: one async Blob retry issue found, fixed in `8849821a`, re-review completed with no findings

### Live testnet evidence

A one-epoch upload completed end to end using wallet address balances, then the payload was read back through the Walrus client:

- Blob ID: `rwPd-5ESC_b3SS8ChY1TF-J6iygqMy5BFNh_RBU9r6g`
- Blob object: `0x8b697e6c360eac8204e788d8d006be976b8833090ceddbce8e06af0c4e58e482`
- Certification transaction: `A5mcrj2EtNiKtbvAvrE544sYUD2nGmtXXpk6KYzVk4u3`
- Read-back payload: `MemWal final address-balance testnet upload 1785814830`

Testnet WAL was moved from owned coin storage into the writer address balance before the upload:

- Funding transaction: `DWU8L6GAxcqehaiovDyeJABtNb53wEB43m4vdLvcjm6g`

## Deployment note

This is not an API schema change, but it is an operational behavior change. Writer wallets must have sufficient SUI and WAL address balances; owned coin fallback is intentionally rejected. The current mainnet wallet pool must be funded through the address-balance distribution flow before deployment.

### DIFF
--- a/services/server/scripts/__tests__/sidecar-query-helpers.test.ts
+++ b/services/server/scripts/__tests__/sidecar-query-helpers.test.ts
@@ -1,7 +1,7 @@
 import test from "node:test";
 import assert from "node:assert/strict";
 import { Ed25519Keypair } from "@mysten/sui/keypairs/ed25519";
-import { Transaction, TransactionDataBuilder } from "@mysten/sui/transactions";
+import { Inputs, Transaction, TransactionDataBuilder } from "@mysten/sui/transactions";
 import {
     assertCompletionBlobObject,
     assertCompletionBlobResponse,
@@ -263,6 +263,14 @@ test("address-balance uploads fail instead of falling back to owned coins", asyn
                     hasNextPage: false,
                 };
             },
+            async getObjects({ objectIds }: { objectIds: string[] }) {
+                return {
+                    objects: objectIds.map((objectId) => ({
+                        objectId,
+                        type: `0x2::coin::Coin<${walType}>`,
+                    })),
+                };
+            },
         },
     };
     await assert.rejects(
@@ -272,6 +280,95 @@ test("address-balance uploads fail instead of falling back to owned coins", asyn
     await assert.rejects(
         transaction.prepareForSerialization({ client: client as never }),
         /Address-balance upload resolved owned coin objects/,
+    );
+});
+
+test("address-balance enforcement permits newly resolved non-coin objects", async () => {
+    const signer = new Ed25519Keypair();
+    const walType = `0x${"2".repeat(64)}::wal::WAL`;
+    const objectId = `0x${"6".repeat(64)}`;
+    const transaction = new Transaction();
+    transaction.setSender(signer.toSuiAddress());
+    transaction.coin({ type: walType, balance: 1n, useGasCoin: false });
+    enforceAddressBalanceCoinIntents(transaction);
+    transaction.addSerializationPlugin(async (transactionData, _options, next) => {
+        transactionData.addInput("object", Inputs.ObjectRef({
+            objectId,
+            version: "1",
+            digest: "33333333333333333333333333333333",
+        }));
+        await next();
+    });
+
+    const client = {
+        core: {
+            async getBalance() {
+                return {
+                    balance: {
+                        balance: "1",
+                        addressBalance: "1",
+                        coinBalance: "0",
+                    },
+                };
+            },
+            async listCoins() {
+                return { objects: [], cursor: null, hasNextPage: false };
+            },
+            async getObjects() {
+                return {
+                    objects: [{
+                        objectId,
+                        type: `${WALRUS_PACKAGE_ID}::blob::Blob`,
+                    }],
+                };
+            },
+        },
+    };
+
+    await assert.doesNotReject(
+        transaction.prepareForSerialization({ client: client as never }),
+    );
+});
+
+test("address-balance uploads reject owned gas resolved during build", async () => {
+    const signer = new Ed25519Keypair();
+    const gasObjectId = `0x${"4".repeat(64)}`;
+    const transaction = new Transaction();
+    transaction.setSender(signer.toSuiAddress());
+    transaction.setGasBudget(1n);
+    transaction.setGasPrice(1n);
+    transaction.transferObjects(
+        [transaction.objectRef({
+            objectId: `0x${"5".repeat(64)}`,
+            version: "1",
+            digest: "11111111111111111111111111111111",
+        })],
+        signer.toSuiAddress(),
+    );
+    enforceAddressBalanceCoinIntents(transaction);
+
+    const client = {
+        core: {
+            resolveTransactionPlugin() {
+                return async (
+                    transactionData: TransactionDataBuilder,
+                    _options: unknown,
+                    next: () => Promise<void>,
+                ) => {
+                    transactionData.gasData.payment = [{
+                        objectId: gasObjectId,
+                        version: "1",
+                        digest: "22222222222222222222222222222222",
+                    }];
+                    await next();
+                };
+            },
+        },
+    };
+
+    await assert.rejects(
+        transaction.build({ client: client as never }),
+        /Address-balance upload resolved gas from an owned coin/,
     );
 });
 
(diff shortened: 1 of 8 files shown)


services/server/scripts/walrus-upload.ts
```diff
@@ -154,6 +159,8 @@ async function main() {
 
     // Step 4: Certify blob on Sui → returns a Transaction
     const certifyTx = flow.certify();
+    enforceAddressBalanceCoinIntents(certifyTx);
+    certifyTx.setGasPayment([]);
```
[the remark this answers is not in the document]
ducnmm: Fixed in commit efc30428: the standalone CLI now calls certifyTx.setSenderIfNotSet(signerAddress) before address-balance enforcement. This makes certify object resolution explicit and consistent with the sidecar signing path. The full 175-test suite and TypeScript check pass.

services/server/scripts/sidecar/address-balance.ts
```diff
@@ -0,0 +1,73 @@
...
+            .map((input) => input.Object?.ImmOrOwnedObject?.objectId)
+            .filter((objectId): objectId is string => (
+                typeof objectId === "string" && !initialOwnedObjectIds.has(objectId)
+            ));
+        if (newlyResolvedOwnedObjects.length > 0) {
+            throw new Error(
+                `Address-balance upload resolved owned coin objects: ${newlyResolvedOwnedObjects.join(", ")}`,
+            );
+        }
+        if (transactionData.gasData.payment && transactionData.gasData.payment.length > 0) {
```
harrymove-ctrl, line 97: This check can't fire in practice — `gasData.payment` isn't resolved yet at this point. `build()` runs `prepareForSerialization()` (this plugin's phase) fully to completion *before* `#prepareBuild()` even starts, and gas-coin resolution (`resolveTransactionPlugin`) only happens in that second phase. See the top-level review comment for the SDK source citation.

Suggest moving this specific check into `transaction.addBuildPlugin(...)` so it runs after gas resolution instead.

services/server/scripts/__tests__/sidecar-query-helpers.test.ts
```diff
@@ -212,6 +229,74 @@ test("the pinned Sui SDK resolves WAL payment and relay SUI tip from address bal
     assert.equal(withdrawals.length, 2, "WAL payment and relay SUI tip both use address balances");
 });
 
+test("address-balance uploads fail instead of falling back to owned coins", async () => {
```
harrymove-ctrl, line 232: This test's title says "fail instead of falling back to owned coins", but it only calls `transaction.prepareForSerialization(...)` (lines 268-274), never `.build(...)`. Gas-coin resolution only happens inside `#prepareBuild`, which `prepareForSerialization` never reaches — so this test can only exercise the coin-intent-resolved-to-owned-object case (which does correctly reject today), not the gas-payment fallback case the PR summary claims is covered.

Suggest adding a case that mocks gas resolution to return an owned coin and asserts `transaction.build({ client })` rejects the same way. With the current code, that new test would fail — that's the actual bug this PR needs to fix.

harrymove-ctrl (changes requested):
## Request changes — gas-payment enforcement doesn't actually work

Verified the testnet evidence in this PR's description independently via `sui client` — it's genuine (certify tx `A5mcrj2EtNiKtbvAvrE544sYUD2nGmtXXpk6KYzVk4u3` succeeded using address-balance/accumulator gas: `Gas Payment:` empty, `Gas Object: 0x000...000`). `pnpm test` also reproduces 175/175 clean.

But that only proves the happy path (sufficient address balance -> works). It doesn't prove the PR's core claim — "fail closed before the SDK can fall back to owned coin objects" — for **gas payment** specifically.

### Root cause

`enforceAddressBalanceCoinIntents` (`address-balance.ts`) registers via `transaction.addSerializationPlugin(...)`. Its post-`next()` check (line 69, `gasData.payment.length > 0`) is meant to catch owned-coin gas fallback, but it can't: per the SDK source (`node_modules/@mysten/sui/src/transactions/Transaction.ts:796-798`), `build()` calls `prepareForSerialization()` and `#prepareBuild()` as two separate, sequential `await`s, not one chain. Serialization plugins (where this check lives) only run in the first phase. Gas-coin resolution (`resolveTransactionPlugin`) only runs in the second phase (`#prepareBuild`, `Transaction.ts:819-824`), strictly after our check has already returned. So by the time line 69 checks `gasData.payment`, gas hasn't been resolved yet — the check is structurally unable to observe an owned-coin gas fallback.

Since certify and metadata-transfer transactions carry no `CoinWithBalance` intents at all, this gas check is their *only* protection — meaning both are currently unprotected. This is reachable today: `DURABLE_WALLET_FALLBACK_POLICY.directSignIfUnconfigured = true` (`wallet.ts:33`) means any environment without `ENOKI_API_KEY` unconditionally hits the unprotected `executeDirectSignedTransaction` path (`enoki.ts:218-233`), and the standalone `walrus-upload.ts` CLI script has the same gap.

### Suggested fix

Split the check: keep the owned-object/coin-intent rejection (lines 59-68, works correctly today) as a serialization plugin, but move the gas-payment check to `transaction.addBuildPlugin(...)` instead. Build plugins run before `resolveTransactionPlugin` in the plugin array, but since `next()` unwinds through the rest of the chain (`resolveTransactionPlugin` is appended last), a check placed *after* `await next()` in a build plugin will correctly observe the resolved `gasData.payment` — same pattern this file already uses, just registered in the phase that actually contains gas resolution.

### Test gap that let this ship green

See inline comment below — the existing "fail instead of falling back to owned coins" test only calls `prepareForSerialization()`, never `build()`, so it never reaches the phase where the real bug lives. 175/175 passing didn't (and structurally couldn't) catch this.

### Other findings (lower severity, not blocking but worth addressing alongside the fix above)
- `enoki.ts`: `fallbackPolicy.gasMode` is optional with no type-level default forcing address-balance mode. A future caller that omits `fallbackPolicy` compiles clean and silently reintroduces owned-coin gas fallback — no test would catch it.
- `address-balance.ts:12`: owned-object detection is a presence-diff (before/after `next()`), not coin-type-aware. Not exploitable today, but it's the only guard certify/metadata-transfer would have even after the gas-phase fix, so worth hardening alongside it.

Scope note against the linked task ("review all upload-flow logic that depends on object-based balance"): register is solidly hardened; certify and metadata-transfer are the ones that still need it.

Happy to re-review once the gas-phase check is moved.

hungtranphamminh (approved):
Verified `4bf0de1e` addresses the review. The gas-payment guard is moved into a build plugin (`addBuildPlugin`, checked after `await next()`), so it runs in the `#prepareBuild` phase where `resolveTransactionPlugin` resolves gas — it can now actually observe an owned-coin gas payment. The new `"...reject owned gas resolved during build"` test mocks gas resolution to an owned coin and asserts `build()` rejects, closing the test gap. Both secondary items are handled too: `gasMode` is now required + defaults to `addressBalance`, and the newly-resolved-object check is coin-type-aware (only rejects `0x2::coin::Coin`).

Also confirmed the enforcement reaches certify and metadata-transfer at runtime — `enforceAddressBalanceCoinIntents` is applied at the top of `executeWithEnokiSponsor` before any branch, so it covers the direct-sign fallback paths as well as the sponsor path.

LGTM.

harrymove-ctrl (approved):
LGTM, approving.

ducnmm:
Closing — no longer needed. Address-balance upload work is not going forward on this branch.

Issue #413, opened 2026-07-13
##  Bug: auto_save can silently fail to save in short scripts


triumphsta:
**What happens**
When auto_save runs, it saves the memory in a background thread. It doesn't wait for that thread to finish before your program can exit. If your script or app closes quickly after sending a message, the save gets killed halfway through — before it reaches Walrus. No error shows up. It just doesn't save.

**How to see it**
1. Write a short script that sends one message through a memwal-wrapped client
2. Let the script end right after
3. Check the dashboard — nothing was saved, and there was no warning

**What would help**
- Say clearly in the docs that this only works safely in apps that stay running (like a web server), not short scripts
- Or add a way to tell the SDK "wait until the save is done before I close"
- Or print a warning if the program is about to exit while a save is still in progress

**Why it matters**
This wasted a lot of time. The chat reply looked normal, no errors showed up, but nothing was actually saved. It only worked properly once tested inside a long-running server, which isn't obvious from the docs.

harrymove-ctrl:
Thanks @triumphsta ! We've confirmed and reproduced this against the codebase (`withMemWal`'s `autoSave` calls `analyze()` as a fire-and-forget operation with no wait mechanism, which matches your description exactly).

This turns out to be the exact same root cause as #359, which has a more detailed investigation thread (including a clean experiment proving that an explicit `remember()` + `waitForRememberJob()` is 100% reliable, isolating the bug specifically to the fire-and-forget `autoSave` path). 

We are going to consolidate tracking over there, so I'm closing this as a duplicate. However, your report was a very helpful signal to confirm that this bug is independently affecting multiple users. 

**Closed as a duplicate of #359.** *(We are tracking the fix internally).*

Issue #359, opened 2026-07-05
withMemWal autoSave produces inconsistent recall results across separate process runs — likely async timing issue


oooo-o-pixel:
## Summary
When using the `withMemWal` middleware with `autoSave: true`, memory persistence appeared inconsistent across separate script executions — the same type of interaction sometimes resulted in the fact being recallable in a later run, and sometimes not, with no errors thrown in either case.

## Environment
- `ai@7.0.15`
- `@mysten-incubation/memwal/ai` (withMemWal)
- `@ai-sdk/openai`, model: gpt-4o-mini
- Node.js v22.19.0, Windows

## What happened
Using this pattern:
```ts
const model = withMemWal(openai("gpt-4o-mini"), {
  key: process.env.MEMWAL_PRIVATE_KEY!,
  accountId: process.env.MEMWAL_ACCOUNT_ID!,
  serverUrl: process.env.MEMWAL_SERVER_URL,
  namespace: "chatbot-prod",
  maxMemories: 5,
  autoSave: true,
});
```
I ran the script once, stating a fact ("My name is Alex and I'm building a football stats agent"). In a subsequent, completely separate process run, asking "What do you know about me?" sometimes correctly recalled the fact, and sometimes returned no information at all — with identical code and no errors logged in either case.

## Suspected root cause
`autoSave` appears to persist memories asynchronously in the background rather than blocking until the write is confirmed. Since each test run was a short-lived CLI script, the Node process likely exited before the background save completed in the "failed" runs — silently dropping the memory with no error surfaced to the caller.

By contrast, switching to explicit `memwal.remember()` + `memwal.waitForRememberJob()` calls (bypassing the `withMemWal` middleware entirely) produced 100% consistent, reliable results in the same script structure.

## Suggestion
- Document that `autoSave` is fire-and-forget and may not complete before a short-lived process exits, with a recommendation to use explicit `remember()` + `waitForRememberJob()` for CLI/script contexts where the process may exit quickly
- Consider exposing a way to flush/await pending `autoSave` writes before process exit (e.g. a `memwal.flush()` or similar), since long-running server processes wouldn't have this issue but short scripts do
- Alternatively, surface a warning/log when a process exits with pending unsaved `autoSave` writes

## Impact
This was confusing to diagnose since there were no errors — the memory simply wasn't there sometimes, which looked like a recall/search issue rather than a write-timing issue, until directly comparing against the explicit remember/wait pattern.

ducnmm:
[...]

PR #534, opened 2026-08-04
fix(sdk): drain fire-and-forget auto-save before process exit (WALM-306)

harrymove-ctrl:
## Summary
`withMemWal`'s `autoSave` (TS) and `with_memwal_openai`/`with_memwal_langchain`'s `auto_save` (Python) scheduled the post-call `analyze()` write as pure fire-and-forget — no strong reference kept on the asyncio Task/Promise, no way to await it. A short-lived process (CLI script, serverless handler) exiting before the background write finished lost it silently: no error, no log, exit code 0. Confirmed live on Mainnet via an external integration report (io.net), independently reproducing GH #359 / #413.

- **TS** (`packages/sdk/src/ai/middleware.ts`): tracks pending `analyze()` promises in a closure-scoped set; `withMemWal()`'s returned model now exposes `await model.flush()`.
- **Python** (`packages/python-sdk-memwal/memwal/middleware.py`): new `_PendingSaves` tracker holds strong refs to `asyncio.Task`s (closes the weak-ref GC hazard the asyncio docs warn about) and background `Thread`s across all 4 fire-and-forget call sites (async/sync × OpenAI/LangChain). Wrapped clients now expose `._memwal` (previously unreachable), `.memwal_flush()` (async), and `.memwal_flush_sync()`. A cancelled save now logs a warning instead of vanishing.

Fixes WALM-306 (GH #359, #413).

## Test plan
- [x] TDD: wrote a failing test first for both SDKs, confirmed RED for the right reason (`flush is not a function` / `AttributeError`-equivalent), then implemented to GREEN.
- [x] TS: `npm test` in `packages/sdk` — 17/17 pass, including new `withMemWal-flush.test.mjs`.
- [x] Python: `pytest tests/` in `packages/python-sdk-memwal` — 110 passed, 16 skipped (integration tests needing a live server), 0 regressions.
- [x] `ruff check` clean on touched Python files.
- [x] Re-verified both suites pass against a fresh `origin/dev` checkout (isolated worktree), not just the working branch.

- `packages/sdk`: `npm test` (or just `node --test test/withMemWal-flush.test.mjs` for the new test alone).
- `packages/python-sdk-memwal`: `python3 -m pytest tests/test_middleware.py -k memwal_flush -v`.

## Risks and dependencies
None. Purely additive API (`flush()` / `memwal_flush()` / `memwal_flush_sync()` / `._memwal`) — no existing signatures changed, no breaking changes. CHANGELOG.md updated in both packages under `Unreleased`; version bump/publish left for a separate release step.

### DIFF
--- a/packages/python-sdk-memwal/CHANGELOG.md
+++ b/packages/python-sdk-memwal/CHANGELOG.md
@@ -1,11 +1,5 @@
 # memwal
 
-## Unreleased
-
-### Fixed
-
-- `with_memwal_openai`/`with_memwal_langchain` auto-save ran fire-and-forget with no strong reference kept on the asyncio Task (or the daemon Thread, for sync callers) and no way to await it — a short-lived process could have its save cancelled or killed before it reached Walrus, silently, with no error and no log. Pending saves are now tracked and drainable via `client.memwal_flush()` (async) / `client.memwal_flush_sync()` (sync), and a cancelled save now logs a warning instead of vanishing. (GH #413, WALM-306)
-
 ## 0.1.5
 
 ### Added
--- a/packages/python-sdk-memwal/CHANGELOG.md
+++ b/packages/python-sdk-memwal/CHANGELOG.md
@@ -1,5 +1,11 @@
 # memwal
 
+## Unreleased
+
+### Fixed
+
+- `with_memwal_openai`/`with_memwal_langchain` auto-save ran fire-and-forget with no strong reference kept on the asyncio Task (or the daemon Thread, for sync callers) and no way to await it — a short-lived process could have its save cancelled or killed before it reached Walrus, silently, with no error and no log. Pending saves are now tracked and drainable via `client.memwal_flush()` (async) / `client.memwal_flush_sync()` (sync), and a cancelled save now logs a warning instead of vanishing. (GH #413, WALM-306)
+
 ## 0.1.5
 
 ### Added
--- a/packages/python-sdk-memwal/memwal/middleware.py
+++ b/packages/python-sdk-memwal/memwal/middleware.py
@@ -162,25 +162,9 @@ class _PendingSaves:
     def flush_sync(self) -> None:
         """Join every pending thread. If async tasks are also pending
         (possible when a sync call site shares state with an async one),
-        run them to completion on a fresh loop first.
-
-        A pending Task belongs to the loop that created it (e.g. an earlier
-        `await llm.ainvoke(...)` call); gathering it under a *different*
-        loop here raises RuntimeError. That's a caller-error case (mixing
-        sync and async entry points on the same wrapped client) this
-        cleanup path can't recover from — log and continue draining
-        threads rather than raise out of what's usually shutdown code.
-        """
+        run them to completion on a fresh loop first."""
         if self._tasks:
-            try:
-                asyncio.run(asyncio.gather(*list(self._tasks), return_exceptions=True))
-            except RuntimeError:
-                logger.warning(
-                    "Walrus Memory flush_sync() could not await %d pending task(s) "
-                    "bound to a different event loop (mixing sync and async calls "
-                    "on the same wrapped client?) — those saves may be lost.",
-                    len(self._tasks),
-                )
+            asyncio.run(asyncio.gather(*list(self._tasks), return_exceptions=True))
         self._join_threads()
 
     def _join_threads(self) -> None:
--- a/packages/python-sdk-memwal/memwal/middleware.py
+++ b/packages/python-sdk-memwal/memwal/middleware.py
@@ -162,9 +162,25 @@ class _PendingSaves:
     def flush_sync(self) -> None:
         """Join every pending thread. If async tasks are also pending
         (possible when a sync call site shares state with an async one),
-        run them to completion on a fresh loop first."""
+        run them to completion on a fresh loop first.
+
+        A pending Task belongs to the loop that created it (e.g. an earlier
+        `await llm.ainvoke(...)` call); gathering it under a *different*
+        loop here raises RuntimeError. That's a caller-error case (mixing
+        sync and async entry points on the same wrapped client) this
+        cleanup path can't recover from — log and continue draining
+        threads rather than raise out of what's usually shutdown code.
+        """
         if self._tasks:
-            asyncio.run(asyncio.gather(*list(self._tasks), return_exceptions=True))
+            try:
+                asyncio.run(asyncio.gather(*list(self._tasks), return_exceptions=True))
+            except RuntimeError:
+                logger.warning(
+                    "Walrus Memory flush_sync() could not await %d pending task(s) "
+                    "bound to a different event loop (mixing sync and async calls "
+                    "on the same wrapped client?) — those saves may be lost.",
+                    len(self._tasks),
+                )
         self._join_threads()
 
     def _join_threads(self) -> None:
--- a/packages/python-sdk-memwal/memwal/middleware.py
+++ b/packages/python-sdk-memwal/memwal/middleware.py
@@ -130,14 +130,72 @@ def _format_memories(
     ])
 
 
-def _fire_and_forget(coro: Any) -> None:
-    """Schedule an async coroutine as fire-and-forget.
+class _PendingSaves:
+    """Tracks in-flight fire-and-forget auto-save work (asyncio Tasks and
+    background Threads) so a caller can drain it deterministically via
+    :meth:`flush`/:meth:`flush_sync` instead of losing it silently when the
+    process exits before it completes. Tasks scheduled via
+    ``loop.create_task()`` with no other reference are only weakly held by
+    the event loop and can be cancelled or garbage-collected before they
+    run — this keeps a strong reference until each one is done.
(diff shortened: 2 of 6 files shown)


harrymove-ctrl:
Ran /simplify + a correctness pass on this diff:

- **Simplify:** \`with_memwal_openai\`/\`with_memwal_langchain\` each repeated the same 3-line "attach \`_memwal\`/\`memwal_flush\`/\`memwal_flush_sync\`" block. Extracted \`_expose_memwal_controls()\`.
- **Correctness:** \`with_memwal_langchain\` patches both async and sync generate paths onto the same \`llm\`, sharing one \`_PendingSaves\`. If a caller mixes them — an earlier async call populates \`self._tasks\`, then \`memwal_flush_sync()\` runs from sync code — \`asyncio.run()\` can't gather a Task bound to a different loop and raises \`RuntimeError\`, which is the worst place for an unhandled exception since flush is typically called during cleanup. Now logs a warning and keeps draining threads instead of raising.

Both pushed as follow-up commits. Full suites green (TS 17/17, Python 110/110 + 16 skipped pre-existing).

harrymove-ctrl:
Reverting the previous commit (\`37f742c7\`) — an earlier pass by a subagent I dispatched added it without my review, and independent verification (fresh clone, actual reproduction) found it doesn't do what it claims:

\`asyncio.gather(...)\` returns a \`_GatheringFuture\`, not a coroutine. \`asyncio.run()\` requires an actual coroutine and raises \`ValueError\` — not \`RuntimeError\` — for anything else. So \`asyncio.run(asyncio.gather(*list(self._tasks), ...))\` in \`flush_sync()\` has *always* raised \`ValueError\` whenever \`self._tasks\` is non-empty, for any reason (not just the cross-loop case the commit targeted), and the \`except RuntimeError:\` added never catches it. Verified by direct reproduction against the exact Python version this repo's CI uses. The bug pre-dates this PR; the "fix" was dead code.

Reverted rather than leaving broken/misleading code in place. A correct fix (wrap in an actual async function before passing to \`asyncio.run()\`, plus a regression test using a real cross-loop scenario) is a separate follow-up, not included here.

ducnmm (changes requested):
Found three blocking/runtime issues on the current head (`2c434c75`):

1. **High — LangChain wrapping fails on real Pydantic-backed models.** `packages/python-sdk-memwal/memwal/middleware.py:182` assigns `memwal_flush` as an arbitrary attribute. Supported LangChain models reject undeclared fields, so `with_memwal_langchain()` fails during wrapping. The tests use `MagicMock`, which masks this. Please reproduce with a real `FakeListChatModel` or equivalent and expose the helper without normal Pydantic field assignment.

2. **High — `memwal_flush_sync()` cannot drain pending async saves.** `middleware.py:167` passes the `Future` returned by `asyncio.gather()` to `asyncio.run()`, which requires a coroutine and raises when `_tasks` is non-empty. Those tasks are also bound to their original event loop, so a wrapper coroutine alone does not solve cross-loop draining. Please define and test the supported loop/lifecycle behavior.

3. **Medium — completed threads accumulate indefinitely.** `middleware.py:152-154` retains every completed background `Thread` until explicit flush. Long-lived clients that keep fire-and-forget behavior and never flush leak one thread object per auto-save. Add completion cleanup analogous to the task cleanup.

Verification: full Python suite passed (110 passed, 16 skipped), focused middleware tests passed, Ruff passed, TS SDK tests/build passed. I independently reproduced all three issues.

harrymove-ctrl:
Thanks for the thorough review, Henry — all 3 reproduced and fixed, TDD'd (failing test first, confirmed it failed for the reported reason, then fixed):

**1. LangChain/Pydantic crash** — confirmed with a real \`FakeListChatModel\`: \`obj.memwal_flush = ...\` raised \`ValueError: "FakeListChatModel" object has no field "memwal_flush"\`, exactly as you found. \`_expose_memwal_controls\` now uses \`object.__setattr__\` to bypass Pydantic's field validation while still landing in the instance's \`__dict__\` (so normal attribute reads still work). New test wraps a real model instead of \`MagicMock\`.

**2. \`flush_sync()\` cross-loop draining** — you're right that a wrapper coroutine alone doesn't fully solve it. Fixed the immediate bug (\`asyncio.gather()\` returns a Future, not a coroutine, so \`asyncio.run()\` was always raising \`ValueError\` regardless of loop — this predates the PR) by wrapping it in an actual async function. For the genuinely-cross-loop case (a Task really is bound to a different, e.g. still-running, event loop), that's caught and logged rather than raised — there's no way to synchronously await a task on another loop, so the defined behavior is: drain what you can, warn about what you can't, never crash out of cleanup code. New test reproduces a real cross-loop pending Task via a background thread running its own loop.

**3. Thread accumulation** — added \`_PendingSaves.spawn_thread()\`, which untracks itself on completion instead of only being cleared by \`flush()\`/\`flush_sync()\`. New test confirms a completed thread is gone from tracking without ever calling flush.

Full suite: 113 passed (+3 new regression tests), 16 skipped (pre-existing, live-infra only), ruff clean. Pushed as a new commit on top of the current head.

hungtranphamminh (commented):
Verified all three of @ducnmm's earlier issues against the current head (`9a5fed93`) — all resolved:

- **1 (Pydantic `__setattr__`)** — fixed. `_expose_memwal_controls` uses `object.__setattr__`, which writes into `__dict__` and is retrievable under Pydantic v1 & v2 (`extra="forbid"` / `__slots__` don't affect it — they only govern Pydantic's own setattr path). The new `FakeListChatModel` test exercises a real Pydantic model, not MagicMock. 👍
- **3 (thread leak)** — fixed. `spawn_thread`'s self-untrack is race-free (append-before-`start()` ordering; `_join_threads` swaps under lock and joins outside it → no deadlock), covered by `test_completed_threads_are_untracked_without_flushing`.
- The TS `flush()` is correct (no `pendingSaves` leak — same promise ref added/deleted; `.catch()` means no unhandledRejection), and `withMemWal-flush.test.mjs` genuinely proves it (asserts not-done before / done after `flush()` — fails against a no-op).

A few follow-ups, only one of which I'd hold merge for:

1. **Medium — the cancelled-save warning has no test.** `_warn_if_cancelled` (`packages/python-sdk-memwal/memwal/middleware.py`) is the "a cancelled save now logs a warning instead of vanishing" behavior from the CHANGELOG — arguably the headline silent-data-loss fix — but no test covers it (nothing cancels a tracked task and asserts the `logger.warning`; confirmed no `caplog`/`CancelledError` reference in `tests/test_middleware.py`). A refactor dropping the wrapper or swallowing the re-raise would pass the whole suite. Could we add a `caplog`-based test that cancels a pending tracked task and asserts the warning fires?

2. **Nit (docs) — `flush_sync()` doesn't drain cross-loop tasks; the name/CHANGELOG imply it does.** `flush_sync` (`middleware.py`) is honest in the *code* (it logs "may be lost"), but I reproduced on CPython 3.9 that `asyncio.run(gather(<task bound to another loop>))` raises immediately and drains nothing — so a caller doing `await smart.ainvoke(...)` then `smart.memwal_flush_sync()` at shutdown silently loses that save apart from the log line. The CHANGELOG lists `memwal_flush_sync()` as a drain path with no caveat. Suggest one line in the CHANGELOG/docstring: it drains threads + same-loop tasks, but **cannot** drain tasks created on a different event loop (mixed async→sync usage) — those are warned, not flushed. (Also minor: the `except RuntimeError` is broad enough to also catch "asyncio.run() cannot be called from a running event loop" and log it as "cross-loop"; and daemon fire-and-forget threads are killed at interpreter exit, so the safety net only helps if the caller explicitly flushes — both fine given the explicit-flush contract, just flagging.)

3. **Low — `_tasks` is read off-thread without the lock.** `middleware.py`: `_lock` guards `_threads` but not `_tasks` (`track_task` `.add`/`.discard` on the loop thread vs `list(self._tasks)` in `flush`/`flush_sync` from another thread). I checked — this is **safe on CPython** (`list(a_set)` is a single atomic C call under the GIL; it does not raise), so no live bug. It's an unsynchronized shared set, a portability/no-GIL nit — optionally a one-line comment on the CPython-atomicity assumption, or fold `_tasks` under the same `_lock` for uniformity. Non-blocking.

4. **Low (TS, non-blocking) — `flush()` is only reachable via `any`.** `withMemWal` returns `any` (from `wrapLanguageModel(...) as any`), so typed consumers get no `.flush()` autocomplete/type-safety. Pre-existing and additive, so not a regression — but a `withMemWal(...): LanguageModelV2 & { flush(): Promise<void> }` return annotation (the type is already imported) would make the new API real for TS users.

Verification: read the PR head blobs (not the dev diff); reproduced the cross-loop `asyncio.run(gather(...))` behavior and the `list(set)`-vs-`for`-over-set distinction on CPython 3.9; confirmed the real-`FakeListChatModel` and thread-cleanup tests are present and the cancelled-save path is untested. Core fix (TS + Python) is sound; 1 is the only item I'd want before merge.

ducnmm (approved):
Re-reviewed current head. The prior runtime findings are resolved, and the latest change only defers version/changelog metadata to the combined SDK release in #542. LGTM.

PR #535, opened 2026-08-04
docs(mcp): unified quickstart, Claude Code setup rewrite, Workers style pass

reemsabawi-mystenlabs:
## Description

The medium-priority Aug 7 docs batch, three tickets in one PR:

**BEDU-1012 — Unified MCP quickstart.** New `docs/mcp/quickstart.md`: one per-client setup table for Claude Code, Claude Desktop, Cursor, Codex, OpenCode, and Antigravity, with the three shared config-block shapes (JSON `mcpServers`, Codex TOML, OpenCode JSON), namespace guidance, and a three-step verification. Config paths and snippets are lifted verbatim from the existing per-client pages, which stay as the detail layer. Added to `docs.json` after the MCP overview. This targets the fragmentation signal (22 Kapa questions from the claude-code page alone).

**BEDU-854 — Claude Code page rewrite.** Prerequisites now cover the Node/`npx` check, plugin-support detection, and where credentials live; verification is a three-layer sequence (server connected → relayer reachable → end-to-end recall); and a new troubleshooting FAQ addresses the common failure modes, each grounded in the MCP package source: failed-server diagnosis (`--help` in a terminal, `MEMWAL_MCP_DEBUG=1`), headless/SSH sign-in via copying `~/.memwal/credentials.json`, account switching with `memwal_logout`, namespace-scoped recall misses, `memwal_restore`, hooks-only-with-plugin, credential expiry after delegate-key revocation, and proxy allowlisting.

**BEDU-845 — Cloudflare Workers page.** The page already existed and is substantively complete; this is a style and accuracy pass: Seal casing, the "this guide" preamble, and the manual Next steps footer folded into a See also block per house style.

## Code example sourcing (review follow-up)

Per review discussion about importing examples from existing code: Mintlify cannot import code from repository files at build time, so every snippet in `docs/` is necessarily a copy. This PR makes that relationship enforced rather than implicit, with a single check that covers the **whole docs tree** rather than one per section.

**`scripts/check-docs-code-sync.mjs`** derives canonical facts from the workspace and validates every `docs/**/*.md` page against them:

| Check | Canonical source |
| --- | --- |
| Import and `require` specifiers in the `@mysten-incubation` scope name a package that exists, and any subpath is a declared export of it | each `packages/*/package.json` (`name` and `exports`) |
| `npm install` / `npx` commands in the same scope name a package that exists | the same `package.json` files |
| MCP client config blocks (JSON `mcpServers` / `mcp`, Codex TOML `[mcp_servers.memwal]`) run the canonical command with the canonical arguments | `packages/mcp/README.md` |

Extra arguments such as `--namespace` are allowed, and URL-based Streamable HTTP entries are exempt as a different transport. Current run: **85 docs pages scanned — 32 import specifiers, 7 install commands, 9 MCP config blocks, all matching** across the three workspace packages.

A **`docs-code-sync`** job ("Docs / Code references match packages") in `test.yml` runs it on every PR, mirroring the existing compatibility-contract job. Failures exit non-zero so drift is fixed in the same PR rather than tracked as follow-up work. The quickstart cites `packages/mcp` as the canonical source.

The Cloudflare Workers snippets have no existing source to import: no app in this repository runs on Workers. They use only the SDK's public API (`MemWal.create`, `health`) and Cloudflare's documented `nodejs_compat` flag.

## Verification

- Style audit clean on all three pages (mechanical and voice gates passed on push).
- `docs.json` validates as JSON with the new entry; all internal links target existing pages.
- Troubleshooting claims checked against `packages/mcp/src/index.ts` (env vars, credentials path, `--logout` behavior, debug flag).
- `check-docs-code-sync.mjs` passes on the current tree; verified it fails with file, line, and reason output when drift is injected into any of the three block shapes; workflow YAML validates.


scripts/check-mcp-docs-sync.mjs
```diff
@@ -0,0 +1,145 @@
...
+ *     mcpServers, parsed from the `command` and `args` lines.
+ *
+ * Run: node scripts/check-mcp-docs-sync.mjs
+ */
+import { readFileSync, readdirSync } from "node:fs";
+import path from "node:path";
+import process from "node:process";
+
+const README = "packages/mcp/README.md";
+const DOCS_DIR = "docs/mcp";
```
jessiemongeon1: What if pages in other sections include code snippets for this section, i.e., a quickstart or other general page outside of this dir? if we're implementing this sort of script check, it should be as complete as possible

for all code samples, could this snippet be repurposed and generalized, so it checks packages and then all docs against it? ideally, we shouldn't have to create individual scripts like this for every section to validate code and prevent drift

scripts/check-mcp-docs-sync.mjs
```diff
@@ -0,0 +1,145 @@
...
+            checkEntry(file, block.line, "mcp_servers.memwal", command, args);
+        }
+    }
+}
+
+if (failures.length > 0) {
+    console.error(`\n${failures.length} MCP snippet(s) drifted from ${README}:\n`);
+    for (const failure of failures) console.error(`  - ${failure}`);
+    console.error(
+        `\nUpdate the docs block(s) to match the package README, or update the README first if the package interface changed.`,
```
jessiemongeon1: does this just log the error, or is there a follow up we need to track to actually address this?

docs/mcp/quickstart.md
```diff
@@ -0,0 +1,115 @@
...
+  opencode.json, and Antigravity accepts the plugin or a JSON block. After adding the server,
+  restart the client and ask the agent to run memwal_login.
+---
+
+Every supported client runs the same local server, `npx -y @mysten-incubation/memwal-mcp`, and differs only in where the configuration lives. Pick your client below, add the server, restart, and sign in.
+
+See also:
+
+- [Overview](/mcp/overview): how the local server bridges tool calls to the relayer
+- [Reference](/mcp/reference): every tool with full parameters and transports
```
jessiemongeon1:
```suggestion
```

docs/mcp/quickstart.md
```diff
@@ -0,0 +1,115 @@
...
+{
+  "mcp": {
+    "memwal": {
+      "type": "local",
+      "command": ["npx", "-y", "@mysten-incubation/memwal-mcp"],
+      "enabled": true
+    }
+  }
+}
+```
```
jessiemongeon1: this section should use tabs so you can see all the examples side by side

docs/sdk/cloudflare-workers.md
```diff
@@ -31,7 +31,13 @@ answer: >-
...
 ---
 
-MemWal runs on the [Cloudflare Workers](https://developers.cloudflare.com/workers/) runtime, but the edge environment differs from Node.js in a few ways that affect bundling and reliability. This guide covers the configuration and patterns that make it work cleanly.
+MemWal runs on the [Cloudflare Workers](https://developers.cloudflare.com/workers/) runtime, but the edge environment differs from Node.js in a few ways that affect bundling and reliability. The configuration and patterns below make it work cleanly.
+
+See also:
+
+- [Walrus Memory client](/sdk/usage/memwal): the default relayer-backed client used below
+- [Public relayer](/relayer/public-relayer): managed Mainnet and Testnet relayer endpoints
+- [API Reference](/sdk/api-reference): full method signatures and config fields
```
jessiemongeon1:
```suggestion
```

jessiemongeon1 (changes requested):

jessiemongeon1:
## Style Guide Audit

scripts/check-docs-code-sync.mjs
```diff
@@ -0,0 +1,224 @@
+#!/usr/bin/env node
+/**
+ * Checks code in the docs against the packages it documents.
+ *
+ * Mintlify cannot import code from repository files at build time, so every
```
jessiemongeon1, line 5: I'm not sure I understand this script, therefore I don't know if this is the best solution for what is asked for here

docs/mcp/quickstart.md
```diff
@@ -0,0 +1,118 @@
+---
+title: MCP Quick Start
```
jessiemongeon1, line 2: engineering needs to review this

scripts/check-docs-code-sync.mjs
jessiemongeon1: IMO we should reach out to Production Engineering and ask their opinion on this script/CI job and if there is a better workflow

jessiemongeon1 (changes requested):

ducnmm (changes requested):
Engineering review:

1. The login behavior is described inaccurately in both the quickstart and Claude Code prerequisites. An unauthenticated memory-tool call does not open the browser or trigger `memwal_login`; `auth-required.ts` returns `LOGIN_INSTRUCTION` and the client/agent must then call `memwal_login` explicitly. Please change this to “ask the agent to run `memwal_login`” (or equivalent).

2. `check-docs-code-sync.mjs` treats the MCP README block as canonical without validating that block against `packages/mcp/package.json`. If the package is renamed, or the README command drifts, README + all docs can remain mutually consistent but wrong and CI still passes. Please derive the package argument from the MCP manifest (or at minimum validate the extracted README command/args against that manifest) before using it as the docs baseline.

Non-blocking: the headless guidance copies a file containing the raw delegate private key. It should explicitly require a secure transfer method and `0600` permissions on the destination.

ducnmm (approved):
Re-reviewed the latest revision. The login flow now correctly requires an explicit `memwal_login` call, the docs-sync baseline is anchored to the MCP package manifest, and the headless credential-copy guidance now calls out secure transfer and `0600` permissions. I also ran `node scripts/check-docs-code-sync.mjs` successfully (85 pages, 9 MCP config blocks). No remaining engineering blockers.

scripts/check-docs-code-sync.mjs
jessiemongeon1: still needs production engineering to advise

jessiemongeon1 (changes requested):

docs/mcp/quickstart.md
ebmifa: No objection to the CI job — a few seconds, no secrets, and it follows
the pattern ducnmm already established with check-compatibility-contract.

Two bugs to fix before merge. Both make it pass when it shouldn't:
1. It silently skips any config block using ```jsonc, containing a
   `// comment`, or using 4-backtick fences. I pasted in a deliberately
   wrong block and it reported success.
2. Nothing asserts it still finds anything. If a docs change breaks its
   search, it checks zero blocks and stays green indefinitely.

On "is this the best approach" — the script polices 9 copies across 7
pages. The better lever is fewer copies. Native Mintlify snippets would
collapse the JSON blocks (3 real variants: bare, +env, +--namespace)
into one parameterized snippet. Worth doing here if the page conversion
to .mdx is small; split it out if not.

Later option: a small generator emitting those snippets from
packages/mcp, committed, with CI checking the diff is clean. That works
fine with hosted Mintlify and would likely be shorter than the current
238-line validator.

Two things I'd not do: port Sui's docs injection system (custom
Docusaurus infra sized for 800+ usages — wrong scale here), and treat
the existing MCP integration job as a config smoke test. It isn't one;
a real one tests the published npx invocation, so it belongs on a
nightly rather than every PR.

jessiemongeon1:
## Style Guide Audit

jessiemongeon1 (approved):

PR #538, opened 2026-08-05
fix(server): signal restore() truncation, including sidecar candidate-cap hits (WALM-317, WALM-319)

harrymove-ctrl:
## Summary
`RestoreRequest.limit` defaults to 10, and `restore()` silently sliced the on-chain missing-blob list to that limit with no signal to the caller that a namespace was only partially recovered. Same failure class as `recall()`'s known `limit=10` truncation (WALM-119/170), but never addressed for `restore()` -- and worse there, since restore() is the recovery path: a silent partial result compounds whatever data-loss scenario the caller was already trying to fix.

Adds `RestoreResponse.truncated: bool`, computed from `all_missing.len() > limit` before that list is sliced to `limit`. Purely additive -- no existing field's meaning changes, no existing caller ignoring the new field breaks.

Fixes WALM-317.

## Test plan
- [x] TDD: added a test referencing the new field first, confirmed it fails to *compile* (`E0609: no field 'truncated'`) -- the Rust-appropriate form of RED -- then implemented to green.
- [x] `cargo test --bin memwal-server`: 414 passed, 0 failed, 30 ignored (pre-existing, live-infra integration tests).
- [x] `cargo fmt --check` clean.
- [x] Verified all 4 `RestoreResponse` construction sites in `restore()` (empty-chain, empty-missing, empty-download, success) are wired to the new field.

`cd services/server && cargo test restore` -- the two new tests (`restore_truncated_flag_matches_missing_vs_limit`, `restore_response_carries_truncated_field`) pin the exact predicate and response shape.

## Risks and dependencies
None. Additive field only. SDK/dashboard changes to actually surface `truncated` to end users are a separate follow-up, not required for this fix to be safe to merge.

### DIFF
--- a/docs/relayer/api-reference.md
+++ b/docs/relayer/api-reference.md
@@ -352,6 +352,9 @@ Rebuild missing vector entries for one namespace. Queries onchain blobs by owner
   "skipped": 7,
   "total": 10,
   "namespace": "demo",
-  "owner": "0x..."
+  "owner": "0x...",
+  "truncated": false
 }
 ```
+
+`truncated` is `true` when this restore is known-incomplete: either more on-chain blobs were missing locally than `limit` allowed this call to restore, or the sidecar's raw on-chain candidate fetch (bounded per owner, shared across all of the owner's namespaces, hard-capped independent of `limit`) hit its own cap before this namespace's blobs were even filtered out of that set. The second case can produce `truncated: true` even when `total` is `0` for this namespace, since a cap hit elsewhere can starve this namespace's fetch entirely. Raising `limit` only helps with the first case — past the sidecar's cap, only a cursor/pagination-based restore would.
--- a/packages/python-sdk-memwal/memwal/client.py
+++ b/packages/python-sdk-memwal/memwal/client.py
@@ -794,16 +794,20 @@ class MemWal:
                 hierarchy semantics (see SKILL.md "Namespace Semantics").
             limit: Max blobs the relayer will *inspect* this call (default:
                 10, matches the server-side default and the TypeScript SDK).
-                The relayer fetches blobs newest-first and caps the work at
-                this number; it does **not** cap ``restored`` separately.
+                The relayer selects up to this many missing blobs from the
+                candidates it sees, in unspecified order; it does **not** cap
+                ``restored`` separately.
 
         Returns:
             :class:`RestoreResult`.
 
         Notes:
-            * **No pagination cursor** — restore is single-shot. To rebuild a
-              large namespace, call repeatedly with a growing ``limit`` or
-              prune the local index first.
+            * **No pagination cursor** — restore is single-shot. A larger
+              ``limit`` may help when ``truncated`` is caused by the request
+              limit, but candidate discovery also has a per-owner source cap
+              shared across namespaces. Hitting that cap requires cursor/
+              pagination support for guaranteed full recovery; increasing
+              ``limit`` or retrying cannot guarantee it.
             * **Performance** scales linearly in ``limit``: up to 10 Walrus
               downloads in parallel, then 3 SEAL decrypts in parallel, then
               embeddings. Expect seconds-per-blob on cold caches.
--- a/packages/python-sdk-memwal/memwal/client.py
+++ b/packages/python-sdk-memwal/memwal/client.py
@@ -784,6 +784,10 @@ class MemWal:
           do **not** count as either restored or skipped.
         * ``total`` — count of on-chain blobs the relayer saw for
           ``(owner, namespace)`` before the limit was applied.
+        * ``truncated`` — True when this restore is known-incomplete (limit
+          truncation or an upstream on-chain candidate-fetch cap; see
+          :class:`~memwal.types.RestoreResult`). Defaults to ``False`` when
+          talking to a relayer older than WALM-319 that omits the field.
 
         Args:
             namespace: Namespace to restore. Exact match — no prefix or
@@ -814,6 +818,10 @@ class MemWal:
             total=data["total"],
             namespace=data["namespace"],
             owner=data["owner"],
+            # Relayers older than WALM-319 omit `truncated` entirely --
+            # treat "not present" as "not known to be truncated" rather
+            # than require every relayer version to send it.
+            truncated=data.get("truncated", False),
         )
 
     async def health(self) -> HealthResult:
--- a/packages/python-sdk-memwal/memwal/types.py
+++ b/packages/python-sdk-memwal/memwal/types.py
@@ -201,6 +201,17 @@ class RestoreResult:
     total: int
     namespace: str
     owner: str
+    #: True when this restore is known-incomplete: either more on-chain
+    #: blobs were missing locally than ``limit`` allowed this call to
+    #: restore, or the sidecar's raw on-chain candidate fetch hit its own
+    #: cap before this namespace's blobs were even filtered out of that
+    #: set (WALM-319) -- this can be ``True`` even when ``total == 0``,
+    #: since a cap hit elsewhere can starve this namespace's fetch
+    #: entirely. Raising ``limit`` only helps with the first case.
+    #:
+    #: Relayers older than WALM-319 don't send this field at all; the SDK
+    #: defaults it to ``False`` in that case rather than requiring it.
+    truncated: bool = False
 
 
 @dataclass
--- a/packages/python-sdk-memwal/tests/test_client.py
+++ b/packages/python-sdk-memwal/tests/test_client.py
@@ -630,6 +630,7 @@ class TestRestore:
                     "total": 7,
                     "namespace": "my-app",
                     "owner": "0xowner",
+                    "truncated": False,
                 },
             )
         )
@@ -641,6 +642,57 @@ class TestRestore:
         assert body["limit"] == 100
         assert result.restored == 5
         assert result.skipped == 2
+        assert result.truncated is False
+
+    @respx.mock
+    async def test_restore_preserves_truncated_true(
+        self, memwal_client: MemWal
+    ) -> None:
+        """truncated=True from the relayer must survive into RestoreResult
+        instead of being dropped when the dataclass is reconstructed."""
+        mock_seal_session_prereqs()
+        respx.post(f"{_TEST_SERVER}/api/restore").mock(
+            return_value=httpx.Response(
+                200,
+                json={
+                    "restored": 5,
+                    "skipped": 2,
+                    "total": 7,
+                    "namespace": "my-app",
+                    "owner": "0xowner",
+                    "truncated": True,
+                },
+            )
+        )
+
+        result = await memwal_client.restore("my-app", limit=100)
+
+        assert result.truncated is True
+
+    @respx.mock
+    async def test_restore_truncated_defaults_false_when_omitted(
+        self, memwal_client: MemWal
+    ) -> None:
+        """Relayers older than WALM-319 omit `truncated` entirely — the
+        SDK must default it to False rather than requiring the field."""
+        mock_seal_session_prereqs()
+        respx.post(f"{_TEST_SERVER}/api/restore").mock(
+            return_value=httpx.Response(
+                200,
+                json={
+                    "restored": 5,
+                    "skipped": 2,
+                    "total": 7,
+                    "namespace": "my-app",
+                    "owner": "0xowner",
+                    # no "truncated" key — simulates an older relayer
+                },
+            )
+        )
+
+        result = await memwal_client.restore("my-app", limit=100)
+
+        assert result.truncated is False
 
 
 class TestHealth:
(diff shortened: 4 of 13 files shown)


hungtranphamminh:
Thanks for the quick turnaround. I traced the full `restore()` path plus the sidecar `query-blobs` / `listBlobObjectsGrpc` before commenting, so a few of these go beyond the diff. **Net: this is a real improvement over the silent drop and it's cleanly additive — but as written the signal only catches part of the truncation, and the tests don't actually exercise the fix.** Details so you can decide what to fold in here vs. a follow-up.

## 1. The flag is correctly wired for the limit-slice case ✅

`truncated = all_missing.len() > limit` is computed before `.take(limit)` consumes `all_missing`, and all four `RestoreResponse` sites set it (the new non-defaulted field makes that compiler-guaranteed). I checked whether it could be dead-`false` — it can't: `query-blobs` doesn't slice its result to `limit`, it returns all namespace/package-filtered matches (`res.json({ blobs, total: blobs.length })`), so `all_missing` can exceed `limit` and the flag does fire. Good.

## 2. But the real truncation is upstream — a cross-namespace ~100 candidate cap 🟠 (main issue)

The candidate bound in `query-blobs` is applied at the **fetch** stage, before namespace/package filtering:

```
desiredMatches = min(limit, 500)
cap            = min(desiredMatches * 5, 100)          // when no explicit blobIds
rawObjs        = listBlobObjectsGrpc(owner, BLOB_TYPE, cap)   // stops paginating at `cap` RAW blobs
// …only THEN filter rawObjs by namespace + packageId
```

Two consequences worth calling out:

- **The cap bounds raw blobs across ALL namespaces, not per-namespace.** For `limit=10` it fetches ≤50 raw blobs total, then keeps the ones matching the requested namespace. So if the owner has memories spread across several namespaces, the target namespace can be silently truncated even when it holds well under 50 blobs — the raw-fetch quota is consumed by other namespaces first. `truncated` (computed post-filter, against `limit`) won't reflect that.
- **Beyond ~100 raw blobs the endpoint can't fully recover regardless of `limit`** (cap is hard-capped at 100). So for a large owner, `truncated=true` tells the caller "raise `limit` and retry", but raising it only grows the cap to `min(limit*5, 100)` → still 100. And once inside the cap, `truncated` reports `false` while blobs are still being dropped.

So `truncated=false` is **not** a "fully recovered" guarantee, and the implied remedy ("raise limit / call again") doesn't hold past the cap. Options: (a) also surface when the fetch hit the candidate cap (so large/multi-namespace owners are flagged too), or (b) at minimum tighten the field's doc comment so it doesn't imply retry-restores-everything — with a note that true full recovery needs a cursor/pagination path (the bigger separate work you already scoped).

## 3. Ordering assumption in the surrounding code looks wrong 🟠

`restore()` has this comment right where the slice happens:

> "query-blobs returns newest-first for restore's recent transaction path, so keep the first N missing blobs."

But the path actually used here is `listBlobObjectsGrpc`, whose own doc comment says the opposite:

> "Unlike the tx scan the order is **unspecified rather than newest-first**, so `cap` bounds work, not recency."

So `all_missing.into_iter().take(limit)` isn't "the N most recent missing blobs" — it's an **arbitrary N** in unspecified order. That's independent of this PR, but it compounds Finding 2: a truncated restore doesn't even deterministically return the newest memories, so `truncated=true` + "retry with higher limit" can keep returning an overlapping/arbitrary subset. Worth at least correcting the misleading comment; ideally the slice should be over a defined ordering.

## 4. `total` under-reports for large owners 🟡 (pre-existing)

`total` = blobs `query-blobs` returned (bounded by the candidate cap), not the true on-chain total. So a caller can't fall back on `total` to detect they're past the cap either. Not introduced here, but it's why `truncated` alone can't be a complete signal — worth a line in the follow-up.

## 5. The tests re-derive the predicate rather than exercising `restore()` 🔴 (please address)

```rust
let truncated = missing_len > limit;      // same expression as production
assert_eq!(truncated, expected_truncated);
```

This asserts `(a > b) == <hand-picked value>` — it re-implements the production predicate inside the test and checks it against a table. It never calls `restore()`, never goes through `query_blobs_by_owner`, and never checks the flag is wired to the right variable at each response site. Concretely: **if one of the four `RestoreResponse { … }` sites set `truncated: false`, or compared against `total` instead of `limit`, these tests would still pass.** (`restore_response_carries_truncated_field` is a struct-construction smoke test only.)

For a fix whose whole point is "stop silently dropping," I'd like a test that drives the path and proves the flag flips on truncated input — e.g. factor "compute `truncated` + slice" into a small pure helper and unit-test it with `all_missing > limit` and `≤ limit`, or stub `query_blobs_by_owner` to return N > limit and assert `restore()`'s response has `truncated=true`. As written the tests wouldn't catch a wiring regression.

## 6. Minor: ticket id in a code comment

The inline `// WALM-317:` in `admin.rs` — our convention is to keep Linear IDs out of code comments (describe the bug, not the ticket). Non-blocking.

---

**Overall:** glad the silent drop is getting a signal, and the mechanical change is clean and additive. The two I'd treat as worth resolving (here or a fast follow-up) are **Finding 2** (the cap — the main truncation is cross-namespace and >cap, which this flag doesn't see, so `truncated=false` can still be a silent partial restore) and **Finding 5** (a test that actually exercises the fix). **Finding 3** is a real but separate correctness/comment issue; **Findings 4/6** are notes. If you'd rather land this as-is and track Findings 2/3/5 separately, that's fine — just want the "`truncated=false` isn't safe" gap captured so it isn't forgotten.

harrymove-ctrl:
Thanks for the deep trace — this is right, and I'd missed the connection despite already knowing about the sidecar's raw-candidate cap (I'd used it to justify *not* raising the default limit, but didn't check whether `truncated` itself could see past it — it can't). Went through each point:

**1. ✅** Agreed, confirmed correct as-is.

**2 (main issue) + 4 (total under-reports):** Agreed, and this is more than a doc nit — `truncated=false` on the disaster-recovery path shouldn't be able to lie. Filed as a proper follow-up rather than folding into this PR, since fixing it for real means threading a cap-hit signal from the sidecar's `query-blobs` through `query_blobs_by_owner` into `RestoreResponse` — bigger than this ticket's scope: [WALM-319](https://linear.app/mysten-labs/issue/WALM-319/restorequery-blobs-candidate-cap-can-silently-under-restore-a). In the meantime, tightened `truncated`'s doc comment on this PR so it doesn't imply a false-recovered guarantee.

**3.** Fixed — corrected the comment to match `listBlobObjectsGrpc`'s actual documented order (unspecified, not newest-first).

**5.** Fair, and a real gap — the old tests re-derived the same expression instead of exercising anything `restore()` calls. Extracted `paginate_missing_blobs()` so the slice + flag are computed (and now tested) as one unit that `restore()` actually calls, rather than a hand-copied predicate. Didn't reach for a full `restore()`-level test (would need mocking the sidecar HTTP call, DB, Walrus aggregators, and SEAL — no existing test infra in this crate for that, and disproportionate for a Low-priority fix) — this closes most of the gap without that lift, though it's fair to note it still wouldn't catch a future copy-paste bug at one of the 4 `RestoreResponse` call sites specifically.

**6.** Fixed — dropped the ticket ID from the code comment.

Pushed as a follow-up commit. Given WALM-319 stays open, `truncated=false` is documented as non-authoritative rather than presented as a guarantee — figured that's an acceptable state to merge this in (a real improvement over the prior fully-silent drop), but happy to hold if you'd rather WALM-319 land first.

harrymove-ctrl:
Pushed a third commit that fully addresses Finding 2 (the main issue) instead of just documenting around it — folding WALM-319 into this PR rather than leaving it as a separate follow-up, since it depends on \`truncated\` already existing here.

**What changed:** the sidecar's \`query-blobs\` now reports \`sourceCapped\` (true when \`listBlobObjectsGrpc\`'s raw-candidate fetch hit its cap, detected via a new testable \`isSourceCapped()\`). That's threaded through \`query_blobs_by_owner\` (now returns \`(blobs, source_capped)\`) and OR'd into \`RestoreResponse.truncated\` at every response site in \`restore()\` — including the \`total == 0\` case, since a cap hit fetching *other* namespaces can starve this one's fetch before it's even filtered out.

\`truncated\` should no longer be able to read \`false\` while a namespace is silently under-restored for this reason. It's still not a full fix for large/multi-namespace owners past the ~100 cap (that needs cursor/pagination — separate, bigger work, not attempted here), but the signal itself is honest now rather than caveated.

Tests: 2 new (sidecar `isSourceCapped`, Rust `QueryBlobsResponse` deserialization incl. default-when-absent for rolling-deploy safety) + full suites green (Rust 416/416, sidecar 26/27 with the 1 failure being the pre-existing unrelated pinned-SDK issue).

Linear WALM-319 (filed for this) closed as folded into this PR rather than left as a separate ticket.

harrymove-ctrl:
Ran /simplify + a correctness pass on the full diff (including the sourceCapped commit). Nothing further to flag — the OR-ing logic, the total==0 edge case, and the #[serde(default)] rolling-deploy compatibility are all correct as implemented, and there's no duplicated logic worth extracting beyond what's already factored into `paginate_missing_blobs`/`isSourceCapped`.

ducnmm (changes requested):
Found one blocking API issue and one truncation edge case on the current head (`5dd17573`):

1. **High — official SDK callers cannot consume the new `truncated` signal.** TypeScript `RestoreResult` omits it (`packages/sdk/src/types.ts:299-305`). Python also omits it and reconstructs the response without the field (`packages/python-sdk-memwal/memwal/types.py:195-203`, `packages/python-sdk-memwal/memwal/client.py:811-817`), permanently discarding the server value. The API reference remains unchanged (`docs/relayer/api-reference.md:347-356`). Please preserve the additive field in both SDKs while tolerating older relayers that omit it, and document it.

2. **Medium — exact-cap results are reported as definitely truncated.** `services/server/scripts/sidecar/routes/walrus-query.ts:129-137` treats `fetchedCount >= cap` as capped without checking whether the final page had `hasNextPage=false`. Exactly 50 candidates at the default cap therefore produce `sourceCapped=true` and ultimately `truncated=true` even when discovery completed. Fetching one extra candidate or propagating page completion would distinguish true truncation.

Verification: complete diff and related SDK/restore paths reviewed; GitHub checks and diff check pass; focused Rust/Python tests pass. No route-level test currently covers sidecar pagination through `sourceCapped` into `/api/restore`.

harrymove-ctrl:
Both findings fixed, TDD'd (failing test first for each):

**1. Both SDKs dropped \`truncated\`** — TS: \`RestoreResult\` now declares the field; both \`MemWal.restore()\` and \`MemWalManual.restore()\` (they each have their own implementation, both had the bug) normalize the response with \`truncated: result.truncated ?? false\` so older relayers still get a defined value. Python: the dataclass now declares \`truncated: bool = False\`, and \`client.py\` passes it through via \`data.get("truncated", False)\` instead of dropping it during reconstruction. Documented in \`docs/relayer/api-reference.md\`.

**2. Exact-cap miscounted as truncated** — implemented your suggested fix directly: the sidecar now probes with \`cap + 1\` instead of \`cap\`, slices back to \`cap\` for actual use, and \`isSourceCapped\` compares the unsliced probe count with \`>\` instead of \`>=\` — so hitting precisely \`cap\` blobs (discovery actually complete) no longer reads as truncated.

Tests: TS SDK 20/20 (4 new), Python SDK 111 passed + 16 pre-existing skips (2 new), sidecar 26/27 (1 pre-existing unrelated pinned-SDK failure, same as before this PR). Lint clean.

hungtranphamminh (approved):
Re-reviewed after the revisions — this is a big step up, thanks for going past documenting the cap and actually fixing it. Confirmed against source:

- WALM-319 `sourceCapped` probe (`cap + 1` then slice, `fetched > cap`) is correct at the boundaries — exact-cap reads as not-truncated (the case Henry flagged), `cap + 1`/over reads as capped. ORed into `truncated` including the `total == 0` early return, so a cap hit that starves this namespace is now visible. This closes the main gap from the last pass.
- Tests now call `paginate_missing_blobs` (the real function) rather than re-deriving the predicate, so a wiring bug actually fails the test — plus `isSourceCapped` boundary tests and `sourceCapped` serde round-trips.
- Both SDKs (TS `MemWal`/`MemWalManual` + Python) now carry `truncated` through and default it to false for older relayers; rolling-deploy skew (an older sidecar/relayer omitting the field) parses safely via `#[serde(default)]` / `?? false`. Verified the one `query_blobs_by_owner` caller handles the new tuple, and the `includeStorageLease` path correctly omits `sourceCapped` since that path can't be capped.

LGTM — approving.

ducnmm (approved):
Re-reviewed current head after the SDK restore documentation correction. Prior findings are resolved, the source-cap guidance now matches the implementation, targeted TypeScript/Python tests pass, and all CI checks are green. LGTM.

Issue #401, opened 2026-07-12
Quick Start example is missing values for required fields (key, accountId, serverUrl)


shmalik06:
**What I did**
Followed the Quick Start snippet in the README to understand how to initialize MemWal:

const memwal = MemWal.create({
  key: "your-delegate-key-hex",
  accountId: "your-walrus-memory-account-id",
  serverUrl: "https://your-relayer-url.com",
  namespace: "demo",
});

**What I expected**
A pointer (link or short note) explaining where a new user is supposed to get
`key`, `accountId`, and `serverUrl` from — e.g. a dashboard, CLI command, or
account creation step.

**What I got**
The Quick Start section jumps straight into code with placeholder values and
no link to where these values actually come from. For someone new to the
project, it's not obvious whether this requires a separate signup step, a CLI
tool, or manual account creation on Sui/Walrus.

**Suggestion**
Add a one-line link in the Quick Start section to the account/setup docs
(e.g. "SDK API Reference" or a dedicated "Getting your credentials" page)
so the snippet is runnable end-to-end without guessing.

PR #542, opened 2026-08-05
docs(WALM-318): make health() credential-blind behavior explicit

harrymove-ctrl:
## Summary
WALM-318 bundled 4 small SDK/relayer DX gaps from an io.net integration report. This PR resolves what's safely fixable; the rest either doesn't reproduce or is already resolved elsewhere.

- **Item 1 (stale Python `ENV_PRESETS`):** doesn't reproduce — Python's presets already point at the correct `*.memory.walrus.xyz` hosts. No change.
- **Item 2 (unauthenticated `health()`):** confirmed, by design. Made the implication explicit in both SDK API references and the Rust handler doc comment — a passing `health()` only means the relayer is reachable, not that credentials are valid.
- **Item 3 (401 doesn't name the mismatched network):** confirmed as a real DX gap, but **deliberately not implementing the originally-suggested fix** (server detects and names the account's actual network). `verify_signature()` routes all rejections through `constant_time_reject` specifically so a caller can't distinguish "account doesn't exist here" from "wrong key" via timing — revealing the account's real network in the response body would leak that same distinction through content instead, turning the endpoint into an account-existence-per-network oracle for anyone probing account IDs. Instead, both SDKs' `AUTH_REJECTED` message (which already names "staging/mainnet mismatch" as a likely cause) now links directly to `docs/troubleshooting/overview.md`'s existing network-mismatch triage table — safe (client-side only), no new server surface.
- **Item 4 (Python middleware doesn't expose `MemWal` for `close()`):** already resolved by #534 (WALM-306), which added `client._memwal`/`llm._memwal`. No separate change needed here.

## Testing
- Items 1/2/4: docs + a doc-comment only — no behavior change. `cargo build --bin memwal-server` clean.
- Item 3: TDD — new tests written first, confirmed RED (message lacked the docs URL), then GREEN.
  - TS: `npm test` in `packages/sdk` — 17/17 pass (new: `sanitize-server-error.test.mjs`).
  - Python: `pytest tests/` in `packages/python-sdk-memwal` — 110 passed, 16 skipped (pre-existing live-infra), 0 regressions (new: `test_auth_rejected_message.py`). `ruff check` clean.

## Risks and dependencies
None. No server behavior changes, no new server code path. Depends on #534 having landed for item 4's resolution to be complete (referenced, not duplicated, here).

### DIFF
--- a/docs/python-sdk/api-reference.md
+++ b/docs/python-sdk/api-reference.md
@@ -183,7 +183,7 @@ RestoreResult(restored: int, skipped: int, total: int, namespace: str, owner: st
 
 ### `health() -> HealthResult`
 
-Check relayer health. No authentication. Raises `MemWalError` on non-200.
+Check relayer health. No authentication — a successful response confirms the relayer is reachable, not that your `key`/`account_id` are valid. A signed call (e.g. `remember()`, `recall()`) can still fail with `401` immediately after a passing `health()`. Raises `MemWalError` on non-200.
 
 ```python
 HealthResult(
--- a/docs/sdk/api-reference.md
+++ b/docs/sdk/api-reference.md
@@ -172,7 +172,7 @@ Rebuild missing indexed entries for one namespace from Walrus. Incremental — o
 
 ### `health(): Promise<HealthResult>`
 
-Check relayer health. Does not require authentication.
+Check relayer health. Does not require authentication — a successful response confirms the relayer is reachable, not that your `key`/`accountId` are valid. A signed call (e.g. `remember()`, `recall()`) can still fail with `401` immediately after a passing `health()`.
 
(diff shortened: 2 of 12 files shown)


hungtranphamminh (approved):
LGTM. Verified: item 3 correctly keeps the network distinction out of the server response — the diff doesn't touch `verify_signature`/`constant_time_reject`, so the constant-time-reject oracle protection is intact, and the enrichment is purely the client-side `AUTH_REJECTED` message. The linked troubleshooting anchor (`#401-auth_rejected-errors`) matches the heading, and both new tests assert the URL is present. Item 2's health() clarification is accurate (it's genuinely public/unauthenticated).

One small note: item 4 says "already resolved by #534" — #534 is still open/unmerged, so item 4 is pending that landing rather than done today. Doesn't affect this PR (nothing here touches `_memwal`), just worth wording it as a dependency. Approving.

PR #546, opened 2026-08-06
WALM-297: Owner-scoped bearer token auth for the read API (Phase 1)

harrymove-ctrl:
## Summary

Console can never call WALM-295's owner-scoped read API (`GET /v1/owners/{owner}/{namespaces,memories,agents}`) under the existing Ed25519 signed-request scheme, because it structurally never holds a delegate key — WALM-298's identity-link flow proves control of an owner address entirely on Console's own side and never grants signing capability.

This adds the missing bridge: after Console has proven control of owner `Y` itself, it authenticates as a trusted client via a single shared **service credential** (team decision — service credential over mTLS/signed-client-assertion, for lower implementation cost on both sides) and mints a short-lived, owner-scoped bearer token limited to `memories.read`.

Full contract: `docs/api/owner-token-auth.md`.

## What's here

- `owner_token_auth.rs` — HMAC-signed opaque bearer token mint/verify, mirroring the existing `security_delete_auth.rs` token scheme; a `FromRequestParts` extractor for token-gated routes.
- `routes/owner_token.rs` — `POST /v1/owner-tokens` (issuance) + `GET /v1/owners/{owner}/_token_probe`, a minimal token-gated example route standing in for WALM-295's real handlers, which don't exist on this branch yet (fresh branch off `dev`). Wiring the `OwnerToken` extractor into the real `namespaces`/`memories`/`agents` handlers is the small follow-up once this branch merges with WALM-295.
- `rate_limit.rs` — three independent budgets on issuance: per-IP (throttles credential guessing, runs before the credential check), per-service-credential, per-owner. All fail closed on Redis error.
- `types.rs` / `routes/remember.rs` — new `Config` fields + fixture updates.
- `docs/api/owner-token-auth.md` — full contract for Console: request/response shapes, all distinct 429/503 response shapes across the three limiters, the shared-secret trust-boundary note.

## Testing

Verified end-to-end against a live local build (real HTTP, not just unit tests): credential rejection, cross-owner 403, real token expiry, forged tokens with altered permissions/audience correctly rejected (proving the scope check isn't hardcoded), rate-limit trip, no secret leakage across logs/responses.

Then reviewed via an adversarial multi-dimension pass (security, correctness, API-contract-vs-code, test-coverage). Confirmed findings fixed in a follow-up commit:
- **Major**: issuance had no throttling on guessing the shared credential at all (credential-gate rejected before the rate limiter ever ran, and that limiter was keyed by the guessed value anyway) → added a true outer per-IP limiter.
[...]

PR #543, opened 2026-08-05
Add per-memory storage expiry (end_epoch/expires_at)

harrymove-ctrl:
## Summary

WALM-296: adds `end_epoch`/`expires_at` to every item in `GET /v1/owners/{owner}/memories` (built by WALM-295, PR #537, not yet merged — this PR targets `feat/memory-read-api` directly rather than `dev`, so the diff here is scoped to WALM-296 only).

- New `vector_entries` columns: `end_epoch`, `expires_at`, `expiry_synced_at` (migration 014), plus an index on `expiry_synced_at` (migration 015).
- `end_epoch` is populated at write time on the 2 write paths that have it for free (primary upload + seal-store); the other 3 (metadata-transfer recovery, finalize-recovery, restore) start `NULL`.
- A new periodic background sweep (`main.rs`, mirrors the existing `evict_expired_delegate_keys`/`fail_stale_remember_jobs` pattern) queries the Walrus chain for storage-lease end epochs and converts them to `expires_at` timestamps, batched by owner, never touching the hot listing path.
- Epoch→timestamp conversion (`expires_at_from_epoch`) is anchored on the **current** epoch, not Walrus genesis — `expires_at = now + (end_epoch - current_epoch) * epoch_duration`. This self-corrects every sweep cycle regardless of whether `epoch_duration` ever changes, and avoids extrapolating from a fixed point over potentially hundreds of epochs.
- `status` on `MemoryItem` changes from a hardcoded `"active"` to a real derivation: `"expired"` if `expires_at` is in the past, `"active"` otherwise (including when not yet synced — an intentional optimistic default).

## Migration notes

- 014: single-transaction, metadata-only (3 nullable `ADD COLUMN`s, no rewrite).
- 015: `CREATE INDEX CONCURRENTLY`, its own file/transaction (required — cannot run inside a transaction).

## Test plan

- [x] Full suite green: 440 passed, 0 failed, 30 ignored (`cargo test --bins`, against real local Postgres).
- [x] Every task individually implemented + reviewed (fresh implementer + independent reviewer per task).
- [x] Final whole-branch review (opus) across the full diff — found and fixed: a sweep failure-ordering bug (schedule-fetch failure could mark rows synced with nothing written, invisible for 24h), a wrong index (`NULLS LAST` when the query needs `NULLS FIRST` — verified via `EXPLAIN`, cost dropped from 163 to 5.96), and an arithmetic overflow panic in the epoch-delta computation (function runs in an unsupervised background task with no restart wrapper).
[...]

PR #554, opened 2026-08-07
WALM-295/296/297: owner-scoped memory read API + bearer token auth for Console

harrymove-ctrl:
## Summary

Combines three related tickets that Console's read-only integration depends on, plus the wiring that connects them:

- **WALM-295** — `GET /v1/owners/{owner}/{namespaces,memories,agents}`: owner-scoped, cursor-paginated read API.
- **WALM-296** — per-memory storage expiry (`end_epoch` + computed `expires_at`) on the `/memories` response.
- **WALM-297** — owner-scoped, short-lived bearer token auth for Console (`POST /v1/owner-tokens`), since Console structurally never holds a delegate key and can't use the existing Ed25519 signed-request scheme.
- **Wiring fix** — `auth::verify_read_api_auth`, a combined dispatcher that lets the three real WALM-295 routes accept *either* WALM-297's bearer token (Console) *or* the existing Ed25519 scheme (SDK/dashboard), additively. This is the piece WALM-297's own PR (#546) explicitly called out as its outstanding follow-up, since that branch predates WALM-295's routes.

## Live-verified on Railway dev

Ran the full flow against `relayer.dev.memwal.ai` with real accounts (not synthetic data): mint a token via the service credential, call all three real routes with `Authorization: Bearer <token>`, confirm owner-mismatch → 403, tampered/invalid token → 401, and that a request with no `Authorization` header still falls through to the unmodified Ed25519 path. `/agents` returned real on-chain delegate keys; `/memories` returned real `end_epoch`/`expires_at` values.

## Supersedes

This branch already contains the full history of #537 (WALM-295), #543 (WALM-296), and #546 (WALM-297) plus the wiring fix on top — recommend closing those in favor of this one once reviewed.

## Test plan

- `cargo check --lib --bin` / `cargo test` (348 lib + 487 bin tests) — clean.
- New unit tests for the auth dispatcher's scope-check and `AuthInfo` construction.
- Opt-in `e2e_test.py` coverage for the bearer-token path (success, owner mismatch, invalid token, Ed25519 regression guard) — gated behind `TEST_OWNER_ADDRESS` + `OWNER_TOKEN_SERVICE_CREDENTIAL`.
- Live HTTP verification against Railway dev, described above.


services/server/docs/api/memory-read-api.openapi.yaml
```diff
@@ -0,0 +1,351 @@
...
+          type: integer
+          format: int64
+        storage_used:
+          type: integer
+          format: int64
+          description: Sum of blob_size_bytes across the namespace's memories.
+
+    NamespacesResponse:
+      type: object
+      required: [namespaces, next_cursor, snapshot_version]
```
nikola0x0: `has_more` is missing here and from `MemoriesResponse` below, but the live API returns it on both endpoints and `memory-read-api.md` calls it the only correct end-of-data signal.

```suggestion
      required: [namespaces, next_cursor, has_more, snapshot_version]
```

(the `has_more` property itself needs adding to `properties` too)

services/server/docs/api/memory-read-api.openapi.yaml
```diff
@@ -0,0 +1,351 @@
...
+      properties:
+        namespaces:
+          type: array
+          items:
+            $ref: '#/components/schemas/NamespaceSummary'
+        next_cursor:
+          type: string
+          nullable: true
+          description: Opaque continuation token; null on the last page.
```
nikola0x0: This contradicts `memory-read-api.md`, which says:

> `next_cursor` is **always** returned for a non-empty page — including the final page of a traversal […] So `next_cursor: null` does **not** mean "end of data".

Confirmed live: draining 48 memories, the last page came back with a non-null cursor and `has_more: false`. An empty page is the only case that returns null.

The distinction matters because a client that believes this description will discard a valid watermark on the empty page it fetches past the end.

services/server/docs/api/memory-read-api.openapi.yaml
```diff
@@ -0,0 +1,351 @@
...
+        snapshot_version:
+          type: integer
+
+    MemoryItem:
+      type: object
+      required: [memory_id, namespace_id, blob_id, created_at, size, agent_id, package_id, status]
```
nikola0x0: `end_epoch` and `expires_at` are missing from `MemoryItem` entirely — that is all of WALM-296, which this branch ships. Both are returned by dev today.

Worth documenting that they are nullable **independently of each other**, not together: we observed `end_epoch` set with `expires_at: null` on 4 of 5 rows written in one batch. We had assumed they moved together and had to correct our own types.

services/server/docs/api/memory-read-api.openapi.yaml
```diff
@@ -0,0 +1,351 @@
...
+            Delegate Ed25519 public key (hex) that wrote this memory. Not
+            directly comparable to `agents[].sui_address` from the /agents
+            endpoint — derive the Sui address from this public key
+            (blake2b256-based, one-hop) to correlate the two.
+        package_id:
+          type: string
+          nullable: true
+        status:
+          type: string
+          enum: [active]
```
nikola0x0: `enum: [active]` with "always active in this Phase 1 response" does not match dev — every row on WM's own test account came back `expired`, and the markdown documents the derivation (`expired` when `expires_at` is in the past).

This one is the sharpest of the four: we validate against the set of statuses we can store and fail loudly on anything unknown, so a client written to this spec would have thrown on its very first real page.

services/server/docs/api/memory-read-api.openapi.yaml
harrymove-ctrl: Fixed in e9a654b0 — added `has_more` to `properties` and `required` on both `NamespacesResponse` and `MemoriesResponse`.

services/server/docs/api/memory-read-api.openapi.yaml
harrymove-ctrl: Fixed in e9a654b0 — rewrote the `next_cursor` description on both response schemas to match memory-read-api.md exactly: always returned on a non-empty page (including the final page), null only for an empty page, and `has_more` is the real end-of-data signal, not cursor nullness.

services/server/docs/api/memory-read-api.openapi.yaml
harrymove-ctrl: Fixed in e9a654b0 — added `end_epoch` and `expires_at` to `MemoryItem` (both nullable, both in `required` since the field is always present, just possibly null). Documented the independent-nullability behavior you observed (end_epoch set with expires_at still null) directly in each field's description so a future reader doesn't re-assume they move together. Also caught `updated_at` missing from the same schema (and from `NamespaceSummary`) in the same pass — added both.

services/server/docs/api/memory-read-api.openapi.yaml
harrymove-ctrl: Fixed in e9a654b0 — enum is now `[active, expired]` and the description matches the real derivation from memory-read-api.md (expired if expires_at is in the past, active otherwise, including while end_epoch/expires_at are still null pre-sync). Good catch — this one would have broken a strict client on its first real page.

services/server/src/routes/memory_read.rs
```diff
@@ -0,0 +1,1570 @@
...
+) -> Result<MemoriesResponse, AppError> {
+    let limit = limit.clamp(1, MAX_MEMORIES_LIMIT);
+
+    // See the doc comment on `MemoriesCursor::snapshot_at`: bind this whole
+    // walk to one boundary, captured now on the first page and threaded
+    // forward unchanged from the incoming cursor on every later page.
+    let (after, snapshot_at) = match cursor {
+        Some(raw_cursor) => {
+            let c = decode_cursor(&raw_cursor)?;
+            let snapshot_at = c.snapshot_at.unwrap_or_else(chrono::Utc::now);
```
ducnmm: The terminal `next_cursor` keeps the original `snapshot_at`, so the next poll still applies `updated_at <= old_snapshot_at` and can never return rows written after the first walk. This breaks the WALM-295 incremental-sync contract. Could terminal checkpoints reset/drop the snapshot bound while continuation cursors retain it? The namespaces path has the same issue.

services/server/src/rate_limit.rs
```diff
@@ -609,6 +623,95 @@ pub async fn rate_limit_middleware(
...
+            return next.run(request).await;
+        }
+    };
+
+    let config = &state.config.read_api_rate_limit;
+    let mut redis = state.redis.clone();
+    let now = chrono::Utc::now().timestamp_millis() as f64;
+
+    let weight = endpoint_weight(request.uri().path());
+    let dk_key = format!("rate:read:dk:{}", auth.public_key);
```
ducnmm, line 682: The signed-request path accepts mixed-case public-key hex and preserves the caller-supplied representation in `AuthInfo`, so the same delegate can vary casing to obtain independent Redis buckets here and bypass the WALM-297 read limit. Please canonicalize with `hex::encode(pk_array)` before account resolution/cache/rate-limit identity construction.

services/server/src/storage/db.rs
```diff
@@ -756,6 +1343,66 @@ impl VectorDb {
...
+
+    /// Write back a resolved end_epoch/expires_at for one row. Never
+    /// touches updated_at — see this plan's Global Constraints.
+    pub async fn set_memory_expiry(
+        &self,
+        id: &str,
+        end_epoch: i32,
+        expires_at: chrono::DateTime<chrono::Utc>,
+    ) -> Result<(), AppError> {
+        sqlx::query("UPDATE vector_entries SET end_epoch = $1, expires_at = $2 WHERE id = $3")
```
ducnmm: This changes API-visible `end_epoch`/`expires_at` without advancing the incremental-sync watermark. A client that already passed this row will never receive the populated expiry or later status change, which prevents WALM-296 data from reaching the WALM-295 incremental consumer. Please make expiry changes cursor-visible, either through `updated_at` or a dedicated read-model watermark.

services/server/src/storage/walrus.rs
```diff
@@ -106,6 +156,8 @@ struct WalrusUploadResponse {
     object_id: Option<String>,
     #[serde(default)]
     transfer_status: Option<String>,
+    #[serde(default)]
+    end_epoch: Option<i32>,
```
ducnmm, line 218: The success response preserves `endEpoch`, but the sidecar also includes it in metadata-transfer failure responses and `WalrusUploadErrorResponse` drops it; recovery then finalizes with `end_epoch = None`. Please carry `end_epoch` through `MetadataTransferFailed`, `SetMetadataAndTransfer`, and `FinalizeUploadedBlob` so the WALM-296 value survives this normal recovery path.

ducnmm (commented):
Reviewed against WALM-295/296/297. I found four issues that affect the agreed incremental-sync, rate-limit, and expiry contracts.

services/server/src/routes/memory_read.rs
harrymove-ctrl: Confirmed and fixed in 5df56c23 — good catch, this is real. Reset `snapshot_at` to `None` specifically on the terminal cursor (`has_more: false`); continuation cursors mid-walk keep it as before, so the mid-walk-duplicate protection this field exists for is unaffected. Same fix in `query_owner_memories`. Extended both existing tests to actually poll with the terminal cursor (the realistic client pattern) instead of only asserting the from-scratch `cursor: None` resync case — the old buggy code passed that trivially, which is why it didn't catch this.

services/server/src/rate_limit.rs
harrymove-ctrl: Confirmed and fixed in 5df56c23 — re-encode from the decoded bytes (`hex::encode(pk_array)`) right after decode in `verify_signature`, so every downstream use of `public_key_hex` (cache keys, `AuthInfo.public_key`, the rate-limit bucket, logging) sees one canonical lowercase form regardless of the caller's header casing. Added a unit test pinning that lower/upper/mixed-case input all canonicalize identically.

services/server/src/storage/db.rs
harrymove-ctrl: Confirmed and fixed in 5df56c23. The never-touch-updated_at behavior was deliberate (to stop the ~24h re-verification sweep from making every row reappear in incremental sync every day) but was applied too broadly — it also suppressed the one write that must be cursor-visible: a row's first resolution from unknown to known expiry. Fixed with a conditional bump (`IS DISTINCT FROM`, null-safe): advances updated_at only when end_epoch/expires_at actually change, so the first real write is visible but routine unchanged re-verification still isn't. Replaced the test that encoded the old assumption with one pinning all three cases (first write bumps, unchanged reverify doesn't, a later real change bumps again).

services/server/src/storage/walrus.rs
harrymove-ctrl: Confirmed and fixed in 5df56c23 — threaded end_epoch through the full chain: added it to WalrusUploadErrorResponse (the sidecar does send it on the metadata-transfer failure response, per walrus-upload.ts, it just wasn't declared here so serde silently dropped it), then through UploadBlobError::MetadataTransferFailed, both WalletOperation recovery variants (SetMetadataAndTransfer, FinalizeUploadedBlob), and finally the insert_vector call — mirroring the exact pattern agent_id/package_id already use for this same recovery path. Added tests pinning the deserialize (with and without end_epoch present, for backward compat).

services/server/src/routes/memory_read.rs
```diff
@@ -0,0 +1,1670 @@
...
+    rows.truncate(limit as usize);
+
+    // Same watermark rule as `query_owner_namespaces`: the cursor always
+    // comes from the last row emitted (never the peeked extra row above), so
+    // the final (or only) page still checkpoints the client for its next
+    // incremental poll. Only an empty page has no cursor, because it has no
+    // row to build one from.
+    // `Some(snapshot_at)` while the walk continues, `None` on the terminal
+    // page — see `encode_cursor`'s doc comment.
+    let next_cursor = rows
```
ducnmm: The snapshot reset still fails when a continuation page becomes empty. For example, page 1 emits A and peeks B with `has_more: true`; if B is then updated past `snapshot_at`, page 2 returns no rows, so this produces no replacement cursor. Per the API contract the client keeps the previous cursor, which still contains the old snapshot, and B remains excluded on every subsequent poll. The namespaces path has the same failure mode. Could an empty continuation response return a reset form of the incoming checkpoint, and could we add regression coverage where every remaining row moves beyond the snapshot before the next page?

harrymove-ctrl: Confirmed and fixed in 7f0206c0 (also resolved the merge conflict with dev in this push). Good catch — the terminal-cursor fix only covered a populated last page, not an empty continuation page hitting the same rows.last() == None path. Fixed by re-encoding the incoming cursor's own watermark position (unchanged, since no row was consumed) with snapshot_at reset whenever a page comes back empty but there was a real incoming cursor; has_more is always false there so the reset is unconditional. A brand-new walk that finds nothing at all (no incoming cursor) still correctly returns no cursor. Added regression tests reproducing exactly this race (a peeked row updated past snapshot_at before the next page is requested) for both namespaces and memories, and updated five pre-existing tests that had asserted the old 'empty page always means None' contract.

services/server/docs/api/memory-read-api.openapi.yaml
```diff
@@ -0,0 +1,404 @@
...
+        namespaces:
+          type: array
+          items:
+            $ref: '#/components/schemas/NamespaceSummary'
+        next_cursor:
+          type: string
+          nullable: true
+          description: >-
+            Always returned for a non-empty page — including the final page
+            of a traversal. Null only for an empty page, which has no row to
```
nikola0x0: Stale again after 7f0206c, which changed the behavior but only touched `memory_read.rs`.

An empty page now returns a **non-null** cursor whenever there was an incoming cursor: `memory_read.rs:282-284` re-encodes the incoming watermark with `snapshot_at` reset. Null is left for exactly one case, the first page of a walk that finds nothing at all.

That matters more than the previous wording did. A client that terminates on cursor nullness used to just lose its watermark on the page past the end; now that page hands back a non-null cursor every time, so the loop never terminates.

Same two lines on `MemoriesResponse` at 373-374.

docs/api/memory-read-api.md
```diff
@@ -0,0 +1,296 @@
...
+unchanged value does *not* advance `updated_at` (this would otherwise make
+every synced memory reappear roughly once a day regardless of whether
+anything changed); only a genuine change does.
+
+### Cursor semantics (both paginated endpoints)
+
+`next_cursor` is **always** returned for a non-empty page — including the
+final page of a traversal and a result that fits entirely in one page. It
+is the watermark of the last row in that page, and it is what you pass as
+`updated_after` on your next poll. It is `null` only for an empty page,
```
nikola0x0: Same drift as the OpenAPI schema: "`null` only for an empty page" is no longer true after 7f0206c. An empty page with an incoming cursor returns the reset cursor; null now only happens on a fresh walk that matches nothing.

The "keep the cursor you already had" advice is also worth rewording, since the whole point of the new branch is that the cursor you already hold carries a stale `snapshot_at` and the reset one supersedes it.

services/server/src/storage/db.rs
```diff
@@ -756,6 +1448,93 @@ impl VectorDb {
...
+        id: &str,
+        end_epoch: i32,
+        expires_at: chrono::DateTime<chrono::Utc>,
+    ) -> Result<(), AppError> {
+        sqlx::query(
+            "UPDATE vector_entries
+             SET end_epoch = $1,
+                 expires_at = $2,
+                 updated_at = CASE
+                     WHEN end_epoch IS DISTINCT FROM $1 OR expires_at IS DISTINCT FROM $2
```
nikola0x0: This condition is effectively always true, so the anti-spam half of the fix doesn't hold.

`main.rs:1197-1202` computes `expires_at = now + (end_epoch - current_epoch) * epoch_duration` with `now` captured at sweep time, and `rows_needing_expiry_refresh` (line 1463) re-selects every row once `expiry_synced_at` is 24h old. So each routine re-verification writes an `expires_at` roughly 24h later than the stored one, `expires_at IS DISTINCT FROM $2` fires, and `updated_at = NOW()` lands on every row daily. That is exactly the "every synced memory reappears roughly once a day" outcome `memory-read-api.md:238-241` promises it prevents, and it degrades `updated_after` into a daily full resync.

The new test passes because it calls `set_memory_expiry` twice with the same literal timestamp, so it pins the SQL but not the caller's behavior.

Easiest fix is probably to condition on `end_epoch` alone, since `expires_at` is derived from it and is deliberately approximate anyway. A tolerance compare on `expires_at` would also work if you want a recomputed duration to still be visible.

nikola0x0 (approved):
Only outdated docs and a nit left. LGTM!

jessiemongeon1:
## Style Guide Audit

**16** violation(s) found. All must be fixed before merge.

### `docs/api/memory-read-api.md` (7 violation(s))

_7 violation(s) (7 regex, 0 claude)_

- **Line 113** — "on-chain" is one word → "onchain"
  - Current: `on-chain`
  - Fix: `onchain`
- **Line 115** — "on-chain" is one word → "onchain"
  - Current: `on-chain`
  - Fix: `onchain`
- **Line 131** — "on-chain" is one word → "onchain"
  - Current: `on-chain`
  - Fix: `onchain`
- **Line 167** — Use "through" not "via"
  - Current: `via`
  - Fix: `through`
- **Line 173** — "on-chain" is one word → "onchain"
  - Current: `on-chain`
  - Fix: `onchain`
- **Line 327** — "on-chain" is one word → "onchain"
  - Current: `on-chain`
  - Fix: `onchain`
- **Line 354** — "on-chain" is one word → "onchain"
  - Current: `on-chain`
  - Fix: `onchain`

### `docs/api/owner-token-auth.md` (9 violation(s))

_9 violation(s) (3 regex, 6 claude)_

- **Line 44** — Use "through" not "via"
  - Current: `via`
  - Fix: `through`
- **Line 122** — Use "through" not "via"
  - Current: `via`
  - Fix: `through`
- **Line 147** — Use "through" not "via"
  - Current: `via`
  - Fix: `through`
- **Line 75** — no-italic
  - Current: `*only*`
  - Fix: `**only**`
- **Line 101** — passive-voice
  - Current: `owner is canonicalized to lowercase`
  - Fix: `the handler canonicalizes owner to lowercase`
- **Line 159** — passive-voice
  - Current: `the request is rejected `403``
  - Fix: `the server rejects the request with `403``
- **Line 161** — passive-voice
  - Current: `is rejected `401``
  - Fix: `the server rejects it with `401``
- **Line 161** — passive-voice
  - Current: `An expired, tampered, wrongly-signed, or wrong-audience token is rejected `401``
  - Fix: `The server rejects an expired, tampered, wrongly-signed, or wrong-audience token with `401``
- **Line 276** — passive-voice
  - Current: `a bearer token is meant to be reused across many read requests`
  - Fix: `a bearer token supports reuse across many read requests`

---

nikola0x0:
Following up after approving — this isn't a regression in the branch, it's a contract gap I only hit once Console's consumer existed. Not a change request and I don't think it blocks the merge; mostly I want the choice to be deliberate.

## Problem

**A deletion is invisible to an incremental consumer, permanently.**

`memory-read-api.md` already flags this ("Deletions are not currently reflected… clients that need to detect deletions should periodically do a full resync"), so this is a quantification rather than news. What I hadn't appreciated until measuring is that there is *no* signal on the incremental path — not a weak one.

The mechanism is in `query_owner_namespaces`:

```sql
GROUP BY namespace
HAVING (MAX(updated_at), namespace) > ($2, $3) AND MAX(updated_at) <= $4
```

The cursor is `MAX(updated_at)` over *surviving* rows. Delete the row holding the max and the watermark moves **backwards**, while the keyset advances strictly forward. No surviving row's `updated_at` changes, so nothing can ever push that namespace past a client's cursor again.

## Evidence

Measured on the dev relayer (`relayer for console test`), owner `0x4824ad59…`. I deleted one row, measured, and restored it from a full-row backup — the fixture is back byte-for-byte and `rows=2438 owners=28` is unchanged, in case anything looked odd in the logs.

Deleted `a07f5e12-…` from `notes` (2 rows → 1), which held that namespace's watermark:

```
rollup            count 2 -> 1, storage_used 817 -> 402       CORRECT
watermark         2026-08-08T07:16:55.674631Z -> ...666392Z   moved BACKWARDS
updated_after=<pre-deletion cursor>   -> 0 namespaces
snapshot_version  2 -> 2                                      unchanged
```

I also confirmed separately that a namespace's `updated_at` equals `MAX(memory.updated_at)` to the microsecond, so the backwards move is structural rather than incidental.

`snapshot_version` staying flat matters: that's the one remaining escape hatch a consumer could watch, and it doesn't fire either.

## Why it matters

**The rollup is already correct** — `COUNT(*)` and `SUM(blob_size_bytes)` are recomputed per query, and both updated instantly. So `storage_used` is trustworthy today and anyone using it for quota or display is fine. The broken part is *change detection* only, which narrows the blast radius but makes it invisible: a consumer polling incrementally sees "no changes" and is confidently wrong.

For Console (COMG-571): deletion detection collapses entirely onto our periodic full reconcile. At its current 6-hour interval, a memory a user deletes in WM stays visible in Console for up to 6 hours with storage overstated. Given deletion shipped partly for privacy reasons (WALM-262/264/303), that seems worth an explicit decision. The same applies to rows removed by blob expiry/cleanup, which users never asked for and won't expect to lag.

## Options

**A — Store a per-namespace `last_mutated_at`, written on any mutation including delete.**
Replace the derived `MAX(updated_at)` cursor with a stored watermark that a delete also bumps. The namespace then resurfaces exactly once after a deletion, and the consumer reconciles against the rollup, which is already correct. Doesn't touch row semantics, doesn't require anything new in the response, fixes the root cause. My preference — it looks like the smallest change that actually closes it.

*(Bumping surviving rows' `updated_at` would also work mechanically, but it makes rows lie about when they changed and would push consumers into re-fetching unchanged data. Not suggesting that.)*

**B — Expose deletions explicitly.** Tombstone rows in `/memories`, or a `deleted_since` feed. More faithful and more useful to consumers who want to distinguish "deleted" from "never seen", but a real build and a response-shape change.

**C — Declare "full resync only" the Phase 1 contract.** Legitimate, and cheapest for you. If so it'd help to state the expectation in the doc, so consumers size their resync interval around a deletion-visibility SLA rather than around load.

## What Console will do meanwhile

Regardless of which way you go, we can work around it, so this isn't blocking us: poll `/namespaces` cursor-lessly (cheap — one request, no pagination), compare `memory_count` against our local per-namespace count, and trigger a full drain only when WM's count is *lower* than ours. That turns a 6-hour detection window into our 5-minute sync interval, and the expensive drain only runs for owners that actually diverged.

Flagging it because it leans on the rollup being recomputed per query. If that's likely to change — cached, materialised, eventually-consistent — tell me and we'll do something else.

---

Unrelated and minor, from the same session: `POST /v1/owner-tokens` accepts an unrecognized `scope` and returns 200. It correctly pins `permissions: ["memories.read"]` regardless of what was asked — I tried `memories.write`, `memories.delete` and `memories.admin` and all three came back read-only, which is the right behaviour. A 400 on an unknown scope would just save a future caller from thinking they got something they didn't.

@harrymove-ctrl

harrymove-ctrl:
HI @nikola0x0, Option A is heavier than expected. There's no namespaces table (namespace data is aggregated via GROUP BY..

 Proceeding with Option C (updating docs/api/memory-read-api.md). If namespace watermarks are still needed later, we can track Option A as a separate follow-up ticket. Pls help me resolve conflict too

harrymove-ctrl:
## Merged with `dev`, conflicts resolved, three defects fixed

Pushed `58182703`. The conflict banner is cleared and GitHub now reports `mergeable=true`.

Done as a merge rather than a rebase: this branch's convention is merging `dev` in (7 prior merges), and rebasing 61 commits would force-push away @nikola0x0's approval and the whole review history. Note that #718 is already merged in here (`92a81fc5`), so this PR now carries WALM-295/296/297/363 together.

### Conflict resolution (`d245ca0b`)

Two textual, two semantic that git could not see:

1. **Migration `014` collided.** `dev` added `014_storage_reservations.sql` (WALM-359, #720) while this branch has `014_memory_read_api_columns.sql`. There is no migration version table, every file re-runs through `raw_sql` on boot and is idempotent, so both files coexist. Split the bindings into `migration_014_reservations` and `migration_014_read_api`; both execute. The duplicate number is ugly and worth renumbering later, but renaming 7 files on an approved branch felt like the wrong trade right now.
2. `db.rs` top-level items: kept this branch's pagination-index helpers alongside `dev`'s `release_storage_reservations_with_pool` and `StorageAdmission`.
3. `dev`'s three storage-reservation tests called `insert_vector` with the pre-branch 7-argument signature.
4. `walrus.rs` test imports: kept both sides.

### One thing worth flagging separately

`deleting_the_watermark_row_still_resurfaces_the_namespace` **never compiled**. `cursor` was moved into the first `query_owner_namespaces` call and used again afterwards. That is the only test covering WALM-363, so deletion behaviour has never actually been exercised. CI stayed green because it does not build `--all-targets`. Fixed, and worth adding `cargo check --all-targets` to CI so this class of thing cannot recur.

### Three defects fixed (`58182703`), each with a test

**1. `expires_at` never reached a client that had already synced the row.** `insert_vector` writes `end_epoch` but not `expires_at`, so when the sweep first resolves a row on the mainline path, `end_epoch` is already equal and the `WHEN end_epoch IS DISTINCT FROM $1` guard is false. `expires_at` went NULL to real with no `updated_at` bump, so the row never re-entered incremental sync and Console kept `null` for it indefinitely. That is WALM-296 silently not working on every new memory. The existing test only covered the legacy path where `end_epoch` was NULL at insert, which is why it passed. The guard now also fires while `expires_at IS NULL`, which preserves the anti-spam property the comment argues for.

**2. `must_resync` fired on healthy clients.** `stale_tombstone` compared the cursor's `deleted_at` against the retention cut, but `minted_at` is re-stamped on every cursor while `deleted_at` freezes at the last tombstone the client saw. A client polling every 5 minutes was told to full-resync 30 days after its last observed deletion, recurring, which is roughly 300 requests on a 150k-row owner. The check was also redundant: the tombstone query already floors on `deleted_at >= retention_cut`. Now keyed off `minted_at` alone, with tests pinning both directions.

**3. The docs described a `status` the code cannot emit.** `memory-read-api.md` said `status` is `"deleted"` for tombstoned rows. The match on `expires_at` only produces `"expired"` or `"active"`, and `DeletedMemory` has no `status` field. A Console client following that line branches on an unreachable value, never evicts deleted rows, and leaves them visible, which is exactly what WALM-363 exists to prevent. The OpenAPI enum was already right.

Also cleared the last 3 style-guide violations, so @jessiemongeon1's audit should come back clean now. Its original count of 75 was from before the #718 docs pass.

### Review threads

Every thread on this PR is answered. Two were left open although their fixes shipped, and I re-verified both against the current tree before resolving them: hex canonicalisation lands at `auth.rs:229`, and `end_epoch` is declared on `WalrusUploadErrorResponse` at `walrus.rs:234`.

### Still open, tracked rather than fixed here

Two acceptance criteria are genuinely not met, and I would rather name them than quietly close them:

* **WALM-295 AC5** (load-tested at scale): no perf harness exists, the largest seeded test is 501 rows. Related: the `/namespaces` rollup aggregates the owner's entire partition per page through a `FULL OUTER JOIN`, so no index covers it and cost scales with account size rather than page size. `/memories` and both tombstone queries are index-clean.
* **WALM-363 AC6** (deletion case in the live e2e suite): `walm297_e2e_full.py` does not exist, and `e2e_test.py` has zero deletion coverage. Deletion is covered by Rust unit tests only.

Neither blocks Phase 1 in my view, but both should be follow-up tickets rather than ticked boxes.

One tracker note: WALM-363's description still prescribes `deleted_at` on `vector_entries` and says a deletion feed is not being taken. This PR does the opposite, a separate `memory_tombstones` table plus `deleted[]`, for the reason recorded in `migrations/020`. The rationale is sound; the ticket text just needs to catch up before it closes.

services/server/src/storage/db.rs
harrymove-ctrl: Reopening this briefly because the original fix was incomplete, and you were more right than the reply gave you credit for.

`5df56c23` made `set_memory_expiry` bump `updated_at` via `WHEN end_epoch IS DISTINCT FROM $1`. That closes the case where `end_epoch` was NULL at insert, which is the legacy and recovery path, and it is the only case the regression test covers (it passes `None` as the 10th argument).

On the mainline path it does not close. `insert_vector` writes `end_epoch` at INSERT time but never writes `expires_at`. So when the sweep first resolves that row, `end_epoch` is already equal, the guard is false, and `expires_at` goes from NULL to a real timestamp with no watermark bump. Exactly the failure you described: a client that already passed the row never receives the populated expiry. Every new memory written with a known `end_epoch` was affected, for the up-to-5-minute window before the sweep reached it.

Fixed properly in `58182703`: the guard is now `WHEN end_epoch IS DISTINCT FROM $1 OR expires_at IS NULL`, which keeps the anti-spam property the comment argues for, since a re-verification of an already-populated row still does not bump. Added `set_memory_expiry_bumps_updated_at_when_only_expires_at_appears`, which inserts with `Some(500)` and asserts the bump, so the mainline path is now covered too.

Resolving.

ducnmm (approved):
Second-pass review of `58182703` (includes #718 + merge of `dev` + the three follow-up fixes). Focused on authz/security of the new Console surface, then residual product/ops.

## Security (the new `/v1/owners/*` + `POST /v1/owner-tokens`)

These routes are the WM-C Console door, not the MCP write path. Confidentiality still sits in SEAL; this surface returns metadata the chain already exposes.

- Issuance is gated on `x-service-credential` (constant-time compare, empty config → 401). WM does **not** re-prove that Console controls `Y` — that is the PRD model. A leaked service credential mints a `memories.read` token for any owner that already has a MemWalAccount. Keep the credential server-side only.
- Tokens are HMAC-SHA256, audience `memwal`, TTL 900s, failures collapse to bare 401. `nonce` is not a single-use replay guard (correct for a bearer). No Phase 1 revocation list — a stolen token lives until expiry.
- Path `{owner}` is checked with `same_owner` (403 on mismatch); SQL binds `auth.owner`, never the URL segment. Bearer path sets `delegate_key` / `seal_session` to None, so a Console token cannot drive SEAL decrypt. Listing SELECTs omit `plaintext` and `embedding`.
- Unknown mint scopes are 400. Missing `memories.read` on verify is 403. Unset `OWNER_TOKEN_SECRET` fail-closes (mint 503, bearer 401).
- Rate limits on issuance (IP + credential + per-owner) and on the GET read routes.

Non-blocking nits, not merge blockers:
- `verify_token` does not assert `subject == "console"`. Harmless while only `mint_token` can produce a valid HMAC.
- `constant_time_eq` returns early on length mismatch (leaks secret length).
- `GET /v1/owners/{owner}/_token_probe` is still mounted (smoke leftover). Needs a valid bearer; no extra rate-limit layer.
- Read routes also accept the existing Ed25519 signed-request scheme when `Authorization` is absent. An SDK/dashboard caller can list **their own** metadata. Not cross-tenant.

No IDOR, no decrypt, no cross-owner read found.

## Residual (already filed / accepted)

- WALM-295 AC5 load-test and WALM-363 live deletion e2e → WALM-400. `/namespaces` still aggregates the whole owner partition per page.
- Two `014_*.sql` files (quota reservations from #720 + read-api columns). Boot uses `include_str` in order, not sqlx versions — works, easy to trip over later.
- First boot `backfill_updated_at` can stall health checks on a large table. Console-test Railway already deployed this head successfully; if the main `dev` relayer shares that DB, the backfill is a no-op.
- WALM-364 script ran on the console-test DB (2662/2674 synced; 12 rows on `0x158a78f0…` left because sidecar listing failed).

## Verdict

Approve. Security of the Console door is sound for Phase 1. Ship to `dev`; keep WALM-400 in front of `main`.

PR #562, opened 2026-08-08
Fix: make remember writes idempotent — no duplicate paid Walrus blob

### LINKED ISSUE #477: Write can exceed the confirmation window and report failure while still landing — no idempotency key or definite-state poll
## Problem

A write can take longer than the client's confirmation window (~45s in our integration). When it does, the call reports failure, but the write can still commit to Walrus afterwards. The outcome is ambiguous: an integrator cannot distinguish a true failure from a slow success, which makes safe retry impossible — retrying risks a duplicate, not retrying risks silent data loss.

## Ask

- An idempotency key on writes, so a retry after an ambiguous timeout is deduplicated instead of duplicated.
- A way to poll a pending write to a definite committed / failed state, instead of a timeout that leaves the result unknown.

Related: #412 (confirming a save / getting its blob id) and #433 (indexer lag after `remember`) touch the same area; this issue is specifically about the ambiguous-timeout outcome and idempotent retry. Configurable timeouts (#420, #417) widen the window but do not remove the ambiguity.

hungtranphamminh:
---

### Why

**The core hazard.** A `remember` write is accepted with a `202` and finished asynchronously by a
wallet job. Inside that job, `upload_blob` **mints a paid, irreversible on-chain Walrus blob** — it
spends gas and creates a unique blob object. Everything else in the job (the DB status row, the
vector index) is cheap and replayable; that one call is not. The job was written as if the whole body
were safe to re-run on retry, which is true for every step *except the mint*. So the underlying
defect is a single sentence: **a paid, irreversible external effect is treated like a replayable
step, and nothing makes "have we already minted?" and "mint" atomic.** Every finding below is a
concrete way that non-atomicity leaks into a second paid blob (or, for RC-4/RC-5, into a corrupt or
lost write).

**How a second mint is triggered — two independent doors.**

- **Client door (the reported bug, GH #477).** A write can outlive the client's ~45s confirmation
  window: the call returns failure, but the write still commits afterward. The client now can't
  distinguish a slow success from a real failure — **retry risks a duplicate paid blob; not retrying
  risks silent data loss.** There was no idempotency key, so a client retry was a *new, independent*
  job → a second mint.

- **Server door (sharpened by a commenter on the issue).** No user needs to be involved at all. The
  relayer's own Apalis retry loop re-runs the *entire* job — including `upload_blob` — on any
  retriable error, e.g. a transient DB blip on the step *after* a successful mint. With no guard
  checking the recorded `blob_id`/`status` before re-uploading, that internal retry mints again.

**Why the obvious fix ("add a guard before `upload_blob`") is not enough.** A guard that reads the
job row and skips the upload when a blob is already recorded closes the *simple serial* case — but a
guard is only as good as (a) the state it reads being *durably written*, (b) *no one else minting
concurrently*, and (c) the state it reads being *unambiguous*. Each of those turned out to be false
here, which is why the fix is five parts, not one:

| RC | The assumption that fails | Verifiable evidence | Consequence |
|----|---------------------------|---------------------|-------------|
| **RC-1** | "the job body runs once" | Apalis retries on any retriable `Err` (`MAX_ATTEMPTS=5`); the old code had no guard | retry re-mints — the reported bug |
| **RC-2** | "the status write after a mint always lands" | the `UPDATE status='uploaded'` was `let _ = …` (fire-and-forget); if it errors or the process dies, the row stays `running`/NULL | the guard on the next attempt reads no blob → **re-mints**. The record of an irreversible act was best-effort. |
| **RC-3** | "Apalis runs a job's attempts serially" | **false.** `WALLET_JOB_CONCURRENCY` = **8** workers; apalis `reenqueue_orphaned_after` = **300s** flips a `Running` job back to `Pending` on the *worker's* stale heartbeat (SQL: `WHERE status='Running' AND workers.last_seen < now-300s`) with **no check that the handler finished**; `upload_blob` can run up to **300s** | two attempts of the *same* job run at once, both read `running`/NULL, **both mint** (TOCTOU). The guard read and the mint were not mutually exclusive. |
| **RC-4** | "`uploaded` means ready to index" | `status='uploaded'` is written on **two** routes: full success, **and** the *metadata/transfer-failed* recovery path (blob minted, transfer still pending, a `SetMetadataAndTransfer` recovery enqueued) | a naive "uploaded → index and mark `done`" resume finalizes a blob **never transferred to the user** → a recallable memory the user doesn't own. (This is the regression a first-cut guard *introduced*.) |
| **RC-5** | "an idempotency key identifies a durable result" | a `(owner, key)` lookup returns any row, including `failed` | a same-key retry of a *failed* write collapses onto the dead row forever → the memory is **never written**. A failed attempt isn't a durable result. |

**Why the sidecar doesn't save us.** The Walrus sidecar journals its own multi-step write for
*resume* and dedups an *identical prepared transaction by digest* — but a server retry rebuilds
`upload_blob` from scratch (fresh encrypt/prepare → a **different** transaction digest), so the
sidecar treats it as a new write and mints again. Idempotency has to be enforced on our side.

**The deeper "why it was unhandled."** The `remember_jobs` row was a *reporting* artifact
(best-effort `let _ =` writes) rather than the *authoritative gate* for a paid action; and the
queue's at-least-once + orphan-reaper semantics were never reconciled with the need for an
at-most-once *mint*. The fix turns the row into the authoritative gate (durable, lock-guarded reads)
and adds the missing mutual exclusion.

*(Full trace with file:line and the empirical apalis/PG reproduction is in the methodology trail;
the runtime constants above — 8 / 300s / 300s / 10-conn pool / unpruned table — were read from source
and apalis-sql 0.7.4, not assumed.)*

### What

The client-facing capability (idempotency key, incl. RC-5) + the four internal safety fixes
(RC-1..RC-4), plus the pre-existing poll endpoint stated honestly.

**Client idempotency key (the reporter's ask #1):** optional `idempotency_key` on `RememberRequest`;
a `(owner, idempotency_key)` partial-unique index; lookup-before-insert + `ON CONFLICT DO NOTHING` +
lost-race re-read so a retry (or concurrent race) collapses onto **one** job and spawns exactly one
background write. A same-key retry of a *failed* job restarts it in place (RC-5) instead of pinning
`failed`.

**Definite-state poll (ask #2):** **already exists** — `GET /api/remember/:job_id` returns terminal
`done`(+blob_id) / `failed`(+error), and the SDK carries `job_id` on a 504. No code change; this PR
does not touch it. (The close-out reply should point the reporter at it.)

**Internal double-mint safety (RC-1..RC-4):**
- **Guard** `upload_resume_disposition` before the mint: `done` → no-op; `uploaded` + stored
  `blob_object_id` → resume transfer; else → upload.
- **Durable persist** `persist_uploaded_state`: the uploaded-state write is no longer fire-and-forget
  — a failure returns retriable so the guard always reads accurate state; also records
  `blob_object_id`.
- **Per-job advisory lock** `JobUploadLock` (`pg_try_advisory_lock` on a **dedicated small pool**):
  serializes concurrent attempts of the same job across guard→mint→persist; a contender **defers
  (retriable) without minting**. `lock_outcome` centralizes the "never mint without the lock" rule.
- **Transfer-aware resume** `build_resume_transfer_job`: an `uploaded`-but-pending retry re-enqueues
  the existing `SetMetadataAndTransfer` recovery (finish transfer, never re-mint, never premature
  `done`).

### Solution notes / design decisions

- **Advisory lock, not `SELECT … FOR UPDATE` / txn lock.** `upload_blob` can run up to 300s; holding
  a request-pool connection (or an open txn) across it would starve recall/status under a sidecar
  backlog. So the lock is `pg_try_advisory_lock` on a **dedicated** `wallet_lock_pool` (sized to
  `WALLET_JOB_CONCURRENCY`), reusing the existing `security_delete_background_sui` isolation pattern.
  A contender that can't get the lock returns retriable — it never blocks-and-holds.
- **`blob_object_id` column (migration 011).** Required so an `uploaded`-but-transfer-pending job can
  resume via `SetMetadataAndTransfer` (which needs the object id) without re-minting. It's an
  additive **metadata-only** column (nullable, no default) — cheaper/safer than the Part-2 index.
- **Privacy floor unaffected.** The key and `blob_object_id` are client-/system-owned identifiers on
  the job row (same trust tier as `owner`/`job_id`/`blob_id`) — no new server-readable plaintext
  *content* index.
- **Byte-identical for keyless clients.** `idempotency_key` is serde-default `None`; the whole
  idempotency block is skipped when absent, and the keyless INSERT is unconstrained.

---

## The failure sequences, step by step

The RC table above says *which* assumption fails; these timelines show *exactly how* each turns into
a second paid blob (or the RC-4 corruption). They're the concrete version of the same three cases.

**Sequence 1 — internal retry after a DB blip (RC-1 + RC-2).** Single worker, no client involved:

```
worker attempt #1:
  upload_blob() ─────────────► ✅ PAID BLOB MINTED (on-chain, irreversible)
  UPDATE status='uploaded'  ── (was `let _ =` — fire-and-forget)
        │
        ├─ if this UPDATE errors, OR the process dies here, the row stays
        │  status='running', blob_id=NULL
        ▼
  insert_vector_and_mark_remember_done() ── transient DB error → Err
  → Apalis retries the whole job (MAX_ATTEMPTS=5)

worker attempt #2 (old behavior):
  no guard → upload_blob() ──► ✅ SECOND PAID BLOB   ← the bug
```
Fix: RC-1 guard reads the row first; RC-2 makes the `uploaded` write durable (retriable on failure)
so attempt #2 reliably sees `uploaded`+`blob_id` and resumes instead of re-minting.

**Sequence 2 — concurrent re-dispatch (RC-3, the TOCTOU).** Two of the 8 workers, same job:

```
t0  worker A picks job J → guard SELECT → status='running', blob_id=NULL → proceed
t1  A enters upload_blob() … (slow: sidecar backlog, up to 300s)
t2  A's worker heartbeat goes stale (>300s); apalis reenqueue_orphaned flips J → Pending
t3  worker B picks J → guard SELECT → STILL status='running', blob_id=NULL → proceed
t4  A's upload_blob() ─► ✅ PAID BLOB #1
t5  B's upload_blob() ─► ✅ PAID BLOB #2       ← both mint; guard read was stale
```
Fix: RC-3 per-job `pg_try_advisory_lock` — B can't get the lock A holds, so B **defers (retriable)
without minting**. (Guard-read + mint + persist are all inside the lock.)

**Sequence 3 — the regression the naive fix introduced (RC-4).** An `uploaded` row is ambiguous:
Route A (full success, about to index) and Route B (mint OK, **transfer failed**, recovery pending)
both write `status='uploaded'`. A naive "uploaded → index and mark done" resume finalizes a Route-B
row → a recallable memory pointing at a blob **never transferred to the user**. Fix: RC-4 resumes an
`uploaded` row via `SetMetadataAndTransfer` (needs the new `blob_object_id`), never a bare finalize.

## `remember_jobs` state machine (after the fix)

```
 pending ─► running ─► uploaded ─► done
                │          │
                └► failed ◄┘         (failed carries error_msg; may retain blob_id/object_id)

guard (upload_resume_disposition) on entry to the upload job:
   done                       → no-op (Ok)                        [never re-mint]
   uploaded + blob_object_id  → resume SetMetadataAndTransfer      [never re-mint]
   uploaded + NULL object_id  → Upload (legacy row, under lock)
   running / pending / failed → Upload
   missing / lookup error     → Upload (fail-open, lock-guarded)

failed + same idempotency_key retry → reset row to running (clear blob cols/error), re-spawn once
```

## Per-commit walkthrough (dependency-ordered; each compiles)

1. **`88b808aa` feat(server): schema + wallet lock pool** — migrations `010`/`011`, `db.rs` wiring
   (`db.rs::new`), `RememberRequest.idempotency_key` (`types.rs`), `AppState.wallet_lock_pool`
   (`types.rs:120`) constructed in `main.rs:726-736` (sized to `WALLET_JOB_CONCURRENCY+1`). No
   behavior change yet.
2. **`f66f80ad` fix(server): idempotent upload** (`jobs.rs`) — the internals:
   - `upload_resume_disposition` (`jobs.rs:888`) — the guard (RC-1/RC-4 classification).
   - `JobUploadLock` (`:921`) + `lock_outcome` (`:963`) — RC-3 mutex + the never-mint-without-lock rule.
   - `persist_uploaded_state` (`:987`) — RC-2 durable persist (now `&PgPool`, returns retriable).
   - `build_resume_transfer_job` (`:1016`) / `resume_metadata_and_transfer` (`:1059`) — RC-4 resume.
   - `execute_upload_and_transfer` (`:1116`, outer: takes lock → calls inner → releases on all paths)
     wrapping `execute_upload_and_transfer_locked` (`:1183`, the guard+mint+persist body).
3. **`60d94b64` feat(remember): idempotency-key dedup** (`remember.rs`) — the `remember` handler
   (`:623`) idempotency block: pre-lookup (`find_remember_job_by_key` `:815`), failed-restart via
   `should_spawn_after_reset` (`:788`), `ON CONFLICT DO NOTHING` + lost-race re-read,
   `validate_idempotency_key` (`:796`).
4. **`906da1cf` test(server): live e2e script** — `tests/live_write_idempotency_check.py`.

## Types of Changes

- [x] New feature (non-breaking) — opt-in idempotency key
- [x] Bug fix (non-breaking) — the double-mint
- [ ] Dependency update
- [ ] Documentation
- [x] Test
- [ ] Security / permission-scope change

## Benchmark impact

- Result: N/A
- [ ] Default-off / byte-identical when the flag is off

Write-path only. Recall/answer/extraction paths untouched. (The added advisory-lock acquire is a
per-upload-attempt DB op on a dedicated pool — not on the read path.)

## Testing

- [x] Tested locally
- [x] `cargo test` / harness tests pass

**Two-layer strategy** (deliberate — the mint is an external on-chain effect that can't be asserted
in-process without a live chain, and we did not want a mock seam in the production upload path):

- **Layer 1 — decision unit tests (CI, Postgres-backed).** Each funds-critical decision is a small
  testable unit: `lock_outcome` (contended / pool-error → defer-retriable, free → proceed);
  `persist_uploaded_state` (stores object id; failure → retriable); `build_resume_transfer_job`
  (routes to `SetMetadataAndTransfer`, not an index/finalize); `should_spawn_after_reset`
  (exactly-once); `upload_resume_disposition` per state; `JobUploadLock` mutual-exclusion; the
  idempotency-collapse / keyless / failed-reset / validation tests.
- **Layer 2 — live e2e script** `tests/live_write_idempotency_check.py` (manual/staging, not CI):
  same-key → same job/blob, concurrent → one job, keyless → distinct, optional failed-key restart.

Run: `cargo test --bin memwal-server` against a live Postgres (`DATABASE_URL`, default
`postgresql://memwal:memwal_secret@localhost:5432/memwal`) — the DB-backed tests silently no-op
without one, so CI must provide it. Local run: **full suite 441 passed, 0 failed** (jobs 51, remember
16), build + clippy clean (no new lints vs the pre-existing lib/bin baseline).

**Layer-1 test → the regression it catches** (each would fail if the fix regressed — not a trivial
pass):

| Test (`jobs::tests` / `routes::remember::tests`) | Catches |
|---|---|
| `upload_resume_resumes_transfer_when_uploaded_with_object_id` | guard sending an `uploaded`+object_id job to re-upload (re-mint) or to a bare finalize |
| `upload_resume_reuploads_when_uploaded_without_object_id` | guard trying to resume a legacy NULL-object_id row it can't complete |
| `upload_resume_is_noop_when_already_done` / `..._uploads_for_pending_running_or_missing_blob` | guard skipping a real write, or re-minting a done job |
| `upload_lock_is_mutually_exclusive_per_job` | the advisory-lock primitive failing to serialize same-job attempts |
| `lock_outcome_defers_when_contended_and_fails_closed_on_error` | a contended/pool-error attempt being allowed to **mint** instead of deferring |
| `lock_outcome_proceeds_when_acquired` | a free lock wrongly blocking the legitimate attempt |
| `persist_uploaded_state_stores_object_id_and_is_retriable_on_failure` | reverting to fire-and-forget persist (RC-2 re-mint) or dropping the object_id write (breaks RC-4) |
| `resume_builds_a_set_metadata_and_transfer_job_with_object_id` | resume routing to an index/finalize instead of the transfer recovery (RC-4 premature-done) |
| `find_by_key_returns_existing_and_collapses_same_key_retry` | ON CONFLICT not deduping → a duplicate keyed job |
| `keyless_writes_are_never_deduped` | the partial index wrongly constraining keyless writes |
| `failed_job_reset_flips_only_failed_and_clears_state` | reset firing on a non-failed row, or not clearing stale blob/error |
| `spawn_only_when_reset_flipped_exactly_one_row` | a concurrent double-reset double-spawning (two paid writes) |
| `idempotency_key_{none,empty,over_max,at_max}` | unbounded / malformed key input |

**Honest coverage boundary:** the funds-critical *decision logic* is unit-tested (above); the *full
`execute_upload_and_transfer` / handler thread* is not exercised in-process (no test builds a full
`AppState`, and `upload_blob` is a hard-coded call — we deliberately avoided adding a mock seam to the
prod upload path), so the wiring between the tested decisions is verified by code review + Layer 2.
The *end-to-end* "no second paid blob" under a real crash / forced orphan-reenqueue is only observable
live — Layer 2 covers the client-visible half; injected-crash chaos is a follow-up, not in this PR.

## Checklist

- [x] Docs updated if needed (n/a in-repo; design/root-cause docs live in the methodology trail)
- [x] Respects the privacy floor (no new server-readable plaintext index)

## Related

- Closes #477
- Linear: WALM-320

## Additional Notes

**Migrations (2, both additive) — per-statement lock analysis:**

| Statement | Lock / cost | Notes |
|---|---|---|
| `010`: `ADD COLUMN idempotency_key TEXT` (nullable, no default) | metadata-only (PG 11+) | no table rewrite/scan |
| `010`: `CREATE UNIQUE INDEX … (owner, idempotency_key) WHERE … IS NOT NULL` | **`SHARE` lock — blocks INSERT/UPDATE/DELETE on `remember_jobs` for the build duration** | plain (non-`CONCURRENTLY`) build scans the table; `remember_jobs` is **unpruned/large** → a deploy-time **write stall on the whole remember path**. This is the one operational item to decide. |
| `011`: `ADD COLUMN blob_object_id TEXT` (nullable, no default) | metadata-only (PG 11+) | safe regardless of table size; no new index |

Both files are `IF NOT EXISTS`, re-runnable, and wired into `db.rs::new` (applied in order after
`009`). The `010` ON CONFLICT partial-index semantics were verified end-to-end with a live `psql`
probe (null-key coexist / same-`(owner,key)` no-op / cross-owner allowed). **Recommend `pg_dump`
before running against any shared/prod DB** (project rule).

Options for the `010` index on prod: (a) accept the brief write-stall if `remember_jobs` is small
enough; (b) build the index out-of-band with `CREATE UNIQUE INDEX CONCURRENTLY` **before** shipping
the app that assumes it (note `CONCURRENTLY` can't run inside a txn block). Needs a prod row count to
decide.

**Open decisions for reviewers / the team:**
1. **Deploy plan for the 010 index** — accept the brief write-lock, or pre-build `CONCURRENTLY`
   out-of-band? (Depends on prod `remember_jobs` size.)
2. **`remember_bulk` is intentionally NOT covered** — it inserts `remember_jobs` rows but takes no
   idempotency key. Out of scope for #477 (single-write ambiguous-timeout case); flag as a possible
   follow-up. Do not read this PR as protecting bulk.
3. **SDK** — the server accepts `idempotency_key` but the SDK doesn't send one yet; an SDK change to
   auto-generate/pass a key is a follow-up.

**Rollback:** additive schema + default-off request field → rollback = revert the app; the two
nullable columns can stay harmlessly.

**Commits (4, dependency-ordered, each compiles):** schema+pool plumbing → internal double-mint
safety (RC-1..4) → client idempotency key (+RC-5) → live e2e script.

### DIFF
--- a/packages/python-sdk-memwal/memwal/client.py
+++ b/packages/python-sdk-memwal/memwal/client.py
@@ -30,6 +30,7 @@ import base64
 import json
 import random
 import time
+import uuid
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, Sequence, Tuple, TypeVar, Union
 
@@ -223,6 +224,9 @@ class MemWal:
         self._session_build_task: Optional[asyncio.Task[str]] = None
         self._relayer_version_metadata: Optional[Dict[str, Any]] = None
         self._compatibility_lock: Optional[asyncio.Lock] = None
+        # Preserve a generated key across an ambiguous transport failure. A
+        # subsequent identical call then collapses onto the accepted paid job.
+        self._pending_remember_keys: Dict[str, str] = {}
 
     @classmethod
     def create(
@@ -281,7 +285,10 @@ class MemWal:
     # ============================================================
 
     async def remember(
-        self, text: str, namespace: Optional[str] = None
+        self,
+        text: str,
+        namespace: Optional[str] = None,
+        idempotency_key: Optional[str] = None,
     ) -> RememberAcceptedResult:
         """Submit a remember request and return as soon as the server accepts it.
 
@@ -299,12 +306,27 @@ class MemWal:
             :class:`RememberAcceptedResult` with ``job_id`` and initial
             status (``"pending"``).
         """
+        resolved_namespace = namespace or self._namespace
+        request_identity = f"{resolved_namespace}\0{text}"
+        generated_key = idempotency_key is None
+        resolved_key = idempotency_key or self._pending_remember_keys.get(request_identity)
+        if resolved_key is None:
+            resolved_key = str(uuid.uuid4())
+        if generated_key:
+            self._pending_remember_keys[request_identity] = resolved_key
+
         data = await self._signed_request(
             "POST",
             "/api/remember",
-            {"text": text, "namespace": namespace or self._namespace},
+            {
+                "text": text,
+                "namespace": resolved_namespace,
+                "idempotency_key": resolved_key,
+            },
             accepted_statuses=(200, 202),
         )
+        if generated_key:
+            self._pending_remember_keys.pop(request_identity, None)
         return RememberAcceptedResult(
             job_id=data["job_id"],
             status=data.get("status", "pending"),
@@ -312,9 +334,12 @@ class MemWal:
 
     # Alias for parity with TS SDK ``rememberAsync``.
     async def remember_async(
-        self, text: str, namespace: Optional[str] = None
+        self,
+        text: str,
+        namespace: Optional[str] = None,
+        idempotency_key: Optional[str] = None,
     ) -> RememberAcceptedResult:
-        return await self.remember(text, namespace)
+        return await self.remember(text, namespace, idempotency_key)
 
     async def wait_for_remember_job(
         self,
@@ -380,6 +405,7 @@ class MemWal:
         namespace: Optional[str] = None,
         poll_interval_ms: int = 1500,
         timeout_ms: int = 60_000,
+        idempotency_key: Optional[str] = None,
     ) -> RememberResult:
         """Submit a remember and wait for the background worker to finish.
 
@@ -388,12 +414,24 @@ class MemWal:
         ``failed``). Mirrors TS ``rememberAndWait``.
         """
 
-        accepted = await self.remember(text, namespace)
-        return await self.wait_for_remember_job(
+        resolved_namespace = namespace or self._namespace
+        request_identity = f"{resolved_namespace}\0{text}"
+        generated_key = idempotency_key is None
+        resolved_key = idempotency_key or self._pending_remember_keys.get(request_identity)
+        if resolved_key is None:
+            resolved_key = str(uuid.uuid4())
+        if generated_key:
+            self._pending_remember_keys[request_identity] = resolved_key
+
+        accepted = await self.remember(text, resolved_namespace, resolved_key)
+        result = await self.wait_for_remember_job(
             accepted.job_id,
             poll_interval_ms=poll_interval_ms,
             timeout_ms=timeout_ms,
         )
+        if generated_key:
+            self._pending_remember_keys.pop(request_identity, None)
+        return result
 
     # ============================================================
     # Bulk remember (ENG-1408)
@@ -1285,16 +1323,22 @@ class MemWalSync:
             return asyncio.run(coro)
 
     def remember(
-        self, text: str, namespace: Optional[str] = None
+        self,
+        text: str,
+        namespace: Optional[str] = None,
+        idempotency_key: Optional[str] = None,
     ) -> RememberAcceptedResult:
         """Synchronous version of :meth:`MemWal.remember` (async accept)."""
-        return self._run(self._inner.remember(text, namespace))
+        return self._run(self._inner.remember(text, namespace, idempotency_key))
 
     # Alias for parity with TS SDK ``rememberAsync``.
     def remember_async(
-        self, text: str, namespace: Optional[str] = None
+        self,
+        text: str,
+        namespace: Optional[str] = None,
+        idempotency_key: Optional[str] = None,
     ) -> RememberAcceptedResult:
-        return self._run(self._inner.remember_async(text, namespace))
+        return self._run(self._inner.remember_async(text, namespace, idempotency_key))
 
     def wait_for_remember_job(
         self,
@@ -1317,6 +1361,7 @@ class MemWalSync:
         namespace: Optional[str] = None,
         poll_interval_ms: int = 1500,
         timeout_ms: int = 60_000,
+        idempotency_key: Optional[str] = None,
     ) -> RememberResult:
         """Synchronous version of :meth:`MemWal.remember_and_wait`."""
         return self._run(
@@ -1325,6 +1370,7 @@ class MemWalSync:
                 namespace,
                 poll_interval_ms=poll_interval_ms,
                 timeout_ms=timeout_ms,
+                idempotency_key=idempotency_key,
             )
         )
 
--- a/packages/sdk/src/memwal.ts
+++ b/packages/sdk/src/memwal.ts
@@ -189,6 +189,13 @@ export class MemWal {
     private sessionBuildPromise: Promise<string> | null = null;
     /** Single-flight guard so concurrent requests share one compatibility probe. */
     private compatibilityPromise: Promise<RelayerVersionMetadata> | null = null;
+    /**
+     * Keep a generated idempotency key while a remember request has no
+     * acknowledged response. If the transport times out after the server
+     * accepted the write, the caller's next identical attempt reuses the key
+     * and collapses onto the original paid job.
+     */
+    private pendingRememberKeys = new Map<string, string>();
 
     private constructor(config: MemWalConfig) {
         this.privateKey = typeof config.key === "string" ? hexToBytes(config.key) : config.key;
@@ -227,6 +234,7 @@ export class MemWal {
         this.serverConfig = null;
         this.relayerVersionMetadata = null;
         this.compatibilityPromise = null;
+        this.pendingRememberKeys.clear();
     }
 
     // ============================================================
@@ -236,13 +244,27 @@ export class MemWal {
     /**
      * Submit a remember request and return as soon as the server accepts the job.
      */
-    async rememberAsync(text: string, namespace?: string): Promise<RememberAcceptedResult> {
-        return this.signedRequest<RememberAcceptedResult>(
+    async rememberAsync(
+        text: string,
+        namespace?: string,
+        options: { idempotencyKey?: string } = {},
+    ): Promise<RememberAcceptedResult> {
+        const resolvedNamespace = namespace ?? this.namespace;
+        const requestIdentity = `${resolvedNamespace}\0${text}`;
+        const generatedKey = options.idempotencyKey === undefined;
+        const idempotencyKey = options.idempotencyKey
+            ?? this.pendingRememberKeys.get(requestIdentity)
+            ?? crypto.randomUUID();
+        if (generatedKey) this.pendingRememberKeys.set(requestIdentity, idempotencyKey);
+
+        const accepted = await this.signedRequest<RememberAcceptedResult>(
             "POST",
             "/api/remember",
-            { text, namespace: namespace ?? this.namespace },
+            { text, namespace: resolvedNamespace, idempotency_key: idempotencyKey },
             [200, 202],
         );
+        if (generatedKey) this.pendingRememberKeys.delete(requestIdentity);
+        return accepted;
     }
 
     /**
@@ -340,10 +362,22 @@ export class MemWal {
     async rememberAndWait(
         text: string,
         namespace?: string,
-        opts: { pollIntervalMs?: number; timeoutMs?: number } = {},
+        opts: { pollIntervalMs?: number; timeoutMs?: number; idempotencyKey?: string } = {},
     ): Promise<RememberResult> {
-        const accepted = await this.rememberAsync(text, namespace);
-        return this.waitForRememberJob(accepted.job_id, opts);
+        const resolvedNamespace = namespace ?? this.namespace;
+        const requestIdentity = `${resolvedNamespace}\0${text}`;
+        const generatedKey = opts.idempotencyKey === undefined;
+        const idempotencyKey = opts.idempotencyKey
+            ?? this.pendingRememberKeys.get(requestIdentity)
+            ?? crypto.randomUUID();
+        if (generatedKey) this.pendingRememberKeys.set(requestIdentity, idempotencyKey);
+
+        const accepted = await this.rememberAsync(text, resolvedNamespace, { idempotencyKey });
+        const completed = await this.waitForRememberJob(accepted.job_id, opts);
+        // Clear only after terminal success. A polling timeout/transport failure
+        // keeps the key so retrying the high-level operation reuses the same job.
+        if (generatedKey) this.pendingRememberKeys.delete(requestIdentity);
+        return completed;
     }
 
     /**
@@ -355,8 +389,12 @@ export class MemWal {
      * @param text - The text to remember
      * @param namespace - Optional namespace override
      */
-    async remember(text: string, namespace?: string): Promise<RememberAcceptedResult> {
-        return this.rememberAsync(text, namespace);
+    async remember(
+        text: string,
+        namespace?: string,
+        options: { idempotencyKey?: string } = {},
+    ): Promise<RememberAcceptedResult> {
+        return this.rememberAsync(text, namespace, options);
     }
 
     /**
--- a/packages/sdk/test/remember-idempotency.test.mjs
+++ b/packages/sdk/test/remember-idempotency.test.mjs
@@ -1,0 +1,64 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { MemWal } from "../dist/memwal.js";
+
+const originalFetch = globalThis.fetch;
+
+test.afterEach(() => {
+    globalThis.fetch = originalFetch;
+});
+
+test("rememberAndWait reuses its generated key and job after polling timeout", async () => {
+    const posted = [];
+    let jobPolls = 0;
+    globalThis.fetch = async (url, init = {}) => {
+        const path = new URL(url).pathname;
+        if (path === "/version") {
+            return Response.json({
+                apiVersion: "1.0.0",
+                relayerVersion: "1.0.0",
+                minSupportedSdk: { typescript: "0.0.4" },
+            });
+        }
+        if (path === "/api/config") {
+            return Response.json({ packageId: "0x1", network: "testnet" });
+        }
+        if (path === "/api/remember" && init.method === "POST") {
+            posted.push(JSON.parse(init.body));
+            return Response.json({ job_id: "stable-job", status: "pending" }, { status: 202 });
+        }
+        if (path === "/api/remember/stable-job") {
+            jobPolls += 1;
+            if (jobPolls < 2) return Response.json({ job_id: "stable-job", status: "pending" });
+            return Response.json({
+                job_id: "stable-job",
+                status: "done",
+                blob_id: "blob-1",
+                owner: "0x1",
+                namespace: "default",
+            });
+        }
+        throw new Error(`unexpected request ${path}`);
+    };
+
+    const client = MemWal.create({
+        key: new Uint8Array(32).fill(1),
+        accountId: "0x1",
+        serverUrl: "https://relayer.example",
+    });
+    client.buildSealSession = async () => "test-session";
+
+    await assert.rejects(
+        client.rememberAndWait("same memory", undefined, { pollIntervalMs: 0, timeoutMs: 1 }),
+        /timed out/,
+    );
+    const result = await client.rememberAndWait("same memory", undefined, {
+        pollIntervalMs: 0,
+        timeoutMs: 100,
+    });
+
+    assert.equal(result.job_id, "stable-job");
+    assert.equal(posted.length, 2);
+    assert.equal(posted[0].idempotency_key, posted[1].idempotency_key);
+});
--- a/services/server/migrations/012_remember_write_idempotency.sql
+++ b/services/server/migrations/012_remember_write_idempotency.sql
@@ -1,0 +1,22 @@
+-- Durable idempotency and recovery state for paid remember writes.
+-- All columns are nullable/additive so existing rows remain readable. Legacy
+-- paid rows without an original preparation payload fail closed and require
+-- manual reconciliation rather than unsafe re-encryption.
+ALTER TABLE remember_jobs
+    ADD COLUMN IF NOT EXISTS idempotency_key TEXT,
+    ADD COLUMN IF NOT EXISTS request_fingerprint TEXT,
+    ADD COLUMN IF NOT EXISTS blob_object_id TEXT,
+    ADD COLUMN IF NOT EXISTS upload_wallet_index INTEGER,
+    ADD COLUMN IF NOT EXISTS upload_wallet_address TEXT,
+    ADD COLUMN IF NOT EXISTS upload_execution_identity JSONB,
+    ADD COLUMN IF NOT EXISTS upload_resume_step JSONB,
+    ADD COLUMN IF NOT EXISTS upload_register_transaction JSONB,
+    ADD COLUMN IF NOT EXISTS prepare_claimed_at TIMESTAMPTZ,
+    ADD COLUMN IF NOT EXISTS prepare_claim_token TEXT,
+    ADD COLUMN IF NOT EXISTS recovery_claimed_at TIMESTAMPTZ,
+    ADD COLUMN IF NOT EXISTS recovery_claim_token TEXT,
+    ADD COLUMN IF NOT EXISTS preparation_encrypted_b64 TEXT,
+    ADD COLUMN IF NOT EXISTS preparation_vector JSONB,
+    ADD COLUMN IF NOT EXISTS preparation_importance REAL,
+    ADD COLUMN IF NOT EXISTS preparation_account_id TEXT,
+    ADD COLUMN IF NOT EXISTS preparation_public_key TEXT;
--- a/services/server/migrations/013_remember_write_idempotency_index.sql
+++ b/services/server/migrations/013_remember_write_idempotency_index.sql
@@ -1,0 +1,4 @@
+-- Keep this standalone: CREATE INDEX CONCURRENTLY cannot run in a transaction.
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS uq_remember_jobs_owner_idempotency_key
+    ON remember_jobs (owner, idempotency_key)
+    WHERE idempotency_key IS NOT NULL;
--- a/services/server/scripts/__tests__/sidecar-find-blob-by-job.test.ts
+++ b/services/server/scripts/__tests__/sidecar-find-blob-by-job.test.ts
@@ -1,0 +1,88 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import type { Server } from "node:http";
+
+// Minimal env for booting the default-mode sidecar app. The find-blob-by-job
+// validation path returns 400 before touching the chain, so no client mocking
+// is needed here (end-to-end reconcile behavior is covered by the live script).
+const { Ed25519Keypair } = await import("@mysten/sui/keypairs/ed25519");
+process.env.SERVER_SUI_PRIVATE_KEYS = new Ed25519Keypair().getSecretKey();
+process.env.SIDECAR_AUTH_TOKEN = "find-blob-test-token";
+process.env.SUI_NETWORK = "testnet";
+process.env.SUI_GRPC_URL = "https://find-blob.testnet.example";
+process.env.WALRUS_PACKAGE_ID = `0x${"a".repeat(64)}`;
+
+const { createSidecarApp } = await import("../sidecar/app.js");
+const { sanitizeRequestId } = await import("../sidecar/log.js");
+
+// The reconcile correctness hinges on the register-side tag value
+// (sanitizeRequestId(jobId)) equaling the query-side value (also
+// sanitizeRequestId(jobId)). For the production key space — server-minted UUIDs
+// — that must be an identity, or the tag would never match and every retry would
+// silently re-mint. Pin it.
+test("a remember-job UUID survives sanitizeRequestId unchanged (tag == query)", () => {
+    for (let i = 0; i < 50; i++) {
+        const uuid = crypto.randomUUID();
+        assert.equal(sanitizeRequestId(uuid), uuid, `UUID must round-trip: ${uuid}`);
+    }
+    // And a non-conforming id sanitizes to null on BOTH sides (no silent mismatch).
+    assert.equal(sanitizeRequestId("has spaces"), null);
+    assert.equal(sanitizeRequestId("x".repeat(129)), null);
+});
+
+async function listen(): Promise<{ server: Server; baseUrl: string }> {
+    return await new Promise((resolve) => {
+        const server = createSidecarApp().listen(0, "127.0.0.1", () => {
+            const address = server.address();
+            if (address === null || typeof address === "string") throw new Error("no port");
+            resolve({ server, baseUrl: `http://127.0.0.1:${address.port}` });
+        });
+    });
+}
+
+const headers = {
+    authorization: "Bearer find-blob-test-token",
+    "content-type": "application/json",
+};
+
+test("find-blob-by-job validates owner and jobId", async () => {
+    const { server, baseUrl } = await listen();
+    try {
+        // missing owner → 400
+        const noOwner = await fetch(`${baseUrl}/walrus/find-blob-by-job`, {
+            method: "POST",
+            headers,
+            body: JSON.stringify({ jobId: "job-1" }),
+        });
+        assert.equal(noOwner.status, 400);
+
+        // missing jobId → 400
+        const noJob = await fetch(`${baseUrl}/walrus/find-blob-by-job`, {
+            method: "POST",
+            headers,
+            body: JSON.stringify({ owner: `0x${"1".repeat(64)}` }),
+        });
+        assert.equal(noJob.status, 400);
+
+        // empty body → 400
+        const empty = await fetch(`${baseUrl}/walrus/find-blob-by-job`, {
+            method: "POST",
+            headers,
+            body: "{}",
+        });
+        assert.equal(empty.status, 400);
+
+        // a jobId that can't be a tag value (fails sanitizeRequestId) → 400, so
+        // the query can never silently diverge from the stored tag value.
+        const badJob = await fetch(`${baseUrl}/walrus/find-blob-by-job`, {
+            method: "POST",
+            headers,
+            body: JSON.stringify({ owner: `0x${"1".repeat(64)}`, jobId: "has spaces & symbols!" }),
+        });
+        assert.equal(badJob.status, 400);
+    } finally {
+        await new Promise<void>((resolve, reject) =>
+            server.close((e) => (e ? reject(e) : resolve()))
+        );
+    }
+});
--- a/services/server/scripts/__tests__/sidecar-find-blob-matching.test.ts
+++ b/services/server/scripts/__tests__/sidecar-find-blob-matching.test.ts
@@ -1,0 +1,83 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+// P1: the actual matching logic of findBlobObjectByJob (GH #477 reconcile). We
+// stub the module-level suiClient's listOwnedObjects + getDynamicField so we can
+// drive real fixtures through the scan/tag-match without a chain. This is the
+// funds-critical part: a wrong match (or a swallowed metadata error) → the
+// relayer either re-mints or adopts the wrong user's blob.
+
+const { Ed25519Keypair } = await import("@mysten/sui/keypairs/ed25519");
+const serverKey = new Ed25519Keypair();
+const serverAddr = serverKey.toSuiAddress();
+process.env.SERVER_SUI_PRIVATE_KEYS = serverKey.getSecretKey();
+process.env.SUI_NETWORK = "testnet";
+process.env.SUI_GRPC_URL = "https://match.testnet.example";
+process.env.WALRUS_PACKAGE_ID = `0x${"a".repeat(64)}`;
+
+const { bcs } = await import("@mysten/sui/bcs");
+const { normalizeSuiAddress } = await import("@mysten/sui/utils");
+const { suiClient } = await import("../sidecar/clients.js");
+const { findBlobObjectByJob } = await import("../sidecar/routes/walrus-query.js");
+
+const METADATA_BCS = bcs.struct("Metadata", { metadata: bcs.map(bcs.string(), bcs.string()) });
+const USER = normalizeSuiAddress(`0x${"1".repeat(64)}`);
+
+// blob_id on chain is a U256 decimal; use a value the sidecar can convert.
+function makeBlob(objectId: string, rawBlobId: string) {
+    return { objectId, json: { blob_id: rawBlobId } };
+}
+
+// tags: Map of key→value → BCS bytes the getDynamicField mock returns.
+function metadataField(tags: Record<string, string>) {
+    const map = new Map(Object.entries(tags));
+    const value = METADATA_BCS.serialize({ metadata: map }).toBytes();
+    return { dynamicField: { value: { bcs: value } } };
+}
+
+// Install stubs; each test sets these two closures.
+let ownedByAddr: (owner: string) => any[];
+let metaByObject: (objectId: string) => any;
+(suiClient as any).listOwnedObjects = async ({ owner }: { owner: string }) => ({
+    objects: ownedByAddr(owner),
+    hasNextPage: false,
+    cursor: null,
+});
+(suiClient as any).getDynamicField = async ({ parentId }: { parentId: string }) => metaByObject(parentId);
+
+const BLOB_RAW = "123456789012345678901234"; // long decimal → treated as U256
+
+test("found: a server-wallet blob tagged with this job AND owner → found", async () => {
+    ownedByAddr = (o) => (o === serverAddr ? [makeBlob("0xobjA", BLOB_RAW)] : []);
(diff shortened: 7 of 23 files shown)


ducnmm (changes requested):
Thanks for the thorough investigation. Requesting changes because duplicate-mint paths remain:

1. `persist_uploaded_state()` failure returns `Transient`, but the row still has no blob state; the next Apalis attempt sees `running` and uploads again. A process crash between mint and persist has the same issue. This needs downstream/sidecar idempotency keyed by stable `job_id` or reconciliation of the already-minted transaction.
2. `upload_resume_disposition()` converts every DB lookup error via `.ok().flatten()` into `Upload`. A transient read failure after a successful mint therefore re-mints; lookup errors must fail closed/retry without uploading.
3. A same-key retry of `failed` clears `blob_id`/`blob_object_id` and respawns upload. But recovery handoff failures and the stale sweeper can mark an already-`uploaded` row failed, so this path discards proof of the paid mint and duplicates it. Resume when blob state exists.
4. `uploaded` with `blob_object_id = NULL` also maps to `Upload`, while normal successful sidecar responses allow `object_id: None`; an index failure after such a success can re-mint. This should resume finalization, not upload.

Also recommend storing a request fingerprint and returning 409 when the same idempotency key is reused with different text/namespace. The SDK does not send the new key yet, so the client-facing case in #477 remains incomplete.

hungtranphamminh:
Thanks — this was a genuinely sharp review; all four findings were valid and are now fixed, plus the fingerprint recommendation. Details:

**1 — crash/persist-window re-mint (reconcile the already-minted tx).** Fixed. A `running`+NULL-blob row (a prior attempt was in-flight) now reconciles on-chain before re-uploading: the live register tags the blob with `memwal_job_id` (+ `memwal_owner`), and a new sidecar `find-blob-by-job` endpoint finds it by `(job, owner)` with no blob id. Important detail your comment prompted: the blob is registered **owned by the server signer wallet** and only transferred to the user at the end, so the lookup scans the **server pool wallets** (where a crash-window not-yet-transferred blob lives) plus the user, matching on both `memwal_job_id` **and** `memwal_owner` (so it can't adopt another user's blob). If a match is found → adopt (persist + resume), never re-mint; if the scan can't be conclusive (owner has more blobs than the cap) → 409 → the relayer fails closed (retries), never a false "not minted". Also: fresh jobs now insert `pending` rather than `running`, so this on-chain round-trip only fires on a real retry, not on the happy path.

Honest residual: this reconciles *after* the mint, so it's subject to read-your-writes lag — if the retry's chain read beats the just-minted blob's visibility on the queried replica, it can still re-mint. It removes the common crash-then-retry case and is fail-closed on "can't tell".

**Next step to close the residual to zero (depends on your #527).** The fully race-free close is to make the register mint idempotent per job via deterministic-digest replay: persist the prepared register `{digest, bytes}` *before* submit, and on retry replay the exact bytes so `getTransaction(digest)` short-circuits to the already-minted tx instead of re-minting. The good news is most of this already exists — the `upload-step-v3` journal route already implements exactly that protocol (prepare → persist digest → execute-with-short-circuit, fail-closed on ambiguity). Two things it needs: (1) the relayer currently calls the atomic single-shot `/walrus/upload`, so it would move to the journaled prepare→persist→execute-register flow, plus one schema column to persist the register digest before the mint; and (2) **the register digest is only deterministic under address-balance gas — which is exactly what your #527 introduces** (`enforceAddressBalanceCoinIntents` + `setGasPayment([])` removes the owned-coin gas selection that would otherwise change the digest on replay). So once #527 lands, this becomes a bounded follow-up (medium effort — integration, not new protocol; but paid-mint-path blast radius, so same fail-closed rigor). Noted as a follow-up, with the #527 dependency called out.

**2 — `.ok().flatten()` re-mint on a transient read error.** Fixed. `upload_resume_disposition` no longer collapses a DB lookup error into `Upload`; it returns an `Indeterminate` disposition that the caller turns into a retriable error **without uploading**.

**3 — failed-key retry discarding a paid mint.** Fixed. The failed-restart now only resets a `failed` row with **no** blob (guarded `AND blob_id IS NULL`); a `failed` row that still carries a blob (marked failed by a recovery-handoff failure / stale sweeper) is preserved, and the guard resumes it rather than re-uploading. (The wallet-error path already leaves `blob_id`/`blob_object_id` intact when it marks a row failed, so the proof of the paid mint isn't lost.)

**4 — `uploaded`+NULL `object_id` re-mint.** Fixed. Any `uploaded` row now resumes — with an object id it resumes metadata/transfer, without one it resumes indexing — and never falls through to upload.

**Rec (fingerprint + 409).** Added. A `request_fingerprint` (sha256 of text+namespace) is stored with the idempotency key; a same-key request whose content fingerprint differs now returns **409** instead of silently collapsing onto the original write. Enforced on both the pre-lookup hit and the concurrent lost-race path.

**Rec (SDK doesn't send the key).** Correct — that's the client-facing half of #477 and it's tracked as a follow-up (SDK change to generate/send a key). The server half is here; I've noted in the PR description that #477 is only fully closed once the SDK ships it.

New tests cover each: the reconcile client's fail-closed contract (409/5xx → retry, never re-mint), the sidecar matching (found / wrong-owner-not-adopted / indeterminate / metadata-error-propagates), the disposition variants, and the fingerprint. Re-requesting review.

ducnmm (changes requested):
The latest commits address the previously reported guard/read-error and persisted-blob cases, but I still don't think this is safe to merge. The end-to-end at-most-once guarantee and recovery liveness remain incomplete:

1. **A sidecar timeout can still produce a duplicate paid mint.** The relayer can time out or lose the connection while the Express handler continues processing. The advisory lock is then released. A second attempt can run `find_blob_by_job` before the first registration is visible on-chain, receive `Ok(None)`, and call `upload_blob` again. Tagging and scanning the chain only reconciles an already-visible mint; it does not reserve `job_id` or deduplicate an in-flight sidecar operation. This needs a durable sidecar reservation/journal keyed by the stable job ID (for example `in_progress` / `completed` with the prepared transaction/result), rather than treating one `not_found` scan as proof that no operation is still running.

2. **Reconciliation can become permanently indeterminate once a scanned wallet has more than 500 blobs.** `scanOwnerForJobBlob` only examines `blobs.slice(0, cap)`. If the target is outside that first page, `findBlobObjectByJob` returns `indeterminate`; every retry repeats the same scan and 409, so the write can never reconcile or proceed. Please paginate to exhaustion, maintain a durable job-to-blob index, or otherwise provide a lookup that can eventually return a conclusive answer.

3. **A failed keyed job that already has `blob_id` is preserved but never resumed.** The handler returns the existing `failed` row without spawning preparation or a recovery job. This avoids re-minting but pins the write in `failed` forever. A same-key retry should enqueue a path whose upload guard resumes transfer/indexing while preserving the recorded blob.

4. **A keyed `pending` job can be stranded between the DB insert and `tokio::spawn`.** If the process exits after inserting the row but before `spawn_prepare_remember_job`, a retry finds the existing `pending` row and returns it without restarting any work. The same issue exists when resetting `failed` to `running` before the in-process spawn. Preparation needs a durable queue/claim, or retries need a safe way to reclaim stale pre-wallet states.

5. **The SDK still does not send `idempotency_key`.** Therefore the original client-facing ambiguous-timeout flow in #477 remains unsafe for normal SDK callers: a retry creates a different job ID, and the per-job server guard cannot correlate it. Please include the SDK plumbing (and tests) or explicitly split the issue and keep #477/WALM-320 open until the end-to-end client path is shipped.

Also, the production plan for the non-concurrent partial unique-index migration remains unresolved; please settle that before deployment.

hungtranphamminh:
Thanks — this is the right level of scrutiny for a funds path, and I agree the current state isn't the
end state. Let me lay out where each point lands, because the honest answer is that two of these
(1 and 4) can't be *fully* closed inside this PR without your #527, and I want to make the dependency
legible rather than hand-wave it.

First, the split:

- **1 & 4** — real, and the correct fix is a durable job-keyed journal (prepare → persist → replay the
  same register digest). That fix is **hard-blocked on #527**; explained below.
- **2 & 3** — real liveness bugs we introduced (fail-closed that can never resolve; a `failed` row
  pinned forever). No dependency on anything — fixing in-place in this PR.
- **5** — correct; adding the SDK plumbing to this PR so #477 closes end-to-end.
- **migration** — answering the deployment story below.

---

### Why 1 & 4 need the durable journal, and why the journal needs #527

**What we do today (and why it's only best-effort).** The relayer calls the atomic single-shot
`POST /walrus/upload` (`storage/walrus.rs:250`). That endpoint builds+signs+submits register/certify/
metadata in one HTTP round-trip and returns only the final `blob_id`/`object_id` — there's no
pre-submit anchor. When a prior attempt was in-flight, we take the `running`+NULL-blob row and run
`find_blob_by_job` (`jobs.rs:1315` `ReconcileThenUpload`) to scan the chain for an already-registered
blob tagged with this job id, and adopt it instead of re-minting.

The limitation you identified is exactly right: **tag-scan reconcile can only catch a mint that is
already *visible on chain*.** It cannot deduplicate:

- **(your finding 1) an in-flight sidecar op.** The relayer times out / drops the connection while Express is
  still registering; the advisory lock releases; a second attempt runs `find_blob_by_job` *before* the
  first registration is visible on the queried replica → `Ok(None)` → `upload_blob` again. The scan
  isn't proof "nothing was minted", only "nothing minted is visible *yet*".
- **(your finding 4) the pre-wallet crash window.** Process exits between the DB insert and `tokio::spawn`
  (or between the `failed`→`running` reset and its in-process spawn); a retry finds the existing
  `pending`/`running` row and returns it without restarting work → the write stalls.

**The principled close is what you described** — a durable reservation keyed by the stable job id: the
register transaction is *prepared and persisted before submit*, and any retry *replays the exact same
signed bytes*, so it can only re-submit the identical transaction digest rather than mint a second
one. On the sidecar this already exists: `prepareRegisterTransaction` builds the register tx and
returns `{transactionBytes, signature, digest}`; `executePreparedRegisterTransaction`
(`walrus-upload-journal.ts:320`) does a `getTransaction(digest)` first and short-circuits to the
already-submitted tx if present, otherwise submits the exact signed bytes — with the ambiguous cases
(`NoSideEffectError`, expired-but-unconfirmed) raised rather than re-minted. That's a finished,
funds-safe implementation of at-most-once register. The relayer simply doesn't call it yet.

**So why is this blocked on #527?** Because digest-replay is only sound when the register bytes are
*byte-identical across attempts*, and that is only true under address-balance gas — which is what #527
introduces. Concretely, `assertAddressBalanceRegisterTransaction`
(`walrus-upload-journal.ts:228`) *enforces*: `gasData.payment` empty, no `GasCoin` argument, no
owned-coin inputs — i.e. gas + tip paid from the **address balance** (an amount-based
`FundsWithdrawal`), not a selected owned SUI coin object.

Under the **current owned-coin gas model**, the SDK pays gas by selecting an owned coin and baking its
object-ref into `gasData.payment`. On a retry, if the wallet's coin set has shifted (a coin spent,
merged, or simply a different one selected), the tx bytes differ → the digest differs → the
`getTransaction(digest)` short-circuit misses → **re-mint.** The journal's own assertion literally
*throws* on a non-address-balance register tx, precisely because that's the only shape whose digest is
replay-stable. #527's `enforceAddressBalanceCoinIntents` + `setGasPayment([])` is what makes the live
register path emit that shape.

Net: I can't land the durable-journal close for 1 & 4 in this PR without #527, and I'd rather say that
plainly than ship a journal that only works once #527 exists.

---

### The plan for 1 & 4 (once #527 is in)

1. **Relayer adopts the journal register flow.** Replace the single-shot register in the upload job
   (`jobs.rs` upload path) with prepare → persist → execute-register: call the sidecar to get
   `{digest, signed bytes}` *before* submit, persist them, and on retry replay the same bytes and rely
   on the sidecar's `getTransaction(digest)` dedup — retiring the `ReconcileThenUpload` tag-scan
   (`jobs.rs:1315`) for the register step.
2. **Schema anchor.** One additive migration to persist the prepared register `digest` (and the signed
   bytes / `maxEpoch`) on the job row *before* the mint — the crash-safe reservation. This is what
   closes your finding-4 pre-wallet window: the durable row *is* the claim, so a retry reclaims it instead of
   returning a do-nothing `pending`.
3. **Honor the ambiguity classification.** The relayer must map the journal's `NoSideEffectError` /
   expired-`UNAVAILABLE` outcomes to retry-without-mint vs. re-prepare correctly — getting this wrong
   reopens the funds bug, so it gets its own tests (crash-between-persist-and-submit, retry-after-
   timeout, cross-epoch-expiry).

That's integration of an already-reviewed mechanism, not a new protocol — but it's the paid-mint path,
so it carries the same fail-closed rigor and I'd want it reviewed as carefully as this PR.

---

### 2 & 3 — fixing in-place in this PR (no dependency)

**2 — reconcile permanently indeterminate past 500 blobs.** Agreed, and this one's ours: a fail-closed
that can never become conclusive is a liveness bug. `scanOwnerForJobBlob` only examines
`blobs.slice(0, cap)`, so a target outside the first page latches `indeterminate` → perpetual 409 with
no path forward. Fix: paginate the owner scan to exhaustion so the lookup always returns a conclusive
found / not-found. (This lookup goes away for the register step once the journal lands per above, but
it needs to be correct in the interim and for the non-register scans, so it's worth fixing now.)

**3 — a `failed` row with a `blob_id` preserved but never resumed.** Also ours. We stopped the re-mint
but pinned the write in `failed`. Fix: split the failed-restart path — `failed` + no blob → restart
upload (as now); `failed` + blob present → enqueue a *resume* (transfer/index the recorded blob),
never a fresh upload. So the recorded paid mint is both preserved *and* driven to completion.

---

### 5 — SDK plumbing (adding to this PR)

Correct — without the client half, a normal SDK retry gets a fresh job id and the per-job server guard
can't correlate it, so #477 isn't actually closed for real callers. I'll add the SDK change (generate
+ send a stable `idempotency_key`, plus tests) to this PR so #477 closes end-to-end rather than being
split. The server guard from this PR is the counterpart it plugs into.

---

### Migration (partial unique index) — going concurrent

Agreed this shouldn't lock the table on deploy. `uq_remember_jobs_owner_idempotency_key` is currently a
plain `CREATE UNIQUE INDEX`, which takes `ACCESS EXCLUSIVE` on `remember_jobs` for the build. Plan:

- The index moves to **`CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS`**, built **out-of-band against
  prod before the code deploy** (a documented ops step). `CONCURRENTLY` can't run inside a transaction
  block, and our boot migrations execute in one, so the in-code migration becomes an idempotent
  `IF NOT EXISTS` guard that no-ops once ops has built it (and stays safe on fresh/dev DBs where it
  builds a tiny empty index).
- The additive `ADD COLUMN ... idempotency_key` stays in the normal migration (a nullable column add
  is a fast metadata-only change, no rewrite).
- Rollout order: build the index concurrently on prod → deploy the code (which reads the existing
  index) → the guard confirms it's present. If the concurrent build ever fails midway it leaves an
  `INVALID` index, which we'd drop and rebuild — documented in the deploy notes.

(Branch note: this rebased onto latest dev; the idempotency migration renumbered to 011 since dev took
010 for the restore-blobs work — the concurrent-index change lands in that file.)

---

### Sequencing — your call

2, 3, 5 I'll do in this PR now. For 1 & 4, the clean options are:

- **Land #527 first, then I stack the journal adoption + schema anchor on this branch** — closes 1 & 4
  properly here. (My preference; keeps the whole idempotency story in one place.)
- **Split 1 & 4 into a dedicated follow-up that lands right after #527** — keeps this PR smaller and
  unblocked on the interim reconcile + 2/3/5, with 1 & 4 tracked as the race-free finisher.

Either way the tag-scan reconcile stays as the best-effort interim (with the residual called out) until
the journal lands. Which sequencing do you prefer — and are you good to prioritise #527 accordingly?

ducnmm (changes requested):
I re-reviewed the current head `34a5e644` end to end. The durable pre-submit register journal is a meaningful improvement: the encoded state and exact signed register bytes/digest are persisted before submission, and a retry of the same wallet job replays those bytes. However, two previously raised end-to-end liveness/idempotency gaps remain, so I cannot approve yet.

1. **[P1] Default SDK retries after `rememberAndWait` timeout still create a new idempotency key/job** (`packages/sdk/src/memwal.ts:247-267,362-370`; Python `client.py:287-329,399-418`). Both clients delete the generated key immediately after the initial POST returns 202. If polling then times out—the original ambiguous client-visible failure—the caller retries `rememberAndWait(text)`, generates a new key, creates a second job, and can still purchase a second blob. The durable register journal only deduplicates retries of one stable job ID; it cannot correlate these two jobs. Python `remember_and_wait` also exposes no idempotency-key parameter. Please keep one generated key for the entire remember-and-wait operation (including retries after polling timeout), and expose/forward an explicit key consistently through async and sync APIs. Add a regression that times out after acceptance, retries the high-level call, and verifies the same key/job is reused.

2. **[P1] Existing `pending` and post-mint `failed` keyed rows are still returned without restarting recovery** (`services/server/src/routes/remember.rs:653-730`). A crash after inserting `pending` but before `spawn_prepare_remember_job` leaves a row with no durable preparation worker; every same-key retry returns that `pending` row without spawning anything. Likewise, a `failed` row with `blob_id` bypasses the only restart branch and is returned as failed; no preparation or wallet recovery is enqueued, so the preserved paid blob remains permanently stranded. This is exactly findings 3 and 4 from the prior review. The register journal closes ambiguity once the wallet upload path has prepared a transaction, but it does not make the in-process preparation handoff durable. Please add a race-safe reclaim/queue mechanism for stale pre-wallet states and enqueue the resume path for failed rows with recorded blob state.

Additionally, `git diff --check` reports a trailing blank line in `012_remember_idempotency_key.sql`; non-blocking but should be cleaned.

All 12 CI checks are green, but they do not exercise either failure sequence above. Keeping my review at changes requested until these are closed.

ducnmm (changes requested):
Re-reviewed `e937a271`. The SDK timeout-key fix is correct, the explicit key is now forwarded through Python async/sync APIs, the TS regression covers the ambiguous polling timeout, and the diff-check issue is fixed. The stale `pending` claim also provides a race-safe retry after the 60s lease expires.

One funds/data-integrity blocker remains in the new paid-blob recovery:

**[P1] Recovering `failed + blob_id` re-runs encryption and then applies the new ciphertext state to the old paid blob** (`services/server/src/routes/remember.rs:721-740`). This branch changes the row to `uploaded` and calls `spawn_persisted_remember_preparation`, which re-embeds and re-encrypts the plaintext. When that newly prepared wallet job reaches the upload guard, `uploaded + blob_id` skips upload and resumes transfer/indexing for the existing blob, but it uses the newly generated `encrypted_b64`/vector. SEAL encryption is not deterministic, so those bytes are not the ciphertext stored under the preserved `blob_id`. `resume_metadata_and_transfer` can therefore persist/warm metadata for ciphertext B while transferring/indexing blob A, producing a corrupted recovery rather than safely finishing the original paid write.

Please resume from the original durable wallet/preparation payload (or persist the original encrypted bytes/vector before wallet handoff and reconstruct the recovery job from that state). Do not re-encrypt plaintext when adopting an already-minted blob. Add a regression with deliberately different old/new ciphertext that verifies recovery uses the payload associated with the recorded blob.

The current CI run is still in progress, but this is a correctness issue independent of check status. Keeping changes requested.

ducnmm (changes requested):
Re-reviewed `1b2c3223`. Persisting and rebuilding the wallet operation from the original ciphertext/vector fixes the specific re-encryption mismatch from my last review, and `git diff --check` is clean. However, two recovery races still block approval:

1. **[P1] The 60s preparation lease is not fenced, so a slow original preparation and its replacement can enqueue different ciphertexts for one job** (`services/server/src/routes/remember.rs:150-180, 893-905`). Summarization/embed/SEAL are external operations and can legitimately exceed 60 seconds. After the lease expires, a retry starts preparation B while A is still alive. Both tasks unconditionally overwrite `preparation_*` and enqueue an `UploadAndTransfer` carrying their own non-deterministic ciphertext/vector. The wallet advisory lock only serializes them; it does not make their payloads equal. If A mints blob A and B later observes the recorded blob, B resumes metadata/indexing with ciphertext/vector B, recreating the exact blob/payload mismatch this commit intends to prevent. Please add a fencing token/generation to the preparation claim and condition both persistence and enqueue eligibility on still owning that claim (or use a durable queue with equivalent ownership semantics). Add a regression where A finishes after its lease was taken over and verify only the current claim can hand off a wallet job.

2. **[P1] Paid recovery still has a DB-state → durable-queue crash window** (`services/server/src/routes/remember.rs:736-768`). The handler first changes `failed` to `uploaded`, then calls `enqueue_persisted_paid_recovery`. If the server exits between those operations, no queue item exists. Subsequent same-key requests see `uploaded`, fall through to the generic idempotent return, and never enqueue recovery, so the paid blob is stranded permanently. The error rollback only covers a returned enqueue error, not process death. Please make the claim/handoff durable or allow a race-safe stale reclaim for `uploaded` recovery states; enqueue and state transition should have an eventual-recovery invariant across crashes. Test the persisted `uploaded`/no-queue state directly.

Also please define the behavior for pre-migration paid rows: the new `preparation_*` columns are nullable, so any existing `failed + blob_id` row will fail decoding/loading forever. Failing closed is correct for integrity, but these rows need an explicit compatible recovery/manual-remediation path rather than being silently unrecoverable.

CI for this head is still running. Keeping changes requested independent of check status.

ducnmm (changes requested):
Re-reviewed `d830d3af`. The added token correctly fences **reclaimed** preparation workers at the persistence UPDATE, and paid recovery can now reclaim a stale claim. `git diff --check` is clean. Two concurrency bugs remain:

1. **[P1] The initial preparation worker is still unfenced** (`services/server/src/routes/remember.rs:800-875,156-180`). A fresh row stores `prepare_claimed_at = NOW()` but no `prepare_claim_token`, and the initial spawn passes `None`. Its persistence predicate is `($7 IS NULL OR prepare_claim_token = $7)`, so for that initial worker it is always true. If it runs longer than 60s, retry B can claim a token and persist/enqueue payload B, while initial worker A can still subsequently overwrite the durable preparation and enqueue its different non-deterministic payload. This is the original race for every newly created job. Generate/store a token in the initial INSERT and pass it to the initial worker; do not provide an unfenced bypass for keyed jobs. The regression currently starts both workers through `claim_remember_preparation`, so it does not cover the production initial-spawn path.

   There is also still a smaller persist→enqueue TOCTOU for reclaimed workers: A can pass the fenced persistence UPDATE just before its lease expires, B can then take over/persist/enqueue, and A can enqueue afterward without another ownership check. Ownership must cover the wallet handoff, ideally by atomically coupling the fenced state transition and durable enqueue (or an equivalent protocol), not only the payload UPDATE.

2. **[P1] Publishing recovery can regress a completed job from `done` back to `uploaded`** (`services/server/src/routes/remember.rs:755-764`). The durable wallet request becomes visible as soon as `enqueue_persisted_paid_recovery` returns/pushes it. A worker may process it and mark the remember row `done` before the route executes its final UPDATE. That UPDATE only checks `id`, token, and `blob_id`; it does not guard the current status, so it overwrites `done` with `uploaded`. Later same-key retries can then reclaim/enqueue recovery repeatedly, and polling no longer reports completion. Make the publish transition monotonic (e.g. guard it against terminal status and verify the affected/current row), and add a regression simulating worker completion between enqueue and publish.

Failing closed for legacy paid rows with missing preparation is acceptable provided the manual/on-chain remediation procedure is explicitly operationalized, but it does not resolve the races above. CI is still running; keeping changes requested.

ducnmm (changes requested):
Re-reviewed `709ae9e9`. The initial route now obtains a real token, and the recovery publish predicate is monotonic with respect to `done`. `git diff --check` is clean. One funds-integrity race still blocks approval:

**[P1] The wallet-side fencing check is still a TOCTOU check performed before the upload lock / side effect** (`services/server/src/jobs.rs:487-512`, then `execute_upload_and_transfer` acquires the advisory lock and eventually mints). Sequence:

1. Worker A checks `prepare_claim_token = A` and gets `true`.
2. Before A acquires the upload lock / changes the row out of `pending`, its 60s preparation lease expires and request B successfully reclaims it, writes token B and persists/enqueues ciphertext B.
3. A continues because it never checks token A again under the lock; it can mint blob A.
4. B later runs, sees the recorded paid blob, and resumes using payload B, recreating the ciphertext/blob mismatch.

Passing the token into the durable job is necessary but the current standalone `SELECT EXISTS` does not make “still owns claim” atomic with entry into the paid wallet path. Under the per-job advisory lock, atomically consume the preparation claim / transition the row out of reclaimable `pending` with `WHERE prepare_claim_token = $token` before any upload/reconcile side effect; if zero rows are affected, drop the stale job. The claim path must then be unable to replace the token while that wallet execution owns the job. Please add an interleaving regression where reclaim happens after A's preliminary check but before A enters the locked wallet path.

Related fencing gap: preparation failures still call unconditional `mark_remember_job_failed` (`remember.rs:212-215`). A stale worker that was reclaimed can therefore overwrite B's `pending/running/done` state with `failed`. Failure publication should also be conditional on the worker's claim token (and terminal states must remain terminal).

The `paid_recovery_publish_never_regresses_done` test currently only asserts properties of a duplicated SQL string rather than executing the production query against a `done` row. The production predicate itself is correct, but a DB regression would protect it from drift.

CI is still running. Keeping changes requested until the wallet claim is consumed atomically at the side-effect boundary and stale failure publication is fenced.

ducnmm (changes requested):
Re-reviewed `16b264dd`. The claim is now consumed under the per-job advisory lock, stale preparation failure publication is token-fenced, the terminal-state test uses the DB, all 12 CI checks are green, and `git diff --check` is clean. One P1 liveness bug remains in the consume design:

**[P1] Consuming the claim makes every legitimate retry of that durable wallet job look stale** (`services/server/src/jobs.rs:487-510,1398-1423`). The first execution changes the row to `running` and clears `prepare_claim_token`. If anything after that returns `Transient`—DB/sidecar timeout, journal step retry, congestion reschedule, process crash after consume, etc.—Apalis runs the same durable `UploadAndTransfer` again with its original token. The preliminary `SELECT EXISTS(... prepare_claim_token = $token)` now returns false (and even without it, the consume UPDATE requires `status='pending'`), so the retry returns `Ok(())` without resuming. The remember row remains `running`; same-key API retries do not reclaim `running`, so the write is permanently stranded. The congestion-rescheduled operation also explicitly carries the old token and will be dropped for the same reason.

Please make claim consumption distinguish an authorized already-consumed retry from a stale superseded worker. For example, persist a stable consumed/active preparation generation on the row: atomically transition `pending + token` to `running + same token` under the upload lock; allow later executions with that same token while `running`; and only reject a different token. Since preparation reclaim is restricted to `pending/failed`, keeping the active token on `running` does not reopen takeover. Clear it only on a terminal/safely persisted handoff as appropriate. Add a regression: consume token A, simulate a transient return/process restart, then verify the same queued payload A is accepted on retry while a superseded token is rejected.

This is independent of CI status; keeping changes requested.

ducnmm (approved):
Re-reviewed `e7b3ea27` end to end. The final blocker is resolved: the active preparation generation is retained while the row is `running`, retries carrying the same durable token remain authorized, and superseded generations are still fenced under the per-job upload lock. Terminal upload/done transitions clear the token, so completed work cannot be restarted through the preparation path. The new DB regression covers same-generation retry versus a different token.

I also rechecked the earlier fixes: generated SDK keys survive high-level polling timeouts, explicit keys are forwarded across TS/Python APIs, request fingerprints prevent conflicting key reuse, stale pending preparation is reclaimable with fencing, the register transaction is journaled before submission and replayed by digest, paid-blob recovery uses the original persisted ciphertext/vector, recovery claims are reclaimable, and terminal state cannot regress from `done` to `uploaded`. Legacy rows without an original preparation payload fail closed rather than re-encrypting against a paid blob.

All 12 required checks are green, the branch is mergeable, and `git diff --check origin/dev...origin/pr-562` is clean. No remaining blockers from my review. Approved.

ducnmm (approved):
Re-reviewed the simplified/squashed head `a1c4b53e` after the force-push.

I compared it against the previously approved implementation (`e7b3ea27`). The production behavior is preserved. The substantive simplification is:

- the five additive schema migrations are consolidated into `012_remember_write_idempotency.sql`;
- the concurrent partial unique index remains a standalone `013_remember_write_idempotency_index.sql` and is still executed outside an explicit transaction;
- test setup now loads those two consolidated migrations;
- the standalone live-check script was removed;
- no other runtime files differ from the approved implementation.

Because these migrations have not landed on the target branch, consolidating/renumbering them before merge does not create an upgrade-order regression. All schema changes remain additive and idempotent, and the paid-write fencing/journal/recovery semantics reviewed previously are unchanged.

All 12 checks are green, the branch is mergeable, and `git diff --check origin/dev...origin/pr-562` is clean. No blockers. Approved current head.

Issue #501, opened 2026-07-27
bug(sidecar): restore/query-blobs discovery has no uploader provenance check — anyone can make a target's restore download an attacker's Blob


brilovartur-max:
### Context
`/api/restore` rebuilds a namespace's index by discovering the owner's Walrus `Blob` objects on-chain (`/walrus/query-blobs` in the sidecar), then downloading and decrypting each one that isn't already indexed.

### Problem
Discovery in `services/server/scripts/sidecar/routes/walrus-query.ts` filters candidates by object-change type and *current* address only:
```ts
filter: { ToAddress: owner },
...
if (change?.type !== "transferred" && change?.type !== "created" && change?.type !== "mutated") continue;
const belongsToOwner =
    ownerMatchesRecipient(change.recipient, owner) ||
    ownerMatchesRecipient(change.owner, owner);
```
There's no check on **who created or transferred** the object — only on who currently holds it. Walrus `Blob` objects are freely transferable, so any address can mint (or already own) a `Blob` carrying `memwal_*` namespace metadata and transfer it straight to a victim's address. On the victim's next `restore()`, that object is discovered as if it were one of their own memories.

I reproduced this on Sui testnet: minted a `Blob` under an attacker address with `namespace: "e2e-f9"` metadata, transferred it to a victim address (digest `5qPdZ3uCcLNRP5N91NsT67JNfCxnVryrfUebsre7Xp5Y`), then ran an authenticated `restore()` for that namespace as the victim. `/walrus/query-blobs` returned the attacker's object, restore downloaded it, and attempted to decrypt it. SEAL rejected it (it wasn't encrypted to the victim's key) so nothing got indexed here — but the query → download → decrypt work still ran against an object the victim never created.

### Ask
1. Define a set of trusted uploader/relayer addresses (or verify the object's *first* on-chain event was a creation by the configured server wallet pool, not just a transfer from an arbitrary sender).
2. Apply that provenance check before an object is returned from discovery, on both the JSON-RPC and gRPC query paths.
3. Regression test: an externally-created-and-transferred `Blob` with matching namespace metadata must be excluded from `query-blobs` / `restore` for the recipient.

### Why it matters
[...]

PR #569, opened 2026-08-10
fix(sidecar): uploader-provenance gate for query-blobs (WALM-299)

### LINKED ISSUE #589: MCP memwal_restore tool drops the truncated flag — reports "Restore complete" on a partial restore
# Issue 3: MCP `memwal_restore` tool drops the `truncated` flag — reports "Restore complete" on a partial restore

## Summary
The MCP `memwal_restore` tool surfaces `total` / `restored` / `skipped` counts but silently drops the `truncated` field that the underlying `/api/restore` response and the SDK's `restore()` both provide. Because the tool also defaults `limit` to 10, any namespace with more than 10 missing blobs is only partially re-indexed, yet the tool prints "Restore complete for namespace ..." with no indication that the restore was cut short. An agent/user relying on the tool believes the index is fully rebuilt when it is not.

## Affected
- `services/server/scripts/mcp/tools/restore.ts` L37-49: the tool formats the response from `session.memwal.restore(namespace, limit)` and prints only `total` / `restored` / `skipped` — `result.truncated` is never read or shown.
- Contrast: the SDK `restore()` (`packages/sdk/src/memwal.ts` L805-814) returns `{ restored, skipped, total, truncated }` and explicitly normalizes `truncated` (WALM-319). The Rust `/api/restore` handler computes `truncated = limit_truncated || source_capped` and returns it in the response (WALM-319 / #538).
- The tool's `limit` schema (L11-17) defaults to 10, so the truncation path is the common case for any non-trivial namespace.

## Steps to Reproduce
1. Have a namespace with, say, 25 memories that need re-indexing (fresh relayer / new machine / after switching servers).
2. Invoke `memwal_restore` with the default limit (10).
3. The tool prints:
   ```
   Restore complete for namespace "my-app":
     total=25  restored=10  skipped=15
   ```
   — and says "Restore complete", with no `truncated` warning.
4. `memwal_recall` returns only the 10 restored memories; the other 15 are silently absent until the user manually raises `limit`.

## Impact
Misleading recovery UX / data-loss confusion in the MCP path. The API and SDK already compute and surface the truncation signal (WALM-319 fixed exactly this for the SDK); the MCP tool regresses it by dropping the field, so a partial restore is indistinguishable from a complete one.

## Suggested fix
Read `result.truncated` and append a warning when true, e.g.:
```
Restore complete for namespace "my-app":
  total=25  restored=10  skipped=15  truncated=true
  ⚠️ More blobs remain to restore — increase limit and call again.
```

## Severity
Usability / correctness defect in the MCP restore tool (friction point).

harrymove-ctrl:
## What

Fixes WALM-299 / GH #501 — restore/query-blobs accepted transferred Blob objects based on current ownership without proving trusted uploader provenance.

### Attack vector
An attacker creates a Blob with matching `memwal_*` metadata and transfers it
to a victim. The victim's next restore discovered and downloaded that foreign
Blob before SEAL rejected decryption, causing untrusted resource consumption.

### Fix
`query-blobs` now resolves each Blob's immutable UID-creation transaction via
archival Sui GraphQL and checks that creation sender against the server wallet
pool derived from `SERVER_SUI_PRIVATE_KEYS`. Current ownership remains a
separate required gate.

- **Server-wallet creation sender + matching current owner** → Blob returned
- **External creation sender** → Blob excluded
- **Missing/incomplete history** → fail closed
- **Archival transport failure after retries** → query fails so restore can retry; no silent partial result

Both normal metadata discovery and `includeStorageLease` use the same
provenance filter. Creation senders are cached in a bounded 10,000-entry
process LRU because creation provenance is immutable.

## Follow-up fixes
- Uses archival Sui GraphQL because current-state gRPC fullnodes may prune old transactions/object versions.
- Surfaces `truncated` in `memwal_restore` and warns on partial restores. Fixes #589.


ducnmm (approved):
Final review completed against `dev` at head `4c581399`.

Verified:
- immutable Blob creation provenance uses archival GraphQL and current-owner checks on both query paths
- exhausted transient provenance lookups fail the request instead of silently returning partial success
- archival requests are timeout/retry bounded
- restore processing and call rate are bounded; permanent failures are negative-cached
- migration 010 is applied in the production `VectorDb::new()` startup path
- MCP exposes `truncated` and does not claim completion for known-partial restores
- mainnet local smoke: relayer/sidecar ready, authenticated query succeeds, restore returns `truncated=true` without the prior 500
- all required CI checks pass

Approved.

PR #583, opened 2026-08-10
BEDU-950: agent memory embed, store, and recall diagrams

reemsabawi-mystenlabs:
Adds two interactive diagrams to `fundamentals/concepts/how-agent-memory-works`, covering the embed, store, and recall loop.

## Why two diagrams

Two separate auto-looping diagrams rather than one tabbed diagram or an inline Mermaid flowchart, so neither path is hidden and each gets its own agent-readable figcaption for Kapa.

Deliberately generic to the concept page, with no Seal, Walrus blob, or pgvector, so it does not duplicate the Walrus Memory storage pipeline in BEDU-912.

Built on the Walrus Diagrams design system, self-contained and offline-capable, dark and light themes, orthogonal arrow routing, responsive on mobile.

## Changes

- `docs/diagrams/interactive_agent-memory-write_v1.html` — write path: text → embedding model → vector stored alongside original text.
- `docs/diagrams/interactive_agent-memory-recall_v1.html` — read path: question → same embedding model → nearest-vector search → top matches added to the context window.
- `docs/fundamentals/concepts/how-agent-memory-works.md` — embeds both under `## How long-term memory works`, write path first, each under its own subheading.

## Notes for the reviewer

Three things worth a look, all noted rather than assumed:

1. **A third subheading was added.** Placing the two figures under `### Writing a memory` and `### Recalling a memory` would have left the existing "kinds of long-term memory" paragraph nested under *Recalling a memory*, which is inaccurate for both readers and Kapa ingestion. That paragraph now sits under its own `### Kinds of long-term memory`. No prose in it changed.

2. **Asset path uses a different convention than the existing diagrams.** These iframes use root-relative `/diagrams/…`, verified against `docs.json`: `favicon` and `logo` resolve the same way (`/logo-light.svg` → `docs/logo-light.svg`), there is no `assets`/`static` allowlist to register with, and `diagrams` is absent from `navigation`, so the HTML stays an asset rather than becoming a page. The already-shipped diagrams on `how-storage-works` instead use absolute `https://docs.wal.app/diagrams/…`. Both forms work on the Mintlify site, but the repo now carries two conventions.

   Related: because the shipped pages point at `docs.wal.app`, these two files may also need to land in `walrus/docs/site/static/diagrams/` for the docs.wal.app republish to resolve them. Not done here, since this PR is scoped to MemWal.

3. **Four wording changes to the supplied caption and title text.** The repo's docs style gate rejected the original strings for passive voice, a reduced participial clause, and hard-wrapped prose. Meaning is unchanged; the edits are "is stored" → "the store keeps", "is added" → "the agent adds", "a vector store built to find" → "that finds", plus both iframe `title` attributes rewritten to active voice and both figcaptions unwrapped onto single lines.

Closes BEDU-950.

---


jessiemongeon1:
## Style Guide Audit

ducnmm (approved):
Looks good overall — approving. A couple of non-blocking follow-ups worth considering:

- The diagrams auto-advance every 3.4 seconds without a pause control or support for `prefers-reduced-motion`. Adding Play/Pause would improve accessibility.
- On narrow mobile viewports, the 880px SVG is scaled down to roughly 332px, making its labels very small. A mobile-specific layout or a minimum width with horizontal scrolling would improve readability.

Neither of these should block the merge.

jessiemongeon1 (approved):

PR #594, opened 2026-08-11
fix(sdk): isolate foreign blobs during manual recall

### LINKED ISSUE #588: recallManual throws on a single foreign/corrupt blob, aborting the entire recall — inconsistent with server-side batch isolation
# Issue 2: `recallManual` throws on a single foreign/corrupt blob, aborting the entire recall — inconsistent with server-side batch isolation

## Summary
`MemWalManual.recallManual` (full client-side recall) aborts the ENTIRE recall with an exception if ANY one downloaded blob's SEAL `packageId` does not match the configured immutable package. The server-side `/seal/decrypt-batch` path isolates per-item errors (marks the bad item `CORRUPT_CIPHERTEXT` and continues), so the same foreign-blob condition that the server tolerates silently destroys all results in manual mode. One bad blob = 0 results, even when the other hits are valid.

## Affected
- `packages/sdk/src/manual.ts` L436-453 (`recallManual`): the packageId check at L446-451 `throw`s instead of skipping.
- Contrast: `services/server/scripts/sidecar/routes/seal.ts` `parseSealDecryptBatchItems` — per-item `CORRUPT_CIPHERTEXT` error, continues.

## Steps to Reproduce
1. Have a namespace whose search results include ≥1 blob whose ciphertext `packageId` differs from the configured immutable package (e.g. a corrupted/truncated Walrus blob, a blob written under a different immutable package before a redeploy, or a stray foreign blob referenced in the index).
2. Call `memwal.recallManual(query)`.
3. The loop at manual.ts L438-453 hits the mismatched blob, `EncryptedObject.parse` succeeds, then the packageId check at L447 throws:
   `MemWalManual: ciphertext <blob_id> packageId does not match the configured immutable packageId`
4. The exception propagates out of `recallManual` — no results are returned, even though the other blobs are valid.

Attached repro (`repro_recallmanual_throw.mjs`) demonstrates the divergent behavior with mock blobs: manual mode throws (0/3 results), server batch isolates (2/3 results).

## Impact
Availability / robustness regression in the manual (client-side) recall path. A single corrupt or foreign blob in the index makes recall unusable for the whole namespace, where the server path is designed to degrade gracefully per item. Developers using `MemWalManual` lose all recall results with no way to recover the valid subset.

## Suggested fix
Match the server-side behavior: on packageId mismatch, log + `continue` (skip that blob) rather than `throw`, so the valid subset is still returned. (The decrypt loop below already isolates per-blob failures with try/catch — the packageId check should follow the same pattern.)

## Severity
Robustness / availability defect in the SDK manual recall path.

ducnmm:
## Summary

- skip ciphertexts whose immutable package ID does not match the configured package instead of aborting the entire `recallManual` batch
- preserve fail-closed behavior by never requesting SEAL keys or decrypting the foreign blob
- add regression coverage for both foreign-only and mixed valid/foreign result sets

## Testing

- `pnpm --dir packages/sdk test` (23/23 passed)
- `git diff --check`

Closes #588

### DIFF
--- a/packages/sdk/src/manual.ts
+++ b/packages/sdk/src/manual.ts
@@ -444,10 +444,11 @@ export class MemWalManual {
                 continue;
             }
             if (normalizeSuiAddress(parsed.packageId) !== immutablePackageId) {
-                throw new Error(
-                    `MemWalManual: ciphertext ${blob.blob_id} packageId does not match ` +
+                console.error(
+                    `[MemWalManual] Skipping ciphertext ${blob.blob_id}: packageId does not match ` +
                         "the configured immutable packageId"
                 );
+                continue;
             }
             parsedBlobs.push({ ...blob, parsed });
         }
--- a/packages/sdk/test/manual-seal-identity.test.mjs
+++ b/packages/sdk/test/manual-seal-identity.test.mjs
@@ -1,6 +1,7 @@
 import assert from "node:assert/strict";
 import test from "node:test";
 import { EncryptedObject } from "@mysten/seal";
+import { Inputs } from "@mysten/sui/transactions";
 
 import { MemWalManual } from "../dist/manual.js";
 
@@ -8,6 +9,7 @@ const ACCOUNT_ID = `0x${"aa".repeat(32)}`;
 const PACKAGE_ID = `0x${"33".repeat(32)}`;
 const OWNER = `0x${"11".repeat(32)}`;
 const DELEGATE_SIGNER = `0x${"22".repeat(32)}`;
+const REGISTRY_ID = `0x${"bb".repeat(32)}`;
 const COUNTER_LE_7 = "0700000000000000";
 
 function ciphertext(packageId) {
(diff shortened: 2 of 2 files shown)


harrymove-ctrl (changes requested):
## Review

Accurately fixes the bug (#588) — behavior is now consistent with the server-side batch path. Test coverage is comprehensive for both foreign-only and mixed cases. The approach is solid: fail-closed (does not fetch keys or decrypt foreign blobs).

---

### Suggestion: `core.getObject` mock in `manualWithAccountResponse`

The test helper mock for `core.getObject` currently returns `({ object: { version: "1" } })`, but this structure is not a valid Sui object shape — it is missing required fields like `status`, `type`, and `json`.

If future tests trigger the `createSessionKey` path (valid blob decryption flow), this mock will crash when the code attempts to access the `.json` field. It is better to mock the proper structure from the start:

```javascript
core: {
    getObject: async () => ({
        object: {
            status: "Exists",
            type: `${PACKAGE_ID}::account::MemWalAccount`,
            version: "1",
            objectId: REGISTRY_ID,
            digest: "mockdigest",
        },
    }),
    resolveTransactionPlugin: () => async (transactionData, _options, next) => {
        transactionData.inputs = transactionData.inputs.map((input) =>
            input.UnresolvedObject
                ? Inputs.SharedObjectRef({
                      objectId: input.UnresolvedObject.objectId,
                      initialSharedVersion: "1",
                      mutable: true,
                  })
                : input
        );
        await next();
    },
},
```

This is not a blocker since the current tests do not exercise that specific code path, but fixing it now will prevent future regressions as test coverage expands.

harrymove-ctrl (approved):

PR #595, opened 2026-08-11
fix(sdk): encode large manual payloads safely

### LINKED ISSUE #593: [SDK/Bug] MemWalManual.rememberManual crashes (RangeError: Maximum call stack size exceeded) on memories >~123 KB — base64 encode via argument spread
**[SDK/Bug] `MemWalManual.rememberManual` throws `RangeError: Maximum call stack size exceeded` on memories larger than ~123 KB (base64 encode via argument spread)**

### Summary

`rememberManual()` — the client-managed-encryption path documented as the way an agent "must hold its own keys and never delegate decryption to the relayer" — crashes on any memory whose SEAL ciphertext exceeds the JS engine's maximum argument count (~123 KB of ciphertext on Node 22). The crash happens **client-side, before the network call**, so no memory is stored and no clear error is surfaced to the caller about *why*.

### Location

`packages/sdk/src/manual.ts:352`

```ts
const encryptedBase64 = btoa(String.fromCharCode(...encrypted));
```

`encrypted` is the SEAL ciphertext `Uint8Array` from `sealEncrypt()` (line 347). The spread `...encrypted` passes **every byte as a separate positional argument** to `String.fromCharCode`. Once the byte count exceeds the engine's max stack-allocated argument count, the call throws `RangeError: Maximum call stack size exceeded`.

### Reproduction

```js
// node >= 18, standalone — mirrors the exact expression at manual.ts:352
function encodeLikeMemwal(encrypted) {
  return btoa(String.fromCharCode(...encrypted));
}
for (const n of [64*1024, 100*1024, 123*1024, 128*1024, 256*1024]) {
  const buf = new Uint8Array(n);
  try { encodeLikeMemwal(buf); console.log(n, "ok"); }
  catch (e) { console.log(n, e.constructor.name + ": " + e.message); }
}
```

Output (Node v22.23.1):
```
65536  ok
102400 ok
125952 -> RangeError: Maximum call stack size exceeded   (first crash ~123 KB)
131072 RangeError: Maximum call stack size exceeded
262144 RangeError: Maximum call stack size exceeded
```

The exact threshold is engine/version-dependent (V8 stack size, other runtimes lower), so the failure is also **non-deterministic across environments** — a memory that saves fine in one runtime can crash in another, which is especially bad for a "portable, works-across-tools" memory layer.

### Impact

- The manual/client-side-encryption path silently caps a single memory at ~123 KB of *ciphertext* (less plaintext after SEAL overhead), well under the 1 MiB `memwal_remember` body limit the rest of the stack advertises. A user storing a long document, a large code file, or a batched context blob via the privacy-preserving path gets a stack-overflow instead of a "too large" error.
- It defeats the documented use case (`docs/sdk/agent-storage-loop.md`: "plaintext never leaves the agent process") for exactly the large payloads where client-side encryption matters most.
- The error message (`Maximum call stack size exceeded`) gives the caller no hint that the cause is payload size, so it looks like an SDK bug rather than a limit.

### Fix

Chunk the byte→string conversion, or use a `Buffer`/base64 codec that doesn't rely on argument spread:

```ts
// chunked, no spread
function toBase64(bytes: Uint8Array): string {
  let binary = "";
  const CHUNK = 0x8000;
  for (let i = 0; i < bytes.length; i += CHUNK) {
    binary += String.fromCharCode.apply(null, bytes.subarray(i, i + CHUNK) as unknown as number[]);
  }
  return btoa(binary);
}
// ...
const encryptedBase64 = toBase64(encrypted);
```

Or, where `Buffer` is available (Node): `Buffer.from(encrypted).toString("base64")`. The non-manual path at `memwal.ts:1067-1069` already prefers `Buffer.from(...).toString("base64")` as a fallback — the manual path should use the same safe encoder.

The chunked encoder passes for all sizes tested (64 KB → 512 KB) where the current line throws at ~123 KB.

### Notes
- Verified distinct from open issues #566 (PG restore rollback) and #523 (Walrus testnet publisher HTTP API size limits) — this is a client-side SDK encoding bug in `manual.ts`.
- Environment: `@mysten-incubation/memwal` SDK `packages/sdk/src/manual.ts`, Node v22.23.1.

ducnmm:
## Summary

- replace whole-array argument spreading in `rememberManual` base64 encoding with fixed-size chunks
- keep the encoder portable across browser and Node runtimes
- add a 256 KiB ciphertext regression test that verifies exact byte round-tripping and request construction

## Testing

- `pnpm --dir packages/sdk test` (23/23 passed)
- `git diff --check`

Closes #593

### DIFF
--- a/packages/sdk/src/manual.ts
+++ b/packages/sdk/src/manual.ts
@@ -131,6 +131,16 @@ function sealServerConfigTotalWeight(configs: ResolvedSealServerConfig[]): numbe
     return configs.reduce((sum, config) => sum + config.weight, 0);
 }
 
+/** Encode arbitrary-size bytes without passing the whole payload as function arguments. */
+function bytesToBase64(bytes: Uint8Array): string {
+    const chunkSize = 0x8000;
+    const chunks: string[] = [];
+    for (let offset = 0; offset < bytes.length; offset += chunkSize) {
+        chunks.push(String.fromCharCode(...bytes.subarray(offset, offset + chunkSize)));
+    }
+    return btoa(chunks.join(""));
+}
+
 // ============================================================
 // MemWalManual Client
 // ============================================================
@@ -349,7 +359,7 @@ export class MemWalManual {
 
         // Step 3: Send encrypted bytes (base64) + vector to server.
         // Server will upload to Walrus via upload-relay and return the blob_id.
-        const encryptedBase64 = btoa(String.fromCharCode(...encrypted));
+        const encryptedBase64 = bytesToBase64(encrypted);
         return this.signedRequest<RememberManualResult>("POST", "/api/remember/manual", {
             encrypted_data: encryptedBase64,
             vector,
--- a/packages/sdk/test/manual-seal-identity.test.mjs
+++ b/packages/sdk/test/manual-seal-identity.test.mjs
@@ -144,6 +144,27 @@ test("inactive accounts fail before encryption", async () => {
     assert.equal(encryptCalls.length, 0);
 });
 
+test("manual remember base64-encodes large ciphertext without argument spread overflow", async () => {
+    const { manual } = manualWithAccountResponse({});
+    const encrypted = Uint8Array.from({ length: 256 * 1024 }, (_, index) => index % 256);
+    let request;
+
+    manual.embed = async () => [0.25, 0.75];
+    manual.sealEncrypt = async () => encrypted;
+    manual.signedRequest = async (method, path, body) => {
+        request = { method, path, body };
+        return { blob_id: "large-blob" };
+    };
+
+    await manual.rememberManual("large memory", "large-namespace");
+
+    assert.equal(request.method, "POST");
+    assert.equal(request.path, "/api/remember/manual");
+    assert.equal(request.body.namespace, "large-namespace");
+    assert.deepEqual(request.body.vector, [0.25, 0.75]);
+    assert.deepEqual(Buffer.from(request.body.encrypted_data, "base64"), Buffer.from(encrypted));
+});
+
 test("manual recall skips a foreign ciphertext package before fetching keys", async () => {
     const { manual } = manualWithAccountResponse({});
     let fetchCalls = 0;


harrymove-ctrl (approved):
### *Code Review: PR #595 — `fix(sdk): encode large manual payloads safely`*

**Overall Assessment:** **LGTM with a few minor items to address.**
Great fix addressing the root cause: `btoa(String.fromCharCode(...encrypted))` crashes on payloads $\ge$ ~100KB due to JavaScript engine argument limits on the spread operator. The 32KB chunking approach in `bytesToBase64` cleanly prevents call-stack overflow.

Here are a few findings and suggestions:

---

#### *Finding 1: `Buffer` API is not portable across runtimes* `[Medium]`
* **File:** `packages/sdk/test/manual-seal-identity.test.mjs:138`
* **Issue:** The test file uses the `.mjs` extension (implying universal/portable ESM for both Node and Browser), but `Buffer` is Node.js-only. If tests are run in a browser or edge environment, this assertion will throw instead of comparing.
* **Suggested Fix:** Use a portable byte-to-byte comparison:
  ```javascript
  const decoded = Uint8Array.from(atob(request.body.encrypted_data), c => c.charCodeAt(0));
  assert.deepEqual(decoded, encrypted);
  ```

---

#### *Finding 2: Test execution and payload mismatch* `[Low]`
* **File:** `packages/sdk/test/manual-seal-identity.test.mjs:130-151`
* **Issue:** 
  1. The new test asserts `request.body.namespace === "large-namespace"`, but the current implementation only sends `encrypted_data` and `vector` in the body.
  2. The PR description notes *"23/23 passed"*, but running `pnpm test` locally only shows 22 tests.
* **Action:** Please double-check if the new test is actually picked up by the test runner and passing as expected.

---

#### *Finding 3: `btoa` intermediate memory pressure* `[Low / Nit]`
* **File:** `packages/sdk/src/manual.ts:135`
* **Note:** `return btoa(chunks.join(""));` chunking protects against the `String.fromCharCode` spread argument limit, but `chunks.join("")` still creates an intermediate concatenated string in memory before calling `btoa`.
* **Impact:** Acceptable for current sizes (solves the 123KB crash), but worth keeping in mind if we ever support multi-megabyte payloads in manual seal identity.

---

#### *Finding 4: Missing early-return guard for empty inputs* `[Low / Nit]`
* **File:** `packages/sdk/src/manual.ts:135`
* **Suggestion:** While `bytes.length === 0` resolves to `btoa("") === ""` without throwing, adding an explicit guard makes edge-case handling cleaner:
  ```typescript
  if (bytes.length === 0) return "";
  ```

---

### *Highlights*
* Solves the exact root cause with clean 32KB (`0x8000`) chunking.
* Includes round-trip exact byte verification.
* Clean, concise PR diff (+32 / −1 lines).

---

### *Summary of Action Items*
| # | Finding | Severity | Suggested Action |
|---|---|---|---|
| **1** | `Buffer` API in `.mjs` test | **Medium** | Switch to portable `Uint8Array` comparison |
| **2** | Test execution & assertion check | **Low** | Verify test count & `namespace` body check |
| **3** | `btoa` intermediate string memory | **Nit** | Good for now; keep in mind for multi-MB payloads |
| **4** | Empty `Uint8Array` handling | **Nit** | Add `if (!bytes.length) return ""` early exit |

PR #597, opened 2026-08-11
fix(mcp): refresh bridge credentials after login

### LINKED ISSUE #580: bug(mcp): in-session memwal_login never updates the bridge's credentials, so every later call 401s until the process is killed
## Environment

- `@mysten-incubation/memwal-mcp` 0.0.6 (stdio, via `npx -y`)
- MCP client: Claude Code (desktop app)
- Relayer: https://relayer.memory.walrus.xyz (production mainnet)
- Node.js v22.23.1, macOS 26.5.2

## What happened

Recall started returning nothing after the July 30 migration, so I called `memwal_login` from inside the running session and approved the browser flow. `credentials.json` was updated with the new delegate key. Every memwal call after that returned `401`, including `memwal_recall`, which had been working minutes earlier.

I assumed my account was broken. I logged in again with a second wallet. Still 401. I got far enough down that path to consider deleting my pre-migration blobs, believing the account was unrecoverable.

It was none of that. The credentials on disk were fine the whole time. I confirmed it by driving the relayer directly with the same on-disk credentials, bypassing the MCP:

- `GET /api/mcp/sse` returned 200
- `initialize` and `tools/call` returned 202, and `memwal_recall` came back with a real result over the SSE stream

Then I killed the `memwal-mcp` process. The next tool call respawned it and everything worked immediately, same credentials, same relayer, no re-login.

## Root cause

Credentials are loaded once per process and captured by the bridge for its lifetime.

`packages/mcp/src/index.ts:116`

```ts
let creds = loadCreds();
```

`packages/mcp/src/bridge.ts:494` takes those creds as a parameter, and every request builds its headers from that one binding (`mcpAuthHeaders(creds)` at `bridge.ts:141`, used by both `openSseStream` and `postMessage`).

The in-session `memwal_login` tool is intercepted locally at `bridge.ts:637` and handled by `handleLocalLogin(config)` at `bridge.ts:391`. That function receives only `config: BridgeConfig`. It has no reference to the bridge's `creds` binding, so it cannot update it. On success it only logs (`bridge.ts:400`):

```ts
.then((creds) => {
    log.info("memwal_login.bridge.success", { ... });
})
```

`loginFlow()` writes the new credentials to disk, and the running bridge keeps signing with the old delegate key. Registering the new key appears to retire the old one server-side, so the bridge is now authenticating with a key that no longer works, for the rest of its life.

The tool's own instructions make this worse. `bridge.ts:449` tells the user:

> 3. Once "Connected" appears, retry the previous request

Retrying in the same process is exactly what cannot work.

## Expected

After a successful in-session login, either the bridge uses the new credentials without a restart, or it says plainly that the client must be restarted.

## Repro

1. Start any MCP client with valid credentials at `~/.memwal/credentials.json`.
2. Confirm `memwal_recall` works.
3. Call `memwal_login` and approve in the browser. Credentials on disk are updated.
4. Call `memwal_recall` again. It returns 401, and keeps returning 401.
5. `pkill -f "node .*memwal-mcp"`. Next tool call respawns and works.

## Why it matters

`memwal_login` is the natural response to an auth problem, and it is what the 401 path steers users toward. Doing it converts a recoverable state into a session that can never authenticate again, and the failure looks identical to "the relayer is down" or "my account was lost in the migration". I spent most of a day on this and came close to deleting real data over it.

There is a second cost: every attempt registers another delegate key on-chain. I ended up with six keys on my account from one debugging session, and each login invalidates other clients' live sessions, so fixing one client breaks the others.

## Not a duplicate of

- #354 is the CLI `login` arg path wiping credentials before re-auth. This is the in-session `memwal_login` tool, which does not wipe anything. It writes good credentials that the running process ignores.
- #560 is the lowercase `authorization` header on POST. My direct probe used lowercase `authorization` on POST and succeeded, and the same bridge works after a restart with identical headers, so that is not what I hit.

## Suggested fix

1. Have `handleLocalLogin` return the new credentials, and have `runBridge` swap them in and reopen the SSE stream with the new key. This is the fix that matches what the message at `bridge.ts:449` already promises.
2. If hot-swapping is not wanted, exit cleanly after a successful in-session login so the MCP client respawns the server, which already reloads credentials at `index.ts:116`.
3. At minimum, change the text at `bridge.ts:449` to say the client must be restarted, and have the 401 handler say "credentials changed since this process started" instead of a bare 401.

ducnmm:
## Summary

- adopt credentials returned by an in-session `memwal_login` and immediately reopen the SSE bridge with the new delegate key/account
- make concurrent reconnect callers await one shared reconnect so requests cannot race the stale session URL
- preserve in-flight replay for same-account key rotation, while failing old-account requests explicitly instead of replaying them against a newly selected account
- add an end-to-end regression test that completes the browser preflight/callback flow, verifies the second SSE handshake uses the new credentials, and recalls successfully without restarting the MCP process

## Testing

- `pnpm --dir packages/mcp test` (10/10 passed)
- `git diff --check`

Closes #580

### DIFF
--- a/packages/mcp/src/bridge.ts
+++ b/packages/mcp/src/bridge.ts
@@ -387,7 +387,10 @@ function writeStdoutMessage(msg: RpcMessage): void {
  * mode, but available even when creds already exist (so user can re-login,
  * switch wallets, or refresh). Returns a click-able URL near-instantly;
  * listener stays alive in the background until callback or timeout. */
-async function handleLocalLogin(config: BridgeConfig): Promise<{ text: string; isError: boolean }> {
+async function handleLocalLogin(
+    config: BridgeConfig,
+    onCredentials: (creds: MemWalCredentials) => Promise<void>,
+): Promise<{ text: string; isError: boolean }> {
     const urlReady = new Promise<string>((resolve) => {
         loginFlow({
             relayerUrl: config.relayerUrl,
@@ -397,7 +400,8 @@ async function handleLocalLogin(config: BridgeConfig): Promise<{ text: string; i
             openBrowser: false,
             onUrl: (url) => resolve(url),
         })
-            .then((creds) => {
+            .then(async (creds) => {
+                await onCredentials(creds);
                 log.info("memwal_login.bridge.success", {
                     accountId: creds.accountId,
                     delegateAddress: creds.delegateAddress,
@@ -514,8 +518,10 @@ export async function runBridge(
     log.info("bridge.connected", { relayer: creds.relayerUrl });
 
     let stdinClosed = false;
-    let reconnecting = false;
     let reconnectAttempt = 0;
+    let reconnectPromise: Promise<void> | null = null;
+    let credentialGeneration = 0;
+    let activeCredentialGeneration = 0;
 
     // In-flight requests pending a response. We replay them after a forced
     // reconnect so a server-side session swap doesn't strand a tool call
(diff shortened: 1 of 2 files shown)


ducnmm (commented):
Found one concurrency issue that should be addressed before merge:

**[P1] A credential update can join a reconnect that already opened with the old credentials** (`packages/mcp/src/bridge.ts:541`).

If an existing reconnect has already evaluated `openSseStream(creds.relayerUrl, creds)` with the old credentials and is waiting for the handshake, `adoptCredentials()` assigns `creds = nextCreds` and then calls `reconnect()`. Because `reconnectPromise` already exists, it only awaits that old reconnect. The resulting SSE session is authenticated with the old key, while subsequent POSTs use the new key. The relayer then returns a session auth mismatch (403), and this bridge only triggers another reconnect on 404, so the process can remain broken even though login succeeded.

Please associate reconnects with a credential generation/version (or otherwise discard and reopen a handshake whose credentials became stale), and add a regression test where the SSE handshake is delayed and login completes while that reconnect is in progress.

ducnmm:
Addressed the reconnect/credential race in the latest push.

- reconnect handshakes now capture a credential generation and credentials snapshot
- a handshake that completes after credentials rotate is aborted instead of becoming the active SSE session
- credentials rotating during in-flight replay also force another handshake
- outbound POSTs verify that the active SSE generation matches the current credentials
- the integration test now deliberately delays an old-key SSE handshake, completes `memwal_login` while it is pending, then verifies that the stale handshake is discarded and recall uses the new account/key

Validation: `pnpm --dir packages/mcp test` (10/10 passed) and `git diff --check`.

ducnmm (commented):
Re-reviewed the latest push (`ac747696`). The credential-generation guard addresses the race from my previous review:

- each handshake captures a credentials snapshot and generation;
- stale handshakes are aborted before becoming active;
- rotation during replay forces another handshake;
- outbound POSTs verify the active generation;
- the regression test now reproduces the delayed old-key handshake race.

I also reran `pnpm --dir packages/mcp test`: 10/10 passed, and `git diff --check` is clean. No further findings from me — LGTM.

harrymove-ctrl (approved):
Nice fix for GH-580! Resolves the 403 auth mismatch issue where mid-session memwal_login calls weren't propagating new credentials to the bridge SSE handshake.

The generation snapshotting in bridge.ts (credentialGeneration + reconnectPromise) handles the concurrent reconnect and POST race conditions cleanly, and the mock handshake delay tests in live-login-credentials.test.mjs cover the edge cases nicely.

LGTM! 👍

PR #598, opened 2026-08-11
fix(chatbot): isolate memory by authenticated user

### LINKED ISSUE #587: Chatbot memory layer uses a single shared Walrus Memory account — cross-user memory leak
# Issue 1: Chatbot memory layer uses a single shared Walrus Memory account — cross-user memory leak

## Summary
The chatbot reference app (`apps/chatbot`) stores and recalls ALL users' memories through ONE server-wide Walrus Memory account (`MEMWAL_PRIVATE_KEY` / `MEMWAL_ACCOUNT_ID` env vars), with no per-user scoping. In any multi-user deployment, one user's saved memories are recalled and injected into another user's LLM context — a cross-user privacy leak. This contradicts the recent per-user document-ownership fix (#517 / commit `1fe567c`), which scoped documents to the session user but left the memory layer shared.

## Affected
- `apps/chatbot/lib/ai/tools/save-memory.ts` (L23-37): `saveMemory` writes to the shared account via `MemWal.create({ key, accountId, serverUrl })` — no `userId` / per-user namespace.
- `apps/chatbot/lib/ai/providers.ts` (L76-102): `getMemWalModel` wraps the model with `withMemWal({ key, accountId, ... })` using the same server-wide credentials.
- `packages/sdk/src/ai/middleware.ts` (`withMemWal`): `transformParams` recalls via `memwal.recall(lastUserMessage, maxMemories)` and `wrapGenerate` auto-saves via `memwal.analyze(userMessage)` — both against the single shared account, default namespace.

## Steps to Reproduce
1. Deploy the chatbot with a single `MEMWAL_PRIVATE_KEY` + `MEMWAL_ACCOUNT_ID` (the documented setup).
2. User A chats: "remember that my medical condition is peanut allergy."
3. `saveMemory` writes this to the shared account (default namespace).
4. User B (different session/user) chats: "what do you know about me?"
5. `withMemWal.transformParams` calls `memwal.recall("what do you know about me?", 5)` against the SAME shared account → returns User A's memory → injected into User B's context.

## Impact
Cross-user disclosure of saved memories in any multi-user deployment. The feature is advertised as "the user's personal memory on the blockchain" (save-memory.ts tool description) but is not actually per-user isolated.

## Suggested fix
Scope each user's memory to a per-user namespace (e.g. `namespace = userId`) or a per-user account. The document path already resolves the session user (`getDocumentByIdForUser`); the memory path should derive the same per-user scope.

## Severity
Privacy / data-isolation defect in the reference integration. Not a core relayer/contract exploit, but a real cross-user leak in the shipped example app.

ducnmm:
## Summary

- derive a stable memory namespace from the authenticated chatbot user ID and reject missing/blank identities
- pass that namespace to the AI middleware so automatic recall and auto-save no longer use the shared default namespace
- pass the same namespace to the explicit `saveMemory` tool
- add regression coverage for distinct user namespaces, middleware/tool configuration, route wiring, and fail-closed behavior

Existing memories in the previously shared `default` namespace are intentionally not recalled because they cannot be attributed safely to an individual user.

## Testing

- `pnpm --filter @mysten-incubation/memwal build`
- `pnpm --dir apps/chatbot test:unit` (15/15 passed)
- `pnpm --dir apps/chatbot exec tsc --noEmit`
- `pnpm --dir apps/chatbot exec next build`
- `git diff --check`

Note: the existing `ultracite check` command still hits the documented `--skip`/Biome version mismatch and is soft-failed on `dev` CI.

Closes #587

### DIFF
--- a/apps/chatbot/app/(chat)/api/chat/route.ts
+++ b/apps/chatbot/app/(chat)/api/chat/route.ts
@@ -15,6 +15,7 @@ import { after } from "next/server";
 import { createResumableStreamContext } from "resumable-stream";
 import { auth, type UserType } from "@/app/(auth)/auth";
 import { entitlementsByUserType } from "@/lib/ai/entitlements";
+import { memoryNamespaceForUser } from "@/lib/ai/memory-namespace";
 import { allowedModelIds } from "@/lib/ai/models";
 import { type RequestHints, systemPrompt } from "@/lib/ai/prompts";
 import { getLanguageModel, getMemWalModel } from "@/lib/ai/providers";
@@ -76,6 +77,10 @@ export async function POST(request: Request) {
     }
 
     if (!session?.user) {
+      return new ChatbotError("unauthorized:chat").toResponse();
+    }
+    const memoryNamespace = memoryNamespaceForUser(session.user.id);
+    if (!memoryNamespace) {
       return new ChatbotError("unauthorized:chat").toResponse();
     }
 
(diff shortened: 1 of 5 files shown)


ducnmm (commented):
Reviewed the user-memory isolation changes. The authenticated user namespace is consistently applied to both middleware recall/auto-save and the explicit `saveMemory` tool, and missing identities/namespaces fail closed instead of falling back to `default`.

I also verified the namespace length is within the server's 255-byte limit for UUID user IDs and reran:

- `pnpm --dir apps/chatbot test:unit` — 15/15 passed
- `pnpm --dir apps/chatbot exec tsc --noEmit` — passed
- `git diff --check` — clean

No blocking findings from me — LGTM.

harrymove-ctrl (approved):
Reviewed PR #598. LGTM

PR #599, opened 2026-08-11
feat(sdk): add deterministic offline mock clients

### LINKED ISSUE #581: [Feature] Provide an offline mock client (MemWalMock) for local unit testing and CI pipelines
Description: Currently, using the SDK requires hitting the staging or prod relayers using MemWal.create(env="staging"). This introduces network dependencies and potential flakiness in local unit tests and CI/CD pipelines.

Suggested Solution: Add a MemWalMock class (or an env="local_mock" configuration) for both Python and TypeScript SDKs. It should implement the exact same interface (rememberAndWait, recall, etc.) but store and retrieve vectors from a local in-memory dictionary or local SQLite file instead of making external HTTP calls to the relayer.

Impact: Vastly improves Developer Experience (DX) by allowing developers to build and test their agents quickly and offline without worrying about rate limits or network issues.

ducnmm:
## Summary

- add dependency-free `MemWalMock` for TypeScript and async/sync `MemWalMock` clients for Python
- mirror core production flows: remember/job polling, bulk remember, recall, analyze, embed, health, compatibility, and restore
- isolate records by namespace and provide deterministic token-overlap ranking and embeddings for non-flaky tests
- support deterministic seed data plus mock-only `forget` and `clear` helpers
- export the clients from both SDK entry points and document offline CI usage and mock semantics

The mocks never read credentials or create network clients. `analyze` deliberately stores the full input as one fact instead of pretending to reproduce production LLM extraction, and the relevance score is intended for application behavior tests rather than search-quality benchmarks.

## Testing

- TypeScript SDK: `pnpm --dir packages/sdk test` (26/26 passed)
- Python SDK: `python3 -m pytest -q` (121 passed, 16 skipped)
- Python changed-file lint: `python3 -m ruff check memwal/mock.py tests/test_mock.py memwal/__init__.py`
- `git diff --check`

Closes #581

### DIFF
--- a/packages/python-sdk-memwal/README.md
+++ b/packages/python-sdk-memwal/README.md
@@ -85,6 +85,23 @@ matches = client.recall(RecallParams(query="food allergies"))
 client.close()
 ```
 
+### Offline tests and CI
+
+`MemWalMock` and `MemWalMockSync` implement the common memory API in process. They require no credentials, relayer, chain, or paid storage and use deterministic token-overlap ranking.
+
+```python
+from memwal import MemWalMock, RecallParams
+
+async def test_memory_flow():
+    memwal = MemWalMock.create(namespace="test-user")
+    await memwal.remember_and_wait("The user prefers dark mode")
+
+    result = await memwal.recall(RecallParams(query="display preference"))
+    assert "dark mode" in result.results[0].text
+```
+
+The mock supports remember/job polling, bulk remember, recall, analyze, embed, ask, health, restore, `forget(blob_id)`, and `clear(namespace)`. For deterministic behavior, `analyze` stores its full input as one fact instead of invoking an LLM extractor. Its simple relevance score is for application tests, not production search-quality evaluation.
+
 ### Context Manager
 
 ```python
--- a/packages/python-sdk-memwal/memwal/__init__.py
+++ b/packages/python-sdk-memwal/memwal/__init__.py
@@ -32,6 +32,7 @@ from .client import (
     MemWalSync,
 )
 from .middleware import with_memwal_langchain, with_memwal_openai
+from .mock import MemWalMock, MemWalMockSeed, MemWalMockSync
 from .types import (
     ENV_PRESETS,
     AnalyzedFact,
@@ -72,6 +73,9 @@ __all__ = [
     # Core client
     "MemWal",
     "MemWalSync",
+    "MemWalMock",
+    "MemWalMockSync",
+    "MemWalMockSeed",
     "MemWalError",
     "MemWalCompatibilityError",
     "MemWalRememberJobFailed",
--- a/packages/python-sdk-memwal/memwal/mock.py
+++ b/packages/python-sdk-memwal/memwal/mock.py
@@ -1,0 +1,533 @@
+"""Deterministic, dependency-free in-memory Walrus Memory client for tests."""
+
+from __future__ import annotations
+
+import asyncio
+import math
+import re
+import unicodedata
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Sequence, Union
+
+from .types import (
+    AnalyzedFact,
+    AnalyzeResult,
+    AnalyzeWaitResult,
+    AskMemory,
+    AskResult,
+    EmbedResult,
+    HealthResult,
+    RecallMemory,
+    RecallParams,
+    RecallResult,
+    RememberAcceptedResult,
+    RememberBulkAcceptedResult,
+    RememberBulkItem,
+    RememberBulkItemResult,
+    RememberBulkOptions,
+    RememberBulkResult,
+    RememberBulkStatusItem,
+    RememberBulkStatusResult,
+    RememberJobStatus,
+    RememberResult,
+    RestoreResult,
+)
+
+
+@dataclass
+class MemWalMockSeed:
+    """One optional memory loaded when a mock is created."""
+
+    text: str
+    namespace: Optional[str] = None
+    blob_id: Optional[str] = None
+
+
+@dataclass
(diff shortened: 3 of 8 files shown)


ducnmm (commented):
Found two mock-fidelity issues to address before merge:

1. **[P1] TypeScript `recall` does not mirror the production overload/precedence** (`packages/sdk/src/mock.ts`, around the `recall` implementation). Production supports `recall(query, RecallOptions)` and resolves `topK ?? limit ?? 10`; the mock only accepts a numeric second argument and resolves `limit ?? topK ?? 10`. As a result, valid production code such as `recall("x", { limit: 2 })` throws in the mock, and `{ limit: 1, topK: 2 }` returns one result in the mock but two in production. This prevents using `MemWalMock` as a drop-in test double. Please mirror the production overload and `topK` precedence, with regression tests for both forms.

2. **[P2] Tokenization is locale-dependent despite the deterministic-across-machines contract** (`packages/sdk/src/mock.ts:52`). `toLocaleLowerCase()` without an explicit locale uses the host default locale, so inputs containing characters such as `I`/`İ` can tokenize and hash differently under different CI locales. `toLowerCase()` (or a fixed locale) would keep results deterministic.

I ran `pnpm --dir packages/sdk test`; all 26 tests pass, but the current tests do not cover these cases.

ducnmm:
Addressed both findings in `67504a0e`:

1. **P1 — production-compatible recall**
   - Added the `recall(query, RecallOptions)` overload.
   - Matched production argument normalization for object, positional numeric, and string + options forms.
   - Changed precedence to `topK ?? limit ?? 10`.
   - Added regressions for `recall(query, { limit, namespace })` and conflicting `{ limit, topK }` in object style.

2. **P2 — locale-independent tokenization**
   - Replaced `toLocaleLowerCase()` with locale-independent `toLowerCase()`.
   - Added a regression that makes `toLocaleLowerCase()` throw, then verifies recall ranking and embedding still work.

Validation: `pnpm --dir packages/sdk test` — **28/28 passed**; `git diff --check` passed.

ducnmm (commented):
Re-reviewed `67504a0e`. The two TypeScript findings are fixed correctly and the new regressions cover them. I reran the TypeScript SDK suite: 28/28 passed.

I found one remaining drop-in compatibility issue in the Python sync mock:

**[P1] `MemWalMockSync` drops supported `MemWalSync` parameters** (`packages/python-sdk-memwal/memwal/mock.py:446-501`). For example, production callers can pass polling controls to `wait_for_remember_job` / `remember_and_wait`, `RememberBulkOptions` to `wait_for_remember_jobs` / `remember_bulk_and_wait`, and `occurred_at` (plus `opts` for `analyze_and_wait`) to the analyze methods. The mock signatures omit all of these, so code that swaps `MemWalSync` for `MemWalMockSync` raises `TypeError` before reaching the mock. Please mirror the production sync signatures and forward the arguments to the async mock, with at least one regression exercising each parameter group.

Also, `MemWalSync._run()` supports calls made while an event loop is already running (e.g. notebooks) by using a worker thread, while `MemWalMockSync._run()` directly calls `asyncio.run()` and fails in that same supported context. Reusing the production runner pattern would preserve sync-client parity.

ducnmm:
Addressed the Python sync parity finding in `3a5c53a5`:

- Mirrored `MemWalSync` polling parameters on `wait_for_remember_job` and `remember_and_wait`, forwarding them to the async mock.
- Mirrored and forwarded `RememberBulkOptions` for `wait_for_remember_jobs` and `remember_bulk_and_wait`.
- Mirrored and forwarded `occurred_at` for `analyze`, plus both `opts` and `occurred_at` for `analyze_and_wait`.
- Replaced direct `asyncio.run()` with the production runner pattern: when a loop is already active (Jupyter/notebooks), the coroutine runs in a worker thread with a fresh event loop.
- Added a parameter-contract regression against `MemWalSync`, exercised every parameter group, and added an async test that calls `MemWalMockSync` from inside a running event loop.

Validation:
- Python full suite: **123 passed, 16 skipped**
- Mock-specific tests: **7/7 passed**
- Changed-file Ruff: passed
- `git diff --check`: passed

ducnmm (commented):
Re-reviewed `3a5c53a5`. The Python sync parity finding is addressed correctly:

- polling, bulk, and analyze signatures now match `MemWalSync` and forward all supported options;
- `occurred_at` typing/forwarding is aligned;
- the runner now handles an already-running event loop via a worker thread;
- regressions verify both signature parity and execution from an active loop.

I reran the mock-specific Python suite (7/7 passed), changed-file Ruff (passed), and `git diff --check` (clean). No further findings — LGTM.

harrymove-ctrl:
## Review — GH #581 (MemWalMock)

**Approve.** Verified the mock is genuinely offline (no `fetch`/`httpx`/`requests`/env-var reads in either `mock.py` or `mock.ts`, and tests actively assert this by making the real network path throw), namespace-isolated, and deterministic (pure token-overlap ranking + hash-based embeddings, no `Math.random`/`random.random`/wall-clock-seeded logic; tie-break via a monotonic sequence counter). Exported from both SDK entry points, and method signatures closely mirror the real clients.

One non-blocking gap: `recallManual`/`recall_manual` and `remember_manual` aren't implemented on the mock — a test using them gets a loud `AttributeError`/`TypeError`, not silent divergent behavior, so it's a coverage gap rather than a correctness hazard. Worth a doc note or a tracked follow-up if those flows matter for CI coverage later, not a blocker here.

Depth: targeted (low-risk, additive SDK surface; CI already green 26/26 TS, 121/121+16 skipped Python).

harrymove-ctrl (approved):

PR #584, opened 2026-08-10
feat(mcp): OAuth 2.1 foundation + authorization server for Claude custom connectors (1-2/3)

harrymove-ctrl:
## Summary
Implements the Notion task **Add Claude custom-connector compatibility to remote MCP** for the hosted MemWal MCP endpoint (`/api/mcp`), plus follow-up fixes from dev validation.

Connector URL for Claude custom connector testing:

```text
https://relayer.dev.memwal.ai/api/mcp
```

Related discovery endpoints:

```text
https://relayer.dev.memwal.ai/.well-known/oauth-authorization-server
https://relayer.dev.memwal.ai/.well-known/oauth-protected-resource
```

This PR is scoped to the MemWal relayer and web app OAuth/custom-connector work only. The Claude Code marketplace plugin work is maintained separately in https://github.com/CommandOSSLabs/walrus-memory-plugin and is intentionally not part of this PR.

This PR currently includes:

- OAuth 2.1 authorization-server foundation and routes for Claude custom connectors.
- `/api/mcp` OAuth bearer resolution through the proxy.
- Dev-environment fixes for the hosted Claude connect flow.
- Account-setup resume handling: if a user starts from `/connect/claude` or `/connect/mcp`, has no Walrus Memory account, creates one in `/setup`, the app returns them to the original connector flow instead of dropping them on the dashboard.

## Notion Acceptance Criteria
- [x] Claude can add the hosted MemWal endpoint as a native custom connector using the connector UI. Dev endpoint is exposed at `https://relayer.dev.memwal.ai/api/mcp`; OAuth discovery is live.
- [x] Login and consent complete without manually pasting a delegate private key or custom headers. The consent flow mints/uses server-custodied OAuth delegates; the no-account path now returns users to the connector flow after setup.
- [ ] `tools/list` and representative memory tool calls succeed after authorization. Needs one final live Claude UI smoke test after the requested security changes land.
- [ ] OAuth discovery, refresh, expiry, revocation, and error responses follow the current MCP authorization specification. Security review follow-ups remain open: delegate-reuse wallet proof, scope validation/enforcement, atomic client-bound refresh rotation, and trusted-proxy IP handling.
- [ ] Existing MCP transport and security tests remain green after the follow-up changes and rebase.

## OAuth / Connector Scope
[...]

PR #603, opened 2026-08-12
docs(mcp): Claude custom connector guide, plus OAuth reference corrections

reemsabawi-mystenlabs:
## Summary

Follows #584, which shipped OAuth 2.1 on the hosted MCP endpoint. Adds the user-facing how-to Kimberly asked for in Slack, and corrects two things in the reference page that #584 introduced.

### New page: `docs/mcp/claude-connector.md`

Covers the flow end to end: picking the right auth path per client, the four connector steps, what the consent screen actually grants, the two-step disconnect, and troubleshooting. Navigation entry added after `mcp/claude-desktop` (one-line `docs.json` change).

The disconnect section exists because revoking OAuth tokens does **not** remove the onchain delegate key. `POST /oauth/revoke` revokes the token or, for a refresh token, the whole grant, and stops there. Without the second step a user who removes the connector still has an authorized delegate on their account.

### Corrections to `docs/mcp/reference.md`

1. **The delegate-key flow was backwards.** The configuration section described the browser generating the keypair and the private key never leaving it. `/oauth/authorize` calls `oauth::generate_delegate_key()` server-side and stores the encrypted private key, which is *why* `MCP_OAUTH_DELEGATE_ENCRYPTION_KEY` exists at all. The prose two sections up already said this correctly ("the server generate and custody an encrypted delegate key"), so the page contradicted itself. Rewrote the flow block to match the code and added a warning that this is the one flow where a server holds a delegate private key.
2. **Missing variable.** `MCP_OAUTH_REGISTRATION_TRUSTED_CIDRS` (default `160.79.104.0/21`, Anthropic's connector egress range) was absent from the optional-overrides table.
3. **`consent_url` derivation was described imprecisely.** For a `relayer.` host the code drops the label rather than substituting `memory`.
4. Style-guide pass over the same page: bold table headers, sentence-case headings, active voice, unwrapped prose, body text between stacked headings.

## Sources

| Claim | Source |
| --- | --- |
| The relayer generates the delegate keypair server-side and encrypts it before storage | `services/server/src/oauth.rs` `generate_delegate_key()`; `services/server/src/routes/oauth.rs` `authorize`, which calls it then `encrypt_delegate_private_key` |
| Envelope format `v1.<nonce>.<ciphertext>`, AES-256-GCM | `services/server/src/oauth.rs` `encrypt_delegate_private_key` |
| Scopes `memwal:read memwal:write offline_access` | `services/server/src/oauth.rs` `DEFAULT_SCOPES` |
| Access token 1 h, refresh 30 d, code 5 min, session 15 min | `McpOAuthConfig::from_env` defaults in `services/server/src/oauth.rs` |
| Redirect allowlist is `claude.ai` plus RFC 8252 loopback | `CLAUDE_HOSTED_CALLBACK`, `is_loopback_host`, and the `MCP_OAUTH_ALLOWED_REGISTRATION_HOSTS` default in `services/server/src/oauth.rs` |
| `MCP_OAUTH_REGISTRATION_TRUSTED_CIDRS` default `160.79.104.0/21` | same file, `registration_trusted_cidrs` |
| Revoking a refresh token ends the grant; revoking an access token ends only that token | `services/server/src/routes/oauth.rs` `revoke` |
| Unsetting the encryption key turns the OAuth routes off (404) | `from_env` returns `None`; `require_oauth` returns 404 in `services/server/src/routes/oauth.rs` |
| Consent screen renders server-fetched session data, not URL params | `apps/app/src/pages/ConnectClaude.tsx` header comment and `GET /api/oauth/session/{id}` fetch |
| The `add_delegate_key` transaction is Enoki-sponsored | `apps/app/src/hooks/useSponsoredTransaction.ts` |
| No-account path routes to `/setup` and returns to the connector flow | `ConnectClaude.tsx` `no-account` step; `App.tsx` redirect to `/connect/claude` |
| `consent_url` derivation | `McpOAuthConfig::from_env` in `services/server/src/oauth.rs` |
| Environment URLs in the connector table | `docs/relayer/public-relayer.md` (prod, staging), `docs/relayer/benchmark-ci-setup.md` (dev) |

Prose is written fresh; no content was copied from the Notion task, which the docs integration cannot read.

## Remaining review questions

1. **Server-custodied keys are stated plainly, not softened.** The warning tells users the relayer holds a delegate private key for this flow. Worth a look from whoever owns the security messaging before this goes public alongside the connector submission.
2. **Claude Code over OAuth.** The allowlist supports RFC 8252 loopback, which the page attributes to Claude Code. Worth confirming that path has actually been exercised, since the existing Claude Code page documents header auth.

## Test plan

- Internal links and anchors in both edited pages resolve against the docs tree (checked programmatically, including the `#mcp-oauth-2-1-configuration` and `#streamable-http` targets).
- `docs.json` parses and the new entry sits under the MCP group.
- Local style gate passes on both files.
- Production deployment verification: both OAuth discovery routes return `200`, and an unauthenticated request to `https://relayer.memory.walrus.xyz/api/mcp` returns the expected `401`.

---

### DIFF
--- a/docs/mcp/claude-connector.md
+++ b/docs/mcp/claude-connector.md
@@ -66,8 +66,8 @@ The OAuth path and the header path reach the same tools through the same relayer
 
 | **Environment** | **Connector URL** |
 | --- | --- |
-| Production (mainnet) | `https://relayer.memory.walrus.xyz/api/mcp` |
-| Staging (testnet) | `https://relayer-staging.memory.walrus.xyz/api/mcp` |
+| Production (Mainnet) | `https://relayer.memory.walrus.xyz/api/mcp` |
+| Staging (Testnet) | `https://relayer-staging.memory.walrus.xyz/api/mcp` |
 | Dev | `https://relayer.dev.memwal.ai/api/mcp` |
 
 ## Add the connector
--- a/docs/mcp/claude-connector.md
+++ b/docs/mcp/claude-connector.md
@@ -33,9 +33,9 @@ questions:
   - How do I disconnect the Walrus Memory connector from Claude?
 answer: >-
   Add Walrus Memory in Claude's custom connector settings by pasting the MCP URL
-  of a relayer that has OAuth enabled. Staging and dev serve the discovery routes
-  today; the production relayer has not enabled OAuth yet. Claude
-  discovers the authorization server, registers itself, and opens the Walrus
+  of a relayer that has OAuth enabled. Production, staging, and dev serve the
+  discovery routes today. Claude discovers the authorization server, registers
+  itself, and opens the Walrus
   Memory consent screen, where you connect your Sui wallet and authorize a
   delegate key onchain. The relayer generates that delegate key, encrypts it at rest, and
   custodies it, because Claude cannot hold a Sui wallet key itself. To
@@ -71,11 +71,7 @@ The relayer's redirect allowlist accepts RFC 8252 loopback addresses and unit te
 | --- | --- | --- |
 | Staging (Testnet) | `https://relayer-staging.memory.walrus.xyz/api/mcp` | Discovery routes serve traffic |
 | Dev | `https://relayer.dev.memwal.ai/api/mcp` | Discovery routes serve traffic |
-| Production (Mainnet) | `https://relayer.memory.walrus.xyz/api/mcp` | The operator has not enabled OAuth here yet. Both discovery routes return `404`, so the connector cannot complete. |
-
-<Warning>
-Do not hand the production URL to users until its discovery routes answer. Confirm with `curl -i https://relayer.memory.walrus.xyz/.well-known/oauth-authorization-server` first.
-</Warning>
+| Production (Mainnet) | `https://relayer.memory.walrus.xyz/api/mcp` | OAuth is live and the discovery routes serve traffic |
 
 ## Add the connector
 
@@ -140,7 +136,7 @@ The same split applies to the stdio client: `memwal_logout` clears local credent
 ## Troubleshooting
 
 **Claude reports that it cannot find an authorization server.**
-That relayer has no OAuth configuration, which is the expected result on production today. Check `GET /.well-known/oauth-authorization-server` on the host, and see [MCP OAuth 2.1 configuration](/mcp/reference#mcp-oauth-2-1-configuration) for what an operator sets to enable it.
+That relayer may have no OAuth configuration. Check `GET /.well-known/oauth-authorization-server` on the host, and see [MCP OAuth 2.1 configuration](/mcp/reference#mcp-oauth-2-1-configuration) for what an operator sets to enable it.
 
 **The consent screen rejects the link.**
 The session ID never arrived, or the relayer already expired it. Consent sessions last 15 minutes by default. Start the connector flow again from Claude.
--- a/docs/mcp/claude-connector.md
+++ b/docs/mcp/claude-connector.md
@@ -34,11 +34,11 @@ questions:
 answer: >-
   Add Walrus Memory in Claude's custom connector settings by pasting the MCP URL
   of a relayer that has OAuth enabled. Staging and dev serve the discovery routes
-  today; production is not enabled yet. Claude
+  today; the production relayer has not enabled OAuth yet. Claude
   discovers the authorization server, registers itself, and opens the Walrus
   Memory consent screen, where you connect your Sui wallet and authorize a
-  delegate key onchain. The relayer generates and custodies that delegate key,
-  encrypted at rest, because Claude cannot hold a Sui wallet key itself. To
+  delegate key onchain. The relayer generates that delegate key, encrypts it at rest, and
+  custodies it, because Claude cannot hold a Sui wallet key itself. To
   disconnect, remove the connector in Claude, then remove the delegate key from
   the Walrus Memory dashboard.
 ---
@@ -54,12 +54,12 @@ The connector flow needs a relayer that has OAuth turned on. Confirm the endpoin
 | **Client** | **Flow** | **Where to look** |
 | --- | --- | --- |
 | Claude web and Claude Desktop | Custom connector over OAuth | The steps below |
-| Claude Code | Header authentication, or the same OAuth flow over an RFC 8252 loopback redirect | [Claude Code](/mcp/claude-code) |
+| Claude Code | Header authentication, the verified path | [Claude Code](/mcp/claude-code) |
 | Cursor, Codex, Antigravity, OpenCode | stdio MCP server | [Overview](/mcp/overview) |
 
 The OAuth path and the header path reach the same tools through the same relayer. They differ in who holds the delegate key, which [What you approve](#what-you-approve) explains.
 
-The relayer's redirect allowlist accepts RFC 8252 loopback addresses, so a native-app OAuth client such as Claude Code can complete the same flow against a loopback callback.
+The relayer's redirect allowlist accepts RFC 8252 loopback addresses and unit tests cover that path, so a native-app OAuth client such as Claude Code can complete the same flow against a loopback callback. Nobody has run a Claude Code OAuth connection end to end yet, so header authentication stays the documented path for Claude Code until someone does.
 
 ## Prerequisites
 
@@ -71,7 +71,7 @@ The relayer's redirect allowlist accepts RFC 8252 loopback addresses, so a nativ
 | --- | --- | --- |
 | Staging (Testnet) | `https://relayer-staging.memory.walrus.xyz/api/mcp` | Discovery routes serve traffic |
 | Dev | `https://relayer.dev.memwal.ai/api/mcp` | Discovery routes serve traffic |
-| Production (Mainnet) | `https://relayer.memory.walrus.xyz/api/mcp` | Not enabled yet. Both discovery routes return `404`, so the connector cannot complete. |
+| Production (Mainnet) | `https://relayer.memory.walrus.xyz/api/mcp` | The operator has not enabled OAuth here yet. Both discovery routes return `404`, so the connector cannot complete. |
 
 <Warning>
 Do not hand the production URL to users until its discovery routes answer. Confirm with `curl -i https://relayer.memory.walrus.xyz/.well-known/oauth-authorization-server` first.
@@ -131,7 +131,7 @@ The connector flow puts a delegate private key on the server. Claude cannot hold
 Disconnecting takes two steps, because removing the connector does not remove the onchain delegate.
 
 1. **Remove the connector in Claude.** This stops Claude from using it. Whether Claude also calls the relayer's revoke endpoint is not documented, so do not rely on removal alone to end the grant. On the relayer side, revoking a refresh token ends the whole grant and every access token issued under it, while revoking an access token ends only that token.
-2. **Remove the delegate key in the dashboard.** Open [memory.walrus.xyz](https://memory.walrus.xyz), find the delegate that matches the address the consent screen showed, and remove it. Until you do, the delegate remains authorized on your account onchain.
+2. **Remove the delegate key in the dashboard.** Open [memory.walrus.xyz](https://memory.walrus.xyz), find the delegate that matches the address the consent screen showed, and remove it. Until you do, that delegate keeps its onchain authorization on your account.
 
 <Note>
 The same split applies to the stdio client: `memwal_logout` clears local credentials but leaves the onchain delegate in place. See [Logout semantics](/mcp/reference#logout-semantics).
--- a/docs/mcp/claude-connector.md
+++ b/docs/mcp/claude-connector.md
@@ -43,7 +43,7 @@ answer: >-
   the Walrus Memory dashboard.
 ---
 
-Claude's built-in custom connector flow adds Walrus Memory over OAuth 2.1. You approve access in the browser with your Sui wallet, and Claude never asks you for a delegate private key or a custom header.
+Claude's built-in custom connector flow adds Walrus Memory over [OAuth 2.1](https://oauth.net/2.1/). You approve access in the browser with your [Sui wallet](https://docs.sui.io/guides/developer/wallets/what-is-a-wallet), and Claude never asks you for a delegate private key or a custom header.
 
 <Note>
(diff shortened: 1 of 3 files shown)


jessiemongeon1:
## Style Guide Audit

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,138 @@
...
+1. **Remove the connector in Claude.** Claude revokes its tokens against the relayer. Revoking a refresh token ends the whole grant, so the access tokens issued under it stop working too.
+2. **Remove the delegate key in the dashboard.** Open [memory.walrus.xyz](https://memory.walrus.xyz), find the delegate that matches the address the consent screen showed, and remove it. Until you do, the delegate remains authorized on your account onchain.
+
+<Note>
+The same split applies to the stdio client: `memwal_logout` clears local credentials but leaves the onchain delegate in place. See [Logout semantics](/mcp/reference#logout-semantics).
+</Note>
+
+## Troubleshooting
+
+- **Claude reports that it cannot find an authorization server.** The relayer has no OAuth configuration. Check `GET /.well-known/oauth-authorization-server` on that host, and see [MCP OAuth 2.1 configuration](/mcp/reference#mcp-oauth-2-1-configuration) for what an operator sets.
```
jessiemongeon1: should use the same troubleshooting format you've been using for the other pages with this section - needs to be consistent.

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,138 @@
...
+  MCP URL, for example https://relayer.memory.walrus.xyz/api/mcp. Claude
+  discovers the authorization server, registers itself, and opens the Walrus
+  Memory consent screen, where you connect your Sui wallet and authorize a
+  delegate key onchain. The relayer generates and custodies that delegate key,
+  encrypted at rest, because Claude cannot hold a Sui wallet key itself. To
+  disconnect, remove the connector in Claude, then remove the delegate key from
+  the Walrus Memory dashboard.
+---
+
+Claude's built-in custom connector flow adds Walrus Memory over OAuth 2.1. You approve access in the browser with your Sui wallet, and Claude never asks you for a delegate private key or a custom header.
```
jessiemongeon1: needs review from eng

jessiemongeon1 (changes requested):

ducnmm (dismissed):
Engineering review on exact head `4773edd1d96d6a1818244e6a5a832412fdfd82ca`.

The server-custody correction is accurate and important, and the staging discovery endpoints are now live. I found several claims that must be corrected before these docs are published:

1. **[P1] Do not present production as live yet.** The answer and environment table direct users to `https://relayer.memory.walrus.xyz/api/mcp`, but both production OAuth discovery routes currently return 404. Mark production as pending/not yet enabled, or hold this user-facing guide until the production promotion and smoke test pass. Dev and staging discovery currently return 200.

2. **[P1] Do not claim the Claude Code OAuth loopback flow has been validated.** The implementation has redirect allowlist/unit coverage for RFC 8252 loopback, but we do not have an end-to-end Claude Code OAuth test. The client table should recommend the tested stdio/header-auth path and describe loopback OAuth as implementation-supported but unverified, not as an equivalent documented flow.

3. **[P1] Correct `MCP_OAUTH_REGISTRATION_TRUSTED_CIDRS`.** It is not the set of networks that may register. In `oauth.rs`, these CIDRs are exempt from the per-IP DCR throttle; other source IPs may still register when their redirect URI passes the allowlist and normal rate limits. The current wording materially overstates the security boundary.

4. **[P1] Bring `memwal_restore` reference in sync with the shipped contract.** The guide still says limit `1–500` and “returns counts only.” Current MCP/relayer behavior caps it at `100` and returns `truncated`; callers must know that `truncated=true` means the bounded restore is incomplete.

5. **[P2] Fix delegate reuse wording.** The relayer currently calls `find_reusable_oauth_delegate(account_id)` and can reuse an active OAuth delegate for the account, not specifically a delegate previously minted for “this client.” Avoid implying client-specific delegate custody here.

6. **[P2] Do not state every grant contains all three scopes.** The authorization server validates and grants the requested subset. Describe `memwal:read`, `memwal:write`, and `offline_access` as supported/requestable scopes; only claim all three when the displayed consent request actually contains them.

7. **[P2] Avoid asserting that removing a connector in Claude revokes its tokens unless the Claude UI behavior is documented/tested.** Server behavior is clear: `POST /oauth/revoke` on a refresh token revokes the grant, while revoking an access token revokes only that token. Phrase the client-removal step conservatively and separately document server-side token/delegate revocation semantics.

The existing docs-style requested changes also remain active. Please rebase onto current `dev`, address these engineering findings and the troubleshooting-format comment, then request another exact-head review. Do not merge this docs PR directly into the current production promotion until the environment-status claims match live endpoints.

ducnmm (changes requested):
Revising my earlier review to focus only on whether the documentation matches the current code. Production OAuth is expected to go live separately, so its current deployment status is not a blocker for this docs PR.

The OAuth architecture is represented correctly overall: authorization-code + PKCE, server-generated and encrypted delegate custody, refresh-token rotation, revocation, scope enforcement, and RFC 8252 loopback support all exist in the implementation.

Please correct these remaining code/docs mismatches:

1. `MCP_OAUTH_REGISTRATION_TRUSTED_CIDRS` identifies source networks exempt from the per-IP dynamic-registration throttle. It is not the set of networks allowed to register; other IPs can register subject to redirect validation and rate limits.
2. `memwal_restore` is bounded to `1–100`, not `1–500`, and its result includes `truncated`. Document that `truncated: true` means more eligible work may remain.
3. Delegate reuse currently uses `find_reusable_oauth_delegate(account_id)`, so it may reuse an active OAuth delegate for the account rather than only one previously minted for the same client.
4. `memwal:read`, `memwal:write`, and `offline_access` are supported/requestable scopes. The server grants the validated requested subset, so the docs should not imply every grant always contains all three.

The Claude Code loopback description is supported by the code and unit tests; this review does not require production or end-to-end rollout evidence. The separate docs-style feedback from Jessie still applies.

ducnmm (approved):
Re-reviewed exact head `e926f7a192ee7c4d093e030a4ad9fae40406adca` against the current implementation. The code/documentation mismatches from my previous review are resolved:

- trusted registration CIDRs are correctly described as a rate-limit exemption;
- restore documents the `1–100` bound and `truncated` result;
- delegate reuse is correctly described as account-level;
- scopes are correctly described as the requested subset;
- server-side delegate generation/custody and Claude Code loopback support match the code.

Engineering approval. The production OAuth status text is deployment-specific rather than a code concern; update that row/warning when production is enabled. Jessie's docs review remains separate.

ducnmm (approved):
Re-reviewed exact head `2141c1ad` after the production-status update. Both production OAuth discovery routes return 200, and unauthenticated `/api/mcp` returns the expected 401. The stale production warning and troubleshooting text are removed. Engineering approval.

ducnmm:
@reemsabawi-mystenlabs Production OAuth status is now updated on exact head `2141c1ad`: both discovery routes return `200`, the stale production warning is removed, and engineering re-review is approved. Please take a final look and coordinate the remaining docs approval from @jessiemongeon1 before merge.

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,148 @@
...
+  discovery routes today. Claude discovers the authorization server, registers
+  itself, and opens the Walrus
+  Memory consent screen, where you connect your Sui wallet and authorize a
+  delegate key onchain. The relayer generates that delegate key, encrypts it at rest, and
+  custodies it, because Claude cannot hold a Sui wallet key itself. To
+  disconnect, remove the connector in Claude, then remove the delegate key from
+  the Walrus Memory dashboard.
+---
+
+Claude's built-in custom connector flow adds Walrus Memory over OAuth 2.1. You approve access in the browser with your Sui wallet, and Claude never asks you for a delegate private key or a custom header.
```
jessiemongeon1: needs embedded link to what oauth 2.1 is and a link to docs on sui wallet

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,148 @@
...
+<Note>
+The connector flow needs a relayer that has OAuth turned on. Confirm the endpoint you plan to use answers `GET /.well-known/oauth-authorization-server` before you hand the URL to someone else. A relayer without the OAuth configuration returns `404` on that route and works only with [header authentication](/mcp/reference#streamable-http).
+</Note>
+
+## Pick the right flow
+
+| **Client** | **Flow** | **Where to look** |
+| --- | --- | --- |
+| Claude web and Claude Desktop | Custom connector over OAuth | The steps below |
+| Claude Code | Header authentication, the verified path | [Claude Code](/mcp/claude-code) |
```
jessiemongeon1: unclear what 'the verified path' means here

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,148 @@
...
+| Cursor, Codex, Antigravity, OpenCode | stdio MCP server | [Overview](/mcp/overview) |
+
+The OAuth path and the header path reach the same tools through the same relayer. They differ in who holds the delegate key, which [What you approve](#what-you-approve) explains.
```
jessiemongeon1:
```suggestion
The OAuth path and the header path reach the same tools through the same relayer. They differ in [who holds the delegate key](#what-you-approve).
```

jessiemongeon1: better SEO for linking on direct question statements like this

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,148 @@
...
+
+The relayer's redirect allowlist accepts RFC 8252 loopback addresses and unit tests cover that path, so a native-app OAuth client such as Claude Code can complete the same flow against a loopback callback. Nobody has run a Claude Code OAuth connection end to end yet, so header authentication stays the documented path for Claude Code until someone does.
```
jessiemongeon1:
```suggestion
The relayer's redirect allowlist accepts RFC 8252 loopback addresses and unit tests cover that path, so a native-app OAuth client such as Claude Code can complete the same flow against a loopback callback. 
```

jessiemongeon1: out of scope of our docs, not relevant for us to have to maintain this statement - we can simply add docs for it when it is added or supported

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Prerequisites
+
+- A Sui wallet in the browser you use for the consent screen.
```
jessiemongeon1: needs link to sui wallet docs

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,148 @@
...
+- A Walrus Memory account that the wallet owns. If the wallet owns no account yet, the consent screen sends you through the one-time setup and returns you to the connector flow.
```
jessiemongeon1: link to what walrus memory account means

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,148 @@
...
+- The MCP URL of a relayer that has OAuth turned on.
+
+| **Environment** | **Connector URL** | **OAuth status** |
+| --- | --- | --- |
+| Staging (Testnet) | `https://relayer-staging.memory.walrus.xyz/api/mcp` | Discovery routes serve traffic |
+| Dev | `https://relayer.dev.memwal.ai/api/mcp` | Discovery routes serve traffic |
+| Production (Mainnet) | `https://relayer.memory.walrus.xyz/api/mcp` | OAuth is live and the discovery routes serve traffic |
+
+## Add the connector
```
jessiemongeon1, line 74: have you followed these steps yourself and confirmed that all the button text and things like that are correct and match? This is relevant to the thread from this morning on things changing in the Claude/Codex clients.

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,148 @@
...
+
+Access tokens last 1 hour and refresh tokens last 30 days by default. A self-hosted relayer can change both. See [MCP OAuth 2.1 configuration](/mcp/reference#mcp-oauth-2-1-configuration).
+
+<Warning>
+The connector flow puts a delegate private key on the server. Claude cannot hold a Sui wallet key, so the relayer generates a delegate keypair, encrypts the private key with AES-256-GCM before it stores the key, and decrypts it in memory to sign your MCP calls. The [stdio](/mcp/overview) and header flows keep the delegate key on your own machine instead. Choose the flow whose trust boundary you accept, and use the dashboard to remove a delegate you no longer want.
+</Warning>
+
+## Disconnect
+
+Disconnecting takes two steps, because removing the connector does not remove the onchain delegate.
```
jessiemongeon1, line 125: why does the module above use steps and this one uses a number list? this should be consistent.

also, have you followed these steps yourself as well to confirm correct flow/correct button descriptions, etc?

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,148 @@
...
+
+1. **Remove the connector in Claude.** This stops Claude from using it. Whether Claude also calls the relayer's revoke endpoint is not documented, so do not rely on removal alone to end the grant. On the relayer side, revoking a refresh token ends the whole grant and every access token issued under it, while revoking an access token ends only that token.
+2. **Remove the delegate key in the dashboard.** Open [memory.walrus.xyz](https://memory.walrus.xyz), find the delegate that matches the address the consent screen showed, and remove it. Until you do, that delegate keeps its onchain authorization on your account.
+
+<Note>
+The same split applies to the stdio client: `memwal_logout` clears local credentials but leaves the onchain delegate in place. See [Logout semantics](/mcp/reference#logout-semantics).
```
jessiemongeon1: link to stdio client, otherwise it looks like a typo

docs/mcp/claude-connector.md
```diff
@@ -0,0 +1,148 @@
...
+</Note>
+
+## Troubleshooting
+
+**Claude reports that it cannot find an authorization server.**
```
jessiemongeon1: needs to use the same troubleshooting format used in other pages - using a different format for each isn't ideal, needs consistency

jessiemongeon1 (changes requested):

jessiemongeon1:
## Style Guide Audit

jessiemongeon1:
## Style Guide Audit

jessiemongeon1 (approved):

Issue #567, opened 2026-08-09
[DX] Add Token Budget Pruning & Similarity Thresholds to Vercel AI SDK Middleware


sandman-sh:
### Problem Statement
When wrapping `MemWal` as middleware for Vercel AI SDK (`streamText` / `generateText`), all recalled memories matching the `owner + namespace` boundary are injected into the prompt context. On long-running agent threads, this injects irrelevant historical memories, inflating token count, escalating inference costs, and triggering "lost in the middle" attention decay.

### Proposed Solution
Add configurable token budgeting and similarity score thresholds directly to the middleware initialization options:

```typescript
import { createMemWalMiddleware } from '@mysten-incubation/memwal/ai';

const memoryMiddleware = createMemWalMiddleware({
  owner: userAddress,
  namespace: 'finance-assistant',
  maxMemoryTokens: 1500,       // Truncates memory payload to fit prompt token budget
  minSimilarityScore: 0.78,    // Drops low-relevance vector matches
  topK: 5                      // Retains only top-5 relevant chunks
});
```

Issue #528, opened 2026-08-04
[Bug] recall() silently truncates results with low default limit


gclgcl-87:
Bugs / Unexpected Behaviorrecall() silently truncates results with a low default limit (previously reported as 10). This causes incomplete context and can quietly corrupt accuracy in longer-running agents.
Parameter name mismatch: documentation uses top_k while the SDK expects limit, leading to TypeErrors on first use.
Namespace identifiers are silently truncated to 12 characters by the relayer with no warning or clear documentation. This can cause data loss or unexpected collisions.
No native way to delete individual memories or clear a namespace. Stale or incorrect memories can only be worked around by rotating namespace prefixes.

Missing Features (High Priority)TTL support on remember(). After ~2 weeks of continuous use, a large portion of memories become stale noise and degrade recall quality.
Cursor-based pagination for recall(). Currently only top-K results are returned with no way to page through large memory sets.
A non-semantic list method to retrieve all memories in a namespace (by time or ID). Semantic recall is a poor workaround for building memory panels or audit UIs.
Proper update / upsert capabilities for structured state (e.g. user preferences).
Clearer and more complete management of delegate permissions in the dashboard.

Unclear DocumentationMCP tool names and parameter schemas are scattered. The README and SDK API reference focus on TypeScript methods, making Agent integration harder than necessary.
Relayer self-hosting, cost, latency, and failure modes are under-documented.
Key constraints (namespace length limit, default limit value, restore behavior, embedding details) are not prominently documented.
Error messages are often not actionable enough for quick troubleshooting.

Access & Onboarding IssuesCredential generation requires visiting the Playground/Dashboard (Google login or wallet connection). A fully headless / API-key style flow would greatly improve CI and scripted use cases.
Multiple peer dependencies (@mysten/sui, @mysten/seal, @mysten/walrus, etc.) increase the chance of version mismatches.
The multi-component stack (Sui + Walrus + Seal + Relayer) means any temporary issue in one layer affects the whole experience.

ducnmm:
Thank you for taking the time to share this detailed feedback.

A number of the items listed here, including pagination, non-semantic memory listing, TTL, deletion, and update/upsert support, are feature requests rather than bugs in the current semantic top-K recall design. That said, these capabilities are aligned with the direction we want the project to evolve toward, particularly for long-running agents and memory management workflows.

[...]

PR #605, opened 2026-08-12
Feat: Native token counting + max_tokens/truncation on recall

### LINKED ISSUE #592: [SDK/Feature] Native token counting + max_tokens / truncation on recall
## Summary

Please add **first-class token budgeting** to the MemWal client so apps don’t have to manage context bloat only via heavy system prompts (“recall max 3 hits”, “don’t dump full history”).

Today `memwal_recall` returns raw text hits; agents paste them into the LLM window with no SDK-level sense of **how many tokens** that payload costs. In multi-session agents (`continuity-keeper`, architecture bibles, etc.) broad semantic recall easily overfills the context window → cognitive fatigue / format drift.

Related but different from [#567](https://github.com/MystenLabs/MemWal/issues/567) (Vercel AI middleware pruning). This request is for the **core MemWal / MCP client**: count + truncate at the memory boundary before the model sees the payload.

## Environment / context

- `@mysten-incubation/memwal-mcp` on Walrus Mainnet
- Agent loops that call `memwal_recall` every scene / every decision
- Session 7 Prompt Evolution — Continuity Keeper evolution under namespace `session7-continuity`

## Problem

1. Recall can return long historical notes (full ADR bodies, multi-fact character sheets).
2. Developers only have coarse knobs (`limit` / `k`) — **hit count ≠ token count**.
3. Prompts absorb the burden (“limit to 3 nodes”, “summarize before use”), which is brittle and wastes system-prompt tokens.
4. No standard way to log or assert “this recall cost ≤ N tokens” in CI or agent telemetry.

Also see silent truncation / low default limit discussions: [#528](https://github.com/MystenLabs/MemWal/issues/528), [#271](https://github.com/MystenLabs/MemWal/issues/271).

### 1. `countTokens(text)` (client-side)

Lightweight tokenizer helper (tiktoken or a documented approximation) for payload weight before injection:

```typescript

const hits = await memwal.recall({
  namespace: "session7-continuity",
  query: "Tide Signal rule cost",
  limit: 8,
});

const joined = hits.map(h => h.text).join("\n---\n");
const tokens = memwal.countTokens(joined); // e.g. cl100k_base or model-tagged estimate

if (tokens > 2048) {
  // app decides: re-query with max_tokens, or summarize
}
```

### 2. `max_tokens` + truncation strategy on `recall`

```typescript
const context = await memwal.recall({
  namespace: "session7-continuity",
  query: "Minh Lan Harbor Node Tide Signal",
  limit: 10,
  max_tokens: 2048,
  truncation_strategy: "high-relevance-only", // or "per-hit-cap" | "drop-tail"
});

// Optional metadata for agents / logs
// context.meta.token_estimate === 1893
```

Suggested strategies:

| Strategy | Behavior |
|----------|----------|
| `high-relevance-only` | Keep lowest-distance hits until token budget fills |
| `per-hit-cap` | Truncate each hit’s text to a token ceiling |
| `drop-tail` | Preserve order; cut from the end of the concatenated payload |

### 3. MCP surface (optional but high leverage)

Expose the same knobs on `memwal_recall` for Cursor/Claude agents, e.g. `max_tokens`, and return `token_estimate` in the tool result so prompts can stay thin.

## Why this belongs in the SDK

- Token accounting is a **platform concern** for memory products, not just app prompt craft.
- Moves optimization out of system prompts → better DX and more predictable agent behavior.
- Complements entry metadata ([#434](https://github.com/MystenLabs/MemWal/issues/434)) and job/index status ([#409](https://github.com/MystenLabs/MemWal/issues/409)).

## Acceptance criteria (suggested)

- [ ] Documented `countTokens` (or `estimateTokens`) with model/encoding note
- [ ] `recall({ max_tokens, truncation_strategy })` honored client-side and/or relayer-side
- [ ] Response includes estimate used for truncation decisions
- [ ] MCP `memwal_recall` accepts the same options when applicable

## Impact

- Severity: **Medium–High DX** for production agents
- Workaround today: prompt rules + manual `limit` + app-side tiktoken — duplicated across every MemWal project

Filed during **Walrus Sessions 7 · Prompt Evolution** (Bug Bounty track).

hungtranphamminh:
### Why
Recall returns raw text hits, but the SDK gives callers only coarse knobs (`limit`/`topK`) to control
how much they inject into a model context — hit count ≠ token count. Multi-session agents that recall
broadly overfill the context window, so today apps hand-roll their own tiktoken + truncation layer, or
lean on brittle prompt rules ("recall max 3 hits"). There's no SDK-level way to weigh a recall payload
or assert "this recall cost ≤ N tokens".

### What
First-class, client-side token budgeting on `recall()`:

- **`estimateTokens(text)` / `memwal.countTokens(text)`** — a lightweight token estimator (also
  exported standalone).
- **`recall({ maxTokens, truncationStrategy, countTokens })`** — trims the recalled memories to fit a
  token budget, with three strategies.
- **`RecallResult.meta { tokenEstimate, truncated }`** — the returned payload's estimated cost and
  whether anything was dropped/shortened. Present only when `maxTokens` is supplied.

### Solution
- **Zero-dependency estimator by default.** `estimateTokens` uses a documented character-based
  approximation (~`chars/4`, counting Unicode code points so emoji/CJK don't skew it). It's an
  estimate, not an exact BPE count — deterministic and CI-assertable. Callers needing exactness inject
  their own counter via `recall({ countTokens })` (e.g. a tiktoken binding). The issue explicitly
  sanctions "tiktoken **or** a documented approximation", so this keeps the SDK dependency-free while
  leaving exactness opt-in.
- **Three behaviorally-distinct truncation strategies:**
  - `high-relevance-only` (default) — keep the lowest-distance **whole** hits until the budget fills;
    drop the rest. Never a partial hit.
  - `drop-tail` — preserve order, keep whole hits from the front, then **truncate the single boundary
    hit** that straddles the budget (a partial tail) so the payload fills to the budget.
  - `per-hit-cap` — cap each hit's text to an equal share (`floor(maxTokens / hitCount)`); hits whose
    share rounds to zero are dropped rather than returned empty.
- **Client-side only.** Budgeting runs on already-decrypted hits, at the existing post-response filter
  seam next to `maxDistance`. No relayer, server, or on-chain change; no privacy-floor implication (it
  operates on plaintext the client already holds).
- **Backward compatible.** The budgeting block triggers only on `typeof maxTokens === "number"`.
  Omitting `maxTokens` returns the result byte-identical to before — no `meta` key, `results`/`total`
  untouched.
- **Malformed budgets never lie.** A non-finite budget (`NaN`/`Infinity`) is treated as no-budget
  (`truncated: false`), so a bad input can't report a truncation that didn't happen.

Scope note: this is the **core SDK** ask. Mirroring the same knobs onto the MCP `memwal_recall` tool
is called out in the issue as "optional but high-leverage" and is left as a follow-up. It's independent
of #567 (Vercel AI middleware pruning), which stays separately usable per that issue's triage.

## Types of Changes

- [x] New feature (non-breaking)
- [ ] Bug fix (non-breaking)
- [ ] Dependency update
- [ ] Documentation
- [x] Test
- [ ] Security / permission-scope change

## Benchmark impact

- [x] No benchmark-affecting change (write/infra only — this is a client-side SDK read-shaping option;
  it does not touch the relayer recall/answer/extraction path)
- [x] Default-off / byte-identical when the option is omitted

## Testing

- [x] Tested locally
- [ ] `cargo test` / harness tests pass (N/A — SDK-only change; SDK suite: 61/61 pass)

## Checklist

- [x] Docs updated if needed (JSDoc on all new public API)
- [x] Respects the privacy floor (no server-readable plaintext index — client-side only)

## Related

- Closes #592
- Related to #567 (Vercel AI middleware token pruning — kept independent), #528 / #271 (silent
  truncation / low default limit)
- Linear: none

### DIFF
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -25,6 +25,13 @@ export {
 // Delegate key utilities (no @mysten/sui dependency)
 export { delegateKeyToSuiAddress, delegateKeyToPublicKey } from "./utils.js";
 export {
+    estimateTokens,
+    truncateToTokenBudget,
+    applyTokenBudget,
+    CHARS_PER_TOKEN,
+} from "./tokens.js";
+export type { TokenCounter } from "./tokens.js";
+export {
     createSponsorAuthorization,
     sponsorAuthorizationMessage,
 } from "./sponsor-auth.js";
@@ -43,6 +50,8 @@ export type {
     RecallMemory,
     RecallOptions,
     RecallParams,
+    RecallTokenMeta,
+    TruncationStrategy,
     ScoringWeights,
     EmbedResult,
     AnalyzeOptions,
--- a/packages/sdk/src/memwal.ts
+++ b/packages/sdk/src/memwal.ts
@@ -68,6 +68,7 @@ import {
     assertCompatibleRelayer,
     compatibilityErrorFromStatus,
 } from "./compatibility.js";
+import { applyTokenBudget, estimateTokens } from "./tokens.js";
 
 // ============================================================
 // Ed25519 Signing (lazy-loaded)
@@ -595,6 +596,25 @@ export class MemWal {
      * const result2 = await memwal.recall("food allergies", 5, "profile");
      * ```
      */
+    /**
+     * Estimate the token cost of a string using the SDK's default
+     * character-based approximation (~chars/4, code-point aware). Useful to weigh
+     * a recall payload before injecting it into a model context:
+     *
+     * ```ts
+     * const hits = await memwal.recall({ query, limit: 8 });
+     * const joined = hits.results.map((h) => h.text).join("\n---\n");
+     * if (memwal.countTokens(joined) > 2048) { /* re-query with maxTokens *\/ }
+     * ```
+     *
+     * This is an estimate, not an exact tokenizer count — see `estimateTokens`
+     * for accuracy caveats. For exact counts, pass your own counter to
+     * `recall({ maxTokens, countTokens })`.
+     */
+    countTokens(text: string): number {
+        return estimateTokens(text);
+    }
+
     async recall(params: RecallParams): Promise<RecallResult>;
     /**
      * @deprecated Positional `recall(query, limit, namespace)` is easy to
@@ -640,18 +660,38 @@ export class MemWal {
                 namespace: resolvedNamespace,
             }, { signal: ac.signal });
 
+            let processed = result;
             if (typeof options.maxDistance === "number") {
                 const filtered = result.results.filter(
                     (memory) => memory.distance < options.maxDistance!,
                 );
-                return {
-                    ...result,
+                processed = {
+                    ...processed,
                     results: filtered,
                     total: filtered.length,
                 };
             }
 
-            return result;
+            // Client-side token budgeting: trim the (already distance-sorted)
+            // payload to fit maxTokens per the chosen strategy, and attach the
+            // token estimate + truncated flag. Omitting maxTokens leaves the
+            // result byte-identical to the pre-budget behavior (no meta).
+            if (typeof options.maxTokens === "number") {
+                const { results: budgeted, meta } = applyTokenBudget(
+                    processed.results,
+                    options.maxTokens,
+                    options.truncationStrategy,
+                    options.countTokens,
+                );
+                processed = {
+                    ...processed,
+                    results: budgeted,
+                    total: budgeted.length,
+                    meta,
+                };
+            }
+
+            return processed;
         } finally {
             clearTimeout(tid);
         }
--- a/packages/sdk/src/mock.ts
+++ b/packages/sdk/src/mock.ts
@@ -19,6 +19,7 @@ import type {
     RememberResult,
     RestoreResult,
 } from "./types.js";
+import { applyTokenBudget } from "./tokens.js";
 
 export interface MemWalMockSeed {
     text: string;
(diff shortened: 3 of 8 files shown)


ducnmm (changes requested):
Found two issues to address before merge:

1. **MemWalMock ignores the new recall budget options.** `RecallOptions` is shared by `MemWal` and `MemWalMock`, so mock callers can pass `maxTokens`, `truncationStrategy`, and `countTokens`, but `packages/sdk/src/mock.ts::recall()` still returns the full ranked payload with no `meta`. This makes offline tests pass against behavior that differs from production. Please apply the same `applyTokenBudget` step in the mock and add parity coverage.

2. **Custom token counters can make truncation quadratic.** `truncateToTokenBudget()` shrinks one code point at a time and rebuilds/re-tokenizes the whole prefix on every iteration. With a valid dense counter, a 50k-character memory took ~1.9s locally; larger recalled notes/tokenizers can block the client much longer. Please use a monotonic binary search (or another bounded approach) for the largest fitting prefix, and add a large-input regression test.

The SDK test suite passes (61/61), and the overall API/strategy wiring looks good once these are fixed.

ducnmm (approved):
Re-reviewed after commits 27d16a2b and 83b576d3. The mock now matches production token-budget behavior, custom-counter truncation is bounded with binary search, the default estimator has a direct fast path, and parity/integration regression coverage was added. Independent Pi review found no blockers. Local SDK suite passes 64/64 and all GitHub checks are green.

Olympusxvn:
Glad to see this merged!

PR #631, opened 2026-08-13
docs: add a multi-agent coordination guide

reemsabawi-mystenlabs:
Closes BEDU-1121. Eight builder questions asked how several agents share memory, and no page tied the pieces together.

## The gap

Delegates and namespaces are each documented well on their own concept page, but nothing says how they combine. That leaves the reader to infer which one is an identity boundary and which is a visibility boundary, and getting it backwards is expensive: **any delegate the account authorizes decrypts everything that account owns, across every namespace.** So namespaces organize an account's memory, and they cannot separate users. The guide states that outright, in a warning, rather than leaving it to be deduced from two pages.

## What the page covers

- **One delegate key per agent.** The keys grant identical access, so this is not an access-control boundary. It buys labelling in the dashboard and per-agent revocation.
- **Three sharing patterns**, as a table: shared namespace, per-agent namespace, and the common case of a shared namespace plus a private one per agent.
- **Handoff as a write followed by a recall**, with the two properties of recall that change how you write for one: it matches by meaning rather than by key, and it returns whole memories, so each fact needs to stand alone.
- **Choosing a boundary**, including the case that must not use namespaces: agents serving different users need separate accounts.

## Also fixes the concept page it contradicted

A review pass found that `fundamentals/concepts/memory-space.md` said the opposite of this guide, in the direction that gets people hurt. It called a memory space "the isolated unit of storage" offering "full isolation between users, apps, and deployments", its goal promised namespaces "prevent cross-namespace data leakage", and it closed with "each is fully isolated". A reader arriving there first would conclude namespaces separate users, which is exactly the mistake this guide exists to prevent.

Neither page stated a falsehood; they described different layers. That page's own App ID section already says the vector database scopes queries by `owner + namespace` while encryption and blob discovery scope by package ID, which is precisely why a delegate reads across namespaces. The isolation language just overstated what that buys you.

Reworded its description, goal, answer, and body to say scoping rather than isolation, and added a warning beside the boundary definition that states the decryption behaviour outright and links here. Fixing it inside this PR rather than filing a follow-up, so the contradiction never ships.

That page also carried twelve pre-existing style gate violations, which surface on push once the file is touched at all: eight em dashes, `SEAL` for `Seal`, a future tense, a quoted frontmatter title, and unbolded table headers. Cleared them in a separate commit. They are mechanical and unrelated to the wording change.

## Sources

| Claim | Source |
| --- | --- |
| An account has one owner and any number of delegates; a delegate is a keypair the owner authorizes | `docs/fundamentals/concepts/ownership-and-access.md`, "Delegates" |
| Delegate keys are registered onchain and verified on every request | same page, "Access Control Enforcement" |
| Only the owner and authorized delegates can access the encrypted content | same page, "Ownership" |
| Namespaces group memories within an account, and one user can hold several | `docs/fundamentals/concepts/memory-space.md`, "Namespace" |
| The vector database scopes queries by `owner + namespace`, while encryption and blob discovery scope by package ID | same page, "App ID" — this is what makes the delegate warning true |
| `MemWal.create` fields `key`, `accountId`, `serverUrl`, `namespace` | `MemWalConfig` in `packages/sdk/src/types.ts` |
| `recall({ query, limit })` | `RecallParams` and `RecallOptions` in the same file |
| `rememberAndWait` waits for the background job to complete | `packages/sdk/src/memwal.ts`, the method and its doc comment |
| `remember` returns once the relayer creates the job, with embedding, encryption, upload, and indexing continuing after | `docs/sdk/api-reference.md`, `remember` |

Prose written fresh. The code samples use only fields and methods verified above.

## Verification

Re-run against current `dev` rather than relying on the original checks, since the branch was 182 commits behind when this pass ran:

- The branch merges cleanly into `dev`; this push brings it up to date.
- Style gate passes on both changed pages.
- `node scripts/check-docs-freshness.mjs` passes: `docs freshness OK (94 pages, 44 routes, IDs, URLs, versions, limits)`, which covers the four reference links.
- Every SDK claim above re-checked against `packages/sdk/src` at current `dev`, not against the versions current when the page was drafted.

### DIFF
--- a/docs/guides/multi-agent-coordination.md
+++ b/docs/guides/multi-agent-coordination.md
@@ -1,0 +1,126 @@
+---
+title: Coordinate Multiple Agents
+description: >-
+  Patterns for running several agents against one Walrus Memory account: give each agent its own
+  delegate key, use namespaces to decide what they share, and hand work from one agent to the next.
+keywords:
+  - multi-agent
+  - coordination
+  - shared memory
+  - delegate key
+  - namespace
+  - handoff
+  - MemWal
+  - Walrus
+goal:
+  description: Decide how several agents share or isolate memory, and implement a handoff between them
+  requires:
+    - has_frontmatter:
+        - title
+        - description
+        - keywords
+      label: Has required frontmatter fields
+    - min_words: 100
+      label: Needs enough content to cover the patterns
+    - has_questions: true
+      label: Needs questions for AI search visibility
+    - has_answer: true
+      label: Needs answer summary for AI citation
+questions:
+  - How do I share memory between multiple agents?
+  - Can two agents use the same Walrus Memory account?
+  - How do I hand off context from one agent to another?
+  - Should each agent have its own delegate key?
+answer: >-
+  Give each agent its own delegate key on one account, then use namespaces to decide what they
+  share. Agents that read and write the same namespace share memory, because any delegate the
+  account authorizes can decrypt everything that account owns. Agents that use separate namespaces
+  never read each other's writes, because recall never crosses a namespace. A handoff is one agent writing to a
+  shared namespace and the next recalling from it.
+---
+
+Several agents can work against one Walrus Memory account. Two decisions shape everything else:
+
+1. How many delegate keys you issue.
+2. Which namespaces the agents read and write.
+
+## Give each agent its own delegate key
+
+An account has one owner and any number of delegates. A delegate is a keypair the owner authorizes onchain, and the relayer verifies that authorization against the contract on every request.
+
+Issue one delegate key per agent rather than sharing a single key. The keys grant identical access, so this is not an access-control boundary, but issuing one per agent gives you two things:
+
+1. The dashboard labels each key, so you can see which agent is running.
+2. You can revoke one agent without disturbing the others.
+
+<Warning>
+Any delegate the account authorizes can decrypt everything that account owns, across every namespace. Delegate keys separate identity, not data. If two agents must not read each other's memory, give them separate accounts, not separate namespaces.
+</Warning>
+
+## Decide what the agents share
+
+Namespaces decide what agents see, because recall and restore match a namespace exactly and never cross one.
+
+| **Pattern** | **Setup** | **Result** |
+| --- | --- | --- |
+| Shared memory | Every agent uses the same namespace | Each agent recalls what the others wrote |
+| Isolated memory | Each agent uses its own namespace | No agent recalls another's writes |
+| Shared plus private | A common namespace, plus one per agent | Agents pool findings and keep working notes apart |
+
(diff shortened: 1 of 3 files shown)


jessiemongeon1:
## Style Guide Audit

docs/guides/multi-agent-coordination.md
```diff
@@ -0,0 +1,126 @@
...
+  share. Agents that read and write the same namespace share memory, because any delegate the
+  account authorizes can decrypt everything that account owns. Agents that use separate namespaces
+  never read each other's writes, because recall never crosses a namespace. A handoff is one agent writing to a
+  shared namespace and the next recalling from it.
+---
+
+Several agents can work against one Walrus Memory account. Two decisions shape everything else:
+
+1. How many delegate keys you issue.
+2. Which namespaces the agents read and write.
```
jessiemongeon1: this reads odd and not natural, should revise

docs/guides/multi-agent-coordination.md
```diff
@@ -0,0 +1,126 @@
...
+
+| **Pattern** | **Setup** | **Result** |
+| --- | --- | --- |
+| Shared memory | Every agent uses the same namespace | Each agent recalls what the others wrote |
+| Isolated memory | Each agent uses its own namespace | No agent recalls another's writes |
+| Shared plus private | A common namespace, plus one per agent | Agents pool findings and keep working notes apart |
+
+Shared plus private suits most multi-agent systems. A research agent writes conclusions to `findings` and keeps its own scratch work in `research-agent`, so a planner agent reading `findings` sees the conclusions without wading through intermediate steps.
+
+```ts
```
jessiemongeon1, line 69: what is the source of this code?

docs/guides/multi-agent-coordination.md
```diff
@@ -0,0 +1,126 @@
...
+});
+```
+
+Both clients use the same delegate key and the same account. Only the namespace differs.
+
+## Hand off work between agents
+
+A handoff is a write followed by a recall. The first agent writes what the next one needs into a shared namespace, and the next agent recalls it when it starts.
+
+```ts
```
jessiemongeon1, line 91: what is the source of this code?

docs/guides/multi-agent-coordination.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Choose a boundary
+
+Match the boundary to what you actually need to separate:
+
+- Agents cooperating on one user's work: one account, one delegate key each, namespaces to divide the work.
+- Agents serving different users: one account per user. Never separate users by namespace, because any delegate reads every namespace on the account.
+- Agents running against different deployments: separate app IDs already isolate them, because the package ID scopes encryption and blob discovery.
+
+## References
```
jessiemongeon1: no references section - this gets auto generated

jessiemongeon1 (changes requested):

PR #649, opened 2026-08-14
chore(release): bump MCP package to 0.0.8 & update changelog

ducnmm:
## Summary

- bump `@mysten-incubation/memwal-mcp` from `0.0.7` to `0.0.8`
- add human-readable tool titles and explicit behavior annotations to the npm client's pre-login `tools/list`
- include the remote relayer annotation correction: recall/analyze disclose their possible stale-index cleanup side effect
- synchronize Claude Code, Cursor, Codex, and Antigravity plugin manifests
- update package/docs changelogs and release validation expectations

## Annotation safety

- `memwal_recall`: non-read-only and potentially destructive because a confirmed Walrus 404 triggers stale vector-row cleanup
- `memwal_analyze`: potentially destructive because pre-extraction context lookup uses the same cleanup path
- `memwal_remember`, `memwal_restore`, and `memwal_login`: non-read-only, non-destructive
- remote `memwal_health`: read-only, non-destructive

## Validation

- `node scripts/verify-manual-sdk-release.mjs`
- `pnpm --filter @mysten-incubation/memwal-mcp typecheck`
- `pnpm --filter @mysten-incubation/memwal-mcp test` (10/10)
- remote `tools/list` annotation tests (4/4 focused tests)
- `git diff --check`

## Rollout

The npm release and relayer deployment are separate surfaces. After promotion, verify both pre-login npm discovery and authenticated production remote `tools/list` before connector submission.

PR #650, opened 2026-08-14
fix(mcp): disclose recall cleanup side effects

ducnmm:
## Summary

- mark `memwal_recall` non-read-only and destructive because its hydration path deletes stale vector rows after a confirmed Walrus 404
- mark `memwal_analyze` destructive because pre-extraction context lookup uses the same cleanup path
- keep the exact `tools/list` contract test aligned with the real side effects

## Why

MCP annotations describe possible behavior, not only the primary intent of a tool. Although recall normally reads and analyze normally adds memories, both may invoke `WalrusSealEngine::cleanup_expired_blob` and delete stale index rows. Advertising recall as read-only or analyze as non-destructive is therefore unsafe for client confirmation policy and blocks the 0.0.8 release claim in #649.

## Validation

- `node --test --import tsx mcp/__tests__/tool-annotations.test.ts mcp/__tests__/restore.test.ts` (4/4)
- `git diff --check`

## Dependency

Merge this correction before #649 or any production claim that the annotations are Claude-review ready.

### DIFF
--- a/services/server/scripts/mcp/__tests__/tool-annotations.test.ts
+++ b/services/server/scripts/mcp/__tests__/tool-annotations.test.ts
@@ -34,7 +34,7 @@ test("tools/list publishes safe titles and behavior annotations for every remote
             },
             memwal_analyze: {
                 title: "Analyze and Remember",
-                annotations: { readOnlyHint: false, destructiveHint: false },
+                annotations: { readOnlyHint: false, destructiveHint: true },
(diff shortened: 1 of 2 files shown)


ducnmm:
Superseded by #649: the exact corrective commit was cherry-picked as `c1c6d805`, with package-side metadata and release fixes added in `2afc49d1`.

Issue #607, opened 2026-08-12
[Python SDK/Bug] LangChain sync wrapper skips memory recall inside running event loops


bshuaibu1:
# Python SDK/Bug: LangChain sync wrapper silently skips memory recall inside running event loops

## Summary

`with_memwal_langchain()` advertises recall-before-call behavior, but its sync `_generate()` wrapper silently disables recall when called while an event loop is already running.

That can happen in notebooks and async application hosts. The LLM call still succeeds, but the memory context is omitted without a warning.

## Affected code

Current `packages/python-sdk-memwal/memwal/middleware.py`:

```py
if loop is not None and loop.is_running():
    # Already in async context -- cannot use asyncio.run
    enriched.append(msg_list)
else:
    enriched.append(asyncio.run(_inject_memories(msg_list)))
```

The same file already has `_run_blocking()`, which handles running-loop sync calls by executing the coroutine in a worker thread. The sync OpenAI wrapper uses that helper, but the sync LangChain wrapper does not.

## Reproduction

No live MemWal credentials or relayer are needed. This stubs the tiny LangChain module surface imported by the wrapper and uses a fake MemWal recall response.

```py
import asyncio
import sys
import types
from unittest.mock import MagicMock

messages_mod = types.ModuleType("langchain_core.messages")
outputs_mod = types.ModuleType("langchain_core.outputs")
langchain_mod = types.ModuleType("langchain_core")

class HumanMessage:
    type = "human"
    def __init__(self, content): self.content = content

class SystemMessage:
    type = "system"
    def __init__(self, content): self.content = content

class AIMessage:
    type = "ai"
    def __init__(self, content): self.content = content

class ChatGeneration:
    def __init__(self, message): self.message = message

class ChatResult:
    def __init__(self, generations): self.generations = generations

messages_mod.HumanMessage = HumanMessage
messages_mod.SystemMessage = SystemMessage
messages_mod.BaseMessage = object
outputs_mod.ChatGeneration = ChatGeneration
outputs_mod.ChatResult = ChatResult
sys.modules["langchain_core"] = langchain_mod
sys.modules["langchain_core.messages"] = messages_mod
sys.modules["langchain_core.outputs"] = outputs_mod

from memwal.middleware import with_memwal_langchain
from memwal.types import RecallMemory, RecallResult
import memwal.middleware as mw

class FakeMemWal:
    def __init__(self):
        self.recall_calls = 0

    async def recall(self, *args, **kwargs):
[...]

PR #652, opened 2026-08-15
fix(python-sdk): close the per-call httpx client instead of dropping it

### LINKED ISSUE #606: [Python SDK/Bug] MemWalSync._run drops cached httpx client without closing it
## Summary

`MemWalSync._run()` drops the cached `httpx.AsyncClient` by assigning `self._inner._client = None` before every sync call, but it does not close the old client first. Any previously-created async client becomes unreachable while still open.

This affects repeated Python sync SDK usage, especially scripts and notebooks. The code comment says the reset is intentional so each `asyncio.run()` path recreates the client inside the loop that will use it, but the current implementation skips the lifecycle cleanup step.

## Affected code

Current `packages/python-sdk-memwal/memwal/client.py`:

```py
def _run(self, coro: Any) -> Any:
    import asyncio

    try:
        loop = asyncio.get_running_loop()
    except RuntimeError:
        loop = None

    # Reset the httpx client before every asyncio.run() path so it is
    # recreated inside the loop that will use it. This matters in
    # notebooks/Jupyter where the sync wrapper runs coroutines in worker
    # threads with short-lived event loops.
    self._inner._client = None

    if loop is not None and loop.is_running():
        import concurrent.futures

        with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:
            return pool.submit(asyncio.run, coro).result()
    else:
        return asyncio.run(coro)
```

## Reproduction

This repro does not require live MemWal credentials or a relayer. It uses a fake client object with an `aclose()` method to show that `_run()` discards the existing client without closing it.

```py
from memwal.client import MemWalSync

class FakeClient:
    is_closed = False

    def __init__(self):
        self.closed = False

    async def aclose(self):
        self.closed = True
        self.is_closed = True

class Inner:
    def __init__(self):
        self._client = FakeClient()

async def noop():
    return "ok"

inner = Inner()
old_client = inner._client
sync = MemWalSync(inner)  # type: ignore[arg-type]

print("before", old_client.is_closed, old_client.closed)
print("result", sync._run(noop()))
print("inner_client_after", inner._client)
print("old_client_after", old_client.is_closed, old_client.closed)
```

Observed output:

```text
before False False
result ok
inner_client_after None
old_client_after False False
```

## Expected behavior

Before replacing `_inner._client`, the sync wrapper should close the existing client if it exists and is not already closed, or otherwise avoid creating a client that has to cross event-loop boundaries.

Expected repro output would end with:

```text
old_client_after True True
```

## Why this matters

In normal use, `_inner._client` is an `httpx.AsyncClient`. Repeated sync SDK calls can leave abandoned clients and their underlying connection pools open until garbage collection. In notebooks and long-running scripts, this can produce resource warnings, stale sockets, and unnecessary connection churn.

## Suggested fix

Close the old client before resetting it, while preserving the current goal of not reusing an `httpx.AsyncClient` across event loops. One conservative shape:

```py
old_client = self._inner._client
if (
    old_client is not None
    and hasattr(old_client, "aclose")
    and not getattr(old_client, "is_closed", False)
):
    if loop is not None and loop.is_running():
        import concurrent.futures
        with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:
            pool.submit(asyncio.run, old_client.aclose()).result()
    else:
        asyncio.run(old_client.aclose())
self._inner._client = None
```

Then run the requested coroutine in the existing branch. This keeps the close operation inside an event loop that owns only the close coroutine, rather than trying to await it from the caller's already-running loop. A regression test can use the fake client above to assert `aclose()` is called before `_client` is cleared.

HoangDucBach:
## Ticket

Fixes #606

## What changed?

- New `_with_fresh_http_client()` in `client.py`: closes whatever is cached on entry, and closes the client this call created in `finally`, while the loop owning it is still alive.
- `_discard_http_client()` clears `_client` even when a cross-loop `aclose()` raises, so the next request can always build one in a live loop.
- Both drop sites route through it: `MemWalSync._run()` and `_wrap_sync_openai._run_memwal()`.
- `_SyncRunInner` test stand-in now mirrors `MemWal`'s real `close()` / `aclose()` contract.

## Why is this needed?

Both sync paths cleared `_client` before every `asyncio.run()` so a client is rebuilt inside the loop that uses it, but never closed the one they replaced. Every sync call therefore orphaned an open `httpx.AsyncClient` and its pool until GC — resource warnings, stale sockets and connection churn in notebooks and long-running scripts.

Worth flagging for review: closing on the way **out** is what removes the leak. Closing only on entry, as the issue suggests, still leaves an open client idle between every pair of calls.

## Scope

httpx client lifecycle on the sync entry points. Async paths are untouched — they already own their client for the caller's loop.

### Out of scope

Sync LangChain wrapper skipping recall (#607), branched off this PR.

## How was this tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] End-to-end tests
- [ ] Manual testing
- [ ] Not applicable

3 new tests in `TestMemWalSyncRun`: replaced client closed, per-call client closed, and both when the coroutine raises. Each verified to fail against the pre-fix `_run()`. Suite: 119 passed, ruff clean.

```bash
cd packages/python-sdk-memwal
python3 -m pytest tests/ -q -m "not integration"    # 119 passed
python3 -m ruff check memwal/ tests/test_client.py  # clean
```

To watch the tests catch the bug, restore the pre-fix body of `MemWalSync._run()` (`self._inner._client = None; wrapped = coro`) and rerun with `-k SyncRun` → 3 failures.

## Risks and dependencies

- **#607 branches off this one and must merge after it.**
- `_client` is now always `None` between sync calls. Nothing reads it across calls; `_http` rebuilds lazily.
- `aclose()` failures are swallowed by design — a client from a finished loop cannot always be closed, and cleanup must not mask the coroutine's result. Logged at debug.

## Author checklist

- [ ] CI is green.

### DIFF
--- a/packages/python-sdk-memwal/memwal/client.py
+++ b/packages/python-sdk-memwal/memwal/client.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 import asyncio
 import base64
 import json
+import logging
 import random
 import time
 import uuid
@@ -88,6 +89,9 @@ AUTH_REJECTED_MESSAGE = (
     "and dashboard credentials. Full troubleshooting: "
     "https://docs.wal.app/walrus-memory/troubleshooting/overview#401-auth_rejected-errors"
 )
+
+
+logger = logging.getLogger("memwal")
 
 
 # ============================================================
@@ -1277,6 +1281,38 @@ class MemWalRememberJobTimeout(MemWalError):
         self.timeout_ms = timeout_ms
 
 
+async def _discard_http_client(memwal: MemWal) -> None:
+    """Close the cached httpx client, and drop it either way.
+
+    ``close()`` can fail when the client belongs to an event loop that has
+    already finished — the caller still needs ``_client`` cleared so the next
+    request builds one in a loop that is actually running.
+    """
+    try:
+        await memwal.close()
+    except Exception:
+        logger.debug("Closing the cached HTTP client failed", exc_info=True)
+    finally:
+        memwal._client = None
+
+
+async def _with_fresh_http_client(memwal: MemWal, coro: Any) -> Any:
+    """Run ``coro`` with an httpx client owned by the *current* event loop.
+
+    The sync entry points execute each coroutine in a throwaway ``asyncio.run()``
+    loop, and an ``httpx.AsyncClient`` is bound to the loop that created it, so
+    one can never be reused across calls. Both ends are closed rather than just
+    dropped (GH #606): on the way in for anything an earlier loop left behind —
+    e.g. a caller mixing ``await memwal.recall()`` with the sync wrapper — and in
+    ``finally`` for the client this call created, while its loop is still alive.
+    """
+    await _discard_http_client(memwal)
+    try:
+        return await coro
+    finally:
+        await _discard_http_client(memwal)
+
+
 class MemWalSync:
     """Synchronous wrapper around the async :class:`MemWal` client.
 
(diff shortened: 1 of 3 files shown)


ducnmm (approved):
## Review

**Verdict: ✅ Approve**

Core fix is correct: `_with_fresh_http_client()` closes on both entry (orphan from prior loop) and exit (client created this call) — fully addresses the leak GH #606.

- `_discard_http_client()` with try/finally is defensive and correct
- 3 tests covering replaced client, per-call client, and exception path are thorough
- `_FakeHttpClient` and `_SyncRunInner.close()` mirroring the real contract is clean test design
- No blocking issues. Ready to merge.

PR #661, opened 2026-08-17
ci(python-sdk): build, unit test and e2e workflow against the dev relayer

HoangDucBach:
## Ticket

CI for the Python SDK: build from source, unit test, e2e against the dev relayer.

### Workflow

New `.github/workflows/test-python-sdk.yml`, named to sit alongside `release-python-sdk.yml`:

- **`unit`** — installs from source, runs `ruff`, then the mocked suite across Python 3.9–3.12. Every pull request touching the package.
- **`e2e`** — runs `tests/test_integration.py` against the dev relayer on pushes to `dev`, manual runs, and a weekly cron. Reuses the `benchmark-dev` environment, so no new secrets are needed, and fails if its credentials are missing rather than silently skipping every authenticated test.

### SDK — accept a Sui bech32 private key

> The only change outside CI and tests. Reviewable on its own; happy to split it into a separate PR if you would rather.

`memwal/utils.py` gains `bech32_decode` and `decode_sui_private_key`; `MemWal.__init__` normalises whatever it is handed into hex, so nothing downstream changes.

| | Accepts |
| --- | --- |
| TypeScript SDK | `suiprivkey1...` or hex, via `decodeSuiPrivateKey` |
| Python SDK, before | hex only — `bytes.fromhex()` |

Why it is in this PR: pointing the job at `BENCH_DELEGATE_KEY` failed all sixteen authenticated tests with `non-hexadecimal number found in fromhex() arg at position 0`. That secret is bech32 because its existing consumer, `bench-recall-latency.ts`, is TypeScript. E2E cannot pass without this.

It is worth fixing regardless of CI: the dashboard issues keys in the bech32 form, so the same credential worked in one SDK and not the other, and the error named nothing useful. `encode_sui_private_key` already existed with no inverse — the direction was simply missing. Verified against `@mysten/sui` on five generated keypairs: decode, encode, normalise and the derived Sui address all match.

### Tests

- Every authenticated test now runs in a per-run `sdk-e2e-<uuid>` namespace; `_sync_client` / `_async_client` replace sixteen identical constructions.
- The live `remember_and_wait` timeout goes from the SDK default of 60s to 120s (see Risks).
- Five cases cover the key formats, and the three pre-existing `ruff` offences in `tests/test_signing.py` are fixed — the new lint step would fail on arrival otherwise.

The reformatting in both test files is `ruff`, not a style pass. `line-length` is 100, and the extra `timeout_ms` argument pushes five calls to 106–130 characters; the sixth still fits and stays on one line.

## Why is this needed?

The Python SDK had no CI. `test.yml` only runs the relayer's own e2e script and `release-python-sdk.yml` just publishes, so nothing built the package or ran either suite on a push — 121 mocked and 23 integration tests sat unexecuted.

## What e2e covers

Unauthenticated, runs without any credential:

| Group | Cases |
| --- | --- |
| Health | `/health` returns ok; exposes a version string |
| Auth rejection | unsigned request; signature from a different key; expired timestamp; future timestamp; unregistered key surfaced as `MemWalError` |

Authenticated, needs the environment secrets:

| Group | Cases |
| --- | --- |
| Remember | job id and status; `remember_and_wait` returns blob and owner; default namespace; custom namespace |
| Recall | returns a list; respects `limit`; result field shape |
| Analyze | fact extraction; owner address |
| Ask | non-empty answer; `memories_used` |
| Full flow | remember then recall a unique marker in a per-run namespace; remember then ask |
| Async | health, remember, recall, analyze, ask via the async client |

The auth-rejection group builds its requests with `httpx` rather than through the SDK, so it checks the relayer's own contract: timestamp window and signature verification.

## Scope

The workflow. Both suites already exist and `tests/test_integration.py` already reads `MEMWAL_SERVER_URL` / `MEMWAL_PRIVATE_KEY` / `MEMWAL_ACCOUNT_ID` and self-skips its authenticated cases — this wires that up rather than adding tests.

### Out of scope

Staging and production environments.

## How was this tested?

- [x] Unit tests
- [x] Integration tests
- [ ] End-to-end tests
- [ ] Manual testing
- [ ] Not applicable

Ran what the `unit` job runs: `ruff` clean, 116 passed. Workflow validated with `actionlint`. All four matrix versions pass on this PR, and `e2e` correctly skips on pull requests.

Ran the full integration suite against `https://relayer.dev.memwal.ai` with a real delegate key: **23 passed**. An earlier run lost six tests to `MemWalRememberJobTimeout`, which is what the timeout change addresses.

Also ran `workflow_dispatch` on this branch end to end ([run 32098755482](https://github.com/MystenLabs/MemWal/actions/runs/32098755482)): **23 passed in 5m46s**, summary reports `Authenticated: true`, so the `BENCH_DELEGATE_KEY` → `MEMWAL_PRIVATE_KEY` mapping is confirmed against the real environment. The decode path is cross-checked against `@mysten/sui` on five generated keypairs — decode, encode, normalise and derived Sui address all match.

```bash
cd packages/python-sdk-memwal
python -m pip install -e '.[dev]'
python -m ruff check memwal/ tests/
python -m pytest tests/ -m "not integration" -q

docker run --rm -v "$PWD/../..":/repo -w /repo rhysd/actionlint:latest \
  .github/workflows/test-python-sdk.yml
```

## Risks and dependencies

- **Reuses `benchmark-dev` rather than adding secrets.** `BENCH_DELEGATE_KEY` is the delegate private key the SDK signs with, which `test_integration.py` reads as `MEMWAL_PRIVATE_KEY`; `BENCH_ACCOUNT_ID` maps to `MEMWAL_ACCOUNT_ID`. Sharing the account is safe because the namespaces do not overlap. `BENCH_SERVER_URL` falls back to `https://relayer.dev.memwal.ai` so a missing variable cannot point a run at the wrong relayer. A missing credential fails the job rather than warning, so the run cannot go green having skipped every authenticated test.
- **A live write takes around 44s** on the dev relayer — embed, SEAL encrypt, Walrus upload, on-chain metadata. The SDK's 60s `remember_and_wait` default left ~16s of headroom, and runs failed on timing alone. The tests now pass 120s, matching the SDK's own bulk default, overridable via `MEMWAL_REMEMBER_TIMEOUT_MS`. The 30 min job timeout covers a full suite comfortably.
- **E2E is skipped on pull requests** — environment secrets are withheld from fork PRs, and each authenticated run writes real memories, costing embedding, Walrus storage and gas.
- **Every authenticated test writes into a `sdk-e2e-<uuid>` namespace unique to the run**, so a recurring job leaves nothing in `default` or `sdk-test` on the shared bench account.
- **The dev relayer's Node sidecar occasionally drops a request** (`Sidecar seal/encrypt request failed`), which surfaces as one failed write. It recovered on a rerun. No retry is wired in: a genuine relayer regression should stay visible.

## Author checklist

- [x] CI is green.

### DIFF
--- a/.github/workflows/test-python-sdk.yml
+++ b/.github/workflows/test-python-sdk.yml
@@ -106,15 +106,24 @@ jobs:
 
       - name: Check credentials
         id: config
+        shell: bash
         run: |
-          # `requires_key` self-skips, so a half-configured environment still
-          # gets the unauthenticated coverage instead of nothing.
-          if [ -n "${MEMWAL_PRIVATE_KEY:-}" ] && [ -n "${MEMWAL_ACCOUNT_ID:-}" ]; then
-            echo "authenticated=true" >> "$GITHUB_OUTPUT"
-          else
-            echo "authenticated=false" >> "$GITHUB_OUTPUT"
-            echo "::warning::BENCH_DELEGATE_KEY / BENCH_ACCOUNT_ID are not set on this environment — authenticated e2e tests will be skipped."
+          set -euo pipefail
+
+          # Without these the `requires_key` marker skips every authenticated
+          # test, and the job would report success having never called the
+          # relayer. Fail instead of reporting a green run that proved nothing.
+          missing=0
+          for name in MEMWAL_PRIVATE_KEY MEMWAL_ACCOUNT_ID; do
+            if [ -z "${!name:-}" ]; then
+              echo "::error::${name} is empty — set BENCH_DELEGATE_KEY / BENCH_ACCOUNT_ID on the benchmark-dev environment."
+              missing=1
+            fi
+          done
+          if [ "$missing" -ne 0 ]; then
+            exit 1
           fi
+          echo "authenticated=true" >> "$GITHUB_OUTPUT"
 
       - name: E2E tests
         id: e2e
--- a/.github/workflows/test-python-sdk.yml
+++ b/.github/workflows/test-python-sdk.yml
@@ -1,0 +1,145 @@
+name: Test Python SDK
+
+on:
+  pull_request:
+    paths:
+      - 'packages/python-sdk-memwal/**'
+      - '.github/workflows/test-python-sdk.yml'
+  push:
+    branches:
+      - main
+      - staging
+      - dev
+    paths:
+      - 'packages/python-sdk-memwal/**'
+      - '.github/workflows/test-python-sdk.yml'
+  workflow_dispatch:
+  schedule:
+    # Catches relayer drift with no code change to trigger it.
+    - cron: '17 9 * * 1'
+
+concurrency:
+  group: test-python-sdk-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    working-directory: packages/python-sdk-memwal
+
+jobs:
+  unit:
+    name: Unit / Python ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # Tracks requires-python and the pyproject classifiers.
+        python-version: ['3.9', '3.10', '3.11', '3.12']
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Build and install from source
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e '.[dev]'
+
+      - name: Lint
+        run: python -m ruff check memwal/ tests/
+
+      - name: Unit tests
+        run: python -m pytest tests/ -m "not integration" -q
+
+  e2e:
+    name: E2E / dev relayer
+    runs-on: ubuntu-latest
+    needs: unit
+    timeout-minutes: 30
+
+    # Dev only for now; staging and production follow once their credentials
+    # exist. Pull requests are excluded because environment secrets are
+    # withheld from fork PRs, and every authenticated run writes real memories.
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event_name == 'schedule' ||
+      (github.event_name == 'push' && github.ref_name == 'dev')
+
+    # Reuses the credentials benchmark-live.yml already relies on, so this
+    # needs no new secrets. Safe to share: the benchmark writes to the
+    # `benchmark` namespace while these tests use `sdk-test`, `default` and a
+    # per-run `sdk-e2e-<uuid>`.
+    environment: benchmark-dev
+
+    env:
+      # BENCH_DELEGATE_KEY is the delegate private key the SDK signs with,
+      # which test_integration.py reads as MEMWAL_PRIVATE_KEY.
+      MEMWAL_PRIVATE_KEY: ${{ secrets.BENCH_DELEGATE_KEY }}
+      MEMWAL_ACCOUNT_ID: ${{ secrets.BENCH_ACCOUNT_ID }}
+      # Public URL (docs/relayer/benchmark-ci-setup.md), defaulted so a missing
+      # variable cannot silently point the run at the wrong relayer.
+      MEMWAL_SERVER_URL: ${{ vars.BENCH_SERVER_URL || 'https://relayer.dev.memwal.ai' }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Build and install from source
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e '.[dev]'
+
+      - name: Check credentials
+        id: config
+        run: |
+          # `requires_key` self-skips, so a half-configured environment still
+          # gets the unauthenticated coverage instead of nothing.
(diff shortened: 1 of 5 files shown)


ducnmm (changes requested):
@HoangDucBach thanks for wiring this up — the workflow shape is right (unit on PRs, live e2e only on `dev` / dispatch / cron, reuse `benchmark-dev`). A few changes before merge.

## Requested changes

1. **Fail the e2e job when credentials are missing.**  
   The current `Check credentials` step only warns, then `requires_key` skips every authenticated test. A push to `dev`, the weekly cron, or a manual dispatch can go green without ever calling `remember` / `recall`. Self-skip is fine for local runs; for the job named `E2E / dev relayer`, missing `BENCH_DELEGATE_KEY` or `BENCH_ACCOUNT_ID` should fail the job.

2. **Do not write into the shared `default` namespace.**  
   `test_remember_default_namespace` and `test_async_recall` persist into `default` on the shared bench account. Every `dev` push plus the weekly cron leaves real Walrus blobs / gas behind, and `TestRecall` / `TestAsk` also query `default`. Asserting the default namespace does not require writing there — isolate writes to a per-run `sdk-e2e-<uuid>` (as the full-flow test already does). Same preference for `sdk-test` if this stays on a recurring schedule.

3. **Rebase onto current `dev`.**  
   The branch is ~25 commits behind. The PR checklist says it is up to date; it is not. Conflict risk looks low, but please rebase before merge.

Please also run **workflow_dispatch** on this branch once and confirm the summary shows `Authenticated: true`. E2E is correctly skipped on the PR, so CI here has not actually proven the secret mapping.

## Coverage — this is enough for WALM-339

The live suite does not need to cover the full Python SDK surface. Core happy-path is the right bar:

- `health`
- auth rejection (unsigned / bad signature / timestamp window)
- `remember` / `remember_and_wait`
- `recall` / `analyze` / `ask`
- a few async counterparts

Out of scope for this ticket (unit/mock coverage is fine): bulk remember, `restore`, `embed`, `analyze_and_wait`, `remember_manual` / `recall_manual`, `compatibility`, and the OpenAI/LangChain middleware.

Not blocking: the 120s `remember_and_wait` timeout is justified (live writes ~44s). The `test_signing.py` ruff wraps are required now that lint covers `tests/`. Skipping e2e on pull_request is correct.

HoangDucBach:
> @HoangDucBach thanks for wiring this up — the workflow shape is right (unit on PRs, live e2e only on `dev` / dispatch / cron, reuse `benchmark-dev`). A few changes before merge.
> 
> ## Requested changes
> 1. **Fail the e2e job when credentials are missing.**
>    The current `Check credentials` step only warns, then `requires_key` skips every authenticated test. A push to `dev`, the weekly cron, or a manual dispatch can go green without ever calling `remember` / `recall`. Self-skip is fine for local runs; for the job named `E2E / dev relayer`, missing `BENCH_DELEGATE_KEY` or `BENCH_ACCOUNT_ID` should fail the job.
> 2. **Do not write into the shared `default` namespace.**
>    `test_remember_default_namespace` and `test_async_recall` persist into `default` on the shared bench account. Every `dev` push plus the weekly cron leaves real Walrus blobs / gas behind, and `TestRecall` / `TestAsk` also query `default`. Asserting the default namespace does not require writing there — isolate writes to a per-run `sdk-e2e-<uuid>` (as the full-flow test already does). Same preference for `sdk-test` if this stays on a recurring schedule.
> 3. **Rebase onto current `dev`.**
>    The branch is ~25 commits behind. The PR checklist says it is up to date; it is not. Conflict risk looks low, but please rebase before merge.
> 
> Please also run **workflow_dispatch** on this branch once and confirm the summary shows `Authenticated: true`. E2E is correctly skipped on the PR, so CI here has not actually proven the secret mapping.
> 
> ## Coverage — this is enough for WALM-339
> The live suite does not need to cover the full Python SDK surface. Core happy-path is the right bar:
> 
> * `health`
> * auth rejection (unsigned / bad signature / timestamp window)
> * `remember` / `remember_and_wait`
> * `recall` / `analyze` / `ask`
> * a few async counterparts
> 
> Out of scope for this ticket (unit/mock coverage is fine): bulk remember, `restore`, `embed`, `analyze_and_wait`, `remember_manual` / `recall_manual`, `compatibility`, and the OpenAI/LangChain middleware.
> 
> Not blocking: the 120s `remember_and_wait` timeout is justified (live writes ~44s). The `test_signing.py` ruff wraps are required now that lint covers `tests/`. Skipping e2e on pull_request is correct.

All four done.

1. Missing credentials now fail the job instead of warning.
2. Every authenticated test runs in a per-run `sdk-e2e-<uuid>`; nothing writes to `default` or `sdk-test`. It was three writes, not two — `test_async_remember` as well.
3. Rebased onto `0458df7d`, no conflicts. It was 33 commits behind.
4. [run 32098755482](https://github.com/MystenLabs/MemWal/actions/runs/32098755482): 23 passed, summary shows `Authenticated: true`.

The first dispatch failed all 16 authenticated tests in 7s: `BENCH_DELEGATE_KEY` is `suiprivkey1...` and the Python SDK only accepted hex, while the TypeScript SDK decodes both. Fixed in `2a4cdd2b`, cross-checked against `@mysten/sui`. That commit is the only change outside CI and tests — happy to split it out.

Two smaller things the live runs surfaced: the dev relayer leaves a job as `running` while its record already holds a 409 error, and `remember_and_wait`'s 60s default is too tight for ~44s writes.

ducnmm (approved):
Looks great — all previous review points have been addressed cleanly:

1. Credentials check is now a hard failure on missing secrets rather than a silent skip.
2. E2E writes are properly isolated into per-run `sdk-e2e-<uuid>` namespaces.
3. Clean rebase on current `dev`.
4. Bech32 `suiprivkey1...` private key support (`decode_sui_private_key`, `normalize_private_key`) is a great addition aligning the Python SDK with TypeScript SDK and dashboard credentials.
5. All unit matrix (Python 3.9–3.12) and lint checks are passing.

Approve.

PR #664, opened 2026-08-17
Promote dev to staging

ducnmm:
## Summary

Promote the latest `dev` changes to `staging`.

Includes #663:
- Enoki-sponsored durable Walrus registration.
- Balance monitoring background worker: alerts when WAL or SUI balances drop below configured thresholds before writes fail.
- Slack alert delivery integration for low-balance notifications.

## Validation

- PR #663 merged and verified.
- Sidecar and relayer verified healthy on `dev`.
- Balance monitor verified initializing with configured thresholds.

### DIFF
--- a/apps/app/src/components/AdminConfig.tsx
+++ b/apps/app/src/components/AdminConfig.tsx
@@ -68,6 +68,13 @@ export function AdminConfig({ adminKey, onInvalidKey }: AdminConfigProps) {
         </div>
 
         <div className="admin-config-item">
+          <span className="admin-config-label">Uploader SUI Low Threshold</span>
+          <code className="admin-config-value" title={`${data.uploaderSuiLowThresholdMist} mist`}>
+            {formatTokenAmount(data.uploaderSuiLowThresholdMist)} SUI
+          </code>
+        </div>
+
+        <div className="admin-config-item">
           <span className="admin-config-label">Sponsor SUI Low Threshold</span>
           <code className="admin-config-value" title={`${data.sponsorSuiLowThresholdMist} mist`}>
             {formatTokenAmount(data.sponsorSuiLowThresholdMist)} SUI
--- a/apps/app/src/utils/admin-api.ts
+++ b/apps/app/src/utils/admin-api.ts
@@ -25,6 +25,7 @@ export interface UploadError {
 export interface AdminConfig {
   balanceMonitorIntervalSecs: number
   uploaderWalLowThresholdFrost: bigint
+  uploaderSuiLowThresholdMist: bigint
   sponsorSuiLowThresholdMist: bigint
 }
 
@@ -115,6 +116,7 @@ interface RawWalletsResponse {
   uploader_pool: {
     wallets: RawWalletBalance[]
     wal_threshold: string
+    sui_threshold: string
     last_updated: string
   }
   sponsor_wallet: {
@@ -145,6 +147,7 @@ interface RawUploadErrorsResponse {
 interface RawConfigResponse {
   balance_monitor_interval_secs: number
   wallet_wal_low_threshold_frost: string
+  wallet_sui_low_threshold_mist: string
   sponsor_sui_low_threshold_mist: string
 }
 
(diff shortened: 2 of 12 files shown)


ducnmm:
Hi @harrymove-ctrl, could you please help review and approve this PR to promote the latest dev changes to staging? Thank you!

hien-p:
## Review: content is clean, safe to promote

`dev` is a strict content superset of `staging` — the two-way diff is exactly symmetric at 12 files / +350 / -65, so there's no drift and no back-merge needed. All CI checks and Railway deploy statuses are green. The only thing holding the merge is the required approving review.

Three commits promoted: `ddb0f8c1` (sponsor durable register), `2162d261` (WAL + SUI balance monitor), `af086ee7` (merge of #663).

### What's good

- Dedup key correctly widened to `(wallet_type:token, address)` so WAL and SUI alert independently, with the test updated to match.
- `assertSponsoredRegisterTransaction` checks both directions: sender must equal the wallet **and** the gas owner must not, on top of the shared WAL address-balance assertions.
- Enoki's returned bytes get a canonical base64 round-trip check *before* signing — good, we never sign bytes we didn't verify.
- Digest cross-check after Enoki execute (`executed.digest !== prepared.digest`).
- New test asserts the sponsored path never falls through to direct execution (`directExecutions === 0`).

### One thing I'd like a conscious yes on before merging

**`ENOKI_FALLBACK_TO_DIRECT_SIGN` is now ignored on the durable register path.** `durableRegisterDirectSigningAllowed` renames the param to `_fallbackToDirectSign` and returns `!enokiConfigured`. The fail-closed intent is clear from the comment (avoid silently draining the upload wallet after a sponsor outage), but the non-durable path in `walrus-upload.ts:218` still honours the flag. So if staging has `ENOKI_FALLBACK_TO_DIRECT_SIGN=true` set, durable registers will now hard-fail on an Enoki outage instead of falling back, while regular uploads still fall back.

Is that the intended behaviour for staging? If yes, worth a line in the PR body so it isn't a surprise during the next Enoki incident.

### Non-blocking notes

**Dead guard** — `walrus-upload-journal.ts:298` calls `durableRegisterDirectSigningAllowed(false, ENOKI_FALLBACK_TO_DIRECT_SIGN)` with the first arg hardcoded to `false`, so it always returns `true` and the `"durable register requires Enoki sponsorship"` throw is unreachable. Harmless, but it reads as a live safety check when it isn't.

**Epoch-expiry guard is skipped for sponsored transactions** — `maxEpoch` is only populated when the transaction carries a `ValidDuring` expiration. Enoki-sponsored transactions pay gas with real coins rather than the address balance, so they typically have none, which means both the `NoSideEffectError` and `UNAVAILABLE` branches of the "expired at epoch" check silently no-op. Not a correctness hole — idempotency is still protected by the up-front `getTransaction(digest)` lookup — but we lose the early abort on stale journals.

**Rollback asymmetry** — forward compatibility is fine (`#[serde(default)]`, no `deny_unknown_fields`, and pre-existing journals replay on the legacy path). But rolling *back* would break in-flight sponsored journals: the old sidecar drops `sponsorDigest` in `parsePreparedRegisterTransaction`, then `validatePreparedRegisterTransaction` rejects with `"registerTransaction gas owner does not match the journaled wallet"`. Those rows stay stuck until we roll forward again. Worth knowing, not worth holding the merge.

**Threshold defaults jumped ~50,000× (WAL) and 50× (sponsor SUI)** — from 0.001 WAL to 50 WAL, and 0.1 SUI to 5 SUI. The old defaults were effectively "never alert", so this is the right direction. Just note that if staging doesn't set these explicitly, the first monitor tick may fire alerts for every wallet sitting between the old and new thresholds. The 12h dedup keeps it to one message per wallet per token, so it won't storm.

**Minor coupling** — `main.rs:215` `continue`s on a `suiMist` parse failure *before* the WAL check, so a missing field would silently drop WAL alerting too. Not an issue today (the sidecar on `staging` already emits `suiMist` at `clients.ts:335`, so there's no rolling-deploy gap), but the ordering makes SUI parsing a hard dependency for an unrelated alert.


cc: @ducnmm

harrymove-ctrl (approved):

Issue #611, opened 2026-08-13
memwal_remember_bulk silently drops --namespace: batched facts land in default, unrecallable from work


Wadefak04:
## Summary

The MCP stdio package injects the configured default namespace (`--namespace` / `MEMWAL_NAMESPACE`) into four tools but omits `memwal_remember_bulk`. `NAMESPACE_TOOLS` in `packages/mcp/src/bridge.ts:42`:

```ts
const NAMESPACE_TOOLS = new Set([
    "memwal_remember",
    "memwal_recall",
    "memwal_analyze",
    "memwal_restore",
]);
```

`memwal_remember_bulk` accepts an optional `namespace` on the relayer side — `services/server/scripts/mcp/tools/remember-bulk.ts:14`:

```ts
namespace: z.string().optional().describe(
    "Optional namespace bucket applied to every fact. Defaults to the session's namespace when omitted."
),
```

Because the bulk call is forwarded without a namespace, the relayer files those facts under its own `"default"` namespace while `memwal_recall` — which *does* get the injection — searches the configured one. Facts save successfully, report `blob_id`s, and are then unrecallable.

## Why this is worse than a normal namespace mixup

The tool descriptions actively steer agents toward the broken path. `memwal_remember`: *"To save several facts at once, use memwal_remember_bulk instead."* `memwal_remember_bulk`: *"Prefer this over repeated memwal_remember calls."* So the recommended write path for multi-fact saves is exactly the one that drops the namespace, and it fails silently — `memwal_remember_bulk` returns `Saved 5/5 fact(s)` with real blob ids.

The failure is also asymmetric in a way that hides the cause: single-fact saves work, batched saves vanish. From the user's side "memory is flaky", not "namespace is wrong".

## Reproduction

Node v22.23.1, `@mysten-incubation/memwal-mcp@0.0.6`, commit `9b84c90`.

```
$ cd packages/mcp && npm run build
$ node -e "
import('./dist/bridge.js').then(({ applyDefaultNamespace }) => {
  const mk = (name, args) => ({ jsonrpc:'2.0', id:1, method:'tools/call', params:{ name, arguments:args } });
  for (const t of ['memwal_remember','memwal_recall','memwal_analyze','memwal_restore','memwal_remember_bulk']) {
    const out = applyDefaultNamespace(mk(t, t==='memwal_remember_bulk'?{facts:['a','b']}:{text:'x'}), 'work');
[...]

PR #667, opened 2026-08-17
fix(mcp): inject default namespace for memwal_remember_bulk calls

### LINKED ISSUE #625: memwal_remember_bulk bypasses --namespace injection, silently writing to the relayer default
### Summary

`dist/bridge.js` injects the configured default namespace only for tools listed in
`NAMESPACE_TOOLS`. `memwal_remember_bulk` is not in that set:

```js
/** Memory tools that take a `namespace` argument. `memwal_remember`,
 * `memwal_recall`, and `memwal_analyze` treat it as optional; `memwal_restore`
 * requires it ... */
const NAMESPACE_TOOLS = new Set([
    "memwal_remember",
    "memwal_recall",
    "memwal_analyze",
    "memwal_restore",
]);
```

The plugin shipped in this same package actively steers agents toward the excluded tool —
`plugin/scripts/on_user_prompt.mjs`:

> "Save it with memwal_remember (or **memwal_remember_bulk** for several distinct facts)"

### Why this matters

A user who pins a namespace in their client config reasonably expects every write to land there:

    "args": ["-y", "@mysten-incubation/memwal-mcp", "--namespace", "work"]

Single writes honour it. Bulk writes do not. When an agent calls `memwal_remember_bulk` without an
explicit `namespace`, no default is injected and the call is forwarded without one — so per the
README's own precedence rules, "the relayer applies its own `default` namespace."

The failure is silent and it is data misplacement, not data loss. The write succeeds, a blob_id comes
back, the agent reports success — and the memory is in the wrong namespace. It surfaces later as
recall returning nothing, by which time the blobs are immutable and unenumerable (see companion issue
re: listing blobs). Bulk writes are also, by definition, the calls that misplace the *most* records
at once.

### Two possibilities, both defects

1. `memwal_remember_bulk` accepts a `namespace` argument and was omitted from `NAMESPACE_TOOLS` —
   a one-line fix.
2. It accepts no `namespace` at all — in which case bulk writes cannot be routed and the tool is
   unusable for anyone using namespaces.

I could not call `tools/list` to determine which, because the relayer handshake is rate-limited (see
companion issue). The `NAMESPACE_TOOLS` comment says the set is "memory tools that take a `namespace`
argument", which implies (2).

### Expected

Bulk writes honour the configured default namespace exactly as single writes do.

### Actual

They fall through to the relayer default, silently.

ducnmm:
Resolves #625 (and companion #611)

### Summary
Includes `memwal_remember_bulk` in the set of `NAMESPACE_TOOLS` so the configured default namespace (`--namespace` / `MEMWAL_NAMESPACE`) is injected when an agent calls bulk remember without an explicit namespace.

### Problem
Previously, `NAMESPACE_TOOLS` in `packages/mcp/src/bridge.ts` only included `memwal_remember`, `memwal_recall`, `memwal_analyze`, and `memwal_restore`. When an agent invoked `memwal_remember_bulk` without an explicit `namespace`, no namespace argument was injected and the batch was written into the relayer's fallback `default` namespace, making it unrecallable from the configured project namespace.

### Changes
- Added `"memwal_remember_bulk"` to `NAMESPACE_TOOLS` in `packages/mcp/src/bridge.ts`.
- Added unit tests in `packages/mcp/test/default-namespace.test.mjs` to verify default namespace injection behavior for `memwal_remember_bulk` and other tools.

### Verification
- `pnpm --filter @mysten-incubation/memwal-mcp test` (14/14 tests pass)

### DIFF
--- a/packages/mcp/src/bridge.ts
+++ b/packages/mcp/src/bridge.ts
@@ -38,10 +38,10 @@ export interface BridgeConfig {
 }
 
 /** Memory tools that take a `namespace` argument. `memwal_remember`,
- * `memwal_recall`, and `memwal_analyze` treat it as optional; `memwal_restore`
- * requires it (its upstream schema still lists `namespace` as required, so
- * agents normally pass one — but a configured default is filled in if the
- * agent calls it without). */
+ * `memwal_remember_bulk`, `memwal_recall`, and `memwal_analyze` treat it as
+ * optional; `memwal_restore` requires it (its upstream schema still lists
+ * `namespace` as required, so agents normally pass one — but a configured
+ * default is filled in if the agent calls it without). */
 const NAMESPACE_TOOLS = new Set([
     "memwal_remember",
     "memwal_remember_bulk",
--- a/packages/mcp/test/default-namespace.test.mjs
+++ b/packages/mcp/test/default-namespace.test.mjs
@@ -19,6 +19,8 @@ test("applyDefaultNamespace injects configured namespace into memwal_remember_bu
     const updated = applyDefaultNamespace(msg, "project-alpha");
     assert.equal(updated.params.arguments.namespace, "project-alpha");
     assert.deepEqual(updated.params.arguments.facts, ["fact 1", "fact 2"]);
+    // The call site ignores the return value — in-place mutation is the real contract.
+    assert.equal(msg.params.arguments.namespace, "project-alpha");
 });
 
 test("applyDefaultNamespace respects explicit namespace on memwal_remember_bulk", () => {
@@ -85,8 +87,37 @@ test("applyDefaultNamespace does not touch unrelated tools or non-call RPC messa
         jsonrpc: "2.0",
         id: 5,
         method: "tools/list",
-        params: {},
+        params: { name: "memwal_remember_bulk", arguments: {} },
     };
     const updatedList = applyDefaultNamespace(listMsg, "test-ns");
-    assert.equal(updatedList.params.arguments, undefined);
+    assert.equal(updatedList.params.arguments.namespace, undefined);
+});
+
+test("applyDefaultNamespace is a no-op when no default is configured", () => {
+    const msg = {
+        jsonrpc: "2.0",
+        id: 6,
+        method: "tools/call",
+        params: { name: "memwal_remember_bulk", arguments: { facts: ["fact 1"] } },
+    };
+
+    applyDefaultNamespace(msg, undefined);
+    assert.equal(msg.params.arguments.namespace, undefined);
+});
+
+test("applyDefaultNamespace overrides a blank explicit namespace", () => {
+    for (const blank of ["", "   "]) {
+        const msg = {
+            jsonrpc: "2.0",
+            id: 7,
+            method: "tools/call",
+            params: {
+                name: "memwal_remember_bulk",
+                arguments: { facts: ["fact 1"], namespace: blank },
+            },
+        };
+
+        applyDefaultNamespace(msg, "project-alpha");
+        assert.equal(msg.params.arguments.namespace, "project-alpha");
+    }
 });
--- a/packages/mcp/test/default-namespace.test.mjs
+++ b/packages/mcp/test/default-namespace.test.mjs
@@ -1,0 +1,92 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { applyDefaultNamespace } from "../dist/bridge.js";
+
+test("applyDefaultNamespace injects configured namespace into memwal_remember_bulk", () => {
+    const msg = {
+        jsonrpc: "2.0",
+        id: 1,
+        method: "tools/call",
+        params: {
+            name: "memwal_remember_bulk",
+            arguments: {
+                facts: ["fact 1", "fact 2"],
+            },
+        },
+    };
+
+    const updated = applyDefaultNamespace(msg, "project-alpha");
+    assert.equal(updated.params.arguments.namespace, "project-alpha");
+    assert.deepEqual(updated.params.arguments.facts, ["fact 1", "fact 2"]);
+});
+
+test("applyDefaultNamespace respects explicit namespace on memwal_remember_bulk", () => {
+    const msg = {
+        jsonrpc: "2.0",
+        id: 2,
+        method: "tools/call",
+        params: {
+            name: "memwal_remember_bulk",
+            arguments: {
+                facts: ["fact 1"],
+                namespace: "explicit-scope",
+            },
+        },
+    };
+
+    const updated = applyDefaultNamespace(msg, "project-alpha");
+    assert.equal(updated.params.arguments.namespace, "explicit-scope");
+});
+
+test("applyDefaultNamespace injects into all namespace-aware tools", () => {
+    const tools = [
+        "memwal_remember",
+        "memwal_remember_bulk",
+        "memwal_recall",
+        "memwal_analyze",
+        "memwal_restore",
+    ];
+
+    for (const toolName of tools) {
+        const msg = {
+            jsonrpc: "2.0",
+            id: 3,
+            method: "tools/call",
+            params: {
+                name: toolName,
+                arguments: {},
+            },
+        };
+
+        const updated = applyDefaultNamespace(msg, "shared-namespace");
+        assert.equal(
+            updated.params.arguments.namespace,
+            "shared-namespace",
+            `expected default namespace to be injected for ${toolName}`
+        );
+    }
+});
+
+test("applyDefaultNamespace does not touch unrelated tools or non-call RPC messages", () => {
+    const loginMsg = {
+        jsonrpc: "2.0",
+        id: 4,
+        method: "tools/call",
+        params: {
+            name: "memwal_login",
+            arguments: {},
+        },
+    };
+    const updatedLogin = applyDefaultNamespace(loginMsg, "test-ns");
+    assert.equal(updatedLogin.params.arguments.namespace, undefined);
+
+    const listMsg = {
+        jsonrpc: "2.0",
+        id: 5,
+        method: "tools/list",
+        params: {},
+    };
+    const updatedList = applyDefaultNamespace(listMsg, "test-ns");
+    assert.equal(updatedList.params.arguments, undefined);
+});
--- a/docs/mcp/reference.md
+++ b/docs/mcp/reference.md
@@ -139,7 +139,7 @@ Set `MEMWAL_MCP_DEBUG=1` to enable verbose stderr logging.
 
 ## Default namespace
 
-Set a default memory namespace once in your client config instead of having the agent pass `namespace` on every call. The package injects it into `memwal_remember`, `memwal_recall`, `memwal_analyze`, and `memwal_restore` calls that don't already carry one.
+Set a default memory namespace once in your client config instead of having the agent pass `namespace` on every call. The package injects it into `memwal_remember`, `memwal_remember_bulk`, `memwal_recall`, `memwal_analyze`, and `memwal_restore` calls that don't already carry one.
 
 Precedence, highest first:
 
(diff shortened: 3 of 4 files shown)


packages/mcp/src/bridge.ts
```diff
@@ -35,12 +35,13 @@ export interface BridgeConfig {
 }
 
 /** Memory tools that take a `namespace` argument. `memwal_remember`,
- * `memwal_recall`, and `memwal_analyze` treat it as optional; `memwal_restore`
- * requires it (its upstream schema still lists `namespace` as required, so
- * agents normally pass one — but a configured default is filled in if the
- * agent calls it without). */
+ * `memwal_remember_bulk`, `memwal_recall`, and `memwal_analyze` treat it as
```
nikola0x0, line 41: The JSDoc is updated here, but the same list lives in two user-facing places that still enumerate the pre-fix four:

- `docs/mcp/reference.md:142` — "injects it into `memwal_remember`, `memwal_recall`, `memwal_analyze`, and `memwal_restore` calls that don't already carry one."
- `packages/mcp/README.md:61` — "the agent has to pass it on every `memwal_remember` / `memwal_recall` / `memwal_analyze` call (and `memwal_restore` requires it)."

The README one ships to npm. Since the bug being fixed *was* "users don't know which tools honour `--namespace`", publishing docs that still list the old set invites the same issue to be filed again. `reference.md:33` already advertises eight tools, so both omissions read as stale rather than deliberate.

```diff
--- a/docs/mcp/reference.md
+++ b/docs/mcp/reference.md
@@ -140,4 +140,4 @@
 Set a default memory namespace once in your client config instead of having
 the agent pass `namespace` on every call. The package injects it into
-`memwal_remember`, `memwal_recall`, `memwal_analyze`, and `memwal_restore`
-calls that don't already carry one.
+`memwal_remember`, `memwal_remember_bulk`, `memwal_recall`, `memwal_analyze`,
+and `memwal_restore` calls that don't already carry one.
```

```diff
--- a/packages/mcp/README.md
+++ b/packages/mcp/README.md
@@ -60,3 +60,3 @@
 By default the MCP tool schemas expose an optional `namespace` argument and the
-agent has to pass it on every `memwal_remember` / `memwal_recall` /
-`memwal_analyze` call (and `memwal_restore` requires it). Set a default once in
-your client config instead:
+agent has to pass it on every `memwal_remember` / `memwal_remember_bulk` /
+`memwal_recall` / `memwal_analyze` call (and `memwal_restore` requires it). Set
+a default once in your client config instead:
```

packages/mcp/test/default-namespace.test.mjs
```diff
@@ -0,0 +1,92 @@
...
+        method: "tools/call",
+        params: {
+            name: "memwal_remember_bulk",
+            arguments: {
+                facts: ["fact 1", "fact 2"],
+            },
+        },
+    };
+
+    const updated = applyDefaultNamespace(msg, "project-alpha");
```
nikola0x0, line 19: This asserts on the **returned** object, but the call site throws the return value away:

```ts
// bridge.ts:764
applyDefaultNamespace(msg, config.namespace);
```

What actually prevents #625 is the in-place mutation of `msg` — that is what makes the forwarded copy *and the replayed-on-reconnect copy* carry the namespace, per the comment directly above that line. Today `updated === msg`, so asserting on either one passes.

The failure mode: if someone later refactors `applyDefaultNamespace` into the more idiomatic non-mutating shape —

```ts
return { ...msg, params: { ...params, arguments: { ...args, namespace } } };
```

— all four tests in this file stay green while the message that actually goes over the wire silently loses its namespace. That is exactly the bug this PR fixes, reintroduced with a passing suite.

Cheap to pin, in at least one test:

```diff
     const updated = applyDefaultNamespace(msg, "project-alpha");
     assert.equal(updated.params.arguments.namespace, "project-alpha");
     assert.deepEqual(updated.params.arguments.facts, ["fact 1", "fact 2"]);
+    // The call site ignores the return value — in-place mutation is the real contract.
+    assert.equal(msg.params.arguments.namespace, "project-alpha");
```

packages/mcp/test/default-namespace.test.mjs
```diff
@@ -0,0 +1,92 @@
...
+    assert.equal(updatedLogin.params.arguments.namespace, undefined);
+
+    const listMsg = {
+        jsonrpc: "2.0",
+        id: 5,
+        method: "tools/list",
+        params: {},
+    };
+    const updatedList = applyDefaultNamespace(listMsg, "test-ns");
+    assert.equal(updatedList.params.arguments, undefined);
```
nikola0x0: This assertion passes regardless of what the function does. `listMsg.params` is `{}`, so `params.arguments` is already `undefined` on the way in — nothing has to preserve it.

Concretely: delete the `if (msg.method !== "tools/call") return msg;` guard at `bridge.ts:63` — the line this test presumably exists to cover — and the very next guard (`typeof params.name !== "string"`, line 67) still returns early, so `params.arguments` stays `undefined` and this stays green.

Exercising the method guard needs namespace-tool-shaped params on a non-`tools/call` method:

```diff
     const listMsg = {
         jsonrpc: "2.0",
         id: 5,
         method: "tools/list",
-        params: {},
+        params: { name: "memwal_remember_bulk", arguments: {} },
     };
     const updatedList = applyDefaultNamespace(listMsg, "test-ns");
-    assert.equal(updatedList.params.arguments, undefined);
+    assert.equal(updatedList.params.arguments.namespace, undefined);
```

The `memwal_login` half above it is fine as written — that one does pin the allowlist check.

packages/mcp/test/default-namespace.test.mjs
```diff
@@ -0,0 +1,92 @@
...
+});
```
nikola0x0, line 123: Two documented precedence rules land with no coverage, and both are cheap to add while this file is new:

1. **No default configured** — `applyDefaultNamespace(msg, undefined)` must forward untouched so the relayer applies its own `"default"`. That is the guard at `bridge.ts:62`, documented as rule 4 in `reference.md:151-152` and `README.md:96-97`. Nothing here would catch a hardcoded fallback creeping in.
2. **Blank explicit namespace** — `bridge.ts:73` deliberately treats `namespace: ""` and `"   "` as absent via `.trim() !== ""` and overwrites them with the default. That `.trim()` could be dropped or inverted and the suite would stay green.

```diff
+test("applyDefaultNamespace is a no-op when no default is configured", () => {
+    const msg = {
+        jsonrpc: "2.0",
+        id: 6,
+        method: "tools/call",
+        params: { name: "memwal_remember_bulk", arguments: { facts: ["fact 1"] } },
+    };
+
+    applyDefaultNamespace(msg, undefined);
+    assert.equal(msg.params.arguments.namespace, undefined);
+});
+
+test("applyDefaultNamespace overrides a blank explicit namespace", () => {
+    for (const blank of ["", "   "]) {
+        const msg = {
+            jsonrpc: "2.0",
+            id: 7,
+            method: "tools/call",
+            params: {
+                name: "memwal_remember_bulk",
+                arguments: { facts: ["fact 1"], namespace: blank },
+            },
+        };
+
+        applyDefaultNamespace(msg, "project-alpha");
+        assert.equal(msg.params.arguments.namespace, "project-alpha");
+    }
+});
```

nikola0x0 (approved):
**Pre-approve before the two doc lists are updated** — the fix itself is correct and I verified it end to end.

Root cause and shape are both right. `services/server/scripts/mcp/tools/remember-bulk.ts:16-21` declares `namespace` as an optional **top-level** argument that the handler fans out across every fact (`facts.map((text) => ({ text, namespace }))`), so injecting at the top level rather than per-item is exactly what the relayer expects. `NAMESPACE_TOOLS` is the only *namespace* allowlist, so nothing else needs the same entry — though `auth-required.ts:37` maintains its own hardcoded tool list and has drifted from the relayer in the same way (collapsed section below).

I also checked that the new tests are real regression guards rather than decoration: at `86c14a5`, deleting the one-line `Set` addition and rebuilding fails 2 of them; with the fix, 14/14 pass. CI wiring is sound — `.github/workflows/test.yml:78` runs the suite and the `test` script does `tsc &&` first, so `../dist/bridge.js` resolves under `pnpm test`. Worth flagging that this is the first *static* top-level import of a built module in the suite: three existing files spawn `dist/bin/memwal-mcp.js` as a subprocess and `login-preflight.test.mjs:14` uses a cache-busting dynamic import, so a bare `node --test` or an editor runner hard-errors on this file where the others don't.

### Majors (details inline)

- **Two user-facing copies of the tool list are now stale.** The JSDoc at `bridge.ts:37-41` was updated, but `docs/mcp/reference.md:142` and `packages/mcp/README.md:61` still enumerate the old four — and the README ships to npm. This is the one thing I would fix before merge; two one-line edits, both inline.
- **The tests do not pin the contract that actually prevents #625.** Every assertion is on the returned object, while `bridge.ts:764` discards the return value and depends purely on in-place mutation of `msg` for the forwarded and replayed-on-reconnect copies. A non-mutating refactor would keep all four tests green while reintroducing the bug. Fix inline on the test file.

### Minor (2)

A vacuous `tools/list` assertion and two uncovered precedence branches (no-default, blank namespace). Both inline, both optional.

<details>
<summary><b>Pre-existing / out of diff — not this PR (4)</b></summary>

None of these are regressions from this change and none should block it. Flagging them because the first one is the same class of bug as #625 and is probably worth its own issue.

**1. `auth-required.ts:37` has the identical drift bug.** The pre-login `tools/list` response hardcodes its own tool array, and it advertises only five: `memwal_remember`, `memwal_recall`, `memwal_analyze`, `memwal_restore`, `memwal_login`. Both `memwal_remember_bulk` and `memwal_health` are missing — same root cause as #625, a hand-maintained list that fell behind the relayer's registry. Impact is narrower than it first looks, since `tools/list` hot-hands-off to upstream once credentials exist, but `capabilities: { tools: { listChanged: false } }` (`auth-required.ts:342`) explicitly tells the client the list will never change, which invites caching the pre-login five. Worth reconciling against `services/server/scripts/mcp/tools/` in a follow-up.

**2. `bridge.ts:797` reports every outbound failure as a parse error.** The `catch` wraps the whole IIFE — `JSON.parse`, `postMessage`, and `reconnect` alike — and logs all of them as `bridge.stdin_parse_failed` with the offending line. So a relayer TCP reset mid-tool-call is surfaced as a stdin parse problem, and because no JSON-RPC error envelope is written back for `msg.id`, the client waits forever on a call that already died. Mislabelled event plus a silent drop.

**3. Blank configured namespace is injected verbatim.** `MEMWAL_NAMESPACE="  "` (easy to produce from a stray space in a JSON `env` block) passes the `!namespace` check at `bridge.ts:62` and gets injected as-is, so memories land in a namespace named two spaces. Meanwhile the identical value arriving from the agent is treated as absent by the `.trim() !== ""` guard at line 73. Trimming and validating at resolution time in `index.ts` would make the two ends agree.

**4. `README.md:106` claims the package ships no test runner.** It says "No automated test runner ships with this package (consistent with the rest of the monorepo)" and then gives a manual MCP-client procedure for verifying namespace injection. Already false before this PR — `package.json` defines `test`, five suites exist, and both `test.yml:78` and `release-mcp.yml:51` gate on them. Natural moment to replace that paragraph with `pnpm --filter @mysten-incubation/memwal-mcp test`, since this PR adds the sixth suite to exactly that feature area.

</details>

ducnmm:
Updated! Addressed all review points:
- Added `memwal_remember_bulk` to the tool list in `docs/mcp/reference.md` and `packages/mcp/README.md`.
- Hardened unit tests in `packages/mcp/test/default-namespace.test.mjs` (asserted on in-place mutation of `msg`, corrected non-tool-call test fixture, and added coverage for no-default and blank namespace precedence rules).

All 16 MCP tests pass locally. Ready for merge!

jessiemongeon1:
## Style Guide Audit

PR #672, opened 2026-08-17
fix(relayer): allow batched delegate-key removal in one sponsored tx

harrymove-ctrl:
Fixes bulk delegate-key revoke on the dashboard, which failed with `Sponsor request was rejected. Please try again.` while adding a key worked.

Linear: [WALM-334](https://linear.app/mysten-labs/issue/WALM-334/staging-bulk-delegate-key-removal-fails-sponsorship-allowlist-permits)
Notion: https://app.notion.com/p/3bf86ef3b64e81a1aabcf997d0402ff4

## Root cause

`validate_sponsor_transaction_kind` required **exactly one** command per sponsored transaction:

```rust
let permitted = programmable.commands.len() == 1 && /* allowlisted account call */;
```

`apps/app/src/pages/Dashboard.tsx` builds bulk revoke as one `remove_delegate_key` MoveCall **per selected key** in a single transaction. Selecting 3 keys produced 3 commands, so the relayer rejected it with `AppError::BadRequest("Transaction kind is not permitted for sponsorship")` → HTTP 400, before Enoki was ever contacted.

| Action | Commands | Before |
|---|---|---|
| Add key | 1 | passes |
| Remove 1 key (trash button) | 1 | passes |
| Remove 2+ keys (Remove selected) | N | **400** |

The one-command rule arrived in `cef97282` (2026-07-31). Enoki, its allowlist, credentials, rate limits, package-id mismatch, and `ACCOUNT_MUTATIONS_PAUSED` were all ruled out — no Enoki-side configuration change is involved.

## Changes

**`services/server/src/routes/sponsor.rs`** — batching for delegate-key removal only:

- One command: any allowlisted call (`create_account`, `add_delegate_key`, `remove_delegate_key`). Unchanged.
- Two or more: every command must be `remove_delegate_key`, capped at `MAX_SPONSORED_DELEGATE_REMOVALS = 20`. That matches `MAX_DELEGATE_KEYS` in `services/contract/sources/account.move`, so it is exactly enough to revoke every key at once and no more, and it bounds sponsored gas per rate-limited request.
- Empty command lists are now rejected explicitly — no effects, but they still burn sponsored gas.
- Mixed batches (`add` + `remove`), repeated `add_delegate_key`, repeated `create_account`, and any foreign command inside a batch all stay rejected. Batching is a bulk-revoke affordance, not a general multi-call hole. The per-command package/module/function/type-argument checks are unchanged.

**Diagnosability** — the reason this took a log dig rather than a glance:

- `mask_upstream` now returns a stable `code` (`sponsor_overloaded`, `sponsor_misconfigured`, `sponsor_upstream_error`, `sponsor_rejected`) alongside the generic message. Statuses are unchanged and the upstream body is still never echoed.
- Sponsor error bodies carry `traceId`, and both upstream-error log lines record `request_id`, so a response ties to one log line.
- `useSponsoredTransaction.ts` surfaces the relayer's own message when the relayer rejected the request itself, and uses `code` to tell a misconfigured sponsor apart from a rejected transaction. Previously every 400 collapsed into the generic retry message, which read as a transient blip and hid an actionable server message.

## Testing

- `services/server`: 523 pass. New cases — batched removals at 2/3/20 accepted, 21 rejected, impure batches rejected, empty command list rejected, `mask_upstream` codes asserted stable snake_case.
- `apps/app`: 84 pass, including a new `useSponsoredTransaction.test.ts` (11 cases) pinning the regression that the generic message must not swallow the relayer's reason.
- `cargo build --release` and the `apps/app` Docker image both build. The relayer Docker image could not be built locally: its base images are pinned by digest to `linux/amd64` and rustc segfaults under qemu on arm64, failing at the dependency-cache stub before any project source is compiled. CI on amd64 is the real gate.

## Staging verification after deploy

Delete-flow checklist is in the Notion page (sections A–J). The sharp checks:

- Remove a single key — expected to pass even without this change
- Remove 2, 3, and 20 keys — the actual regression
- Add key, remember, recall, and memory deletion still work

## Separate follow-up, not in this PR

The staging `relayer` service sets `ACCOUNT_MUTATIONS_PAUSED=true`. That gate exists only on the unmerged `commandoss/feat/block-account-creation-migration-staging` (commit `2e30ae7c`), where enabling it makes both `/sponsor` legs return 503 before parsing. The deployed build clearly does not read it, since add key works — but a redeploy from a branch that implements it would kill all sponsorship, including memory writes. Worth removing or setting to `false`.

### DIFF
--- a/apps/app/src/hooks/useSponsoredTransaction.test.ts
+++ b/apps/app/src/hooks/useSponsoredTransaction.test.ts
@@ -1,0 +1,99 @@
+import { describe, expect, it } from 'vitest'
+import { SponsorHttpError, sponsorFailureMessage } from './useSponsoredTransaction'
+
+describe('SponsorHttpError', () => {
+    it('parses the relayer error envelope', () => {
+        const err = new SponsorHttpError(
+            'sponsor',
+            400,
+            JSON.stringify({ error: 'Sponsor request rejected', code: 'sponsor_rejected', traceId: 'req-1' }),
+        )
+        expect(err.detail).toBe('Sponsor request rejected')
+        expect(err.code).toBe('sponsor_rejected')
+        expect(err.traceId).toBe('req-1')
+    })
+
+    it('survives a non-JSON body', () => {
+        const err = new SponsorHttpError('sponsor', 502, '<html>gateway</html>')
+        expect(err.code).toBeUndefined()
+        expect(err.detail).toBeUndefined()
+        expect(err.body).toBe('<html>gateway</html>')
+    })
+
+    it('survives a JSON body that is not an object', () => {
+        const err = new SponsorHttpError('sponsor', 400, '"just a string"')
+        expect(err.code).toBeUndefined()
+        expect(err.detail).toBeUndefined()
+    })
+})
+
+describe('sponsorFailureMessage', () => {
+    const bodyFor = (fields: Record<string, string>) => JSON.stringify(fields)
+
+    it('surfaces the relayer reason when it rejected the request itself', () => {
+        // Regression: an un-sponsorable transaction shape used to collapse into
+        // the generic "please try again", which read as a transient blip and
+        // hid the real cause — a bulk delegate-key revoke exceeding what the
+        // sponsorship allowlist permits.
+        const err = new SponsorHttpError(
+            'sponsor',
+            400,
+            bodyFor({ error: 'Transaction kind is not permitted for sponsorship' }),
+        )
+        expect(sponsorFailureMessage(err)).toBe('Transaction kind is not permitted for sponsorship')
+    })
+
+    it('surfaces the removal-cap reason verbatim', () => {
+        const err = new SponsorHttpError(
+            'sponsor',
+            400,
+            bodyFor({ error: 'Too many delegate key removals in one transaction (max 20)' }),
+        )
+        expect(sponsorFailureMessage(err)).toContain('max 20')
+    })
+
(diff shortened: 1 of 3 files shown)


ducnmm (approved):
Looks good — approve.

The allowlist change is the right shape: one command stays any of create/add/remove; 2+ commands must be pure `remove_delegate_key` and capped at 20 to match `MAX_DELEGATE_KEYS`. Mixed batches, empty command lists, and foreign commands stay rejected, so this is a bulk-revoke affordance rather than a general multi-call hole.

The diagnosability fix is what would have made WALM-334 obvious from the browser. Relayer-own 400s (no `code`) now surface the real reason; masked upstream failures keep a stable `code` + `traceId` without echoing Enoki.

Nits, non-blocking:
- `json_error_response` mints a fresh UUID when `current_request_id()` is missing, while the matching log line uses `unwrap_or_default()`. Fine on the HTTP path (middleware always sets context), but the two fallbacks should be the same so a missing context cannot invent a traceId that is not in the logs.
- `sponsor_overloaded` (503) hits `status >= 500` before the `code` branch, so the UI drops `traceId` on that path.

Staging after merge: Notion checklist B + C (1 key, then 2/3/20), plus add/remember/recall/delete smoke.

Out of scope but still worth doing: staging relayer has `ACCOUNT_MUTATIONS_PAUSED=true`. The current build ignores it; a redeploy from a branch that implements the gate would 503 every `/sponsor` call.

PR #673, opened 2026-08-17
fix(researcher): retired title model broke every new chat — plus tooltip, auto-resume, pyodide, and sprint-export fixes

nikola0x0:
## Summary

Fixes the production P0 where **every new Researcher chat rendered as a failed response**, plus four more defects found while verifying it live on `researcher.demo.memwal.ai`, and one UX gap in saved sprints.

### The P0 (`c031c472`)

`getTitleModel()` hardcoded `google/gemini-2.0-flash-001`, which OpenRouter has retired — it 404s (`No endpoints found`) for our own production key. The unguarded `await titlePromise` inside `createUIMessageStream`'s `execute()` then surfaced that rejection as an `error` part on the stream, so the client showed a fatal chat error even though the real answer had streamed fine underneath.

- Title model moved to a `TITLE_MODEL` constant beside the curated list, pointed at `gemini-2.5-flash`
- Title block wrapped in try/catch — ports the chatbot fix (`334cf915`) that researcher was missed from
- Picker audit with the production key found `claude-3.5-haiku` / `claude-3.5-sonnet` **also retired** (3 of 6 selectable models were dead) — removed, with a regression test pinning all three retired ids

Production impact when found: 11 of 43 chats untitled; the error path live since the 08-14 deploy.

### Also in this branch

- `754327f4` — portal `TooltipContent` (sticky-header stacking context clipped every tooltip to a 2px sliver); single-line suggestion pills; collapse a duplicate unreachable `DEFAULT_SUGGESTIONS`
- `39e102ae` — stop autoResume firing against the stub stream route (always 204): raced in AI SDK 6.0.37 and threw `Cannot read properties of undefined (reading 'state')` when reopening any reply-less chat
- `9db6db10` — delete the template-leftover pyodide `<Script>`: nothing calls it, multi-MB download per chat page, and a React error under Next 16
- `b9b57890` — sprint Copy-as-markdown / Download `.md` / expanded reading view (shared unit-tested `buildSprintMarkdown`); real page title (`Researcher | Walrus Memory`) instead of "Next.js Chatbot Template"

## Verification

- Local A/B against the exact prod condition (retired title model): with the guard the answer streams with no error frame; without it the error frame reproduces
- Live OpenRouter checks with the production key: retired ids → 404, replacement → 200; full e2e with a real key generates a title end-to-end
- Auto-resume fix verified on the repro chat (clean console; previously threw on every load)
- Sprint actions verified in-browser: clipboard content, toast, dialog
- `tsc --noEmit` clean on changed files; `pnpm test:unit` 14/14 (7 new tests)

## Not in this branch / follow-ups

- **Deploy drift:** prod currently runs `ec6a9b81a` from the unmerged `hotfix/noter-mainnet-rpc-cors` branch — this PR can't reach prod through normal promotion until that's reconciled
- Prod has 6 chats with a user message and no persisted assistant row; that symptom did **not** reproduce locally and needs a post-deploy re-check
- `apps/chatbot` shares the dead title model, (probably) the dead Anthropic picker entries, the unportaled tooltip, and the pyodide tag — needs its own branch

### DIFF
--- a/apps/researcher/app/(chat)/chat/[id]/page.tsx
+++ b/apps/researcher/app/(chat)/chat/[id]/page.tsx
@@ -5,7 +5,7 @@ import { Suspense } from "react";
 import { getSession } from "@/lib/auth/session";
 import { Chat } from "@/components/chat/chat";
 import { DataStreamHandler } from "@/components/data/data-stream-handler";
-import { DEFAULT_CHAT_MODEL } from "@/lib/ai/models";
+import { resolveChatModelId } from "@/lib/ai/models";
 import { getChatById, getMessagesByChatId, getSprintsByIds } from "@/lib/db/queries";
 import { convertToUIMessages } from "@/lib/utils";
 
@@ -48,9 +48,7 @@ async function ChatPage({ params }: { params: Promise<{ id: string }> }) {
   const uiMessages = convertToUIMessages(messagesFromDb);
 
   const cookieStore = await cookies();
-  const chatModelFromCookie = cookieStore.get("chat-model");
-
-  const chatModel = chatModelFromCookie?.value ?? DEFAULT_CHAT_MODEL;
+  const chatModel = resolveChatModelId(cookieStore.get("chat-model")?.value);
 
   // Fetch sprint data if chat has associated sprints
   const sprintIds = chat.sprintIds ?? [];
--- a/apps/researcher/app/(chat)/chat/[id]/page.tsx
+++ b/apps/researcher/app/(chat)/chat/[id]/page.tsx
@@ -71,7 +71,13 @@ async function ChatPage({ params }: { params: Promise<{ id: string }> }) {
   return (
     <>
       <Chat
-        autoResume={true}
+        // Resumable streaming is not implemented — api/chat/[id]/stream is a
+        // stub that always returns 204. With autoResume on, reopening a chat
+        // whose last message is from the user fired resumeStream() against
+        // that stub; overlapping attempts (easy under dev double-effects) race
+        // inside AI SDK 6.0.37 and throw "Cannot read properties of undefined
+        // (reading 'state')". Re-enable only once the stream route is real.
+        autoResume={false}
         id={chat.id}
         initialChatModel={chatModel}
         initialMessages={uiMessages}
--- a/apps/researcher/app/(chat)/api/chat/route.ts
+++ b/apps/researcher/app/(chat)/api/chat/route.ts
@@ -307,9 +307,23 @@ export async function POST(request: Request) {
         );
 
         if (titlePromise) {
-          const title = await titlePromise;
-          dataStream.write({ type: "data-chat-title", data: title });
-          updateChatTitleById({ chatId: id, title });
+          try {
+            const title = await titlePromise;
+            dataStream.write({ type: "data-chat-title", data: title });
+            updateChatTitleById({ chatId: id, title });
+          } catch (error) {
+            // Title generation is a non-fatal side effect of the response
+            // already streaming above it (dataStream.merge). Left unguarded,
+            // the rejection reached createUIMessageStream's onError, which
+            // injected an "error" part into the stream — so the client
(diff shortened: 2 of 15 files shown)


apps/researcher/app/(chat)/chat/[id]/page.tsx
```diff
@@ -71,7 +71,13 @@ async function ChatPage({ params }: { params: Promise<{ id: string }> }) {
...
-        autoResume={true}
+        // Resumable streaming is not implemented — api/chat/[id]/stream is a
+        // stub that always returns 204. With autoResume on, reopening a chat
+        // whose last message is from the user fired resumeStream() against
+        // that stub; overlapping attempts (easy under dev double-effects) race
+        // inside AI SDK 6.0.37 and throw "Cannot read properties of undefined
+        // (reading 'state')". Re-enable only once the stream route is real.
+        autoResume={false}
         id={chat.id}
         initialChatModel={chatModel}
```
Comment (ducnmm) [inline apps/researcher/app/(chat)/chat/[id]/page.tsx:80]: `chatModel` is still `cookie ?? DEFAULT_CHAT_MODEL` (line 53). After this PR, `google/gemini-2.0-flash-001` / `anthropic/claude-3.5-haiku` / `anthropic/claude-3.5-sonnet` are not in `allowedModelIds`, so `/api/chat` 400s them. The compact picker already falls back for *display*, so the UI can show Gemini 2.5 Flash while this prop still holds the retired cookie.

Same one-liner on this page and on `app/(chat)/page.tsx` (new chat — not in this diff):

```ts
const raw = cookieStore.get("chat-model")?.value;
const chatModel = raw && allowedModelIds.has(raw) ? raw : DEFAULT_CHAT_MODEL;
```

ducnmm (changes requested):
P0 is right: unguarded `titlePromise` can no longer inject an error frame, `TITLE_MODEL` is on the curated list, and the three retired picker ids are gone. `autoResume={false}` belongs with this fix.

Please coerce the `chat-model` cookie against `allowedModelIds` on both `app/(chat)/page.tsx` and `chat/[id]/page.tsx` before merge. Users who had Gemini 2.0 / Claude 3.5 selected still 400 on send because `/api/chat` rejects unknown ids while the picker falls back for display only.

```ts
const raw = cookieStore.get("chat-model")?.value;
const chatModel = raw && allowedModelIds.has(raw) ? raw : DEFAULT_CHAT_MODEL;
```

Tooltip / pyodide / sprint-export look safe; not blocking.

nikola0x0:
Confirmed and fixed in 302bec60.

Traced it: `currentModelId` seeds from `initialChatModel` (`chat.tsx:80`) and ships as `selectedChatModel`, which `api/chat/route.ts:76` checks against `allowedModelIds`. The picker fallback (`multimodal-input.tsx:547`) only swaps the displayed name, so the trigger reads "Gemini 2.5 Flash" while the retired id still goes on the wire.

Same semantics as your one-liner, but pulled into `lib/ai/models.ts` since both pages read the cookie:

```ts
export function resolveChatModelId(cookieValue: string | undefined): string {
  return cookieValue && allowedModelIds.has(cookieValue)
    ? cookieValue
    : DEFAULT_CHAT_MODEL;
}
```

Called from `app/(chat)/page.tsx:19` and `chat/[id]/page.tsx:51`. Helper form made it testable — 3 cases added for the retired ids, an unknown id, and no cookie. `pnpm test:unit` 17/17, `tsc --noEmit` clean.

Note: this coerces on read, it does not rewrite the cookie (server components cannot set cookies). Harmless now that nothing stale reaches `/api/chat`.

ducnmm (approved):
Re-review after 302bec60.

`resolveChatModelId` now coerces the `chat-model` cookie against `allowedModelIds` on both `app/(chat)/page.tsx` and `chat/[id]/page.tsx`, so a retired cookie can no longer 400 `/api/chat` while the picker shows the default name. The new tests pin the three retired ids and the stale/absent cookie cases. P0 (title-model guard + `TITLE_MODEL` on the curated list + `autoResume={false}`) is unchanged.

Approved.

PR #680, opened 2026-08-18
test(researcher): Playwright e2e suite and CI job

nikola0x0:
## Summary

Closes the "e2e test researcher" half of the task: a 15-test Playwright suite plus a `researcher-e2e` CI job, mirroring the existing `chatbot-e2e` setup.

Researcher had **no automated tests at all** — the app with the only real Walrus writes was verified entirely by hand. That's how the title-model P0 in #673 reached production.

> **Stacked on #673** (base `fix/researcher-title-model-chat-crash`). The P0 regression spec asserts the error-frame guard from that PR, so merge #673 first; this PR then retargets to `dev` automatically.

## The test that would have caught the P0

`e2e/chat-stream.test.ts` asserts the raw SSE frames from `/api/chat`: a failing title model must not put an `error` frame on the stream.

Verified as a **real** regression test, not merely a passing one — with the `try/catch` guard removed it fails with the exact production symptom, while the answer still streams underneath:

```
Expected length: 0
Received length: 1
Received array:  [{"errorText": "Oops, an error occurred!", "type": "error"}]
```

## Mock seams

Four seams, all keyed off the existing `isTestEnvironment`, so a run never reaches OpenRouter, Sui, or the relayer:

| Seam | What it does |
|------|--------------|
| `lib/ai/providers.ts` | Maps picker model ids onto the three registered mock models. **The pre-existing test branch was broken** — it passed the raw id straight through, so `languageModel("google/gemini-2.5-flash")` would have thrown `NoSuchModelError` on the first message. The branch had simply never run. |
| `lib/ai/models.mock.ts` | A `FAIL_TITLE_GENERATION` sentinel in a user message makes the title model reject, reproducing the P0 on demand. |
| `lib/auth/delegate-account.mock.ts` (new) | Fabricates the `MemWalAccount` object for two fixture identities. The real binding validation still runs, so unregistered keys and unknown accounts fail exactly as they do on-chain — both asserted. |
| `lib/sprint/memwal.ts` | `MemWalMock` from the SDK, one instance per process, so remember → recall round-trips in memory and **CI can never write to the production relayer**. |

## Coverage

| Spec | Covers |
|------|--------|
| `auth.setup.ts` | Signs in as account A through the real form (doubles as the happy-path login test) and account B via API; saves both storage states. |
| `e2e/auth.test.ts` | Anonymous redirects; unregistered key, unknown account, malformed key rejections. |
| `e2e/chat.test.ts` | Input/suggestions render; message streams a reply; chat survives reload with both rows. Asserts **zero uncaught page errors**, guarding the auto-resume `TypeError` from `39e102ae`. |
| `e2e/chat-stream.test.ts` | SSE frame protocol + the P0 regression test above. |
| `e2e/visibility.test.ts` | Private/Public boundary across two real accounts. |

## Two behaviors pinned as they are, not as expected

Both found while writing the specs; neither leaks content:

1. **A private chat opened by another user returns HTTP 200**, not 404 — `app/(chat)/chat/[id]/page.tsx` wraps the async component in `<Suspense>`, so the streaming shell is flushed before `notFound()` runs. The user correctly sees the not-found page; only the status code is misleading (matters for monitoring/bots). The spec asserts the rendered page and the absence of the canary.
2. **Anonymous visitors are redirected to `/login` even for public chats**, since `proxy.ts`'s auth check precedes the visibility check — the known "Public — Anyone with the link" product gap. Pinned deliberately, with a comment to update it if that product decision changes.

## Flakiness worked out up front

`global-setup.ts` clears the auth limiter's Redis keys before each run. The limiter allows 10 verify attempts per IP per minute and a run spends five, so without this a second run inside a minute fails at sign-in with a 429 that looks nothing like the real problem. Both identities also sign in once in a setup project rather than per-test, keeping CI's 2 retries clear of the limit.

Also worth knowing for anyone adding specs (documented in `tests/README.md`): `browser.newContext()` **inherits** the project's `use.storageState`, so a context that must be someone else — including an anonymous one — has to pass its own explicitly. Miss it and the test still passes, for the wrong reason. My first draft of the visibility specs hit exactly this.

## Verification

- **15/15 green on three consecutive runs**, ~10s each
- Green from cold with Playwright managing its own webServer, and with `AUTH_SECRET` / `NEXT_PUBLIC_MEMWAL_PACKAGE_ID` unset — so the config defaults CI relies on are exercised
- Negative control: guard removed → P0 spec fails as shown above; restored → green
- `tsc --noEmit` clean; `pnpm test:unit` 14/14

## Notes

- Lint is not wired into the new job: `ultracite`/`biome` version mismatch already breaks `pnpm lint` repo-wide (soft-failed for chatbot in the same workflow). Same follow-up.
- The suite has no live-model or live-Walrus canary by design; the real remember → recall loop against the production relayer stays a manual check.

### DIFF
--- a/apps/researcher/lib/sprint/memwal.ts
+++ b/apps/researcher/lib/sprint/memwal.ts
@@ -1,10 +1,20 @@
 import "server-only";
 
-import { MemWal } from "@mysten-incubation/memwal";
+import { MemWal, MemWalMock } from "@mysten-incubation/memwal";
 import type { RememberResult } from "@mysten-incubation/memwal";
+import { isTestEnvironment } from "@/lib/constants";
 import type { Citation, SourceMeta } from "./types";
 
-function getMemWalClient(key: string, accountId: string) {
+// Shared across requests so a sprint remembered in one Playwright request is
+// recallable in the next — and so test runs can never write to the real
+// relayer configured in MEMWAL_SERVER_URL.
+let testMemWalClient: MemWalMock | null = null;
+
+function getMemWalClient(key: string, accountId: string): MemWal | MemWalMock {
+  if (isTestEnvironment) {
+    testMemWalClient ??= MemWalMock.create({ owner: "playwright" });
+    return testMemWalClient;
+  }
   return MemWal.create({
     key,
     accountId,
--- a/apps/researcher/lib/sprint/memwal.ts
+++ b/apps/researcher/lib/sprint/memwal.ts
@@ -5,15 +5,20 @@ import type { RememberResult } from "@mysten-incubation/memwal";
 import { isTestEnvironment } from "@/lib/constants";
 import type { Citation, SourceMeta } from "./types";
 
-// Shared across requests so a sprint remembered in one Playwright request is
-// recallable in the next — and so test runs can never write to the real
-// relayer configured in MEMWAL_SERVER_URL.
-let testMemWalClient: MemWalMock | null = null;
+// One mock per account, mirroring the real client's per-(key, accountId)
+// construction: a sprint remembered by one identity must not be recallable by
+// another. Held across requests so remember → recall round-trips within a run,
+// and so test runs can never write to the real relayer in MEMWAL_SERVER_URL.
+const testMemWalClients = new Map<string, MemWalMock>();
 
 function getMemWalClient(key: string, accountId: string): MemWal | MemWalMock {
   if (isTestEnvironment) {
-    testMemWalClient ??= MemWalMock.create({ owner: "playwright" });
-    return testMemWalClient;
+    const existing = testMemWalClients.get(accountId);
+    if (existing) return existing;
+
+    const created = MemWalMock.create({ owner: accountId });
+    testMemWalClients.set(accountId, created);
+    return created;
   }
   return MemWal.create({
     key,
--- a/apps/researcher/playwright.config.ts
+++ b/apps/researcher/playwright.config.ts
@@ -1,0 +1,105 @@
+import path from "node:path";
+import { defineConfig, devices } from "@playwright/test";
+import { config } from "dotenv";
+
+// Later files never override earlier ones, matching Next's own precedence.
+config({ path: ".env.local" });
+config({ path: ".env" });
+
+// Match the dev script's default port (next dev, no --port flag).
+const PORT = process.env.PORT || "3000";
+const baseURL = `http://localhost:${PORT}`;
+
+const isCI = !!process.env.CI;
+
+export const STORAGE_STATE_USER_A = path.join(
+  __dirname,
+  "tests/playwright/.auth/user-a.json"
+);
+
+// A second identity for cross-account tests. Signed in once by the setup
+// project rather than per-test: the auth limiter allows only 10 verify
+// attempts per IP per minute, which retries would otherwise exhaust.
+export const STORAGE_STATE_USER_B = path.join(
+  __dirname,
+  "tests/playwright/.auth/user-b.json"
+);
+
+export default defineConfig({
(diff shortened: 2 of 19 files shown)


apps/researcher/playwright.config.ts
```diff
@@ -0,0 +1,105 @@
...
+        storageState: STORAGE_STATE_USER_A,
+      },
+    },
+  ],
+
+  webServer: {
+    command: "pnpm dev",
+    url: `${baseURL}/ping`,
+    timeout: 120_000,
+    reuseExistingServer: !isCI,
```
ducnmm, line 87: **[suggestion]** Local `reuseExistingServer: !isCI` is the same pattern as chatbot/noter, but this app is the one that actually writes to Walrus. If a developer already has `pnpm dev` up **without** `PLAYWRIGHT=True`, Playwright will reuse it and `isTestEnvironment` stays false — `MemWal.create()` then hits `MEMWAL_SERVER_URL`.

CI is safe (`reuseExistingServer` is false there). For local runs, consider `reuseExistingServer: false` always, or refuse to reuse unless the existing server advertises the test seam (e.g. a `/ping` header). Worth a line in `tests/README.md` either way.

apps/researcher/lib/sprint/memwal.ts
```diff
@@ -1,10 +1,20 @@
...
 
-function getMemWalClient(key: string, accountId: string) {
+// Shared across requests so a sprint remembered in one Playwright request is
+// recallable in the next — and so test runs can never write to the real
+// relayer configured in MEMWAL_SERVER_URL.
+let testMemWalClient: MemWalMock | null = null;
+
+function getMemWalClient(key: string, accountId: string): MemWal | MemWalMock {
+  if (isTestEnvironment) {
+    testMemWalClient ??= MemWalMock.create({ owner: "playwright" });
```
ducnmm: **[suggestion]** One process-wide `MemWalMock` with `owner: "playwright"` ignores `key` / `accountId`. Fine for the current suite (no sprint isolation specs). If a later spec remembers as A and recalls as B, both share the same in-memory store and the test will pass for the wrong reason. Key the mock on `accountId` (or pass `owner: accountId`) when you add that coverage.

ducnmm (approved):
## Summary

This is the right test to add. Researcher had no automated coverage, and `chat-stream.test.ts` actually pins the title-model P0: a failing title generation must not put an `error` frame on the SSE stream while the answer still streams. Mock seams are gated on `isTestEnvironment` (`PLAYWRIGHT`), the CI job passed (`Researcher / Unit + Playwright E2E` 2m10s), and the two-account visibility specs correctly pass storage state explicitly instead of inheriting user A.

Two non-blocking notes below. Happy to merge as-is.

## Issue counts

- bugs: 0
- suggestions: 2

apps/researcher/playwright.config.ts
nikola0x0: Went with the seam header rather than disabling reuse. `/ping` now returns `x-researcher-test-mode`, and `global-setup.ts` aborts the run if it isn't `1` — so `PLAYWRIGHT=True pnpm dev` keeps the fast `--ui` loop and anything else is refused. Verified against a plain `pnpm dev`: the run stops before any browser starts. Noted in `tests/README.md`. 593552ac

apps/researcher/lib/sprint/memwal.ts
nikola0x0: Keyed it now rather than waiting for the spec: `Map<accountId, MemWalMock>` with `owner: accountId`, matching the real client's per-`(key, accountId)` construction. No spec covers sprint isolation yet, so it's type-checked only. 593552ac

PR #678, opened 2026-08-18
chore(release): bump MCP package to 0.0.9

ducnmm:
## Summary

`@mysten-incubation/memwal-mcp@0.0.8` is already on npm (2026-08-14). The WALM-321 cold-start fix (#579) landed on `dev` after that publish and was appended under the **0.0.8** heading.

This is the missing manual version bump:

- `0.0.8` → **`0.0.9`** in every MCP / plugin manifest (same set as the 0.0.8 release).
- New `## 0.0.9` changelog in `packages/mcp/CHANGELOG.md` and `docs/mcp/changelog.mdx`.
- The 0.0.8 heading is restored to what actually shipped (tool titles / hints only).
- `TESTING.md` pins updated to `0.0.9-dev.N` / `latest` = 0.0.9 after `main`.

Also syncs `scripts/verify-manual-sdk-release.mjs` to SDK **0.1.3** (already on `dev`) so the checker matches the tree.

No code change. OpenClaw stays 0.0.5.

## Why a separate PR

#677 (`dev` → `staging`) should not carry a version bump of its own. Merge this to `dev` first; #677 will pick it up automatically.

## After merge

- `release-mcp.yml` should publish `@mysten-incubation/memwal-mcp@0.0.9-dev.N` on the `dev` dist-tag.
- `latest` stays **0.0.8** until this reaches `main`.

PR #681, opened 2026-08-18
fix(mcp): send proactive-usage instructions in the initialize handshake

harrymove-ctrl:
Closes WALM-324.

## Problem

Desktop MCP clients and Codex switched to lazy tool loading. Tool schemas no longer enter the model's context until a tool is explicitly loaded, and our entire "when to save / when to recall" contract lived inside those tool descriptions.

Result: the model sees no memory tool, so it either offers its own built-in memory or flatly denies the tool exists. It only works if the user prefixes the prompt with "using the memwal tools...".

## Why it still worked in the demo videos

The plugin install path ships a SessionStart hook (`packages/mcp/plugin/scripts/on_session_start.mjs`) that injects the proactive guidance into the session. Plain MCP installs (`command + args` in the client config) never got that hook, so only plugin users kept working. The videos were recorded on the plugin path.

## Fix

The MCP `instructions` field is delivered with `initialize`, before any `tools/list`, so lazy tool loading cannot strip it. Clients inject it into the model's system prompt.

There are **three** initialize responders, and none of them set it:

1. `services/server/scripts/mcp/server.ts` (relayer) constructed `McpServer` with no options object. Serves the direct HTTP/OAuth connector path.
2. `packages/mcp/src/bridge.ts` `buildLocalInitializeResult` answers `initialize` **locally** at cold start and **suppresses** the relayer's reply. This is the path every signed-in stdio client uses.
3. `packages/mcp/src/auth-required.ts` hand-rolls its own result for the signed-out case.

All three now send `instructions`. The relayer and bridge carry the full recall/remember/recover guidance plus a line countering the reported symptom ("if a memwal_* tool is not currently loaded, load it and use it; never tell the user that memory is unavailable"). The signed-out stub instead points at `memwal_login`, since every memory tool fails without credentials and telling the model to save proactively there would only manufacture errors.

Bridge and auth-required share one definition in the new `packages/mcp/src/instructions.ts`. The relayer keeps its own copy because `services/server/scripts` is the standalone `memwal-server-scripts` npm package with no workspace link to `packages/mcp`; both sides carry a comment pointing at the other.

### How #2 was found

Worth recording, because it invalidated an assumption in the first version of this PR. I originally claimed the bridge forwards `initialize` upstream untouched, so fixing the relayer alone would be enough. End-to-end probing against a deployed relayer disproved it: `/version` reported `build.commit` exactly equal to this branch HEAD, confirming the relayer was running the fix, yet the client handshake still returned `version 0.0.1` with no instructions. That value could only come from `buildLocalInitializeResult`.

So the relayer fix alone was correct but unreachable for every stdio user. Unit tests could not catch this: the relayer test drives the routes directly and the auth-required test runs a different mode. Only the real bridge path exposed it.

## Regression tests

`instructions` is an optional field on both responders, so dropping it fails no type check and breaks no other assertion. That is exactly how this went missing.

* `services/server/scripts/mcp/__tests__/instructions.test.ts` drives the real mounted routes over loopback HTTP.
* `packages/mcp/test/coldstart-init.test.mjs` now asserts the bridge's local initialize carries instructions and a non-stub version. Every other assertion in that file passed while the field was absent.

Confirmed the relayer test actually bites: removing the constructor argument turns it red, restoring it turns it green.

## Drive-by: real `serverInfo.version`

All three handshakes hardcoded `version: "0.0.1"`, matching no real package. Every user's handshake log reported the same fake version, which is part of why triaging this was guesswork, and the components were indistinguishable on the wire. All now report their real version via `packages/mcp/src/version.ts`.

Self-demonstrating: `packages/mcp` was bumped 0.0.5 to 0.0.9 while this branch was open and the handshake picked it up with no code change. It also turned the version into the diagnostic that located bug #2.

`MEMWAL_MCP_COMPATIBILITY_VERSION` stays pinned at 0.0.1 on purpose. It is the relayer-contract baseline compared against `minSupportedSdk.mcp` and checked by `scripts/check-compatibility-contract.mjs`, not a release version.

## Verification

* `npm test` services/server/scripts: 198 pass, 0 fail
* `pnpm test` packages/mcp: 18 pass, 0 fail
* `pnpm typecheck` packages/mcp: clean

Live bridge probe against the deployed dev relayer, before and after the bridge fix:

```
before:  "serverInfo":{"version":"0.0.1"}            instructions: MISSING
after:   "serverInfo":{"version":"0.0.9"}            instructions: PRESENT (1178 chars)
```

Client consumption is confirmed, not assumed: a Claude client's system prompt was observed gaining a `## memwal` section containing this text verbatim, in a session where it had previously been absent.

Deploy safety: replicated the Dockerfile's scripts layer in a temp dir and confirmed `../package.json` resolves to `/app/scripts/package.json` in the container layout.

## Verified end to end on Claude Desktop

All three behavioral cases pass against the patched build, with the relayer deployed from this branch.

| Case | Input | Result |
|---|---|---|
| Proactive save | `My default deploy region is ap-southeast-1` | `memwal_remember` fired unprompted and succeeded |
| Proactive recall | new chat: `What'''s my default deploy region?` | `memwal_recall` fired, returned the fact, answered correctly |
| Negative control | `Run the tests for me` | no `memwal_remember`, correctly silent |

The recall case is the one the ticket title is about. Request/response captured:

```
Request : {"query": "default deploy region deployment preference"}
Response: [score=0.679] User'''s default deploy region is ap-southeast-1 (Singapore)...
```

An earlier recall run looked like a pass and was not: the answer came from Claude'''s built-in memory while the memwal write had 401'''d, so Walrus held nothing. A direct `memwal_recall` confirmed zero matching entries. The rerun above uses a fact that exists only in Walrus, so it cannot be satisfied by any other store.

The 401 was stale environment-scoped credentials, cleared with `memwal_login`. Not a code issue.

### Scope of the claim

* **Proven:** mechanism on Claude Code, by direct observation of the system prompt gaining a `## memwal` section containing this text verbatim.
* **Proven:** behavior on Claude Desktop, all three cases above.
* **Not isolated:** Desktop never displayed an instructions block when asked to quote its system prompt, and the account carries a stored preference naming memwal. The outcome is correct; attribution between the two channels is not cleanly separable there.

## Also included: collapse duplicate recall results

Folded in from #682 (now closed) so both MCP fixes land together.

A `limit=5` recall on a real dev account returned the same fact five times, spending the entire retrieval budget on one preference and crowding out everything else the query should have surfaced:

```
1. [score=0.224] User prefers dark roast coffee
2. [score=0.224] User prefers dark roast coffee
... x5
```

Fixed on the **read** side, not the write side. Storing a fact repeatedly is legitimate: each remember is a distinct event with its own blob and timestamp, and there is deliberately no content-level uniqueness constraint (the only unique index is request idempotency on `remember_jobs (owner, idempotency_key)`, which guards retries). Deduping on write would silently discard a genuine re-statement.

`collapseDuplicates` folds results whose text matches after trimming, whitespace collapsing, and case folding, keeping the highest scoring copy since rows arrive ranked. Matching is exact-after-normalization rather than fuzzy, because merging facts that merely resemble each other would hide real information. The reply reports how many copies were folded rather than quietly returning fewer rows.

7 tests in `mcp/__tests__/recall-dedupe.test.ts`, including two safety cases: distinct facts must pass through untouched, and `"prefers dark roast coffee"` must not collapse with `"prefers dark roast coffee in the morning only"`.

This is relayer-side only (`services/server/scripts`, `"private": true`), so it does not affect the npm package version.

## Release

Manual bump to **0.0.10** following the flow from #678, not changesets: all 11 manifest/changelog/doc files moved together. Verified with the repo checker:

```
$ node scripts/verify-manual-sdk-release.mjs
MCP package 0.0.10: manifests and changelogs synchronized
```

## Not covered by this PR

* ChatGPT / Codex agent surface, untested.
* Codex setup docs are broken independently, since Codex moved into the ChatGPT app and local MCP servers now only apply to the agent surface. Separate docs task.


ducnmm:
LGTM! The fix via `initialize` instructions and the test coverage look solid. 🚀

Quick reminder: please add a changeset (`.changeset/`) or update the changelog for `@mysten-incubation/memwal-mcp` so CI can pick up the version bump & publish on release.

harrymove-ctrl:
Thanks @ducnmm! Changeset added in `ff0d615d`: [`.changeset/gh-681-proactive-instructions.md`](https://github.com/MystenLabs/MemWal/blob/fix/mcp-proactive-instructions/.changeset/gh-681-proactive-instructions.md), patch bump for `@mysten-incubation/memwal-mcp`.


So `release-mcp.yml` will bump 0.0.9 → 0.0.10 and publish on the next `main` release.

Scoped it to the MCP package only, since the published surface here is the bridge and auth-required `initialize` responders. The relayer-side change in `services/server/scripts` ships with the service rather than npm, so it is not part of the bump.

For the same reason, the follow-up #682 (collapse duplicate `memwal_recall` results) deliberately has no changeset: it only touches `services/server/scripts`, which is `"private": true` and never published. Shout if you would rather it had one anyway.

CI is green here, 15/15.

One heads-up: your LGTM came through as a plain comment rather than a submitted review, so GitHub still reports `REVIEW_REQUIRED` and the PR is `BLOCKED` from merging. If you are happy with it, a formal **Approve** on the Files changed tab would unblock it.

harrymove-ctrl:
Redone properly following the manual release flow, per @henry's note that this repo bumps the version + changelog rather than using changesets. Reference: #678 (`42bdd5de`, "bump MCP package to 0.0.9 & update changelog").

`2de2df60` bumps **0.0.9 → 0.0.10** across all 11 files that #678 touched:

* `packages/mcp/package.json`
* `.claude-plugin/marketplace.json`, `.cursor-plugin/marketplace.json`
* `packages/mcp/plugin/plugin.json` and the `.claude-plugin` / `.codex-plugin` / `.cursor-plugin` variants
* `packages/mcp/CHANGELOG.md` and `docs/mcp/changelog.mdx` (new `## 0.0.10` section, plus the "latest MCP package release is…" answer line)
* `packages/mcp/TESTING.md` (prerelease dist-tag references)
* `scripts/verify-manual-sdk-release.mjs` (expected version)

Verified with the repo's own checker rather than by eye:

```
$ node scripts/verify-manual-sdk-release.mjs
MCP package 0.0.10: manifests and changelogs synchronized
```

**I also removed the changeset I added in `ff0d615d`.** Keeping both would have double-bumped: `release-mcp.yml` runs `pnpm changeset version` on `main`, so it would have taken the manual 0.0.10 to 0.0.11 and rewritten the changelog section added here. One mechanism or the other, not both.

@ducnmm this supersedes the changeset from my earlier reply, so that request is now satisfied via the changelog route you offered as the alternative.

Still worth noting: the LGTM is a plain comment, so GitHub reports `REVIEW_REQUIRED` and the PR stays `BLOCKED`. A formal **Approve** would unblock it.

ducnmm (approved):

PR #645, opened 2026-08-14
fix(demos): use official Sui gRPC in browsers

ducnmm:
## Summary
- move Noter and Researcher browser clients to the official Sui gRPC transport
- configure `NEXT_PUBLIC_SUI_GRPC_URL` at image build time
- normalize object, dynamic-field, and transaction-change responses across gRPC and JSON-RPC fallback paths

## Production incident
The default mainnet JSON-RPC endpoint rejects browser CORS preflights. The official Sui gRPC-Web service supports the required CORS paths, so both demos now use `SuiGrpcClient` against `https://fullnode.mainnet.sui.io:443`.

## Validation
- Noter ESLint and TypeScript pass
- Researcher TypeScript passes
- live mainnet gRPC registry/dynamic-field lookup succeeds
- official gRPC-Web CORS preflights succeed
- production Noter bundle contains the official gRPC URL and no third-party RPC URL

PR #684, opened 2026-08-18
fix(noter): gRPC migration, auth race fixes, e2e suite + CI

harrymove-ctrl:
## Summary

Closes the "fix noter" half of WALM-355. Three fixes plus a 22-spec Playwright suite and CI job, mirroring #680's shape for researcher.

**No test coverage existed for noter at all** before this — the on-chain registration flow, delegate-key auth, and note CRUD were verified entirely by hand.

## Bugs fixed

1. **Google sign-in failing with `[enoki-login] Setup failed: TypeError: Failed to fetch`.** Sui's public JSON-RPC fullnodes were deprecated in 2026; `fullnode.testnet.sui.io` now answers every JSON-RPC call with `Method not found ... migrate to gRPC or GraphQL endpoints` and no CORS header, which Chrome reports as a generic CORS failure. `sui-providers.tsx` and `enoki-login-card.tsx` migrate to `SuiGrpcClient`, bypassing dapp-kit's `SuiClientProvider` (still hard-typed to `SuiJsonRpcClient` even in the latest published version) only where it was actually blocking things — `WalletProvider`'s wallet-standard connect/sign is untouched. Verified by decoding a live testnet `MemWalAccount` and matching its stored delegate public key, and round-tripping a registry dynamic-field lookup back to the same account id.

2. **An authenticated user got bounced back to `/` instead of landing on `/note`.** `sessionAtom` was `atomWithStorage(..., { getOnInit: false })`, so a fresh page load (the post-login `window.location.href` redirect, or a plain reload of `/note`) rendered once with `session=null` before the async post-mount hydration caught up — `useAuth`'s effect read that stale null and concluded "logged out" before the real session arrived. `getOnInit: true` is safe here since nothing renders `session` directly; every page branches on the separate `authAtom`, which always starts `isLoading:true` on both server and client.

3. **An invalid delegate key silently collapsed the login form with no error shown.** `connectEnoki`/`connectDelegateKey` flipped the global `authAtom.isLoading` for the duration of their mutation, and `app/page.tsx` renders `<AuthButtonGroup />` only while `!isAuthenticated && !isLoading` — unmounting the form mid-submit and taking its local `error` state with it before `setError()` could run. `isLoginPending` (from the mutation hooks themselves) already tracks per-call pending state for the submit button, so the global flag no longer needs to move during login.

## Test suite

22 specs across app shell, auth, note lifecycle, and the memory API contract, running against a real dev server and fresh Postgres.

**Mock seam for the delegate-account binding check** (already-merged, separate change): `connectDelegateKey` now calls `assertDelegateAccountBinding`, which reads the claimed account off-chain via gRPC and rejects any key not registered in its `delegate_keys` list — a random key/account pair can no longer reach an authenticated session. Mirroring #680's pattern, `delegate-account.ts` branches on `isTestEnvironment` (set by `PLAYWRIGHT=True`) and serves a fixture from `delegate-account.mock.ts` instead of the gRPC read, so the real validation logic still runs meaningfully.

Noter authenticates a fresh identity per test rather than reusing two shared identities (researcher's approach) — with `workers: 2` and ~15 login call sites, two fixed identities would have concurrent tests collide on each other's notes. The mock and the test fixture instead generate the same 24-entry deterministic pool independently, and the fixture hands out a never-yet-used entry per call, interleaved by Playwright's `parallelIndex` so two workers never claim the same one.

One non-obvious bug found while wiring the mock: `public_key` has to be base64, not hex — the binding check's parser tries `fromBase64()` first, and every 64-char hex string is *also* valid-but-wrong base64 (alphabet `0-9a-f` is a subset of base64's, length is always divisible by 4), so a hex value there silently decodes to the wrong bytes instead of ever matching.

No live-Walrus canary in this suite by design, same as #680 documents for researcher — `isTestEnvironment` is unconditionally true for every Playwright run, so even real, on-chain-registered credentials would fail the binding check's mock path at login. The real remember → recall round trip against production Walrus Memory stays a manual check.

## CI

`noter-e2e` mirrors `chatbot-e2e`'s shape (Postgres service container, cached Playwright browsers, report/trace upload on failure). `noter-checks` adds `tsc --noEmit` and a full `next build`.

## Verification

- 22/22 green, twice in a row, from a fully clean environment (no `.env.local`, fresh Postgres, no real credentials) — matching what CI will actually run
- `tsc --noEmit` clean, `next build` clean, both with placeholder-only env
- gRPC BCS decode verified against live testnet data (not just "compiles")
- Rebased onto current `dev`; this branch had gone stale mid-session and needed reconciling against unrelated auth-hardening work (`19434b2d`, `89f22073`) that landed in the meantime — the delegate-account binding check and the removal of the memory-route server-credential fallback both came from that, not from this PR

## Notes

- `services/contract/sources/account.move`'s source has grown migration/import fields (`AccountRegistry.migration_finalized`, `pinned_allowlist_root`, etc.; `MemWalAccount.admin_quarantined`, `legacy_account_id`, etc.) beyond what's currently deployed at the package this app points at. `lib/sui/account-bcs.ts`'s schemas intentionally decode only the leading fields this app reads and rely on the BCS parser stopping there — verified correct against the live, currently-deployed bytecode, but will need updating if/when that contract version is published to this package.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -202,6 +202,7 @@ jobs:
       NEXT_PUBLIC_APP_URL: http://localhost:3002
       PORT: "3002"
       NODE_ENV: test
+      PLAYWRIGHT: "True"
       # NEXT_PUBLIC_* Enoki/Sui vars are inlined at build time — placeholders
       # are fine here since the e2e suite authenticates via delegate key, not
       # the Google/Enoki popup flow (that needs a real OAuth session and stays
@@ -212,11 +213,10 @@ jobs:
       NEXT_PUBLIC_MEMWAL_PACKAGE_ID: "0xcf6ad755a1cdff7217865c796778fabe5aa399cb0cf2eba986f4b582047229c6"
       NEXT_PUBLIC_MEMWAL_REGISTRY_ID: "0xe80f2feec1c139616a86c9f71210152e2a7ca552b20841f2e192f99f75864437"
       NEXT_PUBLIC_MEMWAL_SERVER_URL: https://relayer.dev.memwal.ai
-      # MEMWAL_E2E_PRIVATE_KEY / MEMWAL_E2E_ACCOUNT_ID intentionally unset:
-      # the memory-write specs that need real relayer credentials self-skip
-      # via test.skip() when absent. CI validates auth + note CRUD
-      # deterministically; the real remember round-trip against the relayer
-      # stays a manual check, same caveat #680 documents for researcher.
+      # No live-Walrus canary in this suite: isTestEnvironment flips the
+      # binding check onto the fixture pool, so even a real on-chain key
+      # would fail at login. CI covers auth + note CRUD + the memory API
+      # contract; remember → recall against production stays a manual check.
 
     steps:
       - name: Checkout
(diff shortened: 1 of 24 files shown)


ducnmm (approved):
Reviewed the gRPC migration, the two auth races, and the e2e/CI work.

- Enoki/on-chain reads and `useSignTransaction` correctly bypass dapp-kit's JSON-RPC client.
- `sessionAtom` `getOnInit: true` and not flipping `authAtom.isLoading` during login fix the `/note` bounce and the disappearing error form.
- Fixture mock is fail-closed in production; hex-first `public_key` parse matches researcher.
- Noter Playwright E2E is green (22/22, no retries) after the install-timeout / cache / `PLAYWRIGHT=True` fixes.

Google/Enoki remains a manual check by design. Overlaps with #645 on the same two noter files — rebase that one after this lands.

Approved.

PR #688, opened 2026-08-18
fix(mcp): verify internal headers at the sidecar and fail closed on missing scope

### LINKED ISSUE #659: Security Vulnerability Report: Inbound `X-MemWal-Internal-Oauth-Scope` Header Injection in MCP Proxy Permitting Scope Escalation

### LINKED ISSUE #685: Harden sidecar: verify x-memwal-internal-* instead of trusting it (follow-up to #659)
Follow-up to #659. That issue lists two remediation items; #665 implements the first (drop inbound `x-memwal-internal-*` at the relayer) and this tracks the second:

> **Sidecar Verification**: Ensure the sidecar ignores `x-memwal-internal-oauth-scope` unless received over a secured loopback connection or validated against an active internal session token.

Filing this so the work stays tracked once #665 closes #659.

## Why it still matters after #665

#665 genuinely closes the reported vector — the repro in #659 no longer works. But it makes the relayer the *only* enforcement point for OAuth scope, and two properties turn that into a single point of failure.

**1. The sidecar trusts the header from any caller.**

`services/server/scripts/mcp/auth.ts:111` reads `x-memwal-internal-oauth-scope` with no check on where it came from. `/mcp/*` is mounted *before* `sharedSecretAuthMiddleware` (`services/server/scripts/sidecar/app.ts:46` and `:62`), so unlike every other sidecar route it has no shared-secret gate at all. `SIDECAR_HOST` is env-configurable (`services/server/scripts/sidecar/config.ts:290`, defaulting to `127.0.0.1`).

This is not exploitable in a default single-host deployment. It becomes reachable with a non-loopback bind, a split relayer/sidecar topology, SSRF, or any container-network foothold — in each case bypassing the #665 fix entirely.

**2. A missing scope header fails open.**

`services/server/scripts/mcp/tools/index.ts:19`:

```ts
const unrestricted = session.oauthScope === undefined;
```

An *absent* header grants `remember` / `remember_bulk` / `analyze` / `restore` rather than nothing. So any failure to attach it on the relayer side — a `HeaderValue::from_str` error in `apply_oauth_headers`, a future proxy path that forgets to call it, a rename — escalates a `memwal:read` token to full write instead of degrading safely.

## Suggested scope

- Accept `x-memwal-internal-*` at the sidecar only from a loopback peer, or authenticate it with the sidecar shared secret.
- Make the absent-header case fail closed. The legacy delegate-key path does legitimately need "no scope = unrestricted", so signal that with an explicit relayer-set marker (e.g. `x-memwal-internal-auth-mode: legacy`) rather than inferring it from absence.
- Add a positive-path test. Today `services/server/src/mcp_proxy.rs` only ever asserts the scope header is *dropped* on inbound; nothing asserts `apply_oauth_headers` actually injects it, so a future outbound sanitizer could silently strip it and leave every OAuth session unrestricted with a green suite.

nikola0x0:
Resolves #685 (WALM-372) — the second remediation item from #659, which #665 did not cover.

### Problem

The relayer decides what an MCP caller may do and tells the sidecar with `x-memwal-internal-oauth-scope`. Two defects made the relayer a single point of failure:

1. **The sidecar never verified the header's origin.** `/mcp/*` is mounted before `sharedSecretAuthMiddleware` (because `Authorization` already carries the end user's delegate key), so it had no shared-secret gate at all. #665 stops a client smuggling the header *through* the relayer; it does nothing about a caller reaching the sidecar directly.
2. **A missing scope header failed open.** `tools/index.ts` read `const unrestricted = session.oauthScope === undefined`, so an *absent* header granted full write. The intent was legitimate — legacy delegate-key callers have no OAuth scope and should get everything — but the sidecar could not tell "legacy login" from "the header went missing."

### Changes

**Relayer** — `apply_oauth_headers` → `apply_internal_headers`, now called on all three proxy handlers including the legacy `Passthrough` arm that previously set nothing. It states both internal headers explicitly:

| Header | Value |
| --- | --- |
| `x-memwal-internal-sidecar-token` | `SIDECAR_AUTH_TOKEN` |
| `x-memwal-internal-oauth-scope` | OAuth: the resolved grant. Legacy: `memwal:read memwal:write` |

Both use `insert` (overwrite), so a client-supplied value never survives. Header-build failures return 500 instead of being silently skipped — previously that left an authenticated request with no scope, which under the old fail-open meant full write.

**Sidecar** — new `verifyInternalOrigin()` (`timingSafeEqual`, length-guarded) runs first in `resolveAuth`, so an unverified caller gets 401 before any internal header is read. `registerTools` drops the `unrestricted` branch: absent or empty scope now registers no tools.

### No new environment variables

`SIDECAR_AUTH_TOKEN` is already mandatory (the sidecar `process.exit(1)`s without it), already documented as required in three places, already in both `.env.example` files and CI, and already inherited by the spawned sidecar from the relayer's environment. Nothing for operators to do.

### Verification

| | Result |
| --- | --- |
| Rust `mcp_proxy` | 14/14 (10 on `dev` + 4 new) |
| Rust full unit suite | 373 pass / 17 fail — identical to the `dev` baseline, no regressions |
| TypeScript | 216/216 (205 baseline + 11 new) |
| `rustfmt` | clean on `mcp_proxy.rs` |
| `clippy` | no new warnings |

### Notes for review

- **This closes #659's vector independently of #665.** Both headers are written with `insert`, so a forged inbound value is always overwritten. Based on `dev` rather than stacked on #665, so it can merge in either order — the two touch different functions in `mcp_proxy.rs`.
- **Zero tools means no `tools` capability**, not an empty list. The MCP SDK stops advertising `tools/list` entirely and returns `-32601`. Correct fail-closed behavior, asserted in `tool-scope.test.ts`.
- **`tool-annotations.test.ts` was relying on the fail-open** — it passed `{} as MemWalSession` and got all six tools. It now passes an explicit full scope. That test was green *because* of the bug.
- Three doc comments that claimed `/mcp/*` had no shared-secret gate are updated (`mcp_proxy.rs` module header, `sidecar/app.ts` mount block, `auth.ts` header contract).

### DIFF
--- a/services/server/scripts/mcp/__tests__/auth.test.ts
+++ b/services/server/scripts/mcp/__tests__/auth.test.ts
@@ -1,0 +1,47 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { McpAuthError, resolveAuth } from "../auth.js";
+
+const TOKEN = "test-sidecar-token-0123456789";
+const DELEGATE_KEY = "a".repeat(64);
+const ACCOUNT_ID = `0x${"b".repeat(64)}`;
+const SERVER_URL = "http://relayer.invalid";
+
+function mcpHeaders(overrides: Record<string, string | undefined> = {}): Headers {
+    const base: Record<string, string | undefined> = {
+        authorization: `Bearer ${DELEGATE_KEY}`,
+        "x-memwal-account-id": ACCOUNT_ID,
+        "x-memwal-internal-sidecar-token": TOKEN,
+        ...overrides,
+    };
+    const h = new Headers();
+    for (const [k, v] of Object.entries(base)) {
+        if (v !== undefined) h.set(k, v);
+    }
+    return h;
+}
+
+test("resolveAuth rejects a caller that cannot prove it is the relayer", async () => {
+    process.env.SIDECAR_AUTH_TOKEN = TOKEN;
+
+    await assert.rejects(
+        () => resolveAuth(mcpHeaders({ "x-memwal-internal-sidecar-token": undefined }), SERVER_URL),
+        (err: unknown) => {
+            assert.ok(err instanceof McpAuthError, `expected McpAuthError, got ${err}`);
+            assert.equal(err.status, 401);
+            return true;
+        }
+    );
+});
+
+test("resolveAuth reads the OAuth scope once the origin is verified", async () => {
+    process.env.SIDECAR_AUTH_TOKEN = TOKEN;
+
+    const { session } = await resolveAuth(
+        mcpHeaders({ "x-memwal-internal-oauth-scope": "memwal:read" }),
+        SERVER_URL
+    );
+
+    assert.equal(session.oauthScope, "memwal:read");
+});
--- a/services/server/scripts/mcp/__tests__/auth.test.ts
+++ b/services/server/scripts/mcp/__tests__/auth.test.ts
@@ -1,7 +1,7 @@
 import assert from "node:assert/strict";
 import test from "node:test";
 
-import { McpAuthError, resolveAuth } from "../auth.js";
+import { McpAuthError, normalizeScope, resolveAuth } from "../auth.js";
 
 const TOKEN = "test-sidecar-token-0123456789";
 const DELEGATE_KEY = "a".repeat(64);
@@ -44,4 +44,40 @@ test("resolveAuth reads the OAuth scope once the origin is verified", async () =
     );
 
     assert.equal(session.oauthScope, "memwal:read");
+});
+
+test("normalizeScope is order- and duplicate-insensitive", () => {
+    // The session key embeds this, so two spellings of the same grant must not
+    // open two distinct sessions.
+    assert.equal(
+        normalizeScope("memwal:write memwal:read"),
+        normalizeScope("memwal:read memwal:write")
+    );
+    assert.equal(normalizeScope("memwal:read memwal:read"), "memwal:read");
+    assert.equal(normalizeScope("  memwal:read   memwal:write  "), "memwal:read memwal:write");
+});
+
+test("normalizeScope collapses an absent or blank scope to the empty string", () => {
+    assert.equal(normalizeScope(undefined), "");
+    assert.equal(normalizeScope(""), "");
+    assert.equal(normalizeScope("   "), "");
+});
+
+test("sessionKey is stable across scope orderings but differs by granted scope", async () => {
+    process.env.SIDECAR_AUTH_TOKEN = TOKEN;
+
+    const keyFor = async (scope: string) =>
+        (await resolveAuth(mcpHeaders({ "x-memwal-internal-oauth-scope": scope }), SERVER_URL))
+            .sessionKey;
+
+    assert.equal(
+        await keyFor("memwal:write memwal:read"),
+        await keyFor("memwal:read memwal:write"),
+        "reordering the same grant must not fork the session"
+    );
+    assert.notEqual(
+        await keyFor("memwal:read memwal:write"),
+        await keyFor("memwal:read"),
+        "a narrower grant must not reuse a wider grant's session"
+    );
 });
--- a/services/server/scripts/mcp/__tests__/instructions.test.ts
+++ b/services/server/scripts/mcp/__tests__/instructions.test.ts
@@ -18,6 +18,7 @@
  * IMPORTANT: env vars MUST be set BEFORE importing `mountMcpRoutes` because
  * the rate limiter is constructed at module-load time.
  */
+process.env.SIDECAR_AUTH_TOKEN ??= "instructions-test-sidecar-token";
 process.env.MCP_MAX_TOTAL_SESSIONS = "100";
 process.env.MCP_MAX_SESSIONS_PER_IP = "100";
 process.env.MCP_MAX_NEW_SESSIONS_PER_IP_PER_MIN = "100";
@@ -81,6 +82,9 @@ async function initialize(): Promise<Record<string, any>> {
             authorization: `Bearer ${randomBytes(32).toString("hex")}`,
             "x-memwal-account-id": "0x" + randomBytes(32).toString("hex"),
             "x-forwarded-for": "203.0.113.7",
+            // Stands in for the relayer, which proves its origin with the
+            // sidecar shared secret before any internal header is honoured.
+            "x-memwal-internal-sidecar-token": process.env.SIDECAR_AUTH_TOKEN!,
         },
         body: JSON.stringify({
             jsonrpc: "2.0",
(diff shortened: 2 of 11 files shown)


ducnmm (changes requested):
Thanks for taking this over — the shared-secret header and explicit legacy scope are the right direction. Two blockers before merge:

1. **Bind the granted scope to the MCP session.** `resolveAuth()` still builds `sessionKey` from only `{accountId, delegatePubKey}`. Existing SSE/streamable servers keep the tools registered at session creation, so a later request with the same account/delegate but a missing or narrower scope still passes the session-key check and can drive the original write-capable session. This means missing scope does not actually fail closed after initialization. Include the normalized scope in `sessionKey`, and add a regression test that opens a full-scope session then reuses its session ID with no scope / `memwal:read` and expects 403.

2. **The stdio bridge integration test no longer exercises a successful upstream connection.** Its `/api` test mount bypasses the Rust proxy, so the bridge cannot send `x-memwal-internal-sidecar-token` or the explicit legacy scope. The test can still pass via the local/auth fallback because it only asserts `memwal_login`. Add proxy-simulation middleware on the `/api` router to inject the internal token + full scope, then assert a remote tool such as `memwal_recall` is present.

Everything else looks solid, and CI is green.

nikola0x0:
Both fixed in 422d4c67. Verified each one first — the second turned out to be worse than described.

### 1. Scope bound to the session key

Confirmed and exploitable, not theoretical. I'd previously assumed each grant mints its own delegate keypair, which would have made the session keys differ anyway. That's wrong: `routes/oauth.rs:640` calls `find_reusable_oauth_delegate(&account_id)`, keyed on **account_id alone**, and the comment at `:381` says reuse is deliberate. So two grants of differing scope for the same account share a delegate pubkey and produced an identical session key.

`sessionKey` now includes the normalized scope — deduplicated and sorted, so reordering the same grant doesn't fork a session. Two regression tests: opening with `memwal:read memwal:write` then reusing the session id with `memwal:read`, and with the scope header dropped entirely. Both returned 200 before the fix and 403 after. Plus unit tests pinning the normalization, mutation-checked (removing the sort, and removing the scope from the key, each fail a test).

### 2. Stdio bridge test

Correct, and the assertion couldn't be repaired by swapping in a remote tool name. `auth-required.ts` now advertises all seven tool names locally (`memwal_remember_bulk` and `memwal_health` included), so the bridge's fallback list is name-identical to the relayer's. I confirmed it empirically: forcing the origin check to fail vs. leaving it working produced byte-identical `tools/list` output.

There was a second cause underneath. Even with the proxy simulation added, the test still passed with the scope injection removed — because `tools/list` was being answered by the bridge's **coldstart path while the upstream SSE was still in flight**. The list never came from upstream at all, regardless of auth.

So the fix is three parts:
- The `/api` mount simulates the relayer's `apply_internal_headers`, injecting the sidecar token and the granted scope.
- The mount records response status at header-write time (not `finish` — an SSE response never finishes), and the test waits for `/mcp/sse` and asserts **200**. Before the fix this recorded `[{"path":"/mcp/sse","status":401}]`.
- `tools/list` is sent only **after** that 200, so it is answered upstream, and asserts the relayer-registered set.

Mutation-checked: removing either injected header now fails the test. Previously removing the scope changed nothing.

### Verification

TS sidecar 221/221, `packages/mcp` 18/18, Rust `mcp_proxy` 14/14, Rust full unit suite unchanged from the `dev` baseline (373 pass / 17 pre-existing failures), rustfmt clean.

ducnmm (approved):
LGTM! The session key now properly includes the normalized scope to prevent post-init privilege escalation, and the integration tests properly simulate the Rust proxy headers. All CI checks pass.

PR #690, opened 2026-08-18
fix(mcp): answer a tool call whose response never arrives

HoangDucBach:
## Ticket

[WALM-328](https://linear.app/mysten-labs/issue/WALM-328/issue-sse-bridge-idle-close-orphans-in-flight-remember-calls) — the MCP bridge leaves a tool call hanging when its response is lost.

## What changed?

All in `packages/mcp/src/bridge.ts`.

- **A deadline per in-flight request.** Each entry is stamped with `startedAt` and a sweeper closes out the expired ones. The stamp is never refreshed, not even by a replay, or a reconnect loop would keep pushing the deadline out and reintroduce the same hang.
- **A truthful error.** Expired calls go through the existing `failRequest`, which already adds the id to `closedOutIds` so a late genuine reply is dropped instead of becoming a second response for that id. `failRequest` gains optional wording: its "relayer unavailable" text would be wrong here, since the relayer is fine and only one reply went missing.
- **`MEMWAL_MCP_CALL_TIMEOUT_MS`**, derived in code from the slowest server-side tool deadline (`analyze`) plus headroom, so it moves with that tool instead of being pinned. The sweep period is capped at 5s so a long deadline is honoured closely.
- **Richer reconnect logging.** In-flight ids and methods, not just the count.

## Why is this needed?

The bridge had no deadline of its own. Its only timer watches for silence on the SSE stream, and the relayer's 3s keepalive keeps that stream looking healthy — so a reply lost while the connection is still up trips nothing: no watchdog, therefore no reconnect, therefore no replay. The request sat in `inFlight` forever and the caller could only report a bare timeout.

Two of the three options in the issue were already implemented: the sidecar keepalive (`services/server/scripts/mcp/index.ts:164-184`) and in-flight replay (`bridge.ts:832-857`). The third, an explicit orphaned-call error, is what this adds.

## Scope

The bridge's own timers. No transport, protocol or relayer change.

### Out of scope

- `notifications/progress`, which the MCP SDK honours via `resetTimeoutOnProgress` and would stop the client giving up at its own 60s default. The protocol's proper answer for long calls; worth its own ticket.
- Replay can double-write: `inFlight.set` runs before the send branch, so a call the relayer already executed is re-POSTed verbatim, and no idempotency key crosses the bridge.

## How was this tested?

- [x] Unit tests
- [x] Integration tests
- [ ] End-to-end tests
- [ ] Manual testing
- [ ] Not applicable

New `test/orphaned-call.test.mjs` reproduces the exact gap: the mock relayer keeps heartbeating throughout, answers `initialize`, then swallows the `memwal_remember` reply. It asserts the call is closed out with a retryable error, that `getSseGetCount() === 1` so the watchdog demonstrably did not rescue it, and that a late genuine reply is dropped rather than delivered twice.

Verified the test earns its place: with the sweeper disabled it fails with `timed out waiting for message`, the reported symptom. Suite is 19/19, `tsc` clean.

One caveat worth stating: a single unattributed failure appeared in one run and did not reproduce across 18 further runs on this branch or 8 on `origin/dev`. Most likely machine load, since these tests spawn child processes on real timers, but it is unexplained.

```bash
cd packages/mcp && npm test

# watch it catch the bug: disable the sweeper, then re-run
# (src/bridge.ts, in the orphanSweeper loop)
#   if (elapsedMs <= callTimeoutMs) continue;   ->   if (true) continue;
node --test test/orphaned-call.test.mjs
```

## Risks and dependencies

- **`SLOWEST_SERVER_TOOL_MS` duplicates a value from another package.** `analyze`'s 180s lives in `services/server/scripts/mcp/tools/`, which this package cannot import from. If that tool is granted longer and this is not raised with it, the bridge will start declaring healthy requests orphaned. The constant carries that warning.
- **A call is answered at most one sweep (5s) after its deadline**, so the effective timeout is 240–245s rather than exactly 240s.
- **The default only errs toward waiting longer.** A short override via the env var is the way to exercise the path; values below 1s are ignored.

## Author checklist

- [ ] CI is green.

### DIFF
--- a/docs/reference/environment-variables.md
+++ b/docs/reference/environment-variables.md
@@ -70,6 +70,7 @@ The stdio MCP package reads these environment variables directly. A CLI flag tak
 | `MEMWAL_CLIENT_LABEL` | `--label <text>` | `MCP Client` / `Walrus Memory MCP` | Friendly delegate-key label shown in the dashboard |
 | `MEMWAL_MCP_DEBUG` | none | `0` | Set to `1` for verbose stderr logging |
 | `MEMWAL_MCP_SSE_IDLE_MS` | none | `30000` | Maximum milliseconds of silence on the SSE stream before the bridge treats the session as dead and reconnects. Values below `500` are ignored and fall back to the default. Mainly for tests |
+| `MEMWAL_MCP_CALL_TIMEOUT_MS` | none | `240000` | Maximum milliseconds a single request might wait for its response before the bridge answers with a retryable error. Covers a reply lost while the stream itself stays healthy, which `MEMWAL_MCP_SSE_IDLE_MS` cannot detect. The default is derived in code from the slowest server-side tool deadline plus headroom, so it moves with that tool rather than being pinned here. Values below `1000` are ignored and fall back to the default |
 
 ## Self-hosted relayer
 
--- a/packages/mcp/src/bridge.ts
+++ b/packages/mcp/src/bridge.ts
@@ -185,6 +185,33 @@ function resolveSseIdleMs(): number {
     return n;
 }
 
+/** Longest deadline a server-side tool gives its own work (`analyze`). It
+ * lives in a package this one cannot import from, so raise this whenever that
+ * grows — otherwise the bridge declares healthy requests orphaned while the
+ * relayer is still working. */
+const SLOWEST_SERVER_TOOL_MS = 180_000;
+
+/** 240s as the constants stand. The headroom absorbs the relayer's own
+ * overhead, so expiry means the reply is lost rather than merely late. */
+const DEFAULT_CALL_TIMEOUT_MS = SLOWEST_SERVER_TOOL_MS + 60_000;
+
+/** An override below this is a mistake, not an intent. */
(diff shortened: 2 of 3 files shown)


jessiemongeon1:
## Style Guide Audit

HoangDucBach:
> ## Style Guide Audit
> Audited **1** file(s) against the [Sui Documentation Style Guide](https://docs.sui.io/references/contribute/style-guide).
> 
> **2** violation(s) found. All must be fixed before merge.
> 
> ### `docs/reference/environment-variables.md` (2 violation(s))
> _2 violation(s) (1 regex, 1 claude)_
> 
> * **Line 73** — Use "might" not "may"
>   
>   * Current: `may`
>   * Fix: `might`
> * **Line 73** — word-preference
>   
>   * Current: `a single request may wait`
>   * Fix: `a single request might wait`
> 
> _Automated audit using the Sui Documentation Style Guide._

Fixed

ducnmm (approved):
LGTM! The per-call deadline handling, sweeper with unref/clear, late response suppression in closedOutIds, and regression test in orphaned-call.test.mjs look solid.

PR #693, opened 2026-08-19
chore(release): include WALM-328 fix in MCP 0.0.10 changelog

ducnmm:
Updates MCP package changelog and docs changelog for v0.0.10 to include the WALM-328 (orphaned call deadline & sweeper) fix, and synchronizes the manual SDK release verification script.


docs/mcp/changelog.mdx
```diff
@@ -28,13 +28,14 @@ questions:
   - What changed in the MemWal MCP changelog?
   - When was the automatic memory plugin added to MemWal MCP?
 answer: >-
-  The latest MCP package release is 0.0.10. It sends proactive-usage instructions in the MCP initialize handshake so the model knows when to save and recall without being asked, which lazy tool loading had broken. Version 0.0.9 answered the initialize handshake locally and connected the relayer in the background so a slow cold start no longer trips the client's 30s timeout.
+  The latest MCP package release is 0.0.10. It closes out orphaned tool calls whose responses never arrive with a retryable error, and sends proactive-usage instructions in the MCP initialize handshake so the model knows when to save and recall without being asked. Version 0.0.9 answered the initialize handshake locally and connected the relayer in the background so a slow cold start no longer trips the client's 30s timeout.
```
nikola0x0: The rewrite drops "**which lazy tool loading had broken**" from the end. This field is flagged `has_answer: true` / "Needs answer summary for AI citation" — it gets quoted standalone, and that clause was the only thing explaining *why* the proactive-instructions fix exists. Nothing forced its removal; it fits:

> …so the model knows when to save and recall without being asked, which lazy tool loading had broken.

Minor: "orphaned tool calls whose responses never arrive with a retryable error" parses as the *response* arriving with an error. The retryable error is what the bridge sends back on expiry. The package changelog gets this right because "Answer … with" brackets it; "closes out … whose responses never arrive with" doesn't.

docs/mcp/changelog.mdx
```diff
@@ -28,13 +28,14 @@ questions:
...
 ---
 
 ## 0.0.10
 
 ### Fixed
 
+- Answer orphaned tool calls whose upstream response never arrives with a retryable error instead of hanging indefinitely. The bridge now tracks in-flight request start times and sweeps expired calls via a per-request deadline (`MEMWAL_MCP_CALL_TIMEOUT_MS`, default 240s), drops late responses to avoid duplicate replies, and enriches reconnect logs with pending request IDs and methods. (WALM-328, #690)
```
nikola0x0: Byte-identical to `packages/mcp/CHANGELOG.md:7` — apply the same two fixes here (late-reply attribution, `WALM-328`).

Info, not a request: nothing verifies these two stay in sync. `verify-manual-sdk-release.mjs` only asserts `content.includes("## <version>\n")` in each — heading presence, never entry parity. Fixing a typo in one file would leave npm and the docs site disagreeing, still green.

jessiemongeon1:
## Style Guide Audit

scripts/verify-manual-sdk-release.mjs
```diff
@@ -57,10 +57,10 @@ for (const release of releases) {
 const openClaw = JSON.parse(
     readFileSync("packages/openclaw-memory-memwal/package.json", "utf8"),
 );
-if (openClaw.version !== "0.0.5") {
-    throw new Error(`OpenClaw: expected unchanged 0.0.5, received ${openClaw.version}`);
+if (openClaw.version !== "0.0.6") {
+    throw new Error(`OpenClaw: expected unchanged 0.0.6, received ${openClaw.version}`);
 }
-console.log("OpenClaw 0.0.5: unchanged (no package changes in this promotion)");
+console.log("OpenClaw 0.0.6: unchanged (no package changes in this promotion)");
```
nikola0x0: Reproduced on `dev`: `OpenClaw: expected unchanged 0.0.5, received 0.0.6`. The guard caught a real bump — `1f494495` moved oc-memwal to 0.0.6 with a full `## 0.0.6` section in both its changelogs — so "unchanged (no package changes)" is now false.

OpenClaw is exactly the shape `releases` verifies. Promote it and drop lines 57–63:

```diff
+    {
+        name: "OpenClaw plugin",
+        version: "0.0.6",
+        manifests: [["packages/openclaw-memory-memwal/package.json", "version"]],
+        changelogs: [
+            "packages/openclaw-memory-memwal/CHANGELOG.md",
+            "docs/openclaw/changelog.mdx",
+        ],
+    },
 ];
```

Today nothing verifies OpenClaw's changelogs at all.

packages/mcp/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Answer orphaned tool calls whose upstream response never arrives with a retryable error instead of hanging indefinitely. The bridge now tracks in-flight request start times and sweeps expired calls via a per-request deadline (`MEMWAL_MCP_CALL_TIMEOUT_MS`, default 240s), drops late responses to avoid duplicate replies, and enriches reconnect logs with pending request IDs and methods. (WALM-328, #690)
```
nikola0x0: **"drops late responses to avoid duplicate replies" isn't new here.** `git show 4d6c87d7 -- packages/mcp/src/bridge.ts | grep "^+" | grep closedOutIds` returns nothing; #690's commit message says `failRequest` "**already**" did this. Suggest: "…and reuses the existing late-reply drop so an expired call cannot get a second response."

**`WALM-328` is the first internal ticket ID in any changelog here** — all seven cite GitHub numbers only. Linear is private; `(#690)` is enough.

docs/mcp/changelog.mdx
nikola0x0: Drops "**which lazy tool loading had broken**". This field is flagged for standalone AI citation, and that clause was the only thing saying *why* the fix exists. It fits:

> …without being asked, which lazy tool loading had broken.

docs/mcp/changelog.mdx
nikola0x0: Same two fixes as `packages/mcp/CHANGELOG.md:7`.

nikola0x0 (commented):
**Blocking on the OpenClaw pin.** Filing under `## 0.0.10` is correct — npm `latest` is still `0.0.8`, and `4d6c87d7` is the only commit touching `packages/mcp/` since the bump, so nothing else is missing.

## Majors (details inline)

- **`verify-manual-sdk-release.mjs:63`** — the guard was firing correctly; re-pinning to 0.0.6 means it can never fail for a real reason. Fix inline.
- **`CHANGELOG.md:7`** — the late-reply drop predates #690. `release-mcp.yml:96` builds the GitHub Release body from this section, so the error ships.

## Ops follow-up — needs an owner

Merging publishes `0.0.10-dev.2`, code-identical to `dev.1`: `release-mcp.yml:9` triggers on `packages/mcp/**`, which `CHANGELOG.md` matches, and the dev job publishes unconditionally. Wants `- '!packages/mcp/**/*.md'` on the filter, as its own ticket.

## Pre-existing, out of diff

`docs/mcp/changelog.mdx:31` claims 0.0.10 is the latest release; npm serves 0.0.8. Small separate issue.

ducnmm:
Self-review (GitHub will not let me approve my own PR):

Changelog-only note for MCP 0.0.10. Matches merged #690 / WALM-328:

- `MEMWAL_MCP_CALL_TIMEOUT_MS` default is 240s (`180s` analyze deadline + 60s headroom).
- Package is already `0.0.10`; the new bullet is under the right heading in both changelogs.
- Frontmatter `answer` now mentions the orphaned-call fix without dropping the 0.0.9 cold-start line.
- `verify-manual-sdk-release.mjs` OpenClaw pin `0.0.5` → `0.0.6` matches `packages/openclaw-memory-memwal/package.json`.

CI is green. Needs another reviewer to approve, or merge if you have admin bypass.

scripts/verify-manual-sdk-release.mjs
ducnmm: Done. OpenClaw 0.0.6 is now a first-class entry in `releases` (manifest + both changelogs), and the pin at lines 57–63 is gone. `node scripts/verify-manual-sdk-release.mjs` passes.

packages/mcp/CHANGELOG.md
ducnmm: Done. Reworded to reuse the existing late-reply drop, and dropped `WALM-328` so the note cites `#690` only. Same change in `docs/mcp/changelog.mdx`.

docs/mcp/changelog.mdx
ducnmm: Restored: `…without being asked, which lazy tool loading had broken.`

docs/mcp/changelog.mdx
ducnmm: Matched the CHANGELOG.md wording: `through` (style guide), existing late-reply drop, `#690` only.

ducnmm:
Addressed @nikola0x0 review in 9ce9d6fa:

- **OpenClaw pin** — promoted 0.0.6 into `releases` (manifest + both changelogs) and dropped the “unchanged” guard.
- **0.0.10 notes** — late-reply wording now says we reuse the existing drop; Linear ticket ID removed (`#690` only); `via` → `through`.
- **Docs answer** — restored “which lazy tool loading had broken”.

Ops follow-up (path filter so changelog-only bumps do not republish `0.0.10-dev.N`) left as its own ticket, as requested.

ducnmm:
#667 is now merged (`1bee35ff`). Added its 0.0.10 note here so the release body includes the `memwal_remember_bulk` default-namespace injection alongside #690.

packages/mcp/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Fixed
 
+- Answer orphaned tool calls whose upstream response never arrives with a retryable error instead of hanging indefinitely. The bridge now tracks in-flight request start times and sweeps expired calls through a per-request deadline (`MEMWAL_MCP_CALL_TIMEOUT_MS`, default 240s), reuses the existing late-reply drop so an expired call cannot get a second response, and enriches reconnect logs with pending request IDs and methods. (#690)
+- Inject the configured default namespace (`--namespace` / `MEMWAL_NAMESPACE`) into `memwal_remember_bulk` calls that omit one, so bulk facts land in the project namespace instead of the relayer fallback `default`. (#667)
```
nikola0x0, line 8: This one shipped in 0.0.9, not 0.0.10. `memwal_remember_bulk` was already in `NAMESPACE_TOOLS` as of `0d9bdd40` (ancestor of the 0.0.9 bump `42bdd5de`) — #667's only `bridge.ts` change is the doc comment above that set; the rest is docs + a regression test.

Not blocking, but `release-mcp.yml:96` builds the Release body from this section, and the same claim is in the `docs/mcp/changelog.mdx` frontmatter `answer` (the AI-citation field). Suggest dropping both.

nikola0x0 (approved):
Pre-approve — earlier four are genuinely fixed and `verify-manual-sdk-release.mjs` passes. One non-blocking note inline on the #667 bullet; fine to correct in a follow-up.

PR #699, opened 2026-08-19
fix(mcp): invalidate the live bridge session on memwal_logout

### LINKED ISSUE #616: [Bug][MCP] memwal_logout does not actually log out — bridge keeps serving requests with deleted credentials
### Summary

`memwal_logout` only deletes the local credentials *file*; the live MCP bridge session keeps working with the in-memory credentials and the still-open SSE connection. Any subsequent tool call (`memwal_recall`, `memwal_remember`, …) is still authorized and executed with the user's delegate key.

### Files

- `packages/mcp/src/bridge.ts:462-482` (`handleLocalLogout`)
- `packages/mcp/src/auth.ts:73-81` (`clearCreds`)

### Root cause

`memwal_logout` is intercepted locally and never forwarded to the relayer:

```ts
// bridge.ts:462-482
function handleLocalLogout(): { text: string; isError: boolean } {
    try {
        clearCreds();   // only unlinks ~/.memwal/credentials.json
        ...
    }
}
```

The `runBridge` closure keeps holding the live `creds` object and the open `sse` session (declared at `bridge.ts:498-524`, mutated only by `adoptCredentials` at `bridge.ts:629-671`). There is **no invalidation path tied to logout**: no `sse.abort()`, no credential-generation bump, and `adoptCredentials` is never called with empty credentials.

### Reproduction

1. Log in (so `~/.memwal/credentials.json` exists), start `memwal-mcp --relayer <relayer>`.
2. Call `tools/call` → `memwal_recall` → succeeds.
3. Call `tools/call` → `memwal_logout` → returns `"✅ Signed out. Credentials removed from ..."`.
4. Verify `~/.memwal/credentials.json` no longer exists.
5. Call `tools/call` → `memwal_recall` **again** → **still succeeds** (the relayer still receives the request with the user's delegate key).

I reproduced this with a mock relayer; the request after logout was served with `isError: false`.

### Impact

The tool's own description says "Remove the saved Walrus Memory credentials from this machine" — a user (or agent) calling logout believes access has been cut off, but every subsequent tool call continues to run with the user's delegate private key. `memwal_logout` is the only revocation path in bridge mode, and it is a no-op for the live session. A compromised or prompt-injected agent can even call `memwal_logout` then `memwal_remember` to steer blame away from itself while still writing with the real key.

### Suggested fix

On logout: abort the SSE stream, clear the in-memory `creds`/session state, and reject subsequent tool calls (or force a new login) until credentials are provided again.

nikola0x0:
Resolves #616 (WALM-358).

### Problem

`memwal_logout` deleted `~/.memwal/credentials.json` and stopped there. The bridge held the delegate key in memory and owned an open SSE session, so every later `memwal_recall` / `memwal_remember` was still forwarded and executed under the key the user had just removed.

Logout is the only revocation path in bridge mode, so a user — or a prompt-injected agent calling logout to look harmless — believed access was cut off while writes kept landing on the real account.

### Changes

Logout now tears the session down **before** it replies, so by the time the client is told it is signed out that is actually true: abort the SSE stream, fail anything still in flight, and refuse every memory tool locally until the next login. Only on success — if the credentials file could not be removed, the user is still signed in.

Three things were not visible from the issue:

| | |
| --- | --- |
| **The pump races the abort** | `reconnect()` only bailed on `stdinClosed`, and the pump calls `reconnect("server-pump-eof")` whenever a stream ends. Aborting the SSE alone would immediately re-authorize a fresh session with the same in-memory credentials. `loggedOut` is set *before* the abort; both connect loops honour it. |
| **An in-flight handshake still got published** | The loop guards only run between iterations, so a connect resolving after logout returned an open, authenticated stream holding the deleted key. Both paths now re-check after `openSseStream` resolves. |
| **Exiting the pump is worse than the bug** | Breaking out of the pump loop ends it permanently, so a later `memwal_login` reconnects a session with nothing draining it and the client hangs forever on its next tool call. The pump parks on `logoutPark` and resumes when a new session is published. |

`failRequest` takes an optional explicit message so in-flight calls purged by logout are not told the relayer is unavailable. The `memwal_logout` tool description and success text are corrected too — both previously claimed only that the credentials file was removed, which is what made the tool misleading in the first place.

### Verification

| | Result |
| --- | --- |
| `packages/mcp` suite | 22/22 (18 on `dev` + 4 new) |
| `tsc` | clean |
| Live check vs `relayer.dev.memwal.ai` | 12/12 |

Each new test was watched failing first:

| Test | Failed as |
| --- | --- |
| Post-logout call refused, never reaches the relayer | `isError: false` — it was served |
| SSE session closed on logout | 0 closes |
| Re-login restores service without a client restart | 10s timeout (client hangs) |
| Mid-handshake logout doesn't publish the session | session published |

The live script (`test/live/logout-invalidation.live.mjs`) is opt-in: the `.live.mjs` suffix is outside the `test/**/*.test.mjs` glob, and it exits 0 unless `MEMWAL_LIVE_HOME` / `MEMWAL_LIVE_RELAYER` are set. **CI is unaffected** — no network, no credentials. It covered the two gaps the mock cannot: that the real handshake matches the mock's model (8 upstream tools, a real recall), and that the teardown does something to a real connection.

### Notes for review

- **The live check deliberately does not assert the socket count drops to zero.** The relayer negotiates HTTP/2, so requests multiplex over one pooled TCP connection: `abort()` sends RST_STREAM, ending the stream while undici keeps the socket for reuse. That assertion would fail no matter how correct the teardown is. It asserts nothing *new* opens instead.
- **Logout still does not revoke the on-chain delegate key**, by design and as the tool states. This stops *this client* using it; revocation remains a dashboard action.
- **Not covered live:** re-login (needs a second wallet approval and on-chain registration) and mid-handshake logout (cannot be reliably timed against a sub-second real handshake). Both are green hermetically. Live coverage is dev only.
- Related but out of scope: WALM-331 (bridge doesn't hot-reload the delegate after re-login) builds on the same invalidation machinery and is still unassigned in Backlog.

### Rebase note

Rebased onto `dev` after #690 landed. Both touch `bridge.ts`:

- #690 generalized `failRequest` to take `{ toolText, errorMessage }`. This PR had added a narrower `explicitText` parameter for the same purpose, so I **dropped mine and adopted theirs** — it is the better shape, since it separates the tool-result text from the JSON-RPC error message. Logout now passes both.
- #690 changed `inFlight` to `Map<id, InFlightEntry>`. `invalidateSession()` iterates it, so it reads `entry.msg` now. Silent semantic conflict, not a textual one — worth a second pair of eyes.

Re-verified after the rebase: 23/23 hermetic (my 4 plus #690's new orphaned-call test), `tsc` clean, and 12/12 live against `relayer.dev.memwal.ai`.


packages/mcp/src/bridge.ts
```diff
@@ -1008,6 +1044,47 @@ export async function runBridge(
...
+        loggedOut = true;
+        logoutPark = new Promise<void>((resolve) => {
+            releaseLogoutPark = resolve;
+        });
+        try {
+            sse?.abort();
+        } catch {
+            /* already dead */
+        }
+        sse = null;
```
ducnmm, line 1167: Please also drop in-memory credentials here (`creds = null`, maybe bump `credentialGeneration`). Right now revocation is entirely the `loggedOut` flag; the delegate key stays in the closure. A missed check on a later path would forward with the key the user just deleted.

packages/mcp/src/bridge.ts
```diff
@@ -1184,6 +1284,20 @@ export async function runBridge(
                         });
                         return;
                     }
+                    // Signed out: refuse every remaining tool locally. `login`
+                    // and `logout` returned above, so anything reaching here
+                    // needs the delegate key the user deleted.
+                    if (loggedOut) {
```
ducnmm: This only covers `tools/call`. After `invalidateSession()`, `sse` is null and `reconnect()` no-ops while `loggedOut`, so a subsequent `tools/list` / ping / initialize can sit in `pendingForward` until login (or hang).

Fail-fast every non-login request while signed out, not just memory tool calls. Clients list tools after logout more often than they recall.

ducnmm (commented):
The logout races are handled well (`loggedOut` set first, mid-handshake discard, pump parks instead of exiting). Tests asserting the relayer counter — not just `isError` — are the right ones.

Two residual gaps below; neither blocks the core GH #616 fix but both are cheap and worth taking now.

ducnmm:
Also dropped the leftover .changeset/gh-571-clock-drift-error.md in 04183893 (along with this PR's own changeset in f97a183b).

MCP/SDK versions are bumped manually, so leaving a changeset on dev would double-bump when changeset version runs on main. README + config.json stay. No extra PR needed.

jessiemongeon1:
## Style Guide Audit

packages/mcp/src/bridge.ts
nikola0x0: Done — `creds` is now `MemWalCredentials | null` and `invalidateSession()` nulls it alongside a `credentialGeneration` bump.

Worth flagging what that surfaced. Making it nullable turned the audit into a compile error, and TS listed six paths that read the key. Five were already covered by `loggedOut`. Reconnect's replay loop was not: it awaits a POST per entry against an `inFlight` snapshot and an `openingCreds`, both captured before the logout, so `invalidateSession()` clearing `inFlight` cannot stop a snapshot already taken. A logout landing mid-replay kept POSTing the remaining entries under the deleted key, and only failed safe by accident once `sse` went null and `sse.postUrl` threw. It now breaks explicitly on `loggedOut || openingGeneration !== credentialGeneration`.

So the flag really was one missed check away — it just wasn't on the path either of us was looking at.

packages/mcp/src/bridge.ts
nikola0x0: Broadened — the guard moved out of the `tools/call` branch and now refuses everything via `failRequest`, which already picks the shape per method: tool-result text for `tools/call`, a JSON-RPC error for `ping` and friends, nothing for notifications.

Two corrections from tracing it:

**`tools/list` is already safe.** Its local branch fires on `sse === null`, which logout makes true, so it is answered from `LOCAL_TOOLS_LIST` and returns before reaching the guard. I added an assertion pinning that, including that the signed-out list still advertises `memwal_login`.

**`ping` was the real one, and it was worse than hang-until-login.** If the initial `connectInBackground` is still in flight when logout lands, it hits its own `loggedOut` break and drains `pendingForward` with `relayer unavailable: connection not established before shutdown` — wrong cause, and it makes the bug look answered. Once that connect has returned, nothing drains the queue and the request is never answered at all. The new test does a baseline recall first specifically to force the connect to complete; without it the test passes for the wrong reason.

Also stopped `initialize` arming a suppression while signed out — it returns after its local reply now, since nothing is forwarded to consume the arm.

ducnmm (approved):
Re-review after beb2cf42.

Both residual gaps are closed: `invalidateSession()` now nulls `creds` and bumps `credentialGeneration`, and the signed-out guard moved out of `tools/call` so every subsequent request fails locally via `failRequest` instead of hanging in `pendingForward`. Tests assert the relayer counter stays at zero after logout, plus ping fail-fast, local `tools/list`, SSE teardown, mid-handshake discard, and re-login restore.

Approved.

PR #700, opened 2026-08-19
fix(server): make storage quota admission atomic with insertion

### LINKED ISSUE #532: bug(server): concurrent `/api/remember/manual` requests bypass the per-owner storage quota [Security / Low]
`check_storage_quota` reads an owner's current usage under a PostgreSQL advisory lock, but the lock is released as soon as that read transaction commits — before the caller inserts anything. Any number of concurrent requests from the same authenticated owner acquire the lock one after another, all observe the same pre-insert usage total, and all pass a check that individually would have correctly rejected each of them once the others had already landed. A single account can use this, within its own legal per-minute request budget, to store more encrypted data than its configured quota permits, at the operator's Walrus/Sui expense.

## Affected code

`services/server/src/rate_limit.rs:624-663`:

```rust
// services/server/src/rate_limit.rs, lines 624-663
/// Uses PostgreSQL advisory lock per-owner to prevent
/// TOCTOU race where concurrent requests all pass quota check then
/// all write, collectively exceeding the limit.
pub async fn check_storage_quota(
    state: &AppState,
    owner: &str,
    additional_bytes: i64,
) -> Result<(), AppError> {
    let max_bytes = state.config.rate_limit.max_storage_bytes;

    // 0 or negative means unlimited
    if max_bytes <= 0 {
        return Ok(());
    }

    // Acquire a per-owner PostgreSQL advisory lock.
    // This serializes concurrent quota checks for the same owner,
    // preventing TOCTOU race conditions.
    let lock_key = stable_hash_i64(owner);

    // Use the combined method which uses an explicit transaction and pg_advisory_xact_lock
    let used = state.db.get_storage_used_with_lock(owner, lock_key).await?;
    let projected = used + additional_bytes;

    if projected > max_bytes {
        // ... (formats "Storage quota exceeded: {used}MB used of {max}MB allowed")
        return Err(AppError::QuotaExceeded(/* ... */));
    }

    Ok(())
}
```

`services/server/src/storage/db.rs:688-723` — the function the comment above refers to:

```rust
// services/server/src/storage/db.rs, lines 688-723
/// Acquire an advisory lock and get storage used within a single transaction.
///
/// Using `pg_advisory_lock` with a connection pool causes deadlocks
/// because it's session-level. We use `pg_advisory_xact_lock` inside an explicit
/// transaction so the lock is automatically released on commit/rollback.
pub async fn get_storage_used_with_lock(
    &self,
    owner: &str,
    lock_key: i64,
) -> Result<i64, AppError> {
    let mut tx = self.pool.begin().await?;

    sqlx::query("SELECT pg_advisory_xact_lock($1)")
        .bind(lock_key)
        .execute(&mut *tx)
        .await?;

    let row: (i64,) = sqlx::query_as(
        "SELECT COALESCE(SUM(blob_size_bytes)::BIGINT, 0) FROM vector_entries WHERE owner = $1",
    )
    .bind(owner)
    .fetch_one(&mut *tx)
    .await?;

    tx.commit().await?; // <-- lock released HERE, before the caller inserts anything

    Ok(row.0)
}
```

`pg_advisory_xact_lock` is transaction-scoped: it releases on `tx.commit()`, which happens inside `get_storage_used_with_lock` itself, one call earlier than the `INSERT INTO vector_entries` that `check_storage_quota`'s caller performs afterward. The lock protects the *read*; it does nothing for the interval between "lock released" and "row inserted," which is exactly where every concurrent caller lands.

## Reachability

`owner` in `check_storage_quota` is the caller's own authenticated identity (`AuthInfo.owner`), so this only lets an account exceed *its own* quota — not another user's. The exploitable surface is bounded by what a single account can legally send in one rate-limit window:

- `POST /api/remember/manual` body is capped by `PROTECTED_BODY_LIMIT_BYTES` (`auth.rs:21`) at 2 MiB, applied to all protected routes via `main.rs:1022`; this route has no per-route override (unlike `/api/remember/bulk`, which explicitly re-applies the same 2 MiB cap at `main.rs:1003`).
- `RememberManualRequest` (`types.rs:1139-1144`) is `{ encrypted_data: base64 String, vector: Vec<f32> (1536-dim), namespace: String }`. After the ~20 KiB JSON/vector overhead and base64 decoding (4 chars → 3 bytes), the maximum raw ciphertext one request can carry is **1,557,504 bytes (≈1.485 MiB)**.
- `/api/remember/manual` has `endpoint_weight` 3 (`rate_limit.rs:979`, test assertion); the default per-account budget is 60 weighted requests/minute (`rate_limit.rs:66`, `RateLimitConfig::default()`). `60 ÷ 3 = 20` — the maximum number of `/api/remember/manual` calls one account can legally fire inside a single rate-limit minute.

No jailbreak, forged signature, or abuse of any other control is needed — this is 20 ordinary, individually-authorized requests sent at once.

## Reproduction

Two independent levels, both against unmodified code, no mocking of the vulnerable functions themselves.

**1. Function-level, real Postgres, no HTTP layer.** The full `memwal-server` crate was built from this commit (`cargo build`, stable Rust 1.97.1) and a `#[tokio::test]` was added inside the existing `rate_limit.rs` test module, calling the two unmodified functions that contain the defect — `VectorDb::get_storage_used_with_lock` and `VectorDb::insert_vector` (`services/server/src/storage/db.rs`) — under real `tokio::spawn` concurrency, against a real PostgreSQL 17.10 instance with the real `pgvector` extension and the crate's own migrations 001–009.

Owner pre-filled to `max_storage_bytes − one_request − 1 KiB` (using the real default `max_storage_bytes = 1_073_741_824`, `rate_limit.rs:69`), so a single request would individually pass. 20 tokio tasks then ran the real code path concurrently:

```
$ cargo test --bin memwal-server quota_toctou -- --nocapture

running 1 test
CAND-MEMWAL-QUOTA-TOCTOU real-race result: passed_count=20 total_bytes=1103333376 max_storage_bytes=1073741824 overshoot_bytes=29591552
test rate_limit::tests::test_quota_toctou_real_race ... ok
```

All 20 tasks read the same pre-insert usage total and all passed; all 20 inserted via the real, unmodified `insert_vector`. Final holding: `1,103,333,376` bytes against a `1,073,741,824`-byte quota — a verified overshoot of `29,591,552` bytes (28.2 MiB), on the real default quota, matching the source-derived constants exactly.

**2. Full HTTP-layer, real Ed25519 auth, real on-chain verification — testnet and mainnet.** Self-hosted the same unmodified server (`cargo build`, this commit) and exercised the real `/api/remember/manual` route end-to-end: real signed HTTP requests, the real `auth.rs` middleware, and a real on-chain check that the signing key is a registered delegate on a real Sui `MemWalAccount` (via `SUI_GRPC_URL`), against a real Postgres instance. `RATE_LIMIT_STORAGE_BYTES` was lowered to `300000` only to make the race provable with a few hundred KB instead of gigabytes; `BENCHMARK_MODE=true` (a maintainer-documented flag — `services/server/.env.example`: *"Storage quota + auth still apply"*) swapped the storage backend to the documented `PlaintextEngine` so the PoC needed no real SEAL ciphertext; `check_storage_quota` and `get_storage_used_with_lock` ran unmodified either way, since `store_blob` dispatches to Walrus vs. Postgres only *after* the quota gate this report is about. No request was ever sent to the shared `relayer.memory.walrus.xyz` / `relayer-staging.memory.walrus.xyz` — both self-hosted runs used only the reporter's own wallet and Postgres.

Testnet run — 20 concurrent signed `POST /api/remember/manual`, 80,000 raw bytes each, against the lowered 300,000-byte quota:

```
delegate key verified onchain (gRPC), owner: 0xd889a19af39dbe575a64513b5a36d0b33046226277ae09c4af700ee60b8f53d0
  (repeated for all 20 requests — real signature verification + real on-chain delegate-key check per request)

req#2:  status=200  {"id":"3b56e055-...","blob_id":"3b56e055-...", ...}
req#7:  status=200  {"id":"38a5325b-...", ...}
...(6 more 200s)...
req#0:  status=402  {"error":"Storage quota exceeded: 0.6MB used of 0.3MB allowed"}
...(11 more 402s)...

Admitted (200): 8/20   Bytes accepted: 640,000 vs 300,000-byte quota
```

Confirmed independently against the same Postgres instance:

```sql
SELECT owner, count(*), sum(blob_size_bytes) FROM vector_entries WHERE owner = '0xd889...53d0' GROUP BY owner;
--  rows | total_bytes
--     8 |      640000
```

A follow-up burst against the same, now-over-quota account was correctly rejected 20/20 — confirming the gap is specifically the concurrent-burst-from-a-low-starting-point race, not a broken check.

Mainnet run — same setup, self-hosted against real Sui mainnet state, using a real, already-registered mainnet `MemWalAccount`: 20 concurrent requests, same 300,000-byte quota, admitted **5/20**, **400,000 bytes committed**, confirmed the same way via `SELECT SUM(blob_size_bytes)`. Real on-chain mainnet delegate-key verification confirmed live via the same `delegate key verified onchain (gRPC)` log line for every request. No real Sui gas or Walrus storage was spent for this run: the pool key (`SERVER_SUI_PRIVATE_KEY`) was a freshly generated, never-funded keypair, and `BENCHMARK_MODE` means no Walrus upload transaction is attempted.

## How this is exploited

1. Register an account through the normal (free) onboarding flow.
2. Fill it to `max_storage_bytes − 1.485 MiB` via ordinary `/api/remember/manual` calls.
3. Fire 20 concurrent `/api/remember/manual` requests, each with ~1.485 MiB of `encrypted_data` — exactly one minute's legal weighted-request budget for this endpoint, not an abuse of the rate limiter.
4. All 20 read the account's usage before any of them inserts, all pass, all upload to Walrus (server wallet pays gas and storage) and insert.
5. The account now holds ~1052 MiB against a 1024 MiB limit.

Because `check_storage_quota` keys on the caller's own identity, this does not expose or corrupt another user's data — it is a resource-limit bypass against the operator, not a confidentiality or integrity breach. Once an account is over quota, any single subsequent request is correctly blocked, so repeating the race on the *same* account requires its usage to drop back down first (e.g. via deletion). Nothing in `account::create_account` (`services/contract/sources/account.move`) gates account creation beyond "one account per Sui address, migration finalized" — no payment, no allowlist — so the primitive is repeatable across any number of distinct accounts an attacker operates, each costing only the ordinary Sui gas for `create_account` + `add_delegate_key`.

## Impact

Each successful race stores real ciphertext on Walrus and reserves real Sui storage epochs, both paid from the server's own wallet, not the caller's (`walrus_seal.rs` — the pool key pays gas and storage). This is an operator-cost / quota-integrity defect, not a data-exposure or fund-theft primitive.

**Cost, computed from cited constants, not assumed:**

- Verified overshoot on the real default quota (function-level reproduction above): `29,591,552` bytes = `0.029591552` GB (decimal).
- Default mainnet Walrus retention for a `remember_manual` write: `default_walrus_storage_epochs_for_network("mainnet") = 3` (`types.rs:39-41`) × the Walrus mainnet epoch length of 14 days (`docs.wal.app/docs/system-overview/available-networks`) = 42 days ≈ 1.4 months.
- Walrus's own fixed storage rate: **$0.023/GB/month** (`MystenLabs/walrus`, `docs/content/system-overview/storage-costs.mdx`: *"Walrus offers a fixed, USD-denominated storage cost of $0.023/GB/month"*).
- `0.029591552 GB × $0.023/GB/month × 1.4 months ≈ $0.00095` per single successful race on one account, at the real default 1 GiB quota.

That per-incident figure is small by design (28.2 MiB is a small overshoot relative to a 1 GiB quota), and it recurs per account, per opportunity to race, with no cap on the number of accounts one operator (a single Sui wallet, or a sponsor paying gas on users' behalf) can create. It is an unbounded-multiplier primitive against a fixed-rate cost, not a one-time loss — the ceiling is set by how many accounts an attacker is willing to create and race, not by anything in this code path.

## Prior art checked (re-verified live via GitHub search API, 2026-08-04)

Three PRs touch this exact code area and are not duplicates of this defect:

- [#66](https://github.com/MystenLabs/MemWal/pull/66) — `feat(relayer): multi-layer rate limiting with Redis, storage quota, a…` — introduced the storage-quota tracking mechanism itself (merged 2026-03-31).
- [#72](https://github.com/MystenLabs/MemWal/pull/72) — `fix(server): restore quota accounting and enoki fallback behavior` — restored quota accounting after a regression (merged 2026-04-01).
- [#97](https://github.com/MystenLabs/MemWal/pull/97) — `fix(security): MEM-18 rate limiter fail-closed on Redis error` — added the `pg_advisory_xact_lock` used today, aimed at Redis fail-closed behavior; its own surrounding code comments describe it as closing a TOCTOU, which is the specific claim this report shows is incorrect — the lock covers the read, not the read-to-insert interval (merged 2026-04-14).

No open or closed issue describes the check/insert gap itself. A live full-text search of this repository's issues and PRs for `check_storage_quota` and `advisory_xact_lock` returns zero matches. This is not a duplicate of #296 (docs request to *document* existing quotas), #424 (feature request for a quota-query API), or #492 (a different bug: an unauthenticated request forcing an unbounded on-chain registry scan) — none of those describe or fix the check-then-insert race in `check_storage_quota`.

## Suggested fix

Make the check-then-insert sequence atomic instead of holding a lock only around the read. Two viable approaches:

1. Perform the `SUM` check and the `INSERT INTO vector_entries` inside the **same** transaction, holding the same advisory lock for both statements, so no other request can read a stale total between them.
2. Replace the SUM-on-read pattern with a reserved-usage counter: increment it (inside the lock) at the moment the request is admitted, and decrement it on failure/rollback, rather than recomputing `SUM(blob_size_bytes)` from scratch on every check.

Either removes the gap between "lock released" and "row inserted" that this report demonstrates is currently exploitable, on both testnet and mainnet.

**Environment:** `MystenLabs/MemWal`, `dev` at `e7fd5c5a0d1849a65a93786f05baa98aed34542a`. Body-limit, weight, and rate constants verified directly from `auth.rs`, `types.rs`, and `rate_limit.rs` at this commit. `cargo build` succeeds on this commit with stable Rust 1.97.1. Function-level reproduction: `cargo test --bin memwal-server quota_toctou` against real PostgreSQL 17.10 + `pgvector`, exercising the unmodified `get_storage_used_with_lock` / `insert_vector`. HTTP-layer reproduction: self-hosted, unmodified server against real Sui testnet and real Sui mainnet, real Ed25519-signed requests through the unmodified `auth.rs` middleware, real on-chain delegate-key verification via Sui gRPC, real Postgres. Never sent a request to the shared `relayer.memory.walrus.xyz` / `relayer-staging.memory.walrus.xyz`.

harrymove-ctrl:
Closes WALM-359. Fixes #532.

## The defect

`check_storage_quota` read `SUM(blob_size_bytes)` under a per-owner advisory lock, but `get_storage_used_with_lock` **committed** that transaction, releasing the lock, before the caller inserted anything. `pg_advisory_xact_lock` is transaction scoped, so the lock protected the read and nothing else. Concurrent requests from one owner all read the same pre-insert total, all passed a check that would have rejected them individually, and all wrote.

The comment claiming this closed a TOCTOU race was wrong, and is corrected here.

## Reproduced before fixing

Against the unmodified code on real Postgres 17.9 + pgvector, pool sized like production (`max_connections(10)`), 20 concurrent requests (the legal per-minute budget for `/api/remember/manual` at weight 3), real 1 GiB default quota, seeded so exactly one request fits:

| Scenario | Admitted | Overshoot |
|---|---|---|
| Inline path, no synchronization | 14 to 15 / 20 | +19 to +21 MiB |
| Enqueued path, 150 ms upload gap | 13 to 20 / 20 | +18 to +28 MiB |
| Staggered arrivals 25 ms apart | 7 / 20 | +8.9 MiB |
| Control: atomic check+insert | 1 / 20 | none |

Stable across 6 runs. No barrier or artificial synchronization: requests arriving 25 ms apart still overshoot, and the admitted count matches the predicted `floor(gap / stagger) + 1`. The control row differs only in holding the lock across both statements, which is what makes the harness credible rather than merely loud.

## The fix

Migration `014` adds `storage_quota_reservations`. `try_reserve_storage` holds the advisory lock across **both** the usage read and the reservation insert, so concurrent callers for one owner serialize and each sees the previous one's commitment.

This mirrors a pattern already shipped in this repo: `security_delete_store::claim_blobs` opens one transaction and calls `lock_and_check_cap` then `insert_batch` inside it. The difference was purely the signature, taking `&mut Transaction` rather than owning and committing one.

Reserved bytes count alongside committed rows at admission, and are released when the row lands or the write terminally fails.

**Why a reservation and not just one transaction.** A single transaction closes the inline paths, but three of the six write paths return from the request handler and insert later in the wallet worker, after a Walrus upload that `AppState` documents as taking up to five minutes. You cannot hold a Postgres transaction across that. The reservation covers both groups with one mechanism.

## Failure modes, deliberately

* **Lost release** (process dies between admission and insertion): reservations carry a 15 minute TTL and are swept at the next admission. The cost is temporary headroom, never permanently stranded quota. Covered by `expired_reservation_does_not_strand_quota`.
* **Retryable job failure**: keeps its reservation, because the next attempt still needs those bytes. Only terminal failures release.
* **Release fails**: logged and swallowed, never converted into a failed write, and never allowed to mask the original error on a failure path.
* **Rejected bulk batch**: reserves nothing, so a partial batch is never silently admitted. Covered by `bulk_admission_is_all_or_nothing`.

## Scope

Six call sites, not the three originally reported. `remember()` at `routes/remember.rs:152` was missing from the issue's list and has the same defect.

Inline (handler owns the reservation): `remember_manual`, analyze benchmark branch.
Enqueued (reservation keyed by job id, released by the wallet worker): `remember()`, `remember_bulk`, analyze main path.

Keying enqueued reservations by `remember_job_id` avoids threading a reservation handle through `WalletOperation`, since that id already reaches the insert site.

Releases are wired at every terminal outcome: success in `insert_vector_and_mark_remember_done`, terminal wallet failure in `update_remember_job_after_wallet_error`, `mark_remember_job_failed` in `jobs.rs`, and the two prepare-stage handlers in `routes/remember.rs`.

## Behavior preserved

* `402` status and the exact `"Storage quota exceeded: {used}MB used of {max}MB allowed"` message shape.
* Bulk admission stays all-or-nothing.
* `max_storage_bytes <= 0` still means unlimited, and now skips reserving entirely rather than accumulating rows nothing reads.
* Rate limiting is untouched.

## Verification

* `cargo test --bin memwal-server`: **552 passed, 0 failed**.
* 5 new real-Postgres regression tests, including the 20-way burst with a delayed insert, which is the case that previously reached 20/20 bypass and now admits exactly 1.
* `cargo clippy`: 0 errors.
* `cargo fmt`: clean on all touched files. Note `dev` has pre-existing formatting drift in `auth.rs`, `mcp_proxy.rs`, `routes/sponsor.rs`, and `types.rs`; I reverted those rather than bundling unrelated churn.
* Compiles clean under `cargo +1.88 build --release`, matching the Dockerfile's pinned toolchain.

**Not verified:** the Docker image build. It fails on this arm64 machine at the dependency stage (`qemu: uncaught target signal 11`, amd64 `rustc` segfaulting under emulation), and fails identically on unmodified `dev`, so it is environmental. That step consumes only `Cargo.toml`/`Cargo.lock`, neither of which this PR touches. CI should confirm.

## Deployment note

Migration `014` is additive (new table plus index, `IF NOT EXISTS`), with no change to existing tables and no backfill. Rollback is safe: reverting the code leaves an unused table.

No `RATE_LIMIT_STORAGE_BYTES` override exists in Railway `production`, `dev`, or `staging`, so all three run the 1 GiB default and are affected.

### DIFF
--- a/services/server/src/rate_limit.rs
+++ b/services/server/src/rate_limit.rs
@@ -613,46 +613,68 @@ pub async fn rate_limit_middleware(
 // Storage Quota Check (called from routes, not middleware)
 // ============================================================
 
-/// Check if a user has enough storage quota for a new blob.
+/// How long a storage reservation survives without an explicit release.
 ///
-/// Storage tracking still uses PostgreSQL (it's per-row in vector_entries).
-/// Returns `Ok(())` if within quota, `Err(AppError::QuotaExceeded)` if not.
+/// Only a backstop for a process that dies between admission and insertion.
+/// Sized above the wallet-job upload ceiling described in `AppState`
+/// (`wallet_lock_pool`, up to ~5 minutes) so a slow-but-healthy upload is
+/// never charged twice, while a lost reservation still frees itself.
+const STORAGE_RESERVATION_TTL_SECONDS: i64 = 900;
+
+/// Reserve storage quota for one or more pending writes, atomically.
 ///
-/// Uses PostgreSQL advisory lock per-owner to prevent
-/// TOCTOU race where concurrent requests all pass quota check then
-/// all write, collectively exceeding the limit.
-pub async fn check_storage_quota(
+/// Replaces the former `check_storage_quota`, whose advisory lock covered only
+/// the usage read and was released before the caller inserted, letting
+/// concurrent requests from one owner collectively exceed the quota (GH #532).
+///
+/// Each entry in `items` is `(reservation_id, bytes)`. Admission is
+/// all-or-nothing, preserving the previous behavior where a bulk request
+/// either fully passes or is rejected with a single error.
+///
+/// The caller MUST release each reservation once the row lands or the write
+/// terminally fails, via `release_storage_quota`. Reservations expire on their
+/// own after `STORAGE_RESERVATION_TTL_SECONDS` as a backstop, so a missed
+/// release costs an owner temporary headroom rather than permanent quota.
+pub async fn reserve_storage_quota(
     state: &AppState,
     owner: &str,
-    additional_bytes: i64,
+    items: &[(String, i64)],
 ) -> Result<(), AppError> {
     let max_bytes = state.config.rate_limit.max_storage_bytes;
 
-    // 0 or negative means unlimited
+    // 0 or negative means unlimited. Skip reserving entirely so unlimited
+    // deployments do not accumulate rows they will never read.
     if max_bytes <= 0 {
         return Ok(());
     }
 
-    // Acquire a per-owner PostgreSQL advisory lock.
-    // This serializes concurrent quota checks for the same owner,
-    // preventing TOCTOU race conditions.
-    // We use a stable hash of the owner string as the lock key.
+    // Per-owner advisory lock key. Held across the usage read AND the
+    // reservation insert inside `try_reserve_storage`.
     let lock_key = stable_hash_i64(owner);
 
-    // Use the combined method which uses an explicit transaction and pg_advisory_xact_lock
-    let used = state.db.get_storage_used_with_lock(owner, lock_key).await?;
-    let projected = used + additional_bytes;
+    let outcome = state
+        .db
+        .try_reserve_storage(
+            owner,
+            lock_key,
+            items,
+            max_bytes,
+            STORAGE_RESERVATION_TTL_SECONDS,
+        )
+        .await?;
 
-    if projected > max_bytes {
+    if let Err(used) = outcome {
+        let additional: i64 = items.iter().map(|(_, bytes)| *bytes).sum();
         let used_mb = used as f64 / 1_048_576.0;
         let max_mb = max_bytes as f64 / 1_048_576.0;
         tracing::warn!(
             "storage quota exceeded: owner={} used={:.1}MB + {:.1}MB > max={:.1}MB",
             owner,
             used_mb,
-            additional_bytes as f64 / 1_048_576.0,
+            additional as f64 / 1_048_576.0,
             max_mb
         );
+        // Message shape unchanged so existing clients keep parsing it.
         return Err(AppError::QuotaExceeded(format!(
             "Storage quota exceeded: {:.1}MB used of {:.1}MB allowed",
             used_mb, max_mb
@@ -660,6 +682,39 @@ pub async fn check_storage_quota(
     }
 
     Ok(())
+}
+
+/// Convenience wrapper for the single-write paths.
+pub async fn reserve_storage_quota_one(
+    state: &AppState,
+    owner: &str,
+    reservation_id: &str,
+    additional_bytes: i64,
+) -> Result<(), AppError> {
+    reserve_storage_quota(
+        state,
+        owner,
+        &[(reservation_id.to_string(), additional_bytes)],
+    )
+    .await
+}
+
+/// Release a reservation after the row has landed or the write has failed.
+///
+/// Never propagates an error: a failed release is recoverable (the TTL
+/// backstop reclaims it) and must not turn a successful write into a failed
+/// request, nor mask the original error on a failure path.
+pub async fn release_storage_quota(state: &AppState, reservation_id: &str) {
+    if state.config.rate_limit.max_storage_bytes <= 0 {
+        return;
+    }
+    if let Err(e) = state.db.release_storage_reservation(reservation_id).await {
+        tracing::warn!(
+            "failed to release storage reservation id={} err={} (will expire via TTL)",
+            reservation_id,
+            e
+        );
+    }
 }
 
 /// Compute a stable i64 hash of a string for use as PG advisory lock key.
--- a/services/server/src/routes/analyze.rs
+++ b/services/server/src/routes/analyze.rs
@@ -372,8 +372,16 @@ pub async fn analyze(
     if state.config.benchmark_mode {
         // Quota check on plaintext byte length (benchmark mode has no
         // ciphertext — plaintext is the closest analog).
-        let total_plaintext_bytes: i64 = facts.iter().map(|f| f.text.len() as i64).sum();
-        rate_limit::check_storage_quota(&state, owner, total_plaintext_bytes).await?;
+        let quota_items: Vec<(String, i64)> = facts
+            .iter()
+            .map(|f| {
+                (
+                    format!("bench-{}", uuid::Uuid::new_v4()),
+                    f.text.len() as i64,
+                )
+            })
+            .collect();
+        rate_limit::reserve_storage_quota(&state, owner, &quota_items).await?;
 
         let store_tasks: Vec<_> = facts
             .iter()
@@ -413,6 +421,12 @@ pub async fn analyze(
             .collect();
 
         let store_results = collect_bounded_results(store_tasks, ANALYZE_CONCURRENCY).await;
+        // Release before the `?` below: rows that landed now count on their own,
+        // and a partial failure must not leave the owner charged for facts that
+        // were never stored.
+        for (reservation_id, _) in &quota_items {
+            rate_limit::release_storage_quota(&state, reservation_id).await;
+        }
         let mut stored_facts: Vec<AnalyzeAcceptedFact> = Vec::with_capacity(store_results.len());
         for r in store_results {
             stored_facts.push(r?);
@@ -485,24 +499,33 @@ pub async fn analyze(
 
     let prep_results = collect_bounded_results(prep_tasks, ANALYZE_CONCURRENCY).await;
 
-    // Quota check on total ciphertext size
+    // Storage-quota admission on total ciphertext size
     let mut prepared: Vec<(String, f32, Vec<f32>, Vec<u8>)> =
         Vec::with_capacity(prep_results.len());
-    let mut total_encrypted_bytes: i64 = 0;
     for r in prep_results {
         let (fact_text, importance, vector, encrypted) = r?;
-        total_encrypted_bytes += encrypted.len() as i64;
         prepared.push((fact_text, importance, vector, encrypted));
     }
-    rate_limit::check_storage_quota(&state, owner, total_encrypted_bytes).await?;
+    // Job ids are generated up front so each reservation is keyed by the job
+    // whose wallet worker will release it after the row lands.
+    let pending_job_ids: Vec<String> = prepared
+        .iter()
+        .map(|_| uuid::Uuid::new_v4().to_string())
+        .collect();
+    let quota_items: Vec<(String, i64)> = prepared
+        .iter()
+        .zip(pending_job_ids.iter())
+        .map(|((_, _, _, encrypted), job_id)| (job_id.clone(), encrypted.len() as i64))
+        .collect();
+    rate_limit::reserve_storage_quota(&state, owner, &quota_items).await?;
 
     // Step 3: For each prepared fact — insert remember_jobs row + enqueue WalletJob.
     // Round-robin across wallet pool so facts upload in parallel.
     let mut job_ids: Vec<String> = Vec::with_capacity(prepared.len());
     let mut accepted_facts: Vec<AnalyzeAcceptedFact> = Vec::with_capacity(prepared.len());
-    for (fact_text, importance, vector, encrypted) in prepared {
-        let job_id = uuid::Uuid::new_v4().to_string();
-
+    for ((fact_text, importance, vector, encrypted), job_id) in
+        prepared.into_iter().zip(pending_job_ids.into_iter())
+    {
         // Insert status row
         sqlx::query(
             "INSERT INTO remember_jobs (id, owner, namespace, status) VALUES ($1, $2, $3, 'pending')",
--- a/services/server/src/routes/remember.rs
+++ b/services/server/src/routes/remember.rs
@@ -81,6 +81,11 @@ async fn mark_remember_job_failed(
     .bind(prepare_claim_token)
     .execute(state.db.pool())
     .await;
+
+    // Preparation failed after admission, so the write will never land.
+    // Release the reservation now rather than leaving the owner charged for it
+    // until the TTL backstop expires.
+    rate_limit::release_storage_quota(state, job_id).await;
 }
 
 async fn mark_remember_jobs_failed(state: &AppState, job_ids: &[String], msg: &str) {
@@ -94,6 +99,11 @@ async fn mark_remember_jobs_failed(state: &AppState, job_ids: &[String], msg: &s
     .bind(job_ids)
     .execute(state.db.pool())
     .await;
+
+    // Same as the single-job path: nothing will be inserted for these ids.
+    for job_id in job_ids {
+        rate_limit::release_storage_quota(state, job_id).await;
+    }
 }
 
 // ============================================================
@@ -149,7 +159,15 @@ fn spawn_prepare_remember_job(
                 let vector = vector_result?;
                 let encrypted = encrypted_result?;
 
-                rate_limit::check_storage_quota(&state, &owner, encrypted.len() as i64).await?;
+                // Reservation is keyed by job_id: the wallet worker releases it
+                // once the row lands or the job terminally fails.
+                rate_limit::reserve_storage_quota_one(
+                    &state,
+                    &owner,
+                    &job_id,
+                    encrypted.len() as i64,
+                )
+                .await?;
 
                 let wallet_index = state.key_pool.next_index().ok_or_else(|| {
                     AppError::Internal(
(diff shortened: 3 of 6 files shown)


ducnmm (commented):
The reservation + `pg_advisory_xact_lock` across read-and-insert is the right fix for GH #532, and the burst tests (inline vs delayed insert) are the regression we needed.

Two leak paths where a reservation is taken and then not released except by the 15-minute TTL — that will look like a stuck quota to the owner. Details inline.

services/server/src/routes/remember.rs
```diff
@@ -149,7 +159,15 @@ fn spawn_prepare_remember_job(
...
-                rate_limit::check_storage_quota(&state, &owner, encrypted.len() as i64).await?;
+                // Reservation is keyed by job_id: the wallet worker releases it
+                // once the row lands or the job terminally fails.
+                rate_limit::reserve_storage_quota_one(
+                    &state,
+                    &owner,
+                    &job_id,
+                    encrypted.len() as i64,
+                )
+                .await?;
```
ducnmm, line 170: Lost fenced claim a few lines below returns `Ok(())` after this reserve (`if persisted.rows_affected() != 1 { return Ok(()); }`), so `mark_remember_job_failed` never runs and the reservation sits until TTL.

Release on that path (or return `Err` so the existing failure path releases). Same for any other early-`Ok` after admission.

services/server/src/routes/analyze.rs
```diff
@@ -485,24 +499,33 @@ pub async fn analyze(
...
+    let pending_job_ids: Vec<String> = prepared
+        .iter()
+        .map(|_| uuid::Uuid::new_v4().to_string())
+        .collect();
+    let quota_items: Vec<(String, i64)> = prepared
+        .iter()
+        .zip(pending_job_ids.iter())
+        .map(|((_, _, _, encrypted), job_id)| (job_id.clone(), encrypted.len() as i64))
+        .collect();
+    rate_limit::reserve_storage_quota(&state, owner, &quota_items).await?;
```
ducnmm, line 520: Admission is all-or-nothing, then the loop below inserts/enqueues one-by-one. If job 2's INSERT or `enqueue_wallet_job` fails, we return `Err` without releasing the remaining reservations (or the ones whose jobs were never created).

On that error path, release every id in `quota_items` that will not be handled by a wallet worker — otherwise the owner is charged for facts that were never stored, until TTL.

(The benchmark branch above already releases before `?`; this non-bench path needs the same.)

services/server/src/rate_limit.rs
```diff
@@ -613,46 +613,68 @@ pub async fn rate_limit_middleware(
...
-/// Check if a user has enough storage quota for a new blob.
+/// How long a storage reservation survives without an explicit release.
 ///
-/// Storage tracking still uses PostgreSQL (it's per-row in vector_entries).
-/// Returns `Ok(())` if within quota, `Err(AppError::QuotaExceeded)` if not.
+/// Only a backstop for a process that dies between admission and insertion.
+/// Sized above the wallet-job upload ceiling described in `AppState`
+/// (`wallet_lock_pool`, up to ~5 minutes) so a slow-but-healthy upload is
+/// never charged twice, while a lost reservation still frees itself.
+const STORAGE_RESERVATION_TTL_SECONDS: i64 = 900;
```
ducnmm, line 622: **TTL is shorter than a healthy congested job.**

900s is sized against the ~5 min `wallet_lock_pool` upload, but `MAX_CONGESTION_REQUEUES = 6` with backoff `30+60+120+240+480+600s` ≈ **25.5 min**. Retryable failures keep the reservation (`jobs.rs` only releases when `aborts_retries()`) and never refresh `expires_at`.

Sequence: admit → reservation expires at T+15m → congestion requeues keep the job alive → next `try_reserve_storage` sweeps the row and admits another write into the same slot → original job inserts → quota exceeded. That is GH #532 again, delayed by the TTL.

Refresh `expires_at` on each retry/requeue, or set TTL above the full congestion+retry+upload ceiling.

services/server/src/routes/analyze.rs
ducnmm, line 520: **Partial-fail leak after this reserve.** Admission is all-or-nothing, then the loop below `?`s on the first INSERT/enqueue failure (`map_err` on the job-row insert / enqueue) and does not release remaining `quota_items`.

Facts that already got a wallet job will be released by the worker. Facts that never got a job keep reservation rows until the 15-minute sweep — up to `MAX_ANALYZE_FACTS` ciphertext sizes of stuck quota.

The inline/benchmark path already releases before `?`. On error here, release every id that will not be handled by a worker.

ducnmm (changes requested):
## Review

The original GH #532 TOCTOU is real, and holding `pg_advisory_xact_lock` across the usage read **and** the reservation insert is the right fix. Burst tests and migration 014 look good.

Please fix the two holes below before merge — both re-open overshoot or leave quota reserved until TTL.

Do **not** release on the lost-prepare fence (`return Ok(())` for the same `job_id`). That reservation still belongs to the winner.

ducnmm:
Superseded by #720 (merged).

PR #701, opened 2026-08-19
fix(mcp): resolve credentials per project, and stop losing them on re-login

### LINKED ISSUE #628: ~/.memwal/credentials.json is global — logging in from one project silently replaces the credential another project depends on
### Summary

Credentials live in a single global file at `~/.memwal/credentials.json`. There is no per-project or
per-account scoping, so authenticating from any directory overwrites the credential every other
project is using.

### What happened

Working on project A (a Walrus Sessions submission), I authenticated from project B, which had its
own `--label` configured. The login succeeded and replaced the global credential — different wallet,
different account, different delegate key. Project A's tooling then silently began targeting project
B's account. The only visible signal was the `label` field, and nothing warns at the point of use.

Had I written memories in that state, they would have landed on the wrong account, on immutable
storage, with no delete path.

The CLI does write `credentials.backup-<date>.json` before overwriting, which is genuinely helpful and
made recovery possible — but it is undocumented, easy to miss, and only holds one generation.

### Impact

Anyone working on two Walrus Memory projects on one machine will hit this. There is no way to hold
two accounts simultaneously, and no way to scope a credential to a directory.

### Expected

- Per-project credential resolution — for example `.memwal/credentials.json` in the working directory
  taking precedence over the global file, in the way `.npmrc` and `.git/config` resolve
- Or named profiles: `--profile <name>` writing to `~/.memwal/credentials.<name>.json`
- At minimum, a warning when a login is about to replace a credential for a *different* account,
  showing both the outgoing and incoming account IDs

nikola0x0:
Resolves #628 (WALM-361).

### Problem

Credentials lived in one global `~/.memwal/credentials.json`. Signing in from one project repointed every other project on the machine at a different account and delegate key. The only visible signal was the `label` field, and nothing warned at the point of use — memories written in that state land on the wrong account, on immutable storage, with no delete path.

### Changes

**Resolution.** A `.memwal/credentials.json` in the working directory takes precedence over the global file, the way `.npmrc` and `.git/config` resolve. Presence-based on purpose: creating the local file is the opt-in, so a machine without one behaves exactly as before. Paths resolve per call rather than at module load, since the working directory is not knowable at import time.

**Replacement safety.** A sign-in that would overwrite existing credentials warns *before* the browser opens, naming the account currently saved and the file at risk — that is the last moment the user can back out without having registered a delegate key on-chain. Afterwards, a different-account sign-in copies the outgoing file to `credentials.backup-<timestamp>.json` and prints both account ids. Same-account re-saves overwrite in place without churning backups.

`saveCreds` returns what it did rather than printing: persistence should not own user-facing output, and returning it keeps the wording testable without capturing stdout.

### A data-loss bug found on the way

`memwal-mcp login` called `clearCreds()` up front, purely so the `loadCreds()` below would return null and the sign-in flow would run. **Anyone who ran `login` and then abandoned the browser flow — or whose login timed out or failed — was left with no credentials at all and nothing to recover from.** The delegate key on disk was destroyed before the new one was ever issued.

It also broke this ticket's own feature: with the file already gone, `saveCreds` had nothing to compare against, so the account change was invisible and there was nothing to back up. All unit tests were green; only the live check caught it.

Forcing a fresh sign-in means ignoring what is on disk, not deleting it. `login` now passes `null` instead, so the old file survives until a successful sign-in replaces it. Flagging it separately here because it predates this ticket and affects anyone who has ever abandoned a login.

### Verification

| | Result |
| --- | --- |
| `packages/mcp` suite | 35/35 (25 on `dev` + 10 new) |
| `tsc` | clean, from a scratch rebuild |
| Live check vs the real binary | 10/10 |
| Docs code-sync | passes |

**Project isolation, real binary, same `HOME`, only cwd differs:**

| cwd | account loaded |
| --- | --- |
| has local `.memwal/` | `0xbbbb…bbbb` |
| no local `.memwal/` | `0x997daad…` |

The live check (`test/live/credential-replacement.live.mjs`) drives the real `memwal-mcp login` process end to end — its own listener, preflight validation, callback parsing and save — and asserts on what the process actually printed. Opt-in: the `.live.mjs` suffix is outside the `test/**/*.test.mjs` glob, so **CI is unaffected**. The browser half is driven programmatically, so it needs no wallet approval and registers nothing on-chain.

### Acceptance criteria

- [x] Backwards-compatible resolution strategy
- [x] Prevent silent replacement of another account's credentials
- [x] Show outgoing and incoming account ids before destructive replacement
- [x] Allow multiple projects/accounts to coexist on one machine
- [x] Document precedence, migration, backup and recovery
- [x] Cross-project isolation and compatibility tests — see the caveat below

### On "supported platforms"

The tests are written portably: the sandbox sets `HOME` **and** `USERPROFILE` together (`os.homedir()` reads the latter on Windows), paths go through `join`, and the backup filename strips `:` so it is a legal Windows path. `chmodSync` was already guarded for Windows.

They are nonetheless **unproven** on Windows, because every job in `.github/workflows/` is `ubuntu-latest` — the repository has no Windows runner at all. That is a pre-existing CI gap rather than something this PR introduces, and it is the same gap WALM-360 (sandbox `USERPROFILE` in MCP login tests on Windows) exists to close. Adding a Windows runner here would collide with that work.

### Notes for review

- **Named `--profile` is deliberately not implemented.** Project-local resolution covers the reported problem; profiles can layer on later without conflicting. Recorded on the ticket.
- **The success page used to hardcode `~/.memwal/credentials.json`** — with project-local files that sentence could name a file the login never touched, in exactly the scenario this ticket exists to fix. It now reports the path actually written.
- **Backups are never pruned**, and each holds a delegate private key. Documented, with a `.gitignore` note for `.memwal/`.


jessiemongeon1:
## Style Guide Audit

ducnmm:
@nikola0x0 could you take a look at the last comment? The style-guide audit is blocking merge — two em dashes in `docs/mcp/reference.md` (lines 154 and 227). Once those are rewritten (comma / parentheses / split sentence), this should be ready for review.

packages/mcp/src/index.ts
```diff
@@ -108,9 +108,6 @@ export async function main(argv: string[] = process.argv.slice(2)): Promise<void
         note(`Credentials removed (${credsPath()}).`);
```
ducnmm: `clearCreds()` then `credsPath()` is wrong once resolution is presence-based.

If the process was using `.memwal/credentials.json` and `~/.memwal/credentials.json` also exists:

1. `clearCreds()` deletes the **project** file
2. `credsPath()` now falls through to **global**
3. this message claims the global file was removed (it wasn't)
4. the next MCP start loads the global account — the silent account-switch this PR is fixing

Capture the path *before* deleting, and warn if a fallback file is still present. Same pattern in `handleLocalLogout` in `bridge.ts` (not in this diff, but it becomes buggy the moment this lands).

packages/mcp/src/auth.ts
```diff
@@ -36,18 +44,44 @@ export interface MemWalCredentials {
...
+ * immutable storage, with no delete path (GH #628).
+ *
+ * Presence-based on purpose: creating the local file is the opt-in, so this is
+ * purely additive. Resolved per call rather than at module load, because the
+ * working directory is not knowable at import time.
+ */
 export function credsPath(): string {
-    return CREDS_PATH;
+    const project = projectCredsPath();
+    return existsSync(project) ? project : globalCredsPath();
```
ducnmm: This only checks `process.cwd()/.memwal/credentials.json`, not ancestors. The docs compare this to `.npmrc` / `.git/config`, which walk up.

`cd src && memwal-mcp login` (or an MCP host whose cwd is a subfolder) misses the project file and writes/reads the global one — back to GH #628.

Either walk up to the git/filesystem root, or drop the analogy and document the cwd-only rule explicitly.

docs/mcp/reference.md
```diff
@@ -121,6 +121,61 @@ Remove the saved credentials from this machine (`~/.memwal/credentials.json`). T
...
 </Note>
 
+## Credential locations
+
+Credentials resolve from two places, in order:
+
+1. `.memwal/credentials.json` in the **current working directory**
+2. `~/.memwal/credentials.json` (global, per machine)
+
+The first one that exists wins, the way `.npmrc` and `.git/config` resolve. Whichever file is chosen is the one read, written, and deleted for that run.
```
ducnmm: This sentence is not true of the implementation: git and npmrc walk up, `credsPath()` does not. See the comment on `auth.ts`. Style-guide bot also wants the em dashes on lines 154 and 227 rewritten (comma / parentheses / split sentence) before merge.

ducnmm (changes requested):
WALM-376 (don't delete creds before login) looks correct. The project-scoping half has a couple of real gaps that recreate GH #628 in a different shape.

Please address the inline comments before merge. Also the style-guide audit is still blocking: em dashes on `docs/mcp/reference.md` lines 154 and 227.

packages/mcp/src/index.ts
nikola0x0: Fixed in b3302fe0. `clearCreds()` resolves the path before unlinking and returns `{removedPath, fallbackPath}`, so logout names the file it actually deleted and warns when a surviving file takes over on the next run. `handleLocalLogout` in bridge.ts fixed in the same commit rather than left to break on merge.

packages/mcp/src/auth.ts
nikola0x0: Fixed in b3302fe0 — it walks up now, bounded at the project root (`.git`), home, or the filesystem root.

Worth noting the bound is load-bearing, not tidiness. With only a home stop the walk climbed out of the test sandbox into the real `~/.memwal/credentials.json` and broke sse-idle-watchdog: `homedir()` only halts the walk when it is a genuine ancestor of cwd, which is false whenever HOME is overridden. Same hazard #705 is closing.

Five new tests cover subdirectory resolution and the bound. 40 pass, up from 35.

docs/mcp/reference.md
nikola0x0: Rewritten in b3302fe0. The section now states the walk and its bound explicitly, so the `.npmrc` / `.git/config` comparison matches the implementation. Both em dashes (154, 227) are gone — `grep "—" docs/mcp/reference.md` returns nothing.

ducnmm (approved):
Re-review of `b3302fe0`. Previous comments are addressed.

- `credsPath()` walks up from cwd, stops at `.git` / home / fs root. Subdirectory login no longer silently uses the global file.
- `clearCreds()` snapshots the path before unlink and returns `{removedPath, fallbackPath}`. Logout names the file it deleted and warns when a survivor takes over.
- Docs match the walk + bound. Em dashes gone.

Tests cover subdirectory resolution, the project-root bound, and each `clearCreds` outcome. CI green.

Approve.

Issue #610, opened 2026-08-13
npm test in packages/mcp destroys real ~/.memwal credentials on Windows (HOME vs USERPROFILE)


Wadefak04:
## Summary

`packages/mcp/test/login-preflight.test.mjs` sandboxes the home directory with `process.env.HOME` only. On Windows `os.homedir()` reads `USERPROFILE`, not `HOME`, so the sandbox has no effect and the test writes its fixture credentials into the developer's **real** `~/.memwal/credentials.json`, overwriting a live delegate private key.

Running `npm test` in `packages/mcp` on Windows destroys the developer's Walrus Memory credentials. There is no backup — the delegate key is only stored in that file.

## Affected

- `packages/mcp/test/login-preflight.test.mjs` (v0.0.6, commit `9b84c90`)

`src/auth.ts:39` resolves the credential path once at module load:

```ts
const CREDS_DIR = join(homedir(), ".memwal");
```

`os.homedir()` on Windows returns `USERPROFILE` and ignores `HOME`:

```
$ node -e "process.env.HOME='C:/tmp/fake'; console.log(require('os').homedir())"
C:\Users\USER          # HOME override ignored

$ node -e "process.env.USERPROFILE='C:/tmp/fake'; console.log(require('os').homedir())"
C:/tmp/fake            # USERPROFILE honored
```

The sibling tests are not affected because they spawn a subprocess and pass both variables explicitly — `login-handoff.test.mjs:114` uses `env: { ...process.env, HOME: home, USERPROFILE: home }`. `login-preflight.test.mjs` imports `../dist/login.js` in-process and sets only `HOME`.

## Reproduction

Windows 10, Node v22.23.1, pnpm 9.12.3.

1. Plant a sentinel credential file at `~/.memwal/credentials.json`:

```json
{
  "version": 1,
  "label": "MY-REAL-PRODUCTION-WALLET",
  "delegatePrivateKey": "SENTINEL_REAL_KEY_DO_NOT_LOSE",
  "accountId": "0xabab...",
  "relayerUrl": "https://relayer.memory.walrus.xyz"
}
```

2. `cd packages/mcp && node --test test/login-preflight.test.mjs`

3. Read the file back:

```
label: Test MCP
key:   129d5b8606a6b905c55ac7b5d89cfe59175f0516551009ba8dc6490f9c14729e
*** OVERWRITTEN ***
```

The real delegate key is gone, replaced by the test's ephemeral keypair.

[...]

PR #705, opened 2026-08-19
fix(mcp): keep the login test off the real credential directory

HoangDucBach:
## Ticket

WALM-360 / GH #610 — the MCP login test overwrites the developer's real credentials on Windows.

### `src/auth.ts`

The credential directory is now resolved per call instead of frozen at module load, and `MEMWAL_CREDS_DIR` overrides it. The four sites that used the old constant (`credsPath`, `loadCreds`, `saveCreds`, `clearCreds`) go through the resolver; nothing outside the file touched it.

### `test/login-preflight.test.mjs`

Points `MEMWAL_CREDS_DIR` at a temp directory instead of mutating `HOME`, and asserts `credsPath()` resolves there **before** the login flow can write anywhere. The old version asserted against a hand-built path, so on Windows it failed because the temp directory was empty — after the real file had already been clobbered.

### `test/creds-dir-sandbox.test.mjs` (new)

- The override wins over `HOME` and `USERPROFILE`, and is re-read per call so an already-imported module follows a change.
- A second case fails if any test file sandboxes `HOME` without `USERPROFILE`, naming the offender.

## Why is this needed?

`os.homedir()` reads `USERPROFILE` on Windows and ignores `HOME`, so a `HOME`-only sandbox has no effect. Running `npm test` in `packages/mcp` wrote fixture credentials over `~/.memwal/credentials.json`. The delegate private key exists nowhere else, and the destroyed key stays authorized on-chain since nothing revokes it.

Setting both variables fixes the symptom but leaves the sandbox depending on process-global mutation happening before the first import — a requirement nothing enforces. Because the path was frozen at module load, a second test in that file with its own temp directory would have silently reused the first one's, on any platform.

## Scope

The credential path resolver and the tests that depend on it.

### Out of scope

- Adding `windows-latest` to CI. Every job in the repo is `ubuntu-latest`; that is a repo-wide decision. The meta-test covers this class of mistake on the existing Linux runners instead.
- `saveCreds` overwriting unconditionally with no backup, and the orphaned on-chain key after a loss. Both are real, both are product behaviour rather than test sandboxing.

## How was this tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] End-to-end tests
- [x] Manual testing
- [ ] Not applicable

Suite is 27/27.

Verified both new tests earn their place. Reverting `auth.ts` to the module-level constant makes `login-preflight` fail at the new assertion, before the flow writes anything. Dropping in a file that sets `HOME` without `USERPROFILE` makes the meta-test fail and name it.

Manually confirmed the invariant the ticket is about: `~/.memwal` does not exist on this machine, and after a full suite run it still does not — the tests never reach the real home.

```bash
cd packages/mcp && npm test

# prove the guard works: revert the resolver to a module-level constant
#   const CREDS_PATH = join(homedir(), ".memwal", "credentials.json");
# then re-run — login-preflight fails at the credsPath() assertion
node --test test/login-preflight.test.mjs
```

## Risks and dependencies

- **`MEMWAL_CREDS_DIR` is a new public env var.** Anyone who can set it can redirect where credentials are read and written. That is not a new capability in practice — a caller who controls the environment already controls `HOME`, which fed the same path.
- **The meta-test matches on source text**, so it catches `HOME:` in an env object and `process.env.HOME =`, not an indirect assignment. It raises the floor rather than proving the property.

## Author checklist

- [ ] CI is green.


jessiemongeon1:
## Style Guide Audit

### `docs/mcp/changelog.mdx` (1 violation(s))

_1 violation(s) (0 regex, 1 claude)_

- **Line 36** — stacked-headings
  - Current: `### Fixed

- Resolve the credential...`
  - Fix: `Add a brief introductory sentence between ## Unreleased and ### Fixed`

---

packages/mcp/src/auth.ts
```diff
@@ -36,18 +36,24 @@ export interface MemWalCredentials {
     version: 1;
 }
 
-const CREDS_DIR = join(homedir(), ".memwal");
-const CREDS_PATH = join(CREDS_DIR, "credentials.json");
+/** Resolved per call, not frozen at module load: a sandbox that has to be in
+ * place before the first import is a requirement nothing enforces, and missing
+ * it cost a developer their delegate key (GH #610). `MEMWAL_CREDS_DIR` moves
```
ducnmm: Please keep ticket/issue identifiers out of source comments — describe the failure mode instead. Shipping tracker IDs in the published package reads as internal leakage.

packages/mcp/test/creds-dir-sandbox.test.mjs
```diff
@@ -0,0 +1,72 @@
+/**
+ * Guards the credential-directory sandbox (GH #610 / WALM-360).
```
ducnmm: Same here — please drop the tracker IDs from the file header. A one-line description of the Windows `os.homedir()` / `USERPROFILE` pitfall is enough.

packages/mcp/test/login-preflight.test.mjs
```diff
@@ -5,13 +5,19 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 
 test("login requires a matching localhost preflight before browser approval", async () => {
-    const home = mkdtempSync(join(tmpdir(), "memwal-preflight-"));
-    const previousHome = process.env.HOME;
-    process.env.HOME = home;
+    // Imports the credential writer in-process, so the sandbox has to hold
+    // here rather than in a child. MEMWAL_CREDS_DIR avoids mutating HOME, which
+    // os.homedir() ignores on Windows (GH #610).
```
ducnmm: Same as `auth.ts`: keep the Windows `homedir()` note, omit the issue number.

ducnmm (changes requested):
Two nits before this is ready:

1. Please strip ticket/issue identifiers from source comments (`auth.ts`, the new sandbox test header, and `login-preflight.test.mjs`). Describe the behavior instead — tracker IDs in the published package look like leaked internal process.
2. This adds a user-facing env var (`MEMWAL_CREDS_DIR`) and you documented it in `docs/reference/environment-variables.md`, but there is no manual changelog entry. Please add one under `packages/mcp/CHANGELOG.md` (and the matching `docs/mcp/changelog.mdx` pair) the same way other MCP env/behavior changes are recorded.

The sandbox fix itself looks correct.

packages/mcp/src/auth.ts
HoangDucBach: resolved

packages/mcp/test/creds-dir-sandbox.test.mjs
HoangDucBach: resolved

packages/mcp/test/login-preflight.test.mjs
HoangDucBach: resolved

HoangDucBach:
> Two nits before this is ready:
> 
> 1. Please strip ticket/issue identifiers from source comments (`auth.ts`, the new sandbox test header, and `login-preflight.test.mjs`). Describe the behavior instead — tracker IDs in the published package look like leaked internal process.
> 2. This adds a user-facing env var (`MEMWAL_CREDS_DIR`) and you documented it in `docs/reference/environment-variables.md`, but there is no manual changelog entry. Please add one under `packages/mcp/CHANGELOG.md` (and the matching `docs/mcp/changelog.mdx` pair) the same way other MCP env/behavior changes are recorded.
> 
> The sandbox fix itself looks correct.

Both done

packages/mcp/src/auth.ts
```diff
@@ -36,18 +36,25 @@ export interface MemWalCredentials {
...
 
-const CREDS_DIR = join(homedir(), ".memwal");
-const CREDS_PATH = join(CREDS_DIR, "credentials.json");
+/** Resolved per call, not frozen at module load. A caller that needs the store
+ * elsewhere would otherwise have to change the environment before this module
+ * is first imported, which nothing enforces; miss it and writes land on the
+ * real credentials file. `MEMWAL_CREDS_DIR` moves the store with no
+ * process-global mutation. */
+function credsDir(): string {
+    return process.env.MEMWAL_CREDS_DIR || join(homedir(), ".memwal");
```
ducnmm: **[bug]** This PR is still based on `86f79ee3`. On `origin/dev`, #701 already replaced the module-level `CREDS_PATH` with `credsPath() => projectCredsPath() ?? globalCredsPath()`, where `globalCredsPath()` is `join(homedir(), ".memwal", "credentials.json")` and a project-local `.memwal/credentials.json` at or above cwd wins. `git merge-tree` against `origin/dev` shows `auth.ts` conflicting while `login-preflight.test.mjs` auto-merges to the PR version: it sets only `MEMWAL_CREDS_DIR` and no longer sets `HOME`. A conflict resolution that keeps `origin/dev`'s resolver leaves the new override unread. The pre-write assertion should then fail before `loginFlow()` (so this is not a silent clobber by itself), but the GH #610 fix does not actually land, and `login-preflight` no longer has even the old Unix `HOME` sandbox as backup. Wiring the env var only into `globalCredsPath()` is not enough either: if a developer has opted into #701's project-local file under the repo, `projectCredsPath()` still wins, the assertion compares that path to the temp dir, and a loosened assertion would overwrite the real project credentials. `MEMWAL_CREDS_DIR` has to win over the project walk, because this test does not `chdir` away from the checkout.

**Suggestion:** Rebase onto `origin/dev`. Keep per-call resolution. In `credsPath()`, if `MEMWAL_CREDS_DIR` is set, return `join(process.env.MEMWAL_CREDS_DIR, "credentials.json")` and skip `projectCredsPath()`. Keep the pre-write assertion. Move the changelog line out of the already-shipped 0.0.10 section on `origin/dev` (package is 0.0.11) into the current unreleased section in both `packages/mcp/CHANGELOG.md` and `docs/mcp/changelog.mdx`.

packages/mcp/test/login-preflight.test.mjs
```diff
@@ -5,13 +5,19 @@ import { tmpdir } from "node:os";
...
+    // here rather than in a child. MEMWAL_CREDS_DIR avoids mutating HOME, which
+    // os.homedir() ignores on Windows.
+    const credsDir = mkdtempSync(join(tmpdir(), "memwal-preflight-"));
+    const previousCredsDir = process.env.MEMWAL_CREDS_DIR;
+    process.env.MEMWAL_CREDS_DIR = credsDir;
 
     const webUrl = "https://memory.example";
     const relayerUrl = "https://relayer.example";
     const { loginFlow } = await import(`../dist/login.js?test=${Date.now()}`);
+    const { credsPath } = await import(`../dist/auth.js?test=${Date.now()}`);
```
ducnmm: **[suggestion]** `login.js?test=…` resolves its static `import { saveCreds } from "./auth.js"` to the canonical `auth.js` URL (relative URL resolution drops the query). The test then imports `auth.js?test=…`, which is a second module instance, and asserts *that* `credsPath()`. With per-call env reads and default `node --test` process isolation the two instances currently agree, so this does not clobber today. It does mean the assertion is not observing the module `loginFlow()` will call. If `auth.js` is ever loaded first without the override, or someone re-freezes the path at import time, the queried instance can report the temp path while `saveCreds` still writes to the real home — the exact failure mode this comment says it now fails before.

**Suggestion:** Import the unqueried module first so the assertion uses the same instance `login.ts` binds: `const { credsPath } = await import("../dist/auth.js");` then `const { loginFlow } = await import(\`../dist/login.js?test=${Date.now()}\`);`. Optionally also snapshot `existsSync(join(homedir(), ".memwal", "credentials.json"))` before the flow and assert it is unchanged after, which is the actual GH #610 symptom.

packages/mcp/src/auth.ts
```diff
@@ -36,18 +36,25 @@ export interface MemWalCredentials {
     version: 1;
 }
 
-const CREDS_DIR = join(homedir(), ".memwal");
-const CREDS_PATH = join(CREDS_DIR, "credentials.json");
+/** Resolved per call, not frozen at module load. A caller that needs the store
```
ducnmm: **[suggestion]** The published-package comment restates the implementation (per-call vs module load) and narrates the sandbox incident. Readers of `@mysten-incubation/memwal-mcp` do not need the import-order history; the tests already document the Windows `HOME` / `USERPROFILE` trap.

**Suggestion:** Keep a one-line why, e.g. `/** Read env on every call so MEMWAL_CREDS_DIR can redirect the store after import. */`

docs/reference/environment-variables.md
```diff
@@ -69,6 +69,7 @@ The stdio MCP package reads these environment variables directly. A CLI flag tak
 | `MEMWAL_WEB_URL` | `--web-url <url>` | dashboard default | Dashboard URL used during login |
 | `MEMWAL_CLIENT_LABEL` | `--label <text>` | `MCP Client` / `Walrus Memory MCP` | Friendly delegate-key label shown in the dashboard |
 | `MEMWAL_MCP_DEBUG` | none | `0` | Set to `1` for verbose stderr logging |
+| `MEMWAL_CREDS_DIR` | none | `~/.memwal` | Directory holding `credentials.json`. Resolved on every access, so it can be pointed elsewhere without restarting. Mainly for tests, which must not write into the real credential directory |
```
ducnmm: **[nit]** "Resolved on every access, so it can be pointed elsewhere without restarting" is misleading for operators. Changing the host's environment still requires restarting the MCP process; per-call resolution only helps in-process tests that mutate `process.env` after `auth.js` is loaded.

**Suggestion:** Drop "without restarting". Something like: `Directory holding credentials.json. Overrides the default ~/.memwal. Mainly for tests, so they do not write the real credential file.`

packages/mcp/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - Align the cold-start `memwal_remember` / `memwal_recall` descriptions with the sidecar's proactive wording. The bridge serves this static list before the relayer session is up, and clients that keep the first `tools/list` were told to call remember only when the user explicitly asked.
 - Stop classifying remember vs recall in the UserPromptSubmit hook. The hook injects a decision rubric and the agent chooses the tool from meaning, so Vietnamese, typos, and phrasing that never says "remember" still work.
 - Keep remember proactive for preferences, decisions, constraints, corrections, and identity, but skip one-off tasks, the current file or bug, and small talk.
+- Resolve the credential directory on every access instead of freezing it at module load, and let `MEMWAL_CREDS_DIR` override it. The login test sandboxed the home directory with `HOME` alone, which `os.homedir()` ignores on Windows, so running the package's test suite there wrote fixture credentials over the developer's real `~/.memwal/credentials.json` and destroyed the delegate key stored in it. (#705)
```
ducnmm: **[bug]** The #705 note is still under `## 0.0.10`. Package version on this branch is 0.0.11, and both `@mysten-incubation/memwal-mcp@0.0.10` and `@mysten-incubation/memwal-mcp@0.0.11` are already tagged; origin/dev's 0.0.10 / 0.0.11 sections match those releases. This change is not in 0.0.10, and stuffing it there rewrites a shipped changelog. The merge kept origin/dev's 0.0.11 block on top and left the incoming bullet in the old section.

**Suggestion:** Move the bullet out of 0.0.10 into a new unreleased / next-version section (0.0.11 is already tagged). Do not leave it under 0.0.10.

docs/mcp/changelog.mdx
```diff
@@ -50,6 +50,7 @@ This release makes `memwal_logout` cut the running bridge session so memory tool
 - Align the cold-start `memwal_remember` / `memwal_recall` descriptions with the sidecar's proactive wording. The bridge serves this static list before the relayer session is up, and clients that keep the first `tools/list` were told to call remember only when the user explicitly asked.
 - Stop classifying remember vs recall in the UserPromptSubmit hook. The hook injects a decision rubric and the agent chooses the tool from meaning, so Vietnamese, typos, and phrasing that never says "remember" still work.
 - Keep remember proactive for preferences, decisions, constraints, corrections, and identity, but skip one-off tasks, the current file or bug, and small talk.
+- Resolve the credential directory on every access instead of freezing it at module load, and let `MEMWAL_CREDS_DIR` override it. The login test sandboxed the home directory with `HOME` alone, which `os.homedir()` ignores on Windows, so running the package's test suite there wrote fixture credentials over the developer's real `~/.memwal/credentials.json` and destroyed the delegate key stored in it. (#705)
```
ducnmm: **[bug]** Same mis-placement as the package changelog: the #705 line sits in the shipped 0.0.10 docs section, while the page already presents 0.0.11 as the latest release.

**Suggestion:** Move the bullet to the section for the version that will actually ship this fix, matching `packages/mcp/CHANGELOG.md`.

packages/mcp/test/creds-dir-sandbox.test.mjs
```diff
@@ -0,0 +1,72 @@
...
+    process.env.MEMWAL_CREDS_DIR = first;
+    assert.equal(credsPath(), join(first, "credentials.json"));
+
+    // Resolved per call, so a module already imported still follows the change.
+    // Freezing it at import time is what made the sandbox order-dependent.
+    process.env.MEMWAL_CREDS_DIR = second;
+    assert.equal(credsPath(), join(second, "credentials.json"));
+
+    delete process.env.MEMWAL_CREDS_DIR;
+    assert.equal(credsPath(), join(homedir(), ".memwal", "credentials.json"));
```
ducnmm, line 65: **[bug]** After deleting `MEMWAL_CREDS_DIR`, the test asserts `credsPath()` is always `join(homedir(), ".memwal", "credentials.json")`. Post-#701, `credsPath()` is `projectCredsPath() ?? globalCredsPath()`, and this test never sandboxes cwd. A project-local `.memwal/credentials.json` at or above the working directory (the #701 opt-in, including this repo) makes the assertion fail. The first two asserts still prove the override; this one documents the pre-#701 resolver.

**Suggestion:** After clearing the override, `chdir` into a temp tree with no credentials file (and a `.git` sentinel so the walk cannot climb into the real checkout), then assert the global sandbox path — or assert only that the result is neither `first` nor `second`.

packages/mcp/test/login-preflight.test.mjs
ducnmm: **[suggestion]** `login.js` statically imports canonical `./auth.js`. This test then imports `auth.js?test=…`, which is a different module instance, and asserts sandboxing on that instance. Writes still go through login's canonical `saveCreds` → `credsPath()`. With per-call env resolution both instances agree, so this is not the current write hole — but the early "fail before the flow can write" check still does not observe the module the writer binds. A freeze-at-import regression on the cached canonical module would pass this assert and write elsewhere.

**Suggestion:** Import `../dist/auth.js` without a query string (the instance `login.js` already bound), or assert against `join(credsDir, "credentials.json")` directly and confirm the real home and project-local files were not created.

packages/mcp/src/auth.ts
```diff
@@ -105,8 +105,17 @@ function projectCredsPath(): string | null {
  * Presence-based on purpose: creating the local file is the opt-in, so this is
  * purely additive. Resolved per call rather than at module load, because the
  * working directory is not knowable at import time.
+ *
+ * `MEMWAL_CREDS_DIR` overrides both when set, taking priority over project and
```
ducnmm: **[suggestion]** The new `credsPath()` JSDoc narrates the Windows `HOME` vs `USERPROFILE` incident and import-order history. The load-bearing fact is one sentence: `MEMWAL_CREDS_DIR` wins over project and global paths and is read per call.

**Suggestion:** Trim to that. Keep the incident write-up in the changelog / test, not on the resolver.

docs/reference/environment-variables.md
ducnmm: **[nit]** "Resolved on every access, so it can be pointed elsewhere without restarting" is true for in-process `process.env` mutations (tests) and misleading for operators. A running MCP process does not pick up a shell-level env change until it is restarted.

**Suggestion:** Drop "without restarting". Say it is re-read on every access so tests can redirect it after import, and that it overrides both project-local and `~/.memwal` credentials.

ducnmm (changes requested):
The origin/dev merge landed MEMWAL_CREDS_DIR correctly (it now short-circuits projectCredsPath()). Still blocking: the #705 changelog bullet is under already-shipped 0.0.10, and creds-dir-sandbox.test.mjs asserts a global-only fallback that is false after #701. See inline comments.

packages/mcp/CHANGELOG.md
HoangDucBach: Fixed in 93bc0013.

docs/mcp/changelog.mdx
HoangDucBach: Fixed in 93bc0013.

packages/mcp/test/creds-dir-sandbox.test.mjs
HoangDucBach: Fixed in 93bc0013.

packages/mcp/test/login-preflight.test.mjs
HoangDucBach: Fixed in 93bc0013.

packages/mcp/src/auth.ts
HoangDucBach: Fixed in 93bc0013.

docs/reference/environment-variables.md
HoangDucBach: Fixed in 93bc0013.

docs/mcp/changelog.mdx
```diff
@@ -28,9 +28,15 @@ questions:
...
+  The latest MCP package release is 0.0.11. It makes memwal_logout actually cut the running bridge session, so memory tools stop after sign-out instead of keeping the deleted delegate key in memory. Version 0.0.10 closes out orphaned tool calls whose responses never arrive with a retryable error, injects the configured default namespace into memwal_remember_bulk calls that omit one, and sends proactive-usage instructions in the MCP initialize handshake so the model knows when to save and recall without being asked, which lazy tool loading had broken. An unreleased fix resolves the credential directory on every access and lets MEMWAL_CREDS_DIR override it.
 ---
 
+## Unreleased
+
+### Fixed
```
ducnmm, line 38: **[nit]** `## Unreleased` is followed immediately by `### Fixed` with no body text between them. The same page's 0.0.11 section puts a one-line release summary between those heading levels, and the docs style audit flags this pair as stacked headings.

**Suggestion:** Add a short sentence between `## Unreleased` and `### Fixed` (for example, that the credential directory is now resolved on every access and `MEMWAL_CREDS_DIR` can override it). Leave `packages/mcp/CHANGELOG.md` as-is; Keep a Changelog allows stacked `## Unreleased` / `### Fixed`.

ducnmm (approved):
## Summary

All six previous findings on 38dc762e are addressed on 93bc0013. The #705 changelog bullet now sits under Unreleased while `@mysten-incubation/memwal-mcp` remains 0.0.11; the no-override sandbox assertion chdirs into a `.git`-bounded temp cwd before expecting `~/.memwal/credentials.json`; login-preflight imports the same `auth.js` instance `login.js` binds; the `credsPath()` JSDoc and `MEMWAL_CREDS_DIR` notes were trimmed. `MEMWAL_CREDS_DIR` still short-circuits `projectCredsPath()`. Remaining issue is a docs stacked-heading nit on the new Unreleased section.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 1

packages/mcp/CHANGELOG.md
ducnmm: Addressed on 93bc0013. Bullet is under `## Unreleased` (`packages/mcp/CHANGELOG.md:3` / `:7`). `## 0.0.11` is unchanged.

docs/mcp/changelog.mdx
ducnmm: Addressed. Docs changelog now has `## Unreleased` with the #705 bullet (`docs/mcp/changelog.mdx:34` / `:39`); frontmatter no longer claims it shipped in 0.0.10/0.0.11.

packages/mcp/test/creds-dir-sandbox.test.mjs
ducnmm: Addressed. After clearing `MEMWAL_CREDS_DIR` the test `chdir`s into a temp cwd with a `.git` sentinel so `projectCredsPath()` cannot walk into the real checkout, then asserts the global path (`creds-dir-sandbox.test.mjs:26`, `:64-65`).

packages/mcp/test/login-preflight.test.mjs
ducnmm: Addressed. `login-preflight` now imports `../dist/auth.js` with no query, matching `login.js`s canonical bind (`login-preflight.test.mjs:24`).

packages/mcp/src/auth.ts
ducnmm: Addressed. JSDoc is two lines: override wins, re-read per call (`auth.ts:109`).

docs/reference/environment-variables.md
ducnmm: Addressed. Env docs now say re-read so a test can redirect after import, not that operators can skip restart (`environment-variables.md:72`).

docs/mcp/changelog.mdx
HoangDucBach: Fixed in 065a55e9. Added a one-line Unreleased summary between `## Unreleased` and `### Fixed` in `docs/mcp/changelog.mdx`. `packages/mcp/CHANGELOG.md` is unchanged.

PR #706, opened 2026-08-19
fix(mcp): make Claude Code auto-remember/recall actually fire

ducnmm:
## Summary

Claude Code was not using Walrus Memory on its own.

1. Cold-start `tools/list` still advertised `memwal_remember` as "Call ONLY when the user explicitly asks". Clients that keep the first list ignored SessionStart / UserPromptSubmit.
2. UserPromptSubmit used English regexes to pick remember vs recall. That inverted "how I like to work" questions, and missed Vietnamese / typos / phrasing that never says "remember".
3. Even with hooks firing, Claude Code's built-in `MEMORY.md` lives in the system prompt and can outrank MCP instructions (Henry canary / Harry's review).

This PR:

- Aligns cold-start `memwal_remember` / `memwal_recall` descriptions with the sidecar (proactive, not "only when asked").
- Stops the hook from classifying. Every substantive prompt gets the same decision rubric; the agent chooses the tool from meaning, any language or spelling.
- Keeps remember **proactive** for preferences, decisions, constraints, corrections, and identity (Dio / Henry canary). Skips one-off tasks, the current file or bug, and small talk.
- Requires the Claude Code **plugin** in `skills/setup` (`claude plugin install`, not MCP-only).
- Has setup merge a user-level `~/.claude/CLAUDE.md` routing block, and tells the user to turn off Claude's built-in / native memory so `MEMORY.md` does not win.

## Test plan

- [x] `pnpm --filter @mysten-incubation/memwal-mcp test`
- [x] Local `tools/list` shows PROACTIVELY on remember/recall
- [x] Claude Code (`fable-5`) + local plugin + local MCP `--staging`:
  - Vietnamese facts, no keyword → `memwal_remember_bulk`
  - New session Vietnamese recall → `memwal_recall` returns the canary
  - "What do you remember about how I like to work…" → `memwal_recall` only
  - Hook debug shows the decision rubric, not a classified remember/recall directive
- [ ] Re-run Henry canary on stock sonnet after `~/.claude/CLAUDE.md` routing + built-in memory off

## Follow-ups (not blocking)

- Plugin `.mcp.json` is still unpinned prod (`npx -y @mysten-incubation/memwal-mcp`); no `--staging` pin
- Do not start the turn until `memwal_*` are connected
- Claude Code may still write `MEMORY.md` if the user-level `CLAUDE.md` block is missing or built-in memory stays on


jessiemongeon1:
## Style Guide Audit

Audited **5** file(s) against the [Sui Documentation Style Guide](https://docs.sui.io/references/contribute/style-guide).

### `docs/mcp/reference.md` (1 violation(s))

_1 violation(s) (1 regex, 0 claude)_

- **Line 56** — No em dashes in prose
  - Current: `—`

---

harrymove-ctrl (changes requested):
Requesting changes. The root cause is correctly identified and the fix is at the right layer: the cold-start `tools/list` saying "Call ONLY when the user explicitly asks" while the sidecar said the opposite was a real contradiction, and dropping the English regex classifier is right (`REMEMBER_INTENT` matched `\bi (prefer|like|...)`, so "What do you remember about how I **like** to work" classified as remember).

Items 1 and 2 should land in this PR. The rest can be follow-ups if you prefer.

### 1. Signed-out stubs now advertise proactive remember (blocking)

`packages/mcp/src/auth-required.ts`

The file's own comment states the signed-out path is deliberately non-proactive: "without credentials every memory tool fails, so telling the model to save proactively here would only manufacture errors." This PR rewrites the justification but keeps serving one static list to both cases. A genuinely signed-out user now gets a model that fires `memwal_remember` on every stated preference and collects a stream of auth errors, which is the failure mode the original comment guarded against.

Please pick the description set from whether `~/.memwal/credentials.json` exists, so a signed-in-but-not-yet-connected client gets the proactive wording and a truly signed-out one keeps the conservative wording. If that is out of scope, say so explicitly in the comment rather than reframing the rationale.

### 2. `memwal_recall` is still `destructiveHint: true` (blocking)

`packages/mcp/src/auth-required.ts`

Not introduced here, but this PR is what makes it matter: recall is now advertised as something to call proactively at the start of every task, while the annotation tells clients the call is destructive and not read-only. Clients that gate on `destructiveHint` will prompt or hold back the exact call this PR exists to encourage. A search should be `readOnlyHint: true, destructiveHint: false`. Cheap to fix here.

### 3. The rubric now fires on every prompt, with the once-per-session dedup removed

`packages/mcp/plugin/scripts/on_user_prompt.mjs`

Previously `firstTime("rubric", sessionId)` injected the general text once per session. It now injects ~90 words of identical `DECISION_RUBRIC` on every prompt over 20 characters for the whole session. That is real per-turn context cost, and repeated identical injections tend to lose salience rather than gain it. Consider keeping `firstTime` for the full rubric and a one-line version afterwards, or re-injecting every N turns.

### 4. `firstTime` is now dead code

`packages/mcp/plugin/scripts/lib/hook-io.mjs`

`on_user_prompt.mjs` was its only consumer. Either remove it along with the session state file it writes, or keep it and use it per item 3.

### 5. The 20-character gate is still a gate, and it penalizes terse languages

`packages/mcp/plugin/scripts/on_user_prompt.mjs`

The thesis is "no keyword gate, the agent decides from meaning", but a length threshold survives. "Tui thích pnpm" (14 chars) is a durable preference in Vietnamese and gets no rubric, while "please refactor this component" does. Given the Vietnamese motivation in the description, please lower it, switch to a word/token count, or document it as a deliberate tradeoff.

### 6. `user-prompt-hook.test.mjs` is partly tautological

After `assert.equal(ctx, DECISION_RUBRIC)`, the follow-up `assert.match(ctx, /memwal_recall/)` and friends re-assert properties of the same imported constant, so they can never fail independently; edit the constant and the assertions move with it. The `doesNotMatch` checks on the removed strings are the only real guard.

The invariant worth pinning is that all seven prompts produce byte-identical output regardless of language, spelling, or the presence of "remember". Collect the outputs and assert they are all equal to each other, then drop the substring matches against the constant.

### 7. Editing the user's global `~/.claude/CLAUDE.md` needs an idempotency marker

`apps/app/public/skills/setup`, `docs/mcp/claude-code.md`

The only re-run guard is prose: "If a Walrus Memory / `memwal_*` routing section is already there, leave it." That is fuzzy for an agent, and it makes future updates to the block impossible to apply cleanly. Wrap it in explicit markers, for example `<!-- memwal:start -->` / `<!-- memwal:end -->`, so re-runs replace the block deterministically and the user can remove it in one shot. Also worth stating that the agent must show the block and get the user's go-ahead before writing to a global file outside the project.

### Merge readiness

The last test-plan box is unchecked: "Re-run Henry canary on stock sonnet after `~/.claude/CLAUDE.md` routing + built-in memory off". That is the acceptance test for this PR's central claim, so it should be green before merge rather than after.

packages/mcp/src/auth-required.ts
```diff
@@ -50,7 +50,7 @@ export const TOOL_DEFINITIONS = [
         title: "Remember a Fact",
         annotations: { readOnlyHint: false, destructiveHint: false },
         description:
-            "Save a fact to the user's Walrus Memory personal memory. Call ONLY when the user explicitly asks to remember/save something. Pass the full, detailed text — never summarize.",
+            "Save a durable fact about the user or project to their Walrus Memory. Call this PROACTIVELY whenever the user states a preference, decision, constraint, correction, identity detail, or recurring workflow — even if they did not say 'remember this'. Skip one-off tasks, the current file or bug, and small talk. Pass the full statement; do not summarize. To save several facts at once, use memwal_remember_bulk instead.",
```
harrymove-ctrl: **Blocking.** This file's own comment in `instructions.ts` says the signed-out path is deliberately non-proactive: "without credentials every memory tool fails, so telling the model to save proactively here would only manufacture errors."

This change keeps one static list for both cases, so a genuinely signed-out user now gets a model that fires `memwal_remember` on every stated preference and collects a stream of auth errors, which is exactly the failure mode the original wording guarded against.

Please pick the description set based on whether `~/.memwal/credentials.json` exists: proactive wording for signed-in-but-not-yet-connected, conservative wording for truly signed out. If that is out of scope for this PR, say so explicitly in the comment instead of reframing the rationale.

packages/mcp/src/auth-required.ts
```diff
@@ -87,7 +87,7 @@ export const TOOL_DEFINITIONS = [
         title: "Recall Memories",
         annotations: { readOnlyHint: false, destructiveHint: true },
```
harrymove-ctrl: **Blocking.** `memwal_recall` is a search, but it is annotated `readOnlyHint: false, destructiveHint: true`.

Not introduced by this PR, but this PR is what makes it matter: line 90 now tells the model to call recall proactively at the start of every task, while the annotation tells clients the call is destructive. Clients that gate on `destructiveHint` will prompt or hold back the exact call this PR exists to encourage.

Should be `readOnlyHint: true, destructiveHint: false`. Same question for `memwal_analyze` on line 105, though that one at least writes.

packages/mcp/plugin/scripts/on_user_prompt.mjs
```diff
@@ -1,41 +1,18 @@
...
-// the agent knows the tools exist, when to reach for them, and to prefer them
-// over any built-in memory feature.
-if (parts.length === 0 && firstTime("rubric", sessionId)) {
-    parts.push(
-        "Walrus Memory (the memwal_* tools) is this user's primary memory system — prefer it over any built-in memory. Call memwal_recall proactively when the user references past work, decisions, or preferences; save durable facts with memwal_remember (or memwal_remember_bulk for several) as you learn them — you don't need to be asked."
-    );
-}
-
-emitContext("UserPromptSubmit", parts.join(" "));
+emitContext("UserPromptSubmit", DECISION_RUBRIC);
```
harrymove-ctrl: The old code injected the general rubric once per session via `firstTime("rubric", sessionId)`. This now injects ~90 words of identical `DECISION_RUBRIC` on every prompt over 20 characters, for the whole session.

Two costs: real per-turn context spend, and repeated identical text tends to lose salience rather than gain it. Consider `firstTime` for the full rubric plus a one-line reminder afterwards, or re-injecting every N turns.

packages/mcp/plugin/scripts/on_user_prompt.mjs
```diff
@@ -1,41 +1,18 @@
...
+ * UserPromptSubmit hook — remind the agent that it chooses remember vs
+ * recall from meaning. No keyword classify, no fetch, no network.
  *
- * Heuristic only: no fetch, no network. The agent does the actual tool call.
+ * The agent has the conversation and understands any language or spelling;
+ * a regex cannot. This only injects a decision rubric.
  */
-import { readStdin, emitContext, firstTime } from "./lib/hook-io.mjs";
-import { detectRecall, detectRemember } from "./lib/signals.mjs";
+import { readStdin, emitContext } from "./lib/hook-io.mjs";
```
harrymove-ctrl: `firstTime` is dropped here, and this was its only consumer, so `firstTime` in `plugin/scripts/lib/hook-io.mjs:52` is now dead code along with the session state file it writes. Either remove it or keep it and use it per the comment on line 17.

packages/mcp/plugin/scripts/on_user_prompt.mjs
```diff
@@ -1,41 +1,18 @@
...
+import { DECISION_RUBRIC } from "./lib/decision-rubric.mjs";
 
 const input = readStdin();
 const prompt = (input.prompt || "").toString();
-const sessionId = input.session_id || "default";
 
 // Skip very short prompts (acks, single words) — nothing useful to act on.
 if (prompt.trim().length < 20) process.exit(0);
```
harrymove-ctrl: The thesis of the PR is "no keyword gate, the agent decides from meaning", but this length threshold is still a gate, and it is biased against terse languages.

"Tui thích pnpm" is 14 characters and is exactly the durable Vietnamese preference this PR is meant to catch, so it gets no rubric, while "please refactor this component" (30 chars) does. Please lower it, switch to a word count, or document it as a deliberate tradeoff.

packages/mcp/test/user-prompt-hook.test.mjs
```diff
@@ -0,0 +1,54 @@
...
+];
+
+test("substantive prompts get the decision rubric, not a classified directive", () => {
+    for (const prompt of SUBSTANTIVE) {
+        const ctx = runHook(prompt);
+        assert.equal(ctx, DECISION_RUBRIC, `should inject rubric: ${prompt}`);
+        assert.match(ctx, /memwal_recall/);
+        assert.match(ctx, /memwal_remember/);
+        assert.match(ctx, /any language or spelling/);
+        assert.match(ctx, /skip one-off tasks/i);
```
harrymove-ctrl: These four `assert.match` calls run after `assert.equal(ctx, DECISION_RUBRIC)` on line 41, so they only re-assert properties of the imported constant. Edit the constant and they move with it; they can never fail independently. The `doesNotMatch` checks below are the only real regression guard.

The invariant actually worth pinning is that every prompt in `SUBSTANTIVE` produces byte-identical output regardless of language, spelling, or the presence of "remember". Collect the outputs and assert they are all equal to each other, then drop lines 42-45.

apps/app/public/skills/setup
```diff
@@ -175,28 +186,67 @@ cwd = "/Users/YOUR_USERNAME"
...
 ```bash
 claude mcp add --scope user memwal -- npx -y @mysten-incubation/memwal-mcp
 ```
 
-Verify:
+If you use that fallback, tell the user the automatic-memory hooks are missing.
 
-```bash
-claude mcp list
+Then merge this block into `~/.claude/CLAUDE.md`. Create the file if it does
```
harrymove-ctrl: The only re-run guard here is prose: "If a Walrus Memory / `memwal_*` routing section is already there, leave it." That is fuzzy for an agent editing a global user file, and it makes future updates to the block impossible to apply cleanly.

Wrap the block in explicit markers, e.g. `<!-- memwal:start -->` / `<!-- memwal:end -->`, so re-runs replace it deterministically and the user can remove it in one shot. Also worth stating that the agent must show the block and get the user's go-ahead before writing to a global file outside the project. Same applies to the copy in `docs/mcp/claude-code.md:66`.

harrymove-ctrl (changes requested):
Follow-up with the findings pinned to the exact lines. Items 1 and 2 are blocking; the rest are fine as follow-ups.

ducnmm:
Addressed the two blocking items in 3fe1c9ce:

1. Signed-out `tools/list` (`SIGNED_OUT_TOOL_DEFINITIONS`) keeps conservative remember/recall wording. Signed-in cold start (bridge, `TOOL_DEFINITIONS`) still uses the sidecar proactive wording.
2. `memwal_recall` is now `readOnlyHint: true, destructiveHint: false` on the cold-start lists and the sidecar, so a later `tools/list_changed` does not put the destructive annotation back.

Also dropped the em dash the style audit flagged in `docs/mcp/reference.md`.

Items 3–7 left as follow-ups as you said.

ducnmm:
All review items on this PR are in:

**Blocking (3fe1c9ce)**
1. Signed-out `tools/list` is conservative; signed-in cold start stays proactive.
2. `memwal_recall` is `readOnlyHint: true`, `destructiveHint: false` on cold-start and sidecar.

**Follow-ups (f1c4f932)**
3. Full rubric once per session via `firstTime`; later turns get a one-line nudge.
4. `firstTime` is used again.
5. Quiet gate is 8 chars, so "Tui thích pnpm" is not skipped. Documented as deliberate.
6. Hook tests assert all fresh-session outputs are byte-identical, then that later turns get the nudge.
7. CLAUDE.md block is wrapped in `<!-- memwal:start -->` / `<!-- memwal:end -->`. Setup tells the agent to show the block and get go-ahead before writing the global file.

Style-audit em dash in `docs/mcp/reference.md` is gone.

Not done: the test-plan box for a stock-sonnet canary after CLAUDE.md routing. That still needs a human pass.

harrymove-ctrl (commented):
Re-reviewed at `3fe1c9c`. Both blocking items are fixed properly.

**1. Signed-out vs signed-in split — resolved.** `buildToolDefinitions(proactive)` with `TOOL_DEFINITIONS` / `SIGNED_OUT_TOOL_DEFINITIONS` is the right shape, and the wiring checks out: `handleAuthLine` serves the signed-out list, `bridge.ts:164` serves the proactive one, and `index.ts` only reaches `runBridge` once credentials load. The mid-session login handoff also lands on the proactive list after the client re-lists, which is the behavior you want. Nice that `tool-definitions.test.mjs` now pins both directions rather than just the proactive one.

**2. `memwal_recall` annotations — resolved.** Flipped in `annotations.ts`, the auth-required list, and both annotation tests, so the sidecar and the cold-start list agree.

One correction to my own earlier comment: I asked for `readOnlyHint: true` without accounting for the stale-row cleanup that the old comment described. Strictly, `readOnlyHint: true` claims the tool does not modify its environment, which is not quite true when recall prunes index rows for blobs Walrus has already dropped. `destructiveHint: false` alone gets you past the clients that gate proactive recall. Your new comment does say the cleanup is incidental, so this is a documented choice and I am fine with it as-is; flagging it only so the reasoning is on record if someone audits the annotations later.

### Still open (all non-blocking, as before)

- Item 3: the rubric still fires on every prompt over 20 chars, with the once-per-session `firstTime` dedup gone.
- Item 4: `firstTime` in `plugin/scripts/lib/hook-io.mjs:52` is still dead code now that `on_user_prompt.mjs` dropped it.
- Item 5: the 20-char gate still under-triggers on short Vietnamese statements.
- Item 6: `user-prompt-hook.test.mjs:42-45` still re-asserts substrings of the imported constant.
- Item 7: the `~/.claude/CLAUDE.md` merge in `apps/app/public/skills/setup:229` still has no `<!-- memwal:start/end -->` markers.

Happy for any of these to move to a follow-up issue. If you would rather land them here, say so and I will re-review.

### Before merge

CI for this push is still running, and the last test-plan box is unchecked: "Re-run Henry canary on stock sonnet after `~/.claude/CLAUDE.md` routing + built-in memory off". That canary is the acceptance test for the PR's central claim. Post the result and I will approve.

harrymove-ctrl (commented):
Re-reviewed at `f1c4f93`. Every item from my earlier reviews is now closed, and CI is green across all 14 checks. I ran the hook tests locally too: `user-prompt-hook.test.mjs` + `signals.test.mjs`, 6/6 pass.

- Item 3 — `firstTime("rubric", sessionId)` is back, full rubric once per session then `DECISION_RUBRIC_NUDGE`. That is the right shape.
- Item 4 — moot now, `firstTime` has a caller again.
- Item 5 — 8-char threshold with the reasoning in the comment, and `"terse Vietnamese preference is not skipped"` pins it against regression. Good.
- Item 6 — the rewritten test collects the outputs and asserts they are byte-identical before checking the constant, so it now fails independently if the hook starts branching.
- Item 7 — `<!-- memwal:start -->` / `<!-- memwal:end -->` markers plus "show this block, get the user's go-ahead" in both the skill and the docs. That is exactly what I was asking for.

Three small notes, none of them worth another round:

**1. Changelog does not mention the new hook behavior.** The entries cover the rubric and the signed-out split, but not the 8-char threshold or the per-session nudge. Both change what users see every turn, so one line each would help whoever reads the release notes.

**2. Version bump timing.** `packages/mcp/plugin` is at `0.0.10` in all four manifests plus the three `marketplace.json` files, and this PR does not touch them, which matches the repo's `chore(release): bump MCP package to ...` convention. Just flagging that the hook fix only reaches already-installed users after the next release bump, so this wants to be paired with one. Related: `CommandOSSLabs/walrus-memory-mcp-plugin#5` carries the same hook change on a separate `0.0.7` to `0.0.8` line. Two version lines for the same plugin will confuse whoever cuts the release; worth deciding which repo is the source of truth.

**3. Trivial.** `"ok thanks"` is 9 characters, so it now clears the 8-char gate and gets the nudge. The test moved to `"ok"`, which is correct for the new threshold. Harmless given how short the nudge is.

The only thing still holding my approval is the last test-plan box: "Re-run Henry canary on stock sonnet after `~/.claude/CLAUDE.md` routing + built-in memory off". Post that result and I will approve.

ducnmm:
Henry canary on `sonnet-5` (Claude CLI default switched from `fable-5`; `~/.claude/CLAUDE.md` routing block in place):

**Remember** (no memwal/save in the prompt):
> A few things about how I work: I always use pnpm, TypeScript strict mode, and my coffee order is a dirty chai. My Henry canary nickname is henry-sonnet-21.

- Model: `sonnet-5`
- Called `memwal_remember_bulk` unprompted
- Saved 4/4, including `henry-sonnet-21`
- No `MEMORY.md` write

**Recall** (new session, local MEMORY.md cleared):
> What do you remember about how I like to work, my coffee order, and my Henry canary nickname?

- Called `memwal_recall` first
- Returned **henry-sonnet-21** and dirty chai

Items 3–7 from the earlier review are also in `f1c4f932` if you want to re-review those here.

harrymove-ctrl (approved):
Approving at `f1c4f93`. This supersedes my earlier changes-requested review, which was the only thing leaving the PR in a blocked state.

All seven review items are resolved, CI is green on every check, and I re-ran the hook tests locally (6/6). The two blocking items from the first pass, the signed-out vs signed-in tools/list split and the `memwal_recall` annotations, are both fixed at the right layer with tests pinning them.

Not merging, that is yours to land.

One thing still on you before merge: the last test-plan box, "Re-run Henry canary on stock sonnet after `~/.claude/CLAUDE.md` routing + built-in memory off". Nothing in the diff blocks it, but that canary is the evidence for the PR's central claim, so please run it and post the result rather than letting the box ship unchecked.

The three minor notes from my previous review still stand as follow-ups, none blocking: no changelog line for the 8-char threshold and the per-session nudge, the plugin version bump needing to be paired with a release commit, and the duplicate version line in `CommandOSSLabs/walrus-memory-mcp-plugin#5`.

PR #713, opened 2026-08-20
fix(server): stop leaking sqlx migration locks through the Neon pooler

ducnmm:
Fixes [WALM-378](https://linear.app/mysten-labs/issue/WALM-378/bugserver-orphaned-sqlx-migration-advisory-lock-leaks-through-the-neon).

## What happened

Dev relayer was hard down 07:30–07:44 UTC. It booted, applied main DB migrations, then panicked:

```
Failed to initialize legacy security-delete database:
  legacy migration failed: canceling statement due to lock timeout
```

Railway reported SUCCESS/RUNNING (no healthcheck). The process never bound :3001, so every request was 502.

## Why it looks random, and why staging/prod stayed up

sqlx migrations take a **session-scoped** `pg_advisory_lock`. `LEGACY_DB_URL` and `DATABASE_URL` both go through Neon’s **transaction-mode pooler**. `LOCK` and `UNLOCK` can land on different backends; a client disconnect returns the backend to the pool **with the lock still held**. Apalis then reused that backend for job-queue traffic, so the lock never cleared.

On the next boot the migrator waits on that lock. A leaked `set_config('lock_timeout', '15s', false)` from `init_apalis_pool` (session GUC, also pooler-leaked) cancels the wait after 15s, and `.expect()` takes the process down.

This only fires **on restart**, and only if a previous boot left the lock on a still-live pooled backend:

| Env | Why it looked fine |
|---|---|
| **staging / prod** | Same code path, same pooler. They just weren’t redeployed into a held lock. Staging is on an older SHA and has been up; prod likewise. The next restart can hit this. |
| **dev** | 9 deploys that day. Redeploy at 07:44 recovered because the lock happened to be free — then leaked again. |

Not caused by #711 (7 TypeScript files, not even merged) or #706.

## Fix

1. **Migrations run on the direct compute host** (`*-pooler.*` → `*.`). Runtime query pools keep the pooled URL.
2. **Retry lock contention** (5s `lock_timeout`, 6 attempts, backoff) instead of a single `.expect()`.
3. **Stop setting session GUCs** (`set_config(..., false)`) on the pooled Apalis connections. Boot is still bounded by tokio timeouts. Apalis schema setup, when needed, also uses the direct endpoint.

No Railway env change required — the pooler hostname is rewritten in code.

## Tests

- URL rewrite: Neon pooler → direct, credentials/query preserved, non-pooler URLs untouched
- Lock-timeout / deadlock errors classified as retryable

### DIFF
--- a/services/server/src/main.rs
+++ b/services/server/src/main.rs
@@ -41,6 +41,7 @@ use jobs::{
 use services::{CompositeRanker, Embedder, Extractor, LlmExtractor, OpenAiEmbedder, Ranker};
 use storage::db::VectorDb;
 use storage::legacy_db::LegacyDb;
+use storage::postgres_url::direct_postgres_url;
 use types::{
     AppState, Config, KeyPool, DEFAULT_BLOB_CACHE_MAX_BYTES, DEFAULT_BLOB_CACHE_TTL_SECS,
     DEFAULT_EMBEDDING_CACHE_TTL_SECS,
@@ -471,33 +472,17 @@ async fn init_apalis_pool(
     database_url: &str,
     startup_timeout: std::time::Duration,
 ) -> Result<sqlx::PgPool, String> {
-    let statement_timeout = format!("{}ms", startup_timeout.as_millis().min(300_000));
-    let lock_timeout_ms = (startup_timeout.as_millis() / 3).clamp(1_000, 30_000);
-    let lock_timeout = format!("{}ms", lock_timeout_ms);
-
     tracing::info!(
         "  Apalis: connecting to PostgreSQL (startup_timeout={}s)",
         startup_timeout.as_secs()
     );
+    // Do not `set_config(..., false)` on this pool. Those are session GUCs;
+    // through a transaction-mode pooler they leak onto backends later reused
+    // by unrelated code (legacy sqlx migrate inherited lock_timeout=15s and
+    // panicked — WALM-378). Boot is still bounded by the tokio timeouts below.
     let pool_future = PgPoolOptions::new()
         .max_connections(10)
         .acquire_timeout(startup_timeout)
-        .after_connect(move |conn, _meta| {
-            let statement_timeout = statement_timeout.clone();
-            let lock_timeout = lock_timeout.clone();
-            Box::pin(async move {
-                sqlx::query(
-                    "SELECT set_config('statement_timeout', $1, false), \
-                            set_config('lock_timeout', $2, false), \
-                            set_config('idle_in_transaction_session_timeout', $1, false)",
-                )
-                .bind(statement_timeout)
-                .bind(lock_timeout)
-                .execute(conn)
-                .await?;
-                Ok(())
-            })
-        })
         .connect(database_url);
 
     let pool = match tokio::time::timeout(startup_timeout, pool_future).await {
@@ -529,7 +514,31 @@ async fn init_apalis_pool(
     }
 
     tracing::info!("  Apalis: running PostgreSQL migrations");
-    match tokio::time::timeout(startup_timeout, PostgresStorage::<()>::setup(&pool)).await {
+    let migrate_url = direct_postgres_url(database_url);
+    let setup_pool_future = PgPoolOptions::new()
+        .max_connections(1)
+        .acquire_timeout(startup_timeout)
+        .connect(migrate_url.as_ref());
+    let setup_pool = match tokio::time::timeout(startup_timeout, setup_pool_future).await {
+        Ok(Ok(setup_pool)) => setup_pool,
+        Ok(Err(err)) => {
+            return Err(format!(
+                "connect to PostgreSQL for Apalis migrations: {err}"
+            ))
+        }
+        Err(_) => {
+            return Err(format!(
+                "timed out after {}s connecting to PostgreSQL for Apalis migrations",
+                startup_timeout.as_secs()
+            ))
+        }
+    };
+    let setup_result = match tokio::time::timeout(
+        startup_timeout,
+        PostgresStorage::<()>::setup(&setup_pool),
+    )
+    .await
+    {
         Ok(Ok(())) => {
             tracing::info!("  Apalis: PostgreSQL migrations applied");
             Ok(pool)
@@ -539,7 +548,9 @@ async fn init_apalis_pool(
             "timed out after {}s running Apalis PostgreSQL migrations",
             startup_timeout.as_secs()
         )),
-    }
+    };
+    setup_pool.close().await;
+    setup_result
 }
 
 async fn apalis_schema_ready(pool: &sqlx::PgPool) -> Result<bool, sqlx::Error> {
--- a/services/server/src/main.rs
+++ b/services/server/src/main.rs
@@ -476,13 +476,38 @@ async fn init_apalis_pool(
         "  Apalis: connecting to PostgreSQL (startup_timeout={}s)",
         startup_timeout.as_secs()
     );
-    // Do not `set_config(..., false)` on this pool. Those are session GUCs;
-    // through a transaction-mode pooler they leak onto backends later reused
-    // by unrelated code (legacy sqlx migrate inherited lock_timeout=15s and
-    // panicked). Boot is still bounded by the tokio timeouts below.
+    // Runtime query bounds for this pool (job fetch/update, not migrate).
+    //
+    // `set_config(..., true)` (SET LOCAL) in after_connect is a no-op:
+    // after_connect runs under autocommit, so Postgres discards the GUC when
+    // that statement's implicit transaction ends — runtime queries would
+    // then be unbounded. Session-scoped (`false`) is the only after_connect
+    // form that actually sticks.
+    //
+    // `lock_timeout` is intentionally omitted. A leaked session lock_timeout
+    // on a transaction-mode pooler backend is what aborted sqlx migrate
+    // (15s wait → panic). Migrate now uses the direct host, but we still
+    // don't put lock_timeout on pooled backends. statement_timeout and
+    // idle_in_transaction_session_timeout bound queries without aborting
+    // lock waits; if they leak onto another pooler client they are still
+    // a bound, not a migrate-killer.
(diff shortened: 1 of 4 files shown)


ducnmm (commented):
## Review (Grok)

Direction is right: sqlx’s `pg_advisory_lock` is session-scoped, Neon `-pooler` is transaction-mode, and `UNLOCK`/disconnect does not reliably free the backend. Moving migrate onto a direct `PgConnection` + dropping session `set_config` on the pooled Apalis connections is the actual fix for WALM-378.

### What looks good
- Rewrite is host-only (`-pooler.` → `.`), credentials/query stay put, non-Neon URLs are left alone.
- Legacy migrate uses a single direct connection, then `pg_advisory_unlock_all()` even on failure — needed because sqlx only unlocks on the success path of `run_direct`.
- Apalis *setup* also goes direct; runtime pool stays pooled.
- Lock-timeout errors retry instead of a 15s `.expect()` death.

### Findings

**1. Residual outage if the lock is already leaked (medium)**  
`legacy_db.rs` — advisory locks are **database-wide**. A pooled backend that still holds sqlx’s migration lock will block the *direct* connection too. Retries (5s × 6 + backoff ≈ 45s) then still bubble into `main.rs:758` `.expect()` and take the process down.

This PR **stops new leaks**. It does not steal an already-held lock. First boot after deploy can still die until that pooled backend is recycled / `pg_terminate_backend`. Worth one log line with the `pg_locks` pid if `pg_advisory_lock` times out — the warn text already mentions it, but we never query the holder.

**2. Direct host must be reachable (medium, ops)**  
`postgres_url.rs` / `apply_legacy_migrations_once` — if Neon (or a firewall) only allows the pooler endpoint, **every** boot fails with `legacy db migrate connect failed`, even when no lock is held. That’s a worse failure mode than today’s intermittent one.

Please confirm from Railway that `ep-…c-2.ap-southeast-1.aws.neon.tech` (no `-pooler`) accepts the same user. If connect fails, include `direct_host` in the error so it’s obvious.

**3. Apalis lost its session timeouts (low)**  
`main.rs` `init_apalis_pool` — removing `set_config(..., false)` is correct (that’s how `lock_timeout=15s` leaked). Runtime Apalis queries now have no `statement_timeout` / `idle_in_transaction_session_timeout`. Boot is still tokio-bounded; a stuck job query is not. Acceptable for this PR; don’t put the GUCs back on the pooled connections.

### Not blockers
- `main.rs` still `.expect()`s after retries exhaust. Fine — can’t serve without legacy DB.
- VectorDb still migrates through the pooler via `raw_sql` (no sqlx advisory lock). Out of scope.
- Tests cover rewrite + retry classification; they don’t hit a live pooler, which is fine.

### Verdict
Approve the approach. Before merge, sanity-check that the **direct** Neon host is reachable from the relayer. After merge, one leftover leaked lock can still brick a single boot — terminate that backend if it happens, then this path should stay clean.

services/server/src/main.rs
```diff
@@ -471,33 +472,17 @@ async fn init_apalis_pool(
...
-
     tracing::info!(
         "  Apalis: connecting to PostgreSQL (startup_timeout={}s)",
         startup_timeout.as_secs()
     );
+    // Do not `set_config(..., false)` on this pool. Those are session GUCs;
+    // through a transaction-mode pooler they leak onto backends later reused
+    // by unrelated code (legacy sqlx migrate inherited lock_timeout=15s and
+    // panicked). Boot is still bounded by the tokio timeouts below.
     let pool_future = PgPoolOptions::new()
```
nikola0x0, line 495: **Major (discussion, not blocking):** this removes `statement_timeout` / `idle_in_transaction_session_timeout` from *every* connection this pool ever opens, not just the boot-time migration connection. This `PgPoolOptions` instance (`max_connections(10)`) is also the pool returned for ongoing Apalis job processing (`job_storage`, `remember_job_storage`, `bulk_job_storage` in `main.rs`) — so a stuck job query or an idle-in-transaction session on this pool is no longer bounded by anything Postgres-side once the process is past boot; only the `tokio::time::timeout` calls below cover boot, not runtime job traffic.

Since the leak vector is specifically `set_config(name, value, false)` (session-scoped, so it survives past the statement/transaction and can be inherited by the next pooler client on that backend), switching the third arg to `true` (transaction-local, i.e. `SET LOCAL` semantics) auto-resets each GUC at transaction end and can't leak through the pooler either — while keeping the timeout guard for Apalis's own long-lived runtime queries.

If dropping that protection for runtime job processing (not just migrations) was a deliberate simplification, no action needed — just flagging since it's a behavior change beyond the stated migration-lock fix.

services/server/src/main.rs
```diff
@@ -539,7 +548,9 @@ async fn init_apalis_pool(
             "timed out after {}s running Apalis PostgreSQL migrations",
             startup_timeout.as_secs()
         )),
-    }
+    };
+    setup_pool.close().await;
```
nikola0x0: **Minor:** `setup_pool.close()` isn't wrapped in a timeout. If the migration step above already hit the outer `tokio::time::timeout` because the connection is genuinely wedged (not just slow), this can block boot indefinitely trying to gracefully close it, instead of surfacing the already-computed `setup_result` error. Low severity — only matters once the network is already in a pathological state — but `tokio::time::timeout(startup_timeout, setup_pool.close())` would keep the same bounded-boot guarantee the rest of this function has.

services/server/src/storage/legacy_db.rs
```diff
@@ -44,6 +46,201 @@ impl LegacyDb {
...
+                tracing::warn!(pid, "legacy migration: pg_terminate_backend returned false")
+            }
+            Err(error) => {
+                tracing::warn!(pid, %error, "legacy migration: pg_terminate_backend failed")
+            }
+        }
+    }
+}
+
+fn migration_error_is_lock_contention(error: &AppError) -> bool {
```
nikola0x0: **Minor:** this lowercases and substring-matches the `Display` text of `AppError::Internal(String)`. By the time it's called, `apply_legacy_migrations_once` (line 125) has already discarded the original `sqlx::Error` — including its structured SQLSTATE via `error.as_database_error().and_then(|e| e.code())` — into a plain formatted string.

If a future sqlx/Postgres version rewords the error text, this classifier silently stops matching, the retry loop treats a transient lock-contention error as fatal on attempt 1, and boot goes straight back to the `.expect()` panic this PR is meant to avoid — with no signal beyond "the fix stopped working." Capturing the SQLSTATE code before the `map_err` into `AppError::Internal` (and classifying on that instead of/alongside the message text) would make this independent of exact wording.

nikola0x0 (approved):
**Pre-approve.** No blocking correctness bugs. I built the branch and ran the new tests locally (`cargo test --lib storage::`, `cargo check --bins`) — everything compiles and the 4 new/updated test modules pass. I also independently re-derived sqlx's advisory-lock ID scheme (`generate_lock_id` in sqlx-postgres) and the `pg_locks` bit-reconstruction: `classid`/`objid` are `oid` (unsigned), so the `(classid::bigint << 32) | objid::bigint` reconstruction in `release_orphaned_sqlx_migration_lock` is correct without a mask — I initially suspected a sign-extension bug there and disproved it by checking the Postgres column types and `SET_LOCKTAG_INT64` in the Postgres source. The direct-endpoint rewrite + retry/backoff design correctly addresses the actual leak mechanism (transaction-mode pooler returning a backend to the pool with a session-scoped lock still held).

**Majors (details inline)**

1. `init_apalis_pool` (main.rs:483) — removing `after_connect` entirely also drops `statement_timeout`/`idle_in_transaction_session_timeout` protection from Apalis's *runtime* connections (job processing for the life of the process), not just the migration step that actually caused the incident. Worth confirming that's intentional, since `set_config(name, value, true)` (transaction-local) would close the same pooler-leak vector without giving that up. See inline comment.

**Notes for whoever picks up related work**

- The retry + `pg_terminate_backend` self-healing path (`apply_legacy_migrations`, `release_orphaned_sqlx_migration_lock`) has no integration coverage — only the pure helpers (CRC/lock-id, error classification, URL rewrite) are unit-tested. Reasonable given how hard that is to test against a real pooler, just flagging the gap.
- The pre-existing `.expect("Failed to initialize legacy security-delete database")` at main.rs:782 (outside this diff) still panics if all 6 retry attempts are exhausted (worst case ≈ 45s: 6× up to 5s lock_timeout + ~15.5s of backoff). That's fine for the incident as described (a leaked pooler backend clears well within that), but if Railway's healthcheck/deploy timeout is tighter than that window, a *genuine* prolonged lock (e.g., two pods migrating concurrently) would still reproduce the original 502 symptom, just far less often. Worth a sentence in the PR description if that's already been checked against Railway's config.

**Minor (2)** — details inline: unbounded `setup_pool.close()`, and string-matching vs. structured SQLSTATE for retry classification.

services/server/src/main.rs
ducnmm: Done in bb1ed5db, with one correction: `set_config(..., true)` in `after_connect` is a no-op. after_connect runs under autocommit, so SET LOCAL is discarded when that statement's implicit transaction ends and runtime job queries would still be unbounded.

Restored session-scoped `statement_timeout` and `idle_in_transaction_session_timeout` on the runtime pool. `lock_timeout` stays omitted — that is the GUC whose leak aborted sqlx migrate. If these two remaining GUCs leak onto another pooler client they still bound queries rather than aborting lock waits.

services/server/src/main.rs
ducnmm: Done in bb1ed5db: `tokio::time::timeout(startup_timeout, setup_pool.close())`, warn and continue with `setup_result` if close is wedged.

services/server/src/storage/legacy_db.rs
ducnmm: Done in bb1ed5db. Classify on the `MigrateError` before mapping into `AppError`, using SQLSTATE `55P03` / `40P01` (`as_database_error().code()`). Message-text is only a fallback when the driver drops the code. Tests pin that a reworded `55P03` still retries and that `57014` (statement timeout) does not, even if the Display text says "lock timeout".

PR #718, opened 2026-08-21
fix(read-api): deletions, metadata refill, docs CI on #554

ducnmm:
Stacked on `feat/console-integration-test` ([#554](https://github.com/MystenLabs/MemWal/pull/554)).

## What this adds

**WALM-363 deletions.** Hard delete now (1) upserts `namespace_watermarks.last_mutated_at` so `/namespaces` incremental polling resurfaces the namespace even when `MAX(updated_at)` moved backwards, and (2) writes `memory_tombstones` so `/memories` returns `status: "deleted"`. `snapshot_version` stays a wire-format constant (now 3); it does not bump on data mutations.

**WALM-364 refill script.** `services/server/scripts/backfill-read-api-metadata.sh` plus `metadata_synced_at`. New writes already dual-write `agent_id`/`package_id`; this sweep fills pre-014 rows from sidecar `/walrus/query-blobs`. Rows with no on-chain agent id stay NULL but get `metadata_synced_at` so they are not retried.

**Nits from #554 review.** Unknown `scope` / `permissions` on `POST /v1/owner-tokens` is 400. `importance` is exposed on `MemoryItem`.

**Docs CI.** Colon-form routes so Docs/Freshness passes; frontmatter; no em dashes; deletion SLA (5-minute `/namespaces` poll sees the watermark; keep a 5-6h full reconcile as safety net).

## Intentionally not in this PR

- Rebase of #554 onto current `origin/dev` (124 commits; #700 not merged so 014-019 on #554 still unique vs `dev`). Do that on #554 itself before merging to `dev`.
- WALM-365 preview / title / tags / source URL: parked pending Thanos/George/Henry. Decrypting listings would reverse the Phase 1 no-plaintext AC.

### DIFF
--- a/services/server/scripts/backfill-read-api-metadata.sh
+++ b/services/server/scripts/backfill-read-api-metadata.sh
@@ -1,101 +1,148 @@
 #!/usr/bin/env bash
 # Refill agent_id / package_id on pre-migration vector_entries rows (WALM-364).
 #
-# New writes already dual-write those columns. Rows created before migration
-# 014 stay NULL forever unless this job runs. Batched, resumable, and never
-# on the read hot path.
-#
-# For each owner that still has unsynced rows, POST /walrus/query-blobs on the
-# sidecar (that payload already includes memwal_agent_id / memwal_package_id).
-# Matching blob_id stamps the columns and metadata_synced_at. Blobs with no
-# on-chain agent id stay NULL but still get metadata_synced_at so they are
-# not retried.
+# Safe for large mainnet sets (100k+ rows):
+#   - pages unsynced (owner, blob_id) from Postgres
+#   - asks the sidecar for THAT page only (blobIds), not the whole owner
+#   - stamps metadata_synced_at only on blob_ids requested in this page
+#   - never marks the rest of an owner "done"
+#   - HTTP / parse failures skip the page and leave rows NULL to retry
 #
 # Usage:
 #   DATABASE_URL=postgres://... \
-#   SIDECAR_URL=http://127.0.0.1:3001 \
+#   SIDECAR_URL=http://127.0.0.1:9000 \
 #   SIDECAR_SECRET=... \
 #     bash services/server/scripts/backfill-read-api-metadata.sh
 #
+# Optional: PAGE=100 SLEEP_SEC=0.25
 # Requires: psql, python3, curl.
 
 set -euo pipefail
+exec python3 - "$@" <<'PY'
+import json, os, subprocess, sys, time, urllib.error, urllib.request
 
-: "${DATABASE_URL:?set DATABASE_URL}"
-: "${SIDECAR_URL:?set SIDECAR_URL}"
-SIDECAR_SECRET="${SIDECAR_SECRET:-}"
-BATCH="${BATCH:-50}"
+db = os.environ.get("DATABASE_URL")
+sidecar = (os.environ.get("SIDECAR_URL") or "").rstrip("/")
+secret = os.environ.get("SIDECAR_SECRET") or ""
+if not db or not sidecar:
+    sys.exit("set DATABASE_URL and SIDECAR_URL")
 
-owners="$(psql "$DATABASE_URL" -Atqc "
-  SELECT DISTINCT owner
-  FROM vector_entries
-  WHERE metadata_synced_at IS NULL
-  ORDER BY owner
-  LIMIT ${BATCH};
-")"
+page = int(os.environ.get("PAGE") or "100")
+sleep_sec = float(os.environ.get("SLEEP_SEC") or "0.25")
+page = max(1, min(page, 200))
 
-if [ -z "$owners" ]; then
-  echo "nothing to backfill"
-  exit 0
-fi
-
-auth_args=()
-if [ -n "$SIDECAR_SECRET" ]; then
-  auth_args=(-H "X-Sidecar-Secret: ${SIDECAR_SECRET}")
-fi
-
-sql_escape() {
-  python3 -c 'import sys; print(sys.stdin.read().replace("'\''", "'\'\''"))' <<<"$1"
-}
-
-while IFS= read -r owner; do
-  [ -z "$owner" ] && continue
-  echo "owner ${owner}"
-  payload="$(python3 -c 'import json,sys; print(json.dumps({"owner": sys.argv[1]}))' "$owner")"
-  blobs_json="$(curl -sS "${SIDECAR_URL%/}/walrus/query-blobs" \
-    -H "content-type: application/json" \
-    "${auth_args[@]}" \
-    -d "$payload")"
-
-  python3 - "$DATABASE_URL" "$owner" "$blobs_json" <<'PY'
-import json, subprocess, sys
-
-db, owner, raw = sys.argv[1], sys.argv[2], sys.argv[3]
+def psql(sql: str) -> str:
+    r = subprocess.run(
+        ["psql", db, "-v", "ON_ERROR_STOP=1", "-At", "-c", sql],
+        check=True,
+        capture_output=True,
+        text=True,
+    )
+    return r.stdout
 
 def lit(value):
     if value is None or value == "":
         return "NULL"
     return "'" + str(value).replace("'", "''") + "'"
 
-try:
-    data = json.loads(raw)
-except json.JSONDecodeError:
-    print("sidecar returned non-json, skipping", owner, file=sys.stderr)
-    sys.exit(0)
-
-blobs = data.get("blobs") or []
-stmts = []
-seen = 0
-for blob in blobs:
-    blob_id = blob.get("blobId") or blob.get("blob_id")
-    if not blob_id:
-        continue
-    seen += 1
-    agent = blob.get("agentId") or blob.get("agent_id")
-    package = blob.get("packageId") or blob.get("package_id")
-    stmts.append(
-        "UPDATE vector_entries SET "
-        f"agent_id = COALESCE(agent_id, {lit(agent)}), "
-        f"package_id = COALESCE(package_id, {lit(package)}), "
-        "metadata_synced_at = NOW() "
-        f"WHERE owner = {lit(owner)} AND blob_id = {lit(blob_id)} "
-        "AND metadata_synced_at IS NULL;"
+def fetch_blobs(owner, blob_ids):
+    body = json.dumps({"owner": owner, "blobIds": blob_ids}).encode()
+    headers = {"content-type": "application/json"}
+    if secret:
+        headers["authorization"] = f"Bearer {secret}"
+    req = urllib.request.Request(
+        f"{sidecar}/walrus/query-blobs",
+        data=body,
+        headers=headers,
+        method="POST",
     )
-stmts.append(
-    "UPDATE vector_entries SET metadata_synced_at = NOW() "
-    f"WHERE owner = {lit(owner)} AND metadata_synced_at IS NULL;"
-)
-subprocess.check_call(["psql", db, "-v", "ON_ERROR_STOP=1", "-c", "\n".join(stmts)])
-print(f"  applied {seen} on-chain blob rows")
+    try:
+        with urllib.request.urlopen(req, timeout=120) as resp:
+            raw = resp.read().decode()
+            status = resp.status
+    except urllib.error.HTTPError as e:
+        raise RuntimeError(f"sidecar HTTP {e.code}: {e.read()[:300]!r}") from e
+    except urllib.error.URLError as e:
+        raise RuntimeError(f"sidecar transport: {e}") from e
+    if status < 200 or status >= 300:
+        raise RuntimeError(f"sidecar HTTP {status}: {raw[:300]!r}")
+    try:
+        data = json.loads(raw)
+    except json.JSONDecodeError as e:
+        raise RuntimeError(f"sidecar non-json: {raw[:300]!r}") from e
+    if not isinstance(data, dict) or "blobs" not in data:
+        raise RuntimeError(f"sidecar missing blobs: {raw[:300]!r}")
+    if data.get("sourceCapped") is True:
+        raise RuntimeError("sidecar sourceCapped=true; not stamping this page")
+    by_id = {}
+    for blob in data.get("blobs") or []:
+        blob_id = blob.get("blobId") or blob.get("blob_id")
+        if not blob_id:
+            continue
+        by_id[blob_id] = (
+            blob.get("agentId") or blob.get("agent_id") or None,
+            blob.get("packageId") or blob.get("package_id") or None,
+        )
+    return by_id
+
+updated = 0
+pages = 0
+errors = 0
+last_owner = ""
+last_blob = ""
+
+while True:
+    sql = (
+        "SELECT owner, blob_id FROM vector_entries "
+        "WHERE metadata_synced_at IS NULL "
+        f"AND (owner, blob_id) > ({lit(last_owner)}, {lit(last_blob)}) "
+        "ORDER BY owner, blob_id "
+        f"LIMIT {page}"
+    )
+    # First page: no cursor.
+    if pages == 0 and not last_owner:
+        sql = (
+            "SELECT owner, blob_id FROM vector_entries "
+            "WHERE metadata_synced_at IS NULL "
+            "ORDER BY owner, blob_id "
+            f"LIMIT {page}"
+        )
+    rows = [ln.split("|", 1) for ln in psql(sql).splitlines() if ln]
+    if not rows:
+        break
+    pages += 1
+    by_owner = {}
+    for owner, blob_id in rows:
+        by_owner.setdefault(owner, []).append(blob_id)
+    last_owner, last_blob = rows[-1]
+
+    for owner, blob_ids in by_owner.items():
+        try:
+            found = fetch_blobs(owner, blob_ids)
+        except Exception as e:
+            errors += 1
+            print(f"SKIP owner={owner} n={len(blob_ids)} err={e}", file=sys.stderr)
+            continue
+        stmts = ["BEGIN;"]
+        for blob_id in blob_ids:
+            agent, package = found.get(blob_id, (None, None))
+            stmts.append(
+                "UPDATE vector_entries SET "
+                f"agent_id = COALESCE(agent_id, {lit(agent)}), "
+                f"package_id = COALESCE(package_id, {lit(package)}), "
+                "metadata_synced_at = NOW() "
+                f"WHERE owner = {lit(owner)} AND blob_id = {lit(blob_id)} "
+                "AND metadata_synced_at IS NULL;"
+            )
+        stmts.append("COMMIT;")
+        psql("\n".join(stmts))
+        updated += len(blob_ids)
+        print(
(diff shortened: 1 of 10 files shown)


services/server/scripts/backfill-read-api-metadata.sh
```diff
@@ -0,0 +1,101 @@
...
+        "UPDATE vector_entries SET "
+        f"agent_id = COALESCE(agent_id, {lit(agent)}), "
+        f"package_id = COALESCE(package_id, {lit(package)}), "
+        "metadata_synced_at = NOW() "
+        f"WHERE owner = {lit(owner)} AND blob_id = {lit(blob_id)} "
+        "AND metadata_synced_at IS NULL;"
+    )
+stmts.append(
+    "UPDATE vector_entries SET metadata_synced_at = NOW() "
+    f"WHERE owner = {lit(owner)} AND metadata_synced_at IS NULL;"
```
ducnmm: This final UPDATE stamps every leftover unsynced row for the owner. If sidecar returns an error object or sourceCapped, those rows never retry. Only stamp blob_ids you observed.

services/server/src/storage/db.rs
```diff
@@ -1424,6 +1450,19 @@ impl VectorDb {
...
+             WHERE owner = $1 AND namespace = $2
+             ON CONFLICT (id) DO UPDATE SET deleted_at = NOW()",
+        )
+        .bind(owner)
+        .bind(namespace)
+        .execute(&self.pool)
+        .await
+        .map_err(|e| AppError::Internal(format!("Failed to tombstone namespace: {}", e)))?;
+        bump_namespace_watermark(&self.pool, owner, namespace).await?;
         let result = sqlx::query("DELETE FROM vector_entries WHERE owner = $1 AND namespace = $2")
```
ducnmm: Tombstone insert, watermark bump, and DELETE are three statements. If DELETE fails, UNION ALL in memories returns the same id as both active and deleted. Wrap in a transaction. On live insert, delete the tombstone for that id.

services/server/src/routes/memory_read.rs
```diff
@@ -810,6 +874,69 @@ mod tests {
         cleanup(&pool, &owner).await;
     }
 
+    #[tokio::test]
```
ducnmm, line 978: Test hand-writes the tombstone and watermark. Drive it through VectorDb::delete_by_blob_id or delete_by_namespace or it does not cover the production path.

ducnmm (commented):
Independent review before handing to Harry. Direction is right (WALM-363 watermark + tombstone, WALM-364 refill, token 400, docs freshness). I would not send this as-is.

## Must-fix

1. Backfill script false-completes an owner. After blob matches it always stamps metadata_synced_at for every leftover unsynced row. Sidecar error JSON or sourceCapped then permanently skips the rest of that owner. Only stamp rows whose blob_id you actually saw.

2. Tombstone + watermark + DELETE are not one transaction. A failed DELETE leaves tombstones next to live rows. /memories is UNION ALL, so the same id returns twice. Same if insert_vector reuses the id. Use a transaction; delete the tombstone on live insert.

3. Test does not call delete_by_namespace / delete_by_blob_id. It hand-inserts the tombstone.

4. OpenAPI not updated (importance, status deleted, snapshot_version 3).

## Also fix before Harry

- Docs replaced every em dash with semicolon, including list markers. Unreadable. Rewrite those sentences.
- Unused sql_escape in the refill script.
- insert_vector_plaintext does not bump the namespace watermark.

WALM-365 (preview/title/tags) correctly stays out.

ducnmm:
Follow-up self-review on e646f0eb:

Must-fix from the previous pass are addressed:
- Backfill stamps leftovers only after a complete (not sourceCapped) listing, and only for that owner. HTTP/parse/capped skips with rows left NULL.
- delete_by_namespace / delete_by_blob_id run tombstone + watermark + DELETE in one transaction. Live insert clears the tombstone. UNION ALL anti-joins live ids.
- Test now calls VectorDb::delete_by_blob_id.
- OpenAPI has importance, status=deleted, snapshot_version 3.

150k mainnet: one sidecar /walrus/query-blobs per owner (paging by blobId would re-walk every owned object). Sidecar then fetches metadata at concurrency 5, so a huge owner can take a long time. TIMEOUT_SEC defaults to 3600. SQL applies in chunks of 200. Re-run is safe (metadata_synced_at IS NULL).

Nits left (not blocking Harry): insert_vector_plaintext still not one transaction; docs still have some comma splices from the em-dash strip.

ducnmm:
Review of 2a28686f + must_resync fix (this commit):

Matches Linear WALM-363 approved design:
- memory_tombstones.memory_id PK, CTE DELETE RETURNING INSERT
- deleted[] additive, memories[] live-only, snapshot_version 2
- FULL OUTER JOIN live x tombstones for namespace watermark
- security-delete finalize writes tombstones
- 30-day sweep

Fixed in this commit: must_resync was using live updated_at as fallback, so a quiet account (last write >30d, polled 5 min ago) would must_resync every time. Now only when deleted_at on the cursor is older than retention. v1 cursors (no deleted_at) replay 30 days, no must_resync.

Still not in this PR (Linear AC, can follow):
- walm297_e2e_full.py deletion case
- sweep unit test
- must_resync unit test for an ancient deleted_at cursor

ducnmm:
@harrymove-ctrl could you review this when you have a moment?

Stacked on #554. It implements the WALM-363 design you and Henry signed off on 2026-08-18:

- `memory_tombstones` (separate table, not `deleted_at` on `vector_entries`)
- one CTE `DELETE … RETURNING` + `INSERT`
- additive `deleted[]` + `must_resync`; `memories[]` stays live-only; `snapshot_version` stays 2
- namespace watermark = GREATEST(live MAX(updated_at), tombstone MAX(deleted_at))
- 30-day retention sweep; security-delete finalization also writes tombstones

Also in this PR (so #554 can merge): docs freshness/style, owner-token 400 on unknown scope, WALM-364 refill script.

Linear: https://linear.app/mysten-labs/issue/WALM-363

jessiemongeon1:
## Style Guide Audit

**6** violation(s) found. All must be fixed before merge.

### `docs/api/memory-read-api.md` (3 violation(s))

_3 violation(s) (2 regex, 1 claude)_

- **Line 88** — Use present tense, not future
  - Current: `will`
- **Line 100** — No Latin abbreviations
  - Current: `e.g.`
  - Fix: `for example`
- **Line 100** — no-latin-abbreviations
  - Current: `e.g. `GET``
  - Fix: `for example, `GET``

### `docs/api/owner-token-auth.md` (3 violation(s))

_3 violation(s) (0 regex, 3 claude)_

- **Line 31** — passive_voice
  - Current: `receives a short-lived bearer token whose only permission is`
  - Fix: `Rewrite to active voice, for example: 'and the endpoint returns a short-lived bearer token...'`
- **Line 133** — quotation_marks
  - Current: `{"error": "owner has no MemWalAccount"}`
  - Fix: `Use backticks for error messages: `{"error": "owner has no MemWalAccount"}``
- **Line 134** — quotation_marks
  - Current: `{"error": "Phase 1 only grants memories.read"}`
  - Fix: `Use backticks for error messages: `{"error": "Phase 1 only grants memories.read"}``

---

services/server/src/storage/security_delete_store.rs
```diff
@@ -689,6 +689,25 @@ pub async fn finalize_batch_deleted(
...
     }
+    sqlx::query(
+        "WITH tracked AS (
+            SELECT owner, blob_id FROM delete_blobs_tracking
+            WHERE batch_id=$1 AND state='deleting'
+         ),
+         removed AS (
+            DELETE FROM vector_entries v
+            USING tracked t
+            WHERE v.owner = t.owner AND v.blob_id = t.blob_id
```
harrymove-ctrl: **Blocking — over-deletion across namespaces.**

This joins on `(owner, blob_id)` with no namespace predicate, but `blob_id` is not unique in `vector_entries`: `migrations/001_init.sql:17` creates a plain, non-unique index, and `VectorDb::delete_by_blob_id` deliberately takes `namespace` as a third parameter precisely because `(owner, blob_id)` does not identify a single row.

Scenario: owner `0xA` references the same blob from namespace `work` and namespace `personal`. A security delete is requested for `work`. `tracked` yields `(0xA, blob-1)`, this `DELETE ... USING` matches both rows, and both are destroyed and tombstoned. The `personal` memory is gone with no request behind it and no recovery path.

Fix: carry `namespace` on `delete_blobs_tracking` and add `AND v.namespace = t.namespace`, or join on the memory id if that is available at finalization time.

Second, separate point on the same hunk: the base branch's `security_delete_store.rs` contains no reference to `vector_entries` at all. This commit changes security-delete finalization from bookkeeping into a destructive purge of memory rows. That may well be correct (the blob is gone from Walrus, so the row is dangling), but the PR body describes this only as "security-delete finalization writes tombstones". Please call the purge out explicitly so a reviewer of the description sees it.

services/server/src/routes/memory_read.rs
```diff
@@ -448,22 +529,77 @@ pub(crate) async fn query_owner_memories(
...
+    let next_deleted = deleted_rows
+        .last()
+        .map(|(id, _, t)| (*t, id.clone()))
+        .or(deleted_after.clone());
+
+    let live_mark = rows
+        .last()
+        .map(|r| (r.4, r.0.clone()))
+        .or(after.clone())
+        .unwrap_or((snapshot_at, String::new()));
```
harrymove-ctrl: **Should fix — this fallback reintroduces the snapshot race.**

The `.unwrap_or((snapshot_at, String::new()))` advances the live watermark to the snapshot boundary in a case the guard above does not catch: first poll (`cursor: None`) that returns zero live rows but at least one tombstone. `rows.is_empty() && deleted_rows.is_empty() && after.is_none()` is false because `deleted_rows` is non-empty, so a cursor is issued at `updated_at = snapshot_at` with an empty id.

Now an insert that was in flight during the query commits with `updated_at` earlier than `snapshot_at`. The next poll's predicate is `(updated_at, id) > (snapshot_at, "")`, which that row fails, and it fails every later poll too since the cursor only moves forward. The memory is never delivered until a full resync.

This is the same failure mode the existing regression tests in this file cover; the previous code avoided it by returning `next_cursor: None` so the client restarted from scratch. Suggest widening the reset condition to `rows.is_empty() && after.is_none()` so a tombstone-only first page does not pin the live cursor.

services/server/src/routes/memory_read.rs
```diff
@@ -389,23 +444,46 @@ pub(crate) async fn query_owner_memories(
...
+        None => None,
+    };
+
+    // must_resync only when the tombstone cursor itself is older than
+    // retention: those tombstones have been swept. A v1 cursor (no
+    // deleted_at) replays 30 days instead. Do not use live updated_at;
+    // a quiet account's last live row can be months old while the client
+    // polled 5 minutes ago.
+    if let Some(c) = decoded.as_ref() {
+        if let Some(deleted_at) = c.deleted_at {
```
harrymove-ctrl: **Should fix — `must_resync` cannot fire for a v1 cursor, however stale.**

The guard only runs inside `if let Some(deleted_at) = c.deleted_at`. A cursor minted before this PR deserializes `deleted_at` to `None`, so the branch is skipped entirely and `must_resync` stays `false` no matter how old the cursor is.

Scenario: a Console instance with a v1 cursor goes offline for 60 days. `deleted_after` is `None`, `tombstone_start` falls back to `retention_cut`, and it replays only the last 30 days. Everything deleted between day 31 and day 60 was already removed by `sweep_expired_tombstones`, so no tombstone exists and no signal is sent. Those rows stay visible in Console indefinitely, which is the exact failure WALM-363 exists to close, and it leaves the "retention > max resync interval" acceptance criterion unmet.

The comment above is right to reject live `updated_at` as a staleness clock, but that currently leaves no clock at all on the v1 path. A `minted_at` field in the cursor would give you one: any cursor whose `minted_at` is older than retention, or that has no `minted_at`, sets `must_resync`.

If you would rather defer this, that is defensible for Phase 1, but please link a follow-up rather than letting the AC close silently.

services/server/src/main.rs
```diff
@@ -1206,6 +1206,13 @@ async fn main() {
             if let Err(e) = evict_state.db.prune_unconsumed_oauth_clients().await {
                 tracing::error!("MCP OAuth client pruning failed: {}", e);
             }
+            if let Err(e) = evict_state
+                .db
+                .sweep_expired_tombstones(std::time::Duration::from_secs(30 * 24 * 3600))
```
harrymove-ctrl: **Should fix — retention is defined twice with nothing tying the two together.**

This literal `30 * 24 * 3600` and `TOMBSTONE_RETENTION` at `memory_read.rs:88` are independent constants. They agree today, and that equality is what makes the deletion contract sound.

If someone later shortens the sweep to 7 days to control table growth and does not find `memory_read.rs:88`, the read path still trusts a 30-day window: `retention_cut` is set 30 days back, `must_resync` only fires past 30 days, but tombstones older than 7 days no longer exist. A client polling on day 10 gets an empty `deleted[]` and `must_resync: false`, and permanently keeps everything deleted between day 7 and day 10. No error anywhere.

Pass `TOMBSTONE_RETENTION` (or a shared `const`) into `sweep_expired_tombstones` rather than a literal at the call site, so the two cannot drift.

services/server/src/storage/db.rs
```diff
@@ -1258,6 +1285,11 @@ impl VectorDb {
             started.elapsed(),
         );
         result?;
+        sqlx::query("DELETE FROM memory_tombstones WHERE memory_id = $1")
```
harrymove-ctrl, line 1297: **Non-blocking — inconsistent with `insert_vector` right above.**

`insert_vector` was correctly wrapped in a transaction so the row insert and this tombstone clear commit together. `insert_plaintext` runs the same `DELETE` against `self.pool` with no surrounding transaction.

If the insert commits and this statement then fails on a transient connection error, the function returns `Err` but the row is live in `vector_entries` with its tombstone still present. The next `/memories` poll returns it in `memories[]` and its id in `deleted[]`, and a client that applies deletions after upserts drops a memory that exists.

Low impact since this is the benchmark path, but the asymmetry with the function twelve lines up is the kind that gets copied into the next writer.

services/server/src/routes/memory_read.rs
```diff
@@ -203,17 +216,43 @@ pub(crate) async fn query_owner_namespaces(
...
+            FROM vector_entries
+            WHERE owner = $1
+            GROUP BY namespace
+        ) l
+        FULL OUTER JOIN (
+            SELECT namespace, MAX(deleted_at) AS last_mutated_at
+            FROM memory_tombstones
+            WHERE owner = $1
+            GROUP BY namespace
+        ) w ON l.namespace = w.namespace
```
harrymove-ctrl, line 240: **Non-blocking — `/namespaces` has no staleness signal to match `/memories`.**

The `FULL OUTER JOIN` fixes the backwards-watermark bug, which was the point of the ticket. The remaining gap is the tail: once a namespace is emptied, it surfaces with `memory_count 0` for 30 days (actionable), and then the sweep drops its tombstones and it vanishes from the rollup entirely.

A client whose namespaces cursor predates the deletion never sees the count-0 row and never gets a removal signal, so the namespace stays in its local state with the last counts it saw. `/memories` grew `deleted[]` and `must_resync` for exactly this; `/namespaces` did not. Same root cause as the `must_resync` comment, so probably the same fix.

Minor while I am here: `GREATEST` already ignores NULLs in Postgres, so the nested `COALESCE` pair reduces to `GREATEST(l.last_updated_at, w.last_mutated_at)`.

harrymove-ctrl (changes requested):
## Code review (medium): 6 findings, 1 blocking

Reviewed this branch against its base `feat/console-integration-test`. A separate security-focused pass found nothing exploitable, and I want to record what I actually verified there rather than just saying "looks fine": owner authorization is the first statement in all four new handlers and `path_owner` is never used as a query input (every query binds `auth.owner`); `same_owner` canonicalizes through `Address::from_str`, so hex case, `0x` prefix, and zero-padding cannot slip past it; token verify uses `Mac::verify_slice` with expiry, audience, and scope checked independently, and an unconfigured secret is rejected in all three places; the only `format!`-built SQL interpolates a `const &str`, with every user value bound, including both cursor components; and no response `SELECT` touches a plaintext or embedding column.

On the design itself: moving to a separate `memory_tombstones` table instead of the `deleted_at` column WALM-363 originally specified is the right call, and the migration header makes the case well. Soft-delete would have put a filter obligation on 21 existing queries where a single miss leaks content. Good deviation.

What blocks merge is one finding.

### Blocking

**`security_delete_store.rs`: the new purge has no namespace predicate.** It joins on `(owner, blob_id)` only, but `blob_id` is a plain non-unique index (`001_init.sql:17`), and `delete_by_blob_id` takes namespace as a third parameter precisely because that pair does not identify one row. A security delete scoped to one namespace can destroy the same blob's rows in another. Separately, the base branch's `security_delete_store.rs` has no reference to `vector_entries` at all, so this commit turns finalization into a destructive purge while the PR body describes it only as "writes tombstones". Worth stating explicitly in the description either way.

### Should fix before merge

Two of these are small; the third is an acceptance criterion from the ticket.

* `memory_read.rs`: an empty live page that carries tombstones still advances the cursor to `now`, reintroducing the snapshot race this file already has regression tests for.
* `memory_read.rs`: `must_resync` never fires for a v1 cursor, which leaves WALM-363's "retention > max resync interval" AC open. Fix it or defer it explicitly with a linked ticket, but it should not land silent.
* `main.rs`: retention is two unlinked literals, and only their equality keeps the deletion contract sound.

### Non-blocking

* `db.rs`: `insert_plaintext` clears the tombstone outside a transaction while `insert_vector` twelve lines above was correctly wrapped.
* `memory_read.rs`: `/namespaces` gained monotonic watermarks but no `must_resync` equivalent.

Details inline. Also note this is stacked on #554, which still needs its rebase before either can land on `dev`, so merge readiness is gated there regardless.

ducnmm:
@harrymove-ctrl addressed your review on `3c46b4c2`:

**Blocking — namespace / over-delete.** `delete_blobs_tracking` has no namespace column (PK is `(owner, blob_id)`). The purge now joins `vector_entries` and deletes by **memory id**, so each tombstone carries that row's namespace. A shared `blob_id` across namespaces is still purged in every namespace: security-delete erases the Walrus blob, so those rows are dangling either way. The PR body now calls this purge out (it is a real behavior change vs the base branch). Carrying namespace on tracking would be a separate schema change to security-delete.

**Snapshot race.** Removed the `unwrap_or((snapshot_at, ""))` fallback. A tombstone-only first page returns `next_cursor: None` unless `deleted` still has more, in which case the live keyset stays at epoch so in-flight inserts are not skipped.

**v1 `must_resync`.** Cursors now carry `minted_at`. Missing (v1) or older than retention → `must_resync: true`. One-time full resync for COMG-568 after deploy.

**Retention drift.** One `TOMBSTONE_RETENTION` in `db.rs`; sweep takes no duration argument.

**insert_plaintext.** Same transaction as `insert_vector`.

**Namespaces GREATEST.** Dropped the redundant COALESCE pair. `/namespaces` still has no `must_resync` (your non-blocking note); same `minted_at` follow-up if we want it.

PR #719, opened 2026-08-21
fix(frontend): WALM-326 scope the dashboard no-key notice to this bro…

HoangDucBach:
## Ticket

[WALM-326](https://linear.app/mysten-labs/issue/WALM-326/issue-dashboard-no-delegate-key-in-this-browser-banner-conflicts-with) — the dashboard's "no delegate key in this browser" banner contradicts the delegate-keys list below it.

### `src/pages/Dashboard.tsx`

`isRecoveringExistingAccount` gated the banner on `!delegateKey` alone — the key in this tab's `sessionStorage` — and never read `onChainKeys`. It is replaced by `browserHasNoKey` plus `onChainKeyCount`, so the notice and the list draw from one fetch and cannot disagree. The copy names each scope: this browser has no key bound to it, the account has N for other clients.

A separate branch covers the empty account; the old copy said "revoke an old one below" with nothing to revoke. Render waits on `!isKeyListLoading`, or N shows a stale 0 and then jumps.

The list now carries the boundary itself rather than only describing it: the per-row badge changes from `current` to `This browser`, and the card subtitle gains the count.

### `src/index.css`

New `.dash-alert--info` variant — neutral panel, `Info` icon, 15px/400 instead of 21px/500 — so the notice no longer outweighs the list it points at. `hasMaxDelegateKeys` keeps the yellow `.dash-alert`, which is now reserved for real errors.

## Why is this needed?

Both surfaces said "delegate key" with nothing separating browser scope from account scope — the word "saved" carried all of it. An account with keys registered for other clients got a yellow warning above the fold, styled identically to the delegate-key-limit error, directly above a card listing those very keys. Users read the pair as a broken dashboard.

## Scope

The dashboard's no-key notice and the delegate-keys card.

### Out of scope

The notice says "Create a browser key" while the CTA below says "Create a delegate key". The issue's suggested copy is kept verbatim.

## How was this tested?

- [ ] Unit tests
- [ ] Integration tests
- [ ] End-to-end tests
- [x] Manual testing
- [ ] Not applicable

Manually on testnet. Existing suite 84/84; `tsc -b`, `eslint`, and `vite build` clean.

No tests added. The change is copy plus a CSS variant over three booleans derived from values that already existed — there is no new logic worth pinning.

Verified the two states the issue is about, and that they disagree correctly:

- Browser holds a key → no notice, badge on that row, subtitle reads "1 key registered…".
- `sessionStorage` cleared → notice names "1 delegate key for other clients", list shows the key with no badge.

Also stashed the diff and reloaded, to confirm the old yellow banner reproduces the reported contradiction.

Sign in, create a delegate key, then unbind this tab:

```js
sessionStorage.removeItem('memwal_delegate'); location.reload()
```

The notice should name the key count and no row should carry the `This browser` badge. Create a key again and both invert.

## Risks and dependencies

- **Two states are unverified.** An account with an on-chain Account object but zero keys, and the `This browser` badge below 640px with a long key name — it is wider than `current` and shares the row with an ellipsised name.
- **`.dash-alert--info` is appended at the end of an 11.9k-line stylesheet** and wins on source order over the `.dash-page .dash-alert` blocks above it. A later rule of equal specificity would silently take it back.

### Before (current dev) - Removed cache browser
Before (with cleared cache)

### After - Removed cache browser
After (with cleared cache)

## Author checklist

- [ ] CI is green.
- [ ] The branch is up to date with its target branch.
- [ ] I added or updated tests where appropriate.

### DIFF
--- a/apps/app/src/index.css
+++ b/apps/app/src/index.css
@@ -11930,3 +11930,58 @@ h1, h2, h3 {
     padding: 20px;
   }
 }
+
+.dash-page .dash-alert--info {
+  min-height: 0;
+  align-items: flex-start;
+  gap: 12px;
+  margin-bottom: 28px !important;
+  padding: 14px 18px;
+  border: 1px solid var(--dash-panel-border);
+  border-radius: var(--radius-md);
+  background: var(--dash-panel);
+  color: var(--dash-subtle);
+}
+
+.dash-page .dash-alert--info .dash-alert-icon {
+  width: 20px;
+  height: 20px;
+  margin-top: 2px;
+  color: var(--dash-yellow);
+}
+
+.dash-page .dash-alert--info p {
+  color: var(--dash-subtle);
+  font-size: 15px;
+  font-weight: 400;
+  line-height: 1.5;
+}
+
+.dash-page .dash-alert-link {
+  padding: 0;
+  border: 0;
+  background: none;
+  color: var(--dash-yellow);
+  font: inherit;
+  text-decoration: underline;
+  text-underline-offset: 2px;
+  cursor: pointer;
+}
+
+.dash-page .dash-alert-link:hover {
+  text-decoration-thickness: 2px;
+}
+
+.dash-page .dashboard-key-current-badge {
+  white-space: nowrap;
+}
+
+@media (max-width: 640px) {
+  .dash-page .dash-alert--info {
+    padding: 12px 14px;
+  }
+
+  .dash-page .dash-alert--info p {
+    font-size: 14px;
+  }
+}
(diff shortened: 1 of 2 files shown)


ducnmm:
Do you have a photo, please send it to the PR description @HoangDucBach

HoangDucBach:
> Do you have a photo, please send it to the PR description @HoangDucBach

Yarss, I'm writing now

HoangDucBach:
> Do you have a photo, please send it to the PR description @HoangDucBach

Pls check

ducnmm (approved):
Looks good — the description matches the diff.

The banner now keys off `onChainKeyCount` instead of only the tab's session key, so it can no longer contradict the list. Waiting on `!isKeyListLoading` avoids the 0→N flash. Demoting this to `.dash-alert--info` and renaming the badge to `This browser` is the right hierarchy/scope split for WALM-326.

Noted from the write-up and left as follow-ups, not merge blockers:
- empty-account and sub-640px badge states unverified
- `.dash-alert--info` wins on source order at the end of `index.css`
- "see below" `scrollIntoView` will tuck under the sticky dashboard nav (hash link + `scroll-margin-top` would be cleaner)

Verified the two states the ticket cares about. Approving.

PR #720, opened 2026-08-21
fix(server): make storage quota admission atomic with insertion (WALM-359)

### LINKED ISSUE #532: bug(server): concurrent `/api/remember/manual` requests bypass the per-owner storage quota [Security / Low]

harrymove-ctrl:
Fixes WALM-359. Closes #532.

## The bug

`check_storage_quota` read `SUM(blob_size_bytes)` inside a transaction holding `pg_advisory_xact_lock`, then committed. Advisory *xact* locks are transaction scoped, so the lock died on that commit: the quota comparison and the caller's `INSERT` both ran unprotected. Concurrent requests from one owner all observed the same pre-insert total, all passed, and all inserted.

The comment claiming that lock closed a TOCTOU race was wrong, and is removed.

## Reproduced before fixing

Real PostgreSQL, 1 GiB quota, owner seeded so exactly one more write fits, 20-way burst:

| Path | Admitted | Overshoot |
| -- | -- | -- |
| Inline (check → insert immediately) | 17 / 20 | +16 MiB |
| Enqueued (150 ms upload gap) | 20 / 20 | +19 MiB |
| After this change | **1 / 20** | none |

The 150 ms figure is generous to the old code. The real check-to-insert window on the enqueued paths spans a Walrus upload, documented in `types.rs` as running up to five minutes.

## Approach

Admission now records a **reservation** inside the same transaction that holds the lock, so the intent to write is committed before the lock is released. Usage during admission is `SUM(vector_entries.blob_size_bytes)` plus live reservations, which is what lets a later request in a burst see an earlier one's bytes even though no row exists yet.

A single transaction spanning check and insert would close the inline paths, but not the enqueued ones: those admit, queue a wallet job, upload to Walrus, and insert minutes later, and a Postgres transaction cannot be held across that. A reservation table covers both groups on one mechanism instead of letting them diverge.

**Reservations are keyed by the caller's own id**, not a generated one:

* enqueued paths use the `remember_jobs.id` the vector row will carry, so `jobs.rs` releases from an id it already has. Nothing new is threaded through the serialized `WalletOperation` payloads, so there is no job-payload compatibility risk for in-flight jobs.
* inline paths mint a local UUID and release it a few lines later.

**Release happens after the row is committed, never before.** That briefly double counts the same bytes, which is the safe direction; releasing first would open a window where neither the row nor the reservation counts.

## Not stranding quota

Three layers, so a missed release can never make an account permanently unusable:

1. Explicit release on the success path and on terminal wallet-job failure.
2. A periodic reconcile that drops reservations whose `remember_jobs` row is already `done` or `failed`. This is what keeps the guarantee from depending on nobody ever missing a call site.
3. A 15 minute TTL as the backstop, chosen to exceed `STALE_REMEMBER_JOB_AFTER` (10 min), after which the sweeper fails the job and releases anyway.

Both new sweeps run on the existing 60 s stale-job tick, ordered after it so rows it just failed are reconciled in the same pass.

The worst case is bounded over-counting, never a locked-out account.

## Call sites covered (5)

Inline: `remember_manual`, analyze benchmark branch.
Enqueued: `remember()`, `remember_bulk`, analyze main path.

Batch admission is all-or-nothing, so an over-quota bulk or analyze request is still rejected whole. The `402` / "Storage quota exceeded" response shape is unchanged.

## Migration

`014_storage_reservations.sql` adds the `storage_reservations` table. Additive, `CREATE TABLE IF NOT EXISTS`, no change to existing tables, safe to roll back by leaving the table in place.

## Tests

Six real-Postgres tests in `storage::db::quota_admission_tests`, `#[ignore]`d in line with the existing convention for tests needing external services:

```bash
DATABASE_URL=postgresql://... cargo test quota_admission_tests -- --ignored
```

They cover the 20-way burst on both the inline path and the enqueued path with a delayed insert, plus reservation-counts-before-the-row, release-frees-immediately, TTL expiry, and all-or-nothing batching.

**These were checked against the broken code before being committed**: reintroducing the old read-commit-compare-insert path makes the two burst tests fail with the 17/20 and 20/20 numbers above, so they are not vacuous.

## Verification

* 937 unit tests pass, plus the 6 new ones.
* `cargo clippy` introduces no new real lints. The only delta against `dev` is dead-code `never used` entries for the new public items, which is the same lib-target artifact that already lists `check_storage_quota` on `dev` today.
* `cargo fmt` clean, no reformatting of untouched files.
* Docker build **not verified locally**: it fails on this Apple Silicon host during the dependency prebuild step, where amd64 `rustc` segfaults under QEMU. That step consumes only `Cargo.toml` / `Cargo.lock`, neither of which this PR touches, so the failure is environmental. Worth a look at CI's build result before merging.

## Note on the ticket

The ticket cites `security_delete_store.rs` as in-repo precedent for the fix pattern. That file does not exist on `dev` (it appears to be from an unmerged branch). The pattern is still right, it just is not shipped precedent.

### DIFF
--- a/services/server/migrations/014_storage_reservations.sql
+++ b/services/server/migrations/014_storage_reservations.sql
@@ -1,0 +1,55 @@
+-- Walrus Memory — Storage Quota Reservations
+--
+-- Closes the quota admission race (GH #532 / WALM-359).
+--
+-- Before this table, admission read SUM(blob_size_bytes) under a
+-- pg_advisory_xact_lock and then COMMITTED that transaction. Advisory
+-- *xact* locks are transaction scoped, so the lock died on that commit and
+-- the caller's INSERT happened unprotected. Concurrent requests from one
+-- owner all read the same pre-insert total, all passed, and all inserted.
+--
+-- A reservation is the missing "intent to write" record. Admission now
+-- inserts the reservation inside the same transaction that holds the lock
+-- and reads usage, so check and commit-of-intent are atomic. Usage during
+-- admission is SUM(vector_entries.blob_size_bytes) + SUM(unexpired
+-- reservations.bytes).
+--
+-- Why a table rather than one long transaction: the enqueued paths
+-- (/api/remember, /api/remember/bulk, /api/analyze) admit the request, queue a
+-- wallet job, upload to Walrus, and only then insert the row. That window is
+-- documented in types.rs as running up to five minutes. A Postgres
+-- transaction cannot be held across it. Inline paths could use a single
+-- transaction, but sharing one mechanism keeps both groups on the same
+-- accounting rather than letting them diverge.
+--
+-- `id` is the caller's own identifier, not a generated key:
+--   * enqueued paths use the `remember_jobs.id` the row will be inserted with
+--   * inline paths use a freshly minted UUID
+-- Keying by an id the caller already owns means release sites need no extra
+-- plumbing through the serialized wallet-job payloads, and makes release
+-- idempotent: a retried job that already released simply deletes nothing.
+-- There is deliberately no FK to remember_jobs, since inline reservations
+-- have no job row.
+--
+-- `expires_at` is the backstop required by the acceptance criteria. If a
+-- release is ever missed (process killed between admission and insert, job
+-- lost, task cancelled), the reservation stops counting at expiry. The
+-- failure mode is therefore temporary over-counting, never a permanently
+-- unusable account.
+
+CREATE TABLE IF NOT EXISTS storage_reservations (
+    id         TEXT PRIMARY KEY,
+    owner      TEXT   NOT NULL,
+    bytes      BIGINT NOT NULL,
+    expires_at TIMESTAMPTZ NOT NULL,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+-- Admission reads live reservations for exactly one owner, so lead with owner
+-- and carry expires_at for the range predicate.
+CREATE INDEX IF NOT EXISTS idx_storage_reservations_owner_expires
+    ON storage_reservations (owner, expires_at);
+
+-- The periodic sweeper scans by expiry across all owners.
+CREATE INDEX IF NOT EXISTS idx_storage_reservations_expires
+    ON storage_reservations (expires_at);
--- a/services/server/src/jobs.rs
+++ b/services/server/src/jobs.rs
@@ -196,6 +196,18 @@ async fn update_remember_job_after_wallet_error(
         "running"
     };
 
+    // Terminal means no later attempt will ever insert the row, so the bytes
+    // this job reserved at admission must go back to the owner now rather than
+    // waiting out the TTL. Retryable errors keep the reservation: the next
+    // attempt still intends to write those bytes, and releasing early would let
+    // a concurrent burst overcommit while the retry is in flight.
+    //
+    // Safe to run even when a concurrent attempt already won and released:
+    // release is a delete by id, so a second call is a no-op.
+    if error.aborts_retries() {
+        crate::storage::db::release_storage_reservations_with_pool(pool, &[jid.to_string()]).await;
+    }
+
     // Guard against downgrading a row a concurrent attempt already finished.
     // Every pre-existing caller only runs while holding the per-job upload
     // lock (JobUploadLock), so it was never the current status writer's own
@@ -841,6 +853,12 @@ async fn insert_vector_and_mark_remember_done(
         .bind(jid)
         .execute(state.db.pool())
         .await;
+
+        // Release only now that the vector row is committed. The row carries
+        // the bytes from here on, so holding the reservation would double count
+        // them; releasing before the insert would instead open a window where
+        // neither counts and a concurrent burst could slip past the quota.
+        crate::rate_limit::release_storage_quota_one(state, jid).await;
     }
 
     tracing::info!(
--- a/services/server/src/main.rs
+++ b/services/server/src/main.rs
@@ -1310,6 +1310,28 @@ async fn main() {
             {
                 tracing::error!("Stale remember job sweep failed: {}", e);
             }
+
+            // Storage-quota reservation upkeep, on the same tick.
+            //
+            // Ordered after the stale-job sweep so rows it just marked failed
+            // are reconciled in the same pass instead of waiting a full minute.
+            // The reconcile handles terminal jobs whose release was missed; the
+            // expiry sweep is the last-resort backstop for reservations with no
+            // job row at all (inline paths) or whose job vanished entirely.
+            if let Err(e) = stale_job_state
+                .db
+                .release_reservations_for_terminal_jobs()
+                .await
+            {
+                tracing::error!("Terminal-job reservation reconcile failed: {}", e);
+            }
+            if let Err(e) = stale_job_state
+                .db
+                .sweep_expired_storage_reservations()
+                .await
+            {
+                tracing::error!("Expired reservation sweep failed: {}", e);
+            }
         }
     });
 
--- a/services/server/src/rate_limit.rs
+++ b/services/server/src/rate_limit.rs
@@ -6,10 +6,12 @@ use axum::{
 };
 use percent_encoding::percent_decode_str;
 use std::sync::Arc;
+use std::time::Duration;
 use uuid::Uuid;
 
 use crate::{
     client_ip::canonical_client_ip,
+    storage::db::{StorageAdmission, StorageReservationRequest},
     types::{AppError, AppState, AuthInfo},
 };
 
@@ -618,13 +620,42 @@ pub async fn rate_limit_middleware(
 /// Storage tracking still uses PostgreSQL (it's per-row in vector_entries).
 /// Returns `Ok(())` if within quota, `Err(AppError::QuotaExceeded)` if not.
 ///
-/// Uses PostgreSQL advisory lock per-owner to prevent
-/// TOCTOU race where concurrent requests all pass quota check then
-/// all write, collectively exceeding the limit.
-pub async fn check_storage_quota(
+/// How long an unreleased storage reservation keeps counting against quota.
+///
+/// This is the backstop for a release that never happens: process killed
+/// between admission and insert, task cancelled, job lost. It must be longer
+/// than any window in which the write could still legitimately land, or a slow
+/// upload would have its reservation expire while the row is still coming and
+/// a concurrent burst could slip past the quota.
+///
+/// The longest such window is bounded by `main::STALE_REMEMBER_JOB_AFTER`
+/// (10 minutes), after which the sweeper marks the job failed and releases the
+/// reservation anyway. 15 minutes leaves margin over that without letting a
+/// leak sit around long. Over-counting for at most this long is the intended
+/// failure mode; an account is never permanently stranded.
+pub const STORAGE_RESERVATION_TTL: Duration = Duration::from_secs(15 * 60);
+
+/// Atomically admit pending writes against the owner's storage quota.
+///
+/// Replaces the old `check_storage_quota`, which was not atomic: it read usage
+/// under a `pg_advisory_xact_lock`, committed (releasing the lock, because xact
+/// locks are transaction scoped), compared outside any lock, and left the
+/// caller to insert later still. Concurrent requests from one owner all read
+/// the same pre-insert total, all passed, and all inserted (GH #532 /
+/// WALM-359).
+///
+/// Admission now records the intent to write inside the locked transaction, so
+/// a later request in the same burst sees the earlier one's bytes even though
+/// no row exists yet. Callers must release once the row lands or the write is
+/// known to be dead — see `release_storage_quota`.
+///
+/// Admission is all-or-nothing across `reservations`, preserving the existing
+/// behaviour where a bulk request lands whole or is rejected whole, and the
+/// `402` / "Storage quota exceeded" response shape is unchanged.
+pub async fn reserve_storage_quota(
     state: &AppState,
     owner: &str,
-    additional_bytes: i64,
+    reservations: &[StorageReservationRequest],
 ) -> Result<(), AppError> {
     let max_bytes = state.config.rate_limit.max_storage_bytes;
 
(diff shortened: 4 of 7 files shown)


ducnmm (approved):
The diagnosis is right: `pg_advisory_xact_lock` dying on commit made the old check a no-op against a burst. A reservation table is the correct shape for the enqueued paths (cannot hold a Postgres transaction across Walrus upload).

What I checked:
- Lock, usage read (`vector_entries` + live reservations), compare, and INSERT all sit in one transaction.
- Five call sites reserve then release: `remember`, `remember_bulk`, `remember_manual`, analyze main, analyze benchmark. Release is after the row commits, never before.
- Terminal job failure releases; retryable errors keep the reservation. Reconcile + 15 min TTL sit on the stale-job tick after `fail_stale_remember_jobs`.
- All-or-nothing batch admission; 402 shape unchanged.
- `ON CONFLICT (id)` keeps retried jobs from double-counting themselves.
- Tests were run against the broken path first (17/20 and 20/20), so they are not vacuous.

Non-blocking:
- This migration is `014_storage_reservations.sql`. Console read-API (#554 / #718) also introduces a 014. Fine against current `dev`; renumber when those merge.
- Close #700 as superseded by this PR.

Approve.

PR #721, opened 2026-08-21
fix(sdk): accept a bech32 suiprivkey delegate key, not just hex

ducnmm:
`MemWal.create({ key })` fed the key straight into `hexToBytes`, which only accepts hex. A Sui `suiprivkey1...` string therefore died at construction with "hexToBytes: input contains non-hex characters", before any request went out.

That is what broke the JS SDK e2e job against the dev relayer: all 11 authenticated tests failed while the 7 no-auth ones passed, because the shared BENCH_DELEGATE_KEY secret is stored in bech32. The Python SDK reads that same secret and passes, since it normalizes the key first (`memwal/utils.py` `normalize_private_key`) — and its docstring already claimed the TypeScript SDK took both forms. It did not. This closes that gap rather than working around it in CI, because the two forms are both in circulation: `sui keytool` and wallets hand out bech32, while `generateDelegateKey()` returns hex.

Port `normalize_private_key` to `utils.ts` and call it everywhere a user-supplied delegate key is parsed: both client constructors and the two `delegateKeyTo*` helpers. `account.ts` is untouched — its `hexToBytes` calls take public keys, where suiprivkey does not apply.

The bech32 decoder is hand-rolled rather than imported from `@mysten/sui`, for the reason `u64ToLeHex` gives: the core path keeps `@mysten/*` as peer dependencies so the client works without them, and the constructor is synchronous so it cannot await a dynamic import. The Python SDK hand-rolls it for the same reason. Verified against `@mysten/sui` on 2000 random keys.

### DIFF
--- a/packages/sdk/src/manual.ts
+++ b/packages/sdk/src/manual.ts
@@ -43,6 +43,7 @@ import {
     sha256hex,
     hexToBytes,
     bytesToHex,
+    normalizePrivateKey,
     u64ToLeHex,
     normalizeServerUrl,
     sanitizeServerError,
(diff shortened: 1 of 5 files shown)


ducnmm (commented):
Looks correct. `hexToBytes` was the actual e2e failure mode (bech32 `BENCH_DELEGATE_KEY`), and this matches the Python SDK's `normalize_private_key` instead of special-casing CI.

Call sites are complete: both constructors plus `delegateKeyToPublicKey` / `delegateKeyToSuiAddress`. `account.ts` staying hex-only is right (public keys). Hand-rolled bech32 is justified (sync constructor, `@mysten/*` peer). Tests cover hex / `0x` / bech32 / whitespace / secp256k1 reject / checksum / both clients.

CI green. Approve.

ducnmm:
@nikola0x0 @harrymove-ctrl I cannot approve this PR (GitHub: author is me). Code review is LGTM — need one of you to click Approve so it can merge. CI is green.

nikola0x0 (approved):

PR #725, opened 2026-08-21
fix(app,mcp): WALM-333 explain a failed login hand-off in plain language

HoangDucBach:
## Ticket

[WALM-333](https://linear.app/mysten-labs/issue/WALM-333/issue-login-callback-failure-message-is-opaque-to-first-time-users) — the login callback failure message is opaque to first-time users, and the MCP client gets no signal to retry.

### `apps/app/src/pages/ConnectMcp.tsx`

`callbackDelivered: boolean` becomes a `CallbackOutcome`, so a listener that refused the hand-off is told apart from one that never answered. The failure copy drops "listener", "callback" and the loopback URL, states that the on-chain key is already registered and needs no repeating, names the likely cause per outcome, and links troubleshooting. The URL moves to a detail row for anyone debugging.

The heading follows the outcome too. It previously showed a green tick and "MCP client connected" on failure, one line above the text explaining the failure.

### `packages/mcp/src/auth-required.ts`, `src/bridge.ts`

A background sign-in that never completes records the reason and prepends it to the next tool call's not-signed-in error, plus a `notifications/message` warning. `MEMWAL_MCP_LOGIN_TIMEOUT_MS` makes the 5-minute listener deadline overridable, mostly for tests.

### `docs/`

The troubleshooting section the ticket asks to link — none existed — and the new env var.

## Why is this needed?

Sign-in has two halves: the browser registers a delegate key on chain, then hands it to a listener on `127.0.0.1`. The first half is irreversible; the second is what fails. The old message described the second in listener/callback terms, so a first-time user could not act on it, and the key stays registered and unusable either way.

The MCP side was silent by construction: `memwal_login` returns the URL immediately with `isError: false`, so when the flow later fails there is no pending response to turn into an error. The next tool call is the first place left to say so.

## Scope

The failure surface of the login hand-off, in the browser and in the MCP client.

### Out of scope

Nothing recovers the orphaned on-chain key — the user signs in again and the old key sits there until revoked from the dashboard. That is product behaviour, not this message.

## How was this tested?

- [x] Unit tests
- [x] Integration tests
- [ ] End-to-end tests
- [ ] Manual testing
- [ ] Not applicable

`apps/app` 88/88, `packages/mcp` 44/44, `tsc` and `eslint` clean.

New `ConnectMcp.test.tsx` drives the real page through preflight, consent and signing — stubbing only the wallet and `fetch` — and asserts all three outcomes. New `login-failure-notice.test.mjs` spawns the binary, lets the sign-in time out with nobody opening the URL, and asserts both the warning and the next tool call's text.

Verified both earn their place: restoring the unconditional heading fails exactly the unreachable case, and stubbing `loginFailureNotice()` to `""` fails the MCP test.

Worth checking rather than assuming: `auth-required.ts` already documents that Claude Code drops `notifications/message`, which is why the tool-call text is the primary channel and the notification only secondary.

```bash
cd apps/app && npx vitest run src/pages/ConnectMcp.test.tsx
cd packages/mcp && npm test

# watch it catch the bug: make the SuccessCard heading unconditional,
# then re-run — only the unreachable case fails
```

## Risks and dependencies

- **`MEMWAL_MCP_LOGIN_TIMEOUT_MS` is a new public env var.** A short value shortens the real sign-in window; values below 100 fall back to the default.
- **The MCP notice only lands if the agent makes another tool call.** A user who walks away after a failed sign-in sees nothing; the notification covers clients that surface those, which Claude Code does not.
- **`packages/mcp` has pre-existing flaky tests** in `coldstart-*` and `initialize-id-reuse`. Baseline on this branch failed 5 of them; they pass in isolation and passed on the final full run.

## Author checklist

- [ ] CI is green.


jessiemongeon1:
## Style Guide Audit

apps/app/src/pages/ConnectMcp.tsx
```diff
@@ -548,34 +552,82 @@ function ConsentCard({
...
             )}
-            {callbackDelivered === false && (
-                <p className="setup-classic-description" style={errorTextStyle}>
-                    The on-chain registration succeeded, but the local MCP login listener at{' '}
-                    <code style={codeStyle}>http://127.0.0.1:{port}/callback</code> did not accept the callback. Restart the MCP login command and try again so credentials can be saved locally.
-                </p>
+            {!delivered && callbackOutcome !== null && (
+                <>
+                    <p className="setup-classic-description">
+                        Your key was created and registered to your account, so that part is done
```
ducnmm: **[suggestion]** After rebase, keep `CallbackOutcome` and these two cause paragraphs, and add the #746 revoke line, e.g. "The unused key from this attempt is already on your account — remove it from the dashboard if you are not using it."

"You do not need to repeat it" then "Sign in again" is easy to misread: the second attempt **does** register a **new** key.

apps/app/src/pages/ConnectMcp.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+        })
+        await signIn()
+
+        expect(await screen.findByText(/Almost there/i)).toBeInTheDocument()
+        expect(screen.queryByText(/MCP client connected/i)).not.toBeInTheDocument()
+        expect(screen.getByText(/Nothing answered on your computer/i)).toBeInTheDocument()
+        expect(screen.getByText(/second attempt usually works/i)).toBeInTheDocument()
+    })
+
+    it('tells a refused hand-off apart from an absent one', async () => {
```
ducnmm, line 126: **[suggestion]** Unreachable asserts `/Almost there/i` and that "MCP client connected" is absent — that is the original bug. Refused only checks the cause sentence. If a rebase put the green tick back on HTTP 409, this test would still pass.

Also assert the failure heading and `queryByText(/MCP client connected/i)` is null on 409.

ducnmm (commented):
## Rebase required

GitHub reports this PR **CONFLICTING** with current `dev`. Merge-tree shows three files changed on both sides:

- `apps/app/src/pages/ConnectMcp.tsx` — conflict with **#746** (merged yesterday). #746 rewrote the same `SuccessCard` heading/fail copy ("On-chain key added… remove it from the dashboard").
- `packages/mcp/src/auth-required.ts`
- `packages/mcp/src/bridge.ts`

@HoangDucBach please rebase onto current `dev` before this can merge.

When you resolve `ConnectMcp.tsx`, **keep this PR's `CallbackOutcome` model** (`delivered` / `unreachable` / `rejected`) — it is the better split — and **fold in #746's dashboard-revoke sentence**. Taking #725 wholesale drops "If you are not using it, remove it from the dashboard". Taking #746 wholesale drops the unreachable-vs-rejected copy and the tests that catch the green-tick-on-failure bug.

## Review of the intended change

Direction is right: heading no longer says "MCP client connected" on failure, jargon is out of the prose, MCP prepends `loginFailureNotice()` on the next tool call, and `ConnectMcp.test.tsx` / `login-failure-notice.test.mjs` actually fail if you restore the old heading / empty notice.

One extra note not inlined (line not in the diff): docs present `MEMWAL_MCP_LOGIN_TIMEOUT_MS` as the listener deadline, but only `auth-required.ts` reads it. `handleLocalLogin` in `bridge.ts` still hard-codes `LOGIN_BG_TIMEOUT_MS`. Share `resolveLoginTimeoutMs()` there too.

I will re-review the rebase, not the whole design.

apps/app/src/pages/ConnectMcp.tsx
HoangDucBach: Done. Reworded to drop the misleading "you do not need to repeat it" (the second sign-in does register a new key) and added the revoke line: "The unused key from this attempt is already on your account. Remove it from the dashboard if you are not using it." `ConnectMcp.test.tsx` updated to match.

apps/app/src/pages/ConnectMcp.test.tsx
HoangDucBach: Done. Added both assertions to the 409 case: /Almost there/i present and queryByText(/MCP client connected/i) null. Restored the unconditional heading locally to confirm this test now catches it too (2 failures instead of the previous 0), then reverted.

HoangDucBach:
Also addressed the non-inlined note: moved resolveLoginTimeoutMs() into login.ts and had both auth-required.ts and handleLocalLogin in bridge.ts import it, so MEMWAL_MCP_LOGIN_TIMEOUT_MS now controls both listener deadlines instead of just the auth-required one.

apps/app/src/pages/ConnectMcp.test.tsx
HoangDucBach: fixed

apps/app/src/pages/ConnectMcp.tsx
HoangDucBach: fixed

apps/app/src/pages/ConnectMcp.tsx
```diff
@@ -548,42 +552,86 @@ function ConsentCard({
...
+                <>
+                    <p className="setup-classic-description">
+                        The first half of sign-in succeeded: your key was created and registered
+                        to your account. What did not work is the last step: this page could not
+                        pass the key back to the Walrus Memory app running on your computer, so
+                        your MCP client cannot use it yet.
+                    </p>
+                    <p className="setup-classic-description">
+                        {callbackOutcome === 'unreachable'
+                            ? 'Nothing answered on your computer. Sign-in links stop working after five minutes, so most often the link was simply opened too late.'
```
ducnmm: **[bug]** The unreachable SuccessCard copy says sign-in links die after five minutes, “so most often the link was simply opened too late.” This card only renders after a successful `/preflight` and an on-chain `add_delegate_key`. Opening the URL too late fails preflight and lands on the generic error step, never here. Unreachable at callback means the listener went away *after* the page loaded — wallet prompt sat through the deadline, MCP client restarted, or the login command was cancelled. The linked troubleshooting section repeats the impossible cause.

**Suggestion:** Explain that the app stopped waiting after this tab was already open (slow wallet / client restarted). Move “opened too late” to the preflight error path, or drop it from this card and the SuccessCard symptom in docs.

docs/troubleshooting/overview.md
```diff
@@ -76,6 +76,23 @@ This section covers problems that appear before the memory tools work.
 
 **Fix:** Quit the MCP client fully, then start it again. If you registered the server with `claude mcp add`, run `claude mcp list` to confirm `memwal` is registered before you restart.
 
+### Sign-in succeeds but credentials are not saved
+
+**Symptom:** The sign-in page confirms that your delegate key was registered, but says it could not hand the credentials back to your computer. The agent stays logged out and `~/.memwal/credentials.json` does not appear.
+
+**Cause:** Signing in has two halves. Your browser registers a delegate key onchain, then sends that key back to a short-lived listener the MCP package runs on `127.0.0.1`. The first half succeeded and does not need repeating; the second is what failed. The usual reasons:
```
ducnmm: **[bug]** Cause still says “The first half succeeded and does not need repeating.” That is the sentence removed from `SuccessCard` as misleading: a retry generates a new keypair and another `add_delegate_key`. The failure card now links here, so the dropped UI claim is reintroduced on the page users are sent to.

**Suggestion:** Match the UI: the unused key is already on the account and should be revoked; signing in again is a full new attempt, including the wallet step.

apps/app/src/pages/ConnectMcp.tsx
```diff
@@ -548,42 +552,86 @@ function ConsentCard({
...
+                            : 'The app on your computer turned down the hand-off. That usually means a newer sign-in was started while this tab was still open.'}
```
ducnmm: **[suggestion]** Rejected-hand-off copy says a newer sign-in was started while this tab was still open. After `startOrReuseLoginFlow` (origin/dev, merged in this PR), a concurrent `memwal_login` joins the same listener and URL. A later attempt binds a new port, so the old tab’s POST fails with connection refused (`unreachable`), not HTTP refusal. `login.ts` never returns 409; refusal is 400/403/428/500 (duplicate callback, state mismatch, callback before preflight). The same stale model is in `docs/troubleshooting/overview.md`.

**Suggestion:** Describe a refusal as “the app on your computer rejected this hand-off” (stale tab after a completed attempt, or a mismatch). Do not pin it on a second `memwal_login` replacing the listener.

apps/app/src/pages/ConnectMcp.tsx
```diff
@@ -548,42 +552,86 @@ function ConsentCard({
...
 }: {
     payload: McpCallbackPayload
-    callbackDelivered: boolean | null
+    callbackOutcome: CallbackOutcome | null
     port: string
 }) {
+    const delivered = callbackOutcome === 'delivered'
     return (
         <div className="setup-classic-intro">
+            {/* Heading follows the outcome; it used to say "connected" even on failure. */}
```
ducnmm: **[suggestion]** JSX comments narrate the change: “it used to say ‘connected’ even on failure” and “Demoted out of the prose, kept for debugging the listener”. The test file does the same (`ConnectMcp.test.tsx`).

**Suggestion:** Delete them, or keep a one-line constraint (heading must follow `callbackOutcome`; loopback URL is debug-only).

packages/mcp/test/login-failure-notice.test.mjs
```diff
@@ -0,0 +1,151 @@
...
+            res({ server, base: `http://127.0.0.1:${server.address().port}` });
+        });
+    });
+}
+
+test("a sign-in that never completes is reported on the next tool call", async (t) => {
+    const { server, base } = await startMockRelayer();
+    const home = mkdtempSync(join(tmpdir(), "memwal-test-"));
+
+    const child = spawn(process.execPath, [BIN, "--relayer", base, "--web-url", base], {
```
ducnmm: **[suggestion]** The new test spawns the binary with no credentials, so it only exercises auth-required `handleLoginToolCall`. The merge left `handleLocalLogin` on `LOGIN_BG_TIMEOUT_MS`; `3ad5c395` fixed that, but nothing would fail if bridge stopped passing `resolveLoginTimeoutMs()` again. `loginFlow` in the CLI / TTY path still ignores `MEMWAL_MCP_LOGIN_TIMEOUT_MS` even though the env docs describe it as the listener deadline.

**Suggestion:** Default `loginFlow`’s timeout through `resolveLoginTimeoutMs()` so CLI and both MCP wrappers cannot drift, and cover a signed-in `memwal_login` timeout (bridge notification) in this test or a sibling.

ducnmm (changes requested):
## Summary

HEAD keeps `CallbackOutcome`, does not claim “MCP client connected” on 409 or a dead listener, adds the #746 revoke sentence, and shares `resolveLoginTimeoutMs()` between auth-required and bridge. Both prior comments are addressed. Remaining problems are user-facing copy that does not match the post-preflight / singleflight flow, plus comments that narrate the change and a test that would not catch the merge leftover this PR already had to fix.

## Issue counts by severity

- bugs: 2
- suggestions: 3
- nits: 0

apps/app/src/pages/ConnectMcp.tsx
HoangDucBach: fixed in 14529177

docs/troubleshooting/overview.md
HoangDucBach: fixed in 40a6ffaf

apps/app/src/pages/ConnectMcp.tsx
HoangDucBach: fixed in 7df7f875

apps/app/src/pages/ConnectMcp.tsx
HoangDucBach: fixed in 4e606960

packages/mcp/test/login-failure-notice.test.mjs
HoangDucBach: fixed in a46f2927

apps/app/src/pages/ConnectMcp.tsx
```diff
@@ -548,42 +555,84 @@ function ConsentCard({
...
+                            ? 'Nothing answered on your computer. The app stopped waiting after this tab was already open, which usually means the wallet prompt sat through the deadline, the MCP client restarted, or the login command was cancelled.'
+                            : 'The app on your computer rejected this hand-off. That usually means this tab is leftover from a sign-in that already finished, or the request did not match what the app expected.'}
+                    </p>
+                    <p className="setup-classic-description">
+                        The unused key from this attempt is already on your account. Remove it
+                        from the dashboard if you are not using it.
+                    </p>
+                    <p className="setup-classic-description">
+                        <strong>Sign in again and open the new link straight away.</strong> The
+                        second attempt usually works.{' '}
```
ducnmm: **[suggestion]** After an on-chain `add_delegate_key` and a failed localhost hand-off, the page still says “The second attempt usually works.” The unreachable paragraph on the same card lists an MCP client restart as a usual cause. A restart loop will fail the same way and register another unused delegate key (cap 20). The same promise is in `packages/mcp/src/auth-required.ts:209` (`loginFailureNotice`) and `docs/troubleshooting/overview.md:89`, and `packages/mcp/test/login-failure-notice.test.mjs:209` plus `apps/app/src/pages/ConnectMcp.test.tsx:99` lock it in. Retry is the right next step; promising that it usually works is not, especially when the trigger is a recurring client restart.

**Suggestion:** Keep “sign in again and open the new link straight away,” but drop “usually works.” If the listener died because the MCP client restarted, say a retry only helps once the client is left running through the wallet prompt. Put revoke next to retry (the unused key is already on-chain). Mirror that in `loginFailureNotice` and the troubleshooting Fix; update the two tests that assert the old sentence.

packages/mcp/src/auth-required.ts
```diff
@@ -269,19 +285,30 @@ async function handleLoginToolCall(
...
                 accountId: creds.accountId,
                 delegateAddress: creds.delegateAddress,
             });
         },
     );
+    // startOrReuseLoginFlow's onSuccess callback only covers the success
+    // path. Track failure here so the next tool call can report it (Claude
+    // Code drops notifications/message; Cursor shows them, so this is the
+    // primary channel either way).
+    session.result.catch((err) => {
```
ducnmm: **[suggestion]** Every `memwal_login` call attaches `session.result.catch(...)`. `startOrReuseLoginFlow` returns the in-flight session on a concurrent call, so an agent that retries `memwal_login` while the listener is still bound registers another handler. One timeout then emits multiple `notifications/message` warnings (Cursor shows them) and `log.warn("memwal_login.bg.failed")` once per join. The same attach-on-every-call pattern is in `packages/mcp/src/bridge.ts:594`. `startOrReuseLoginFlow` already observes `result` rejection for logging.

**Suggestion:** Register the user-facing failure side effect once, inside `startOrReuseLoginFlow` (or only when this call actually started the flow). Auth-required can still set `lastLoginFailure` from that single handler. Optionally clear `lastLoginFailure` when a *new* attempt starts so the next memory tool does not keep quoting the previous timeout while a fresh URL is live.

apps/app/src/pages/ConnectMcp.tsx
```diff
@@ -219,10 +225,11 @@ export default function ConnectMcp() {
                     headers: { 'content-type': 'application/json' },
                     body: JSON.stringify(payload),
                 })
-                setCallbackDelivered(res.ok)
+                // A listener answered and refused — different cause than nothing answering.
```
ducnmm: **[nit]** `// A listener answered and refused — different cause than nothing answering.` restates the `rejected` vs `unreachable` assignment. Same class of narrating comment as the JSX notes that were dropped.

**Suggestion:** Delete it. `CallbackOutcome` already names the cases.

ducnmm (changes requested):
## Summary

WALM-333 now distinguishes callback delivery from an HTTP refusal, moves the “opened too late” story onto the preflight path (where the listener is already gone), and surfaces a failed background `memwal_login` on the next tool call plus a bridge warning. `loginFlow` honors `MEMWAL_MCP_LOGIN_TIMEOUT_MS` when `timeoutMs` is omitted, and both auth-required and bridge pass `resolveLoginTimeoutMs()`. The five prior findings landed. The copy still promises that a second attempt usually works, including when the page itself names an MCP client restart as a usual cause — that is the remaining product miss.

### Previous review

1. **Fixed.** SuccessCard unreachable copy no longer says the link died after five minutes / “opened too late”. That sentence is only on the preflight TypeError path; the unreachable SuccessCard talks about the listener going away after the tab was already open. Tests assert both.
2. **Fixed.** Troubleshooting Cause no longer says the first half does not need repeating. It now says the unused key is already on the account and should be revoked, and that signing in again is a full new attempt including the wallet step.
3. **Fixed.** Refused-hand-off copy no longer blames a newer sign-in replacing the listener. It describes a leftover tab or a request that did not match what the app expected.
4. **Fixed.** The SuccessCard JSX comments that narrated the change are gone.
5. **Fixed.** `loginFlow` defaults through `resolveLoginTimeoutMs()`; the suite covers a signed-in bridge timeout warning and the env default when `timeoutMs` is omitted.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 1

apps/app/src/pages/ConnectMcp.tsx
HoangDucBach: fixed in b09c4b64

packages/mcp/src/auth-required.ts
HoangDucBach: fixed in 21bc4169

apps/app/src/pages/ConnectMcp.tsx
HoangDucBach: fixed in 2af9206f

ducnmm (approved):
## Summary

Re-reviewed `2af9206f`. The three follow-ups from the last request-changes land:

1. **b09c4b64** — UI, `loginFailureNotice`, troubleshooting, and tests no longer promise that a second attempt usually works. Retry is still the next step, gated on the client staying up through the wallet prompt, with revoke next to it.
2. **21bc4169** — `onFailure` is attached only when `startOrReuseLoginFlow` actually starts the flow. Concurrent joins cannot emit a second timeout warning. Covered by `login-singleflight.test.mjs`.
3. **2af9206f** — the narrating `rejected` vs `unreachable` comment is gone.

No remaining blockers from my side.

PR #730, opened 2026-08-21
docs: import code snippets from the workspace instead of copying them

reemsabawi-mystenlabs:
Closes BEDU-1280.

## Why

`scripts/check-docs-code-sync.mjs` states the problem in its own header:

> Mintlify cannot import code from repository files at build time, so every snippet in docs/ is a copy. This script is the safety net.

That safety net validates what package metadata can prove — package names, entry points, MCP config blocks — but it cannot tell whether the body of a snippet still matches the declaration it came from. Jessie raised this on #631 ("what is the source of this code?") and confirmed it as a gap worth closing.

## What this adds

`scripts/sync-docs-code.mjs`. A docs page marks a fenced block:

````markdown
<!-- memwal:import packages/sdk/src/types.ts#RememberResult -->
```ts
```
<!-- /memwal:import -->
````

`pnpm docs:sync` fills the fence with the real declaration, doc comments included. The result stays committed, so Mintlify renders it with no build step, it reads correctly on GitHub, and nothing breaks if the script never runs. The script only keeps it honest.

Two selectors, both anchored to a name rather than to a line number, because line numbers go stale on the next refactor with nothing to catch it:

| **Selector** | **Extracts** |
| --- | --- |
| `path#SymbolName` | The named declaration and its doc comment. Requires nothing in the source file, so importing a symbol costs the source nothing. |
| `path#region:name` | The span between `// #region docs:name` and its `#endregion`, for a slice that is not one declaration. |

`pnpm docs:sync:check` exits non-zero when a committed block no longer matches its source, and CI runs it beside the existing code-sync job.

## What it caught immediately

Converting three blocks in `docs/sdk/api-reference.md` corrected a real error. The hand-written copy declared `job_id` as required on `RememberResult`; the SDK has it optional (`job_id?: string`, `packages/sdk/src/types.ts:51`). The imported version also carries the field-level doc comments, which the copy had replaced with shorter inline ones.

## Safety

Importable roots are allowlisted (`packages`, `apps`, `services`, `scripts`, `contract`), so a marker cannot inline an env file or a keystore. Paths are normalized before the root check, so a traversal cannot hide behind an allowed prefix — `packages/../../etc/passwd` is rejected rather than resolved.

Markers inside fenced blocks are skipped, so `docs-workflow.md` documents the syntax without rewriting itself.

## Test plan

Verified by running each case against this branch:

- `pnpm docs:sync` on a clean tree is a no-op, and re-running it is idempotent.
- `pnpm docs:sync:check` passes; adding a field to `RememberAcceptedResult` in the SDK makes it fail, naming `docs/sdk/api-reference.md:64`, and it passes again once the source is reverted.
- Unknown symbol, path traversal, disallowed root, and a missing close marker each exit `1` with a message naming the file and the cause.
- A page with a real import plus the same syntax shown inside a ````` ```` ````` fence rewrites only the real one.
- Region extraction dedents correctly from an indented `#region docs:` block.
- `node scripts/check-docs-code-sync.mjs` and `node scripts/check-docs-freshness.mjs` both still pass.

## Sources

- The problem statement and the "every snippet is a copy" framing: `scripts/check-docs-code-sync.mjs` header, quoted above.
- Design precedent: walrus's `docs/site/src/scripts/generate-import-context.js` and `docs/site/src/shared/js/inline-imports.js`. Their `<ImportContent>` resolves at render time through a React component, which Docusaurus supports and Mintlify does not, so this takes the inverse approach of injecting text at author time. The allowlist idea is taken directly from `ALLOWED_ROOTS` in `generate-import-context.js:15`.
- The CI floor check and its rationale ("a check that finds nothing reports success forever") follow the existing precedent in `check-docs-code-sync.mjs`.
- Converted declarations: `packages/sdk/src/types.ts`, read directly.
- The script itself and the `docs-workflow.md` section are written fresh; no existing source.

## Also in this diff

`docs/sdk/api-reference.md` carried 21 pre-existing style-guide violations that the pre-push gate audits at whole-file scope. Cleaned them in the same pass rather than working around the page: em dashes, `SEAL` → `Seal`, `via` → `through`, `on-chain` → `onchain`, `e.g.`/`vs.` spelled out, unbolded table headers, a quoted frontmatter title, and one pair of stacked headings.

## Scope note

This does not cover composed examples — a snippet showing two clients cooperating, or a call sequence assembled to teach a concept, exists nowhere in the source, so there is nothing to import. Those stay hand-written under `check-docs-code-sync.mjs`, and `docs-workflow.md` says so explicitly. The snippets on #631 are that kind, so this mechanism does not resolve Jessie's question there.

https://claude.ai/code/session_01WoBWxqLzd9hAxrhoqv3iSv

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,6 +55,13 @@ jobs:
       - name: Check docs code references against the packages
         run: node scripts/check-docs-code-sync.mjs
 
+      # Snippets marked <!-- memwal:import path#Symbol --> are extracted from
+      # the workspace rather than copied. This re-extracts them and fails if
+      # the committed block no longer matches its source, so a rename or a
+      # changed field cannot leave a docs page quietly wrong.
+      - name: Check imported docs snippets match their source
+        run: node scripts/sync-docs-code.mjs --check
+
   docs-freshness:
     name: Docs / Freshness
     runs-on: ubuntu-latest
--- a/docs/contributing/docs-workflow.md
+++ b/docs/contributing/docs-workflow.md
@@ -51,10 +51,10 @@ The docs source of truth is the markdown content under `docs/` in this repositor
 Mintlify cannot import code at build time, so snippets in `docs/` start as copies and drift the moment the source changes. When a snippet reproduces a declaration that lives in `packages/`, `apps/`, `services/`, `scripts/`, or `contract/`, import it instead of copying it:
 
 ````markdown
-<!-- memwal:import packages/sdk/src/types.ts#RememberResult -->
+{/* memwal:import packages/sdk/src/types.ts#RememberResult */}
 ```ts
 ```
-<!-- /memwal:import -->
+{/* /memwal:import */}
 ````
 
 Run `pnpm docs:sync` and the fence fills in with the real declaration, doc comments included. You commit the result, so Mintlify renders it with no build step and it reads correctly on GitHub. CI runs `pnpm docs:sync:check`, which fails when a page no longer matches its source, so whoever changes the code fixes the page in the same pull request.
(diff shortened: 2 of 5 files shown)


jessiemongeon1:
## Style Guide Audit

PR #733, opened 2026-08-22
fix(oauth): persist the client loopback redirect URI (GH #619)

### LINKED ISSUE #619: [Bug] OAuth loopback flow (Claude Code / RFC 8252) is internally inconsistent — redirect goes to the wrong port and the token exchange rejects the client's real redirect_uri
### Summary

The loopback OAuth flow matches the redirect URI **port-agnostically** at authorize time, but then stores the *registered* (port-less) URI in the session and code row. Two consequences: (1) the authorization code is redirected to the registered URI (port 80 / no port) instead of the client's ephemeral loopback listener; (2) the token exchange does an exact-match on `redirect_uri`, so the client's real port-bearing URI gets `invalid_grant`.

### Files

- `services/server/src/routes/oauth.rs:325-336` (stores registered URI)
- `services/server/src/routes/oauth.rs:709-713` (code redirect)
- `services/server/src/routes/oauth.rs:935-939` (token exchange exact match)
- `services/server/src/oauth.rs:574-595` (`redirect_uri_matches`, port-agnostic; test at `oauth.rs:956-965` pins `http://localhost:51234/callback` matches `http://localhost/callback`)

### Root cause

```rust
// routes/oauth.rs:325-336 — matching is port-agnostic, but the STORED uri is the registered one
let Some(redirect_uri) = client
    .redirect_uris
    .iter()
    .find(|registered| oauth::redirect_uri_matches(&query.redirect_uri, registered))
    .cloned()                       // <-- clones the REGISTERED uri (e.g. "http://localhost/callback")
else { ... };
```

The port-less URI is carried into the session and the code row, then:

```rust
// routes/oauth.rs:935-939
if code_row.redirect_uri != redirect_uri {
    return Err(OAuthError::invalid_grant(
        "redirect_uri does not match the authorization request",
    ));
}
```

RFC 6749 §4.1.3 requires the client to send the *same* URI it used at authorize time (port-bearing), so the exact match always fails for ephemeral loopback ports — which the code's own comments say clients MUST use.

### Reproduction

1. Register a client with redirect URI `http://localhost/callback`.
2. Start an OAuth flow from `http://localhost:51234/callback` (ephemeral port, per RFC 8252).
3. The code redirect lands at `http://localhost/callback?code=...` (wrong port — the client's listener on 51234 never receives it).
4. Even if the code is captured manually, the token exchange rejects it: `redirect_uri does not match`.

### Impact

The Claude Code / RFC 8252 loopback flow — a primary purpose of this OAuth feature — fails whenever the client uses an ephemeral port, which is exactly what the spec (and this code's own comments) require.

### Suggested fix

Store the client-presented (candidate) redirect URI in the session instead of the registered one, or apply the same port-agnostic matching at token-exchange time.

ducnmm:
Linear: [WALM-382](https://linear.app/mysten-labs/issue/WALM-382/bug-oauth-loopback-flow-claude-code-rfc-8252-is-internally)
Fixes https://github.com/MystenLabs/MemWal/issues/619

## Why

`GET /oauth/authorize` matches loopback redirect URIs port-agnostically (RFC 8252 / Claude Code ephemeral ports), then stored the **registered** URI (`http://localhost/callback`). Two failures follow:

1. The authorization code is redirected to port 80, not the client's listener on e.g. `:51234`.
2. Token exchange exact-matches `redirect_uri` (RFC 6749 §4.1.3), so the real port-bearing URI gets `invalid_grant`.

## What this does

After a successful match, persist the **client-presented** candidate URI on the session (and therefore the code row and the redirect target). HTTPS hosted callbacks are unchanged (registered == candidate).

## Tests

`oauth::tests::authorize_stores_loopback_candidate_not_registered_uri` — stores `http://localhost:51234/callback`, keeps exact HTTPS, rejects a different path.

### DIFF
--- a/services/server/src/oauth.rs
+++ b/services/server/src/oauth.rs
@@ -594,6 +594,23 @@ pub fn redirect_uri_matches(candidate: &str, registered: &str) -> bool {
         && candidate_url.query() == registered_url.query()
 }
 
+/// Redirect URI to persist on the authorize session and code row.
+///
+/// Matching is port-agnostic for RFC 8252 loopback, but RFC 6749 §4.1.3
+/// requires the token request to send the *same* URI used at authorize
+/// time. Store the client-presented candidate (`http://localhost:51234/callback`),
+/// not the registered URI (`http://localhost/callback`), so the code
+/// redirect hits the ephemeral listener and token exchange succeeds.
+pub fn authorize_redirect_uri<'a>(
+    candidate: &'a str,
+    registered_uris: impl IntoIterator<Item = &'a str>,
+) -> Option<&'a str> {
+    registered_uris
+        .into_iter()
+        .any(|registered| redirect_uri_matches(candidate, registered))
+        .then_some(candidate)
+}
+
 /// Decision D4: registration only accepts redirect URIs on a known-safe
 /// host allowlist (Anthropic's own callback domain) or an RFC 8252 loopback
 /// pattern. This is the direct fix for the objection that stalled PR #193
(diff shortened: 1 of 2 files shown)


ducnmm:
@HoangDucBach review request. Linear [WALM-382](https://linear.app/mysten-labs/issue/WALM-382/bug-oauth-loopback-flow-claude-code-rfc-8252-is-internally).

HoangDucBach (approved):
LGTM ! 🚀

PR #734, opened 2026-08-22
fix(sponsor): restore pending binding after sidecar execute failure (GH #617)

### LINKED ISSUE #617: [Bug][Server] /sponsor/execute deletes the pending-sponsorship record BEFORE the sidecar call — a transient sidecar failure permanently burns the sponsored transaction
### Summary

`consume_pending_sponsor` (an atomic Redis GET+DEL) runs **before** the sidecar execution call. If the sidecar call fails for any transient reason (transport error or upstream non-2xx), the `sponsor:pending:{digest}` key is already gone. The client holds a validly signed sponsored transaction it can **never execute**: any retry returns `401 "Invalid or expired sponsor authorization"`, and the client must redo the entire `/sponsor` flow (new wallet-auth signature, new nonce, new sponsored tx — i.e. another gas obligation for the sponsor).

### Files

- `services/server/src/routes/sponsor.rs:437-440` (call order)
- `services/server/src/routes/sponsor.rs:310-330` (`consume_pending_sponsor` Lua)
- `services/server/src/routes/sponsor.rs:324-328` (retry → 401)

### Root cause

```rust
// sponsor.rs:437-440
consume_pending_sponsor(&state, &req.digest, sender).await?;          // DELETES the record
let (upstream_status, resp_body) =
    call_sidecar_sponsor_execute(&state, &req.digest, &req.signature, false).await?; // may fail
```

The Lua script deletes the key even when the subsequent execution never happens:

```lua
redis.call('DEL', KEYS[1])
return 1
```

### Reproduction

1. `POST /sponsor` → receive sponsored `tx_bytes` + pending record.
2. Make the sidecar temporarily unavailable (or force a transport error / non-2xx from `/sponsor/execute`).
3. `POST /sponsor/execute` with a valid signature → masked 502/503, record already consumed.
4. Retry `POST /sponsor/execute` with the same digest+signature → `401 Invalid or expired sponsor authorization`. The sponsored tx can never be executed.

### Impact

Under any transient sidecar/network hiccup, every in-flight sponsored execution is burned: lost transactions plus forced re-sponsorship (extra sponsor gas). Ordering should be: execute → only consume on success (or restore the record on failure).

### Suggested fix

Move `consume_pending_sponsor` to after a successful sidecar execution, or re-store the pending record on failure so retries work.

ducnmm:
Linear: [WALM-380](https://linear.app/mysten-labs/issue/WALM-380/bugserver-sponsorexecute-deletes-the-pending-sponsorship-record-before)
Fixes https://github.com/MystenLabs/MemWal/issues/617

## Why

`POST /sponsor/execute` ran `consume_pending_sponsor` (Redis GET+DEL) **before** the sidecar call. A transport error or 5xx/429 left the client holding a signed sponsored tx it could never execute: retry returned `401 Invalid or expired sponsor authorization` and forced a full re-sponsor.

## What this does

Keep consume-before-execute (so a lost 200 cannot double-submit as easily). If the sidecar call errors at transport, or returns 5xx / 429, **restore** the pending binding. 4xx stays consumed (the signed tx was rejected; the client must re-sponsor).

## Tests

`routes::sponsor::more_tests::restore_pending_on_transient_sidecar_status` — 5xx/429 restore, 4xx/200 do not.

### DIFF
--- a/services/server/src/routes/sponsor.rs
+++ b/services/server/src/routes/sponsor.rs
@@ -275,7 +275,11 @@ fn validate_sponsor_transaction_kind(
         [_] => Ok(()),
         // Batching exists for bulk revoke only. Every command has to be a
         // removal, and the batch stays inside the sponsored-gas bound.
-        calls if calls.iter().all(|function| *function == "remove_delegate_key") => {
+        calls
+            if calls
+                .iter()
+                .all(|function| *function == "remove_delegate_key") =>
+        {
             if calls.len() > MAX_SPONSORED_DELEGATE_REMOVALS {
                 return Err(AppError::BadRequest(format!(
                     "Too many delegate key removals in one transaction (max {MAX_SPONSORED_DELEGATE_REMOVALS})"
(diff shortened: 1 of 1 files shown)


ducnmm:
@HoangDucBach review request. Linear [WALM-380](https://linear.app/mysten-labs/issue/WALM-380/bugserver-sponsorexecute-deletes-the-pending-sponsorship-record-before).

HoangDucBach (approved):
Approved.

Verified the 4xx / 5xx split, consume-before-execute being preserved, and the
restore staying best-effort so a Redis failure still surfaces the original
error.

PR #735, opened 2026-08-22
fix(mcp): require a registered delegate before opening a session

### LINKED ISSUE #624: SSE handshake 429 crashes the client; "ip_active_cap" persists across different public IPs and never clears
### Summary

When the relayer returns HTTP 429 during the SSE handshake, the client does not back off or retry —
it crashes, and on Windows the crash ends in a libuv assertion failure rather than a clean exit.

### Reproduction

Start the MCP server with valid saved credentials while the per-IP active cap is reached:

    npx -y @mysten-incubation/memwal-mcp --namespace <ns>

Output (v0.0.6, Windows 11, identifiers redacted):

    {"ts":"...","level":"info","scope":"memwal-mcp","event":"bridge.connecting",
     "relayer":"https://relayer.memory.walrus.xyz","accountId":"0x…","delegate":"0x…"}
    [memwal-mcp] fatal: Walrus Memory relayer SSE handshake failed: HTTP 429
      {"jsonrpc":"2.0","error":{"code":-32000,
       "message":"MCP rate limit: ip_active_cap. Try again in 30s."},"id":null}
    Error: Walrus Memory relayer SSE handshake failed: HTTP 429
        at openSseStream (.../dist/bridge.js:126:15)
        at async runBridge (.../dist/bridge.js:421:15)
        at async main (.../dist/index.js:207:5)
    Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 76

Reproduced **three times**:

| # | Time (UTC) | Network | Result |
|---|---|---|---|
| 1 | 15:20 | Home ISP | 429 `ip_active_cap` |
| 2 | 15:21 | Home ISP | 429 `ip_active_cap` |
| 3 | 17:18 | **Mobile carrier hotspot — entirely different public IP** | 429 `ip_active_cap` |

Same account and delegate key throughout. Two hours elapsed between the second and third attempt,
and the third was made from an unrelated public IP on a different network operator.

### Three distinct problems

**1. No backoff.** The error carries a retry interval ("Try again in 30s") and the client ignores it,
exiting fatally instead. An MCP server that dies on startup appears to the host as a broken server,
not a throttled one. In Claude Code the result is simply that no `memwal_*` tools exist, with no
indication why.

**2. Unclean shutdown, and the cap is on *active* connections.** The libuv assertion means the
process did not close its handles properly. If a half-open connection is still counted against
`ip_active_cap` server-side, the crash sustains the condition that caused it — each retry crashes,
each crash may leave another phantom connection. That would match what I observed: the 429 persisted
well past the stated interval.

**3. The error message misidentifies the scope, and the limit never clears.** The code is
`ip_active_cap` and the message says "Try again in 30s", but neither holds:

- Switching to a completely different public IP (mobile carrier instead of home ISP) produced the
  identical error. Whatever the cap is counting, **it is not the client's IP address** — it tracks
  the account or the delegate key.
- Two hours passed between attempts 2 and 3. The advertised 30-second interval is off by at least
  two orders of magnitude, and there is no evidence the limit clears on its own at all.

Taken with problem 2, the likely mechanism is that each crashed handshake leaves a connection
registered server-side against the account or delegate, and nothing ever releases it. The account
then sits permanently at its cap, and every further attempt crashes and — if the hypothesis is
right — adds another phantom. There is no client-side remedy: no way to enumerate open connections,
no way to close them, and no documented way to reset the state.

This is currently a hard blocker. A working account with valid credentials cannot connect at all.

If the cap *is* also IP-scoped in some configurations, note additionally that carrier-grade NAT is
the norm across much of Africa, South Asia and Latin America, where many subscribers share one
public IPv4 address — an IP-based cap would make Walrus Memory intermittently unusable for those
developers through no fault of their own.

### Expected

- Honour the interval in the 429 and retry with backoff instead of exiting
- Exit cleanly when giving up, closing handles so no connection is left half-open
- A clear message distinguishing "throttled, retrying" from "misconfigured"
- Reconsider per-IP capping, or raise it substantially for shared-IP ranges

### Actual

Immediate fatal crash, libuv assertion, no retry, and a rate limit that appears not to clear on the
advertised schedule.

### LINKED ISSUE #707: POST /api/mcp accepts any 64-hex Bearer — full tool session with no on-chain check, sessions pin ip_active_cap until DELETE
## Bug

`https://relayer.memory.walrus.xyz/api/mcp` (Streamable HTTP) is advertised as OAuth-protected. Unauthenticated requests get `401` plus `WWW-Authenticate: Bearer resource_metadata=...`.

If `Authorization` is `Bearer` plus **any** 64-character hex string, and `X-MemWal-Account-Id` is **any** `0x` + 64 hex chars, the relayer treats the caller as a legacy delegate-key client and the sidecar opens a real MCP session.

Nothing checks that the key is a registered delegate, that the account object exists, or that an OAuth token was issued. The session is given the full tool set (`memwal_remember`, `memwal_remember_bulk`, `memwal_analyze`, `memwal_restore`, `memwal_recall`, `memwal_health`).

Streamable HTTP sessions stay in memory until `DELETE /api/mcp` or a sidecar restart. There is no idle timeout. Each leftover session holds an `ip_active_cap` slot (default 16 per IP). Repeating `initialize` without `DELETE` eventually returns `429` with `MCP rate limit: ip_active_cap` and blocks further MCP use from that IP, including real clients.

Writes/recalls still fail later on the signed relayer API (`401`) because the key is not on-chain. The MCP layer itself has already accepted the session, listed write tools, and served `memwal_health`.

Relevant code:

- Relayer classifies a 64-hex Bearer as a legacy delegate and skips OAuth (`services/server/src/mcp_proxy.rs`, `is_legacy_delegate_bearer` / `classify_and_resolve`).
- That path stamps full read+write scope (`apply_internal_headers` when `identity` is `None`).
- Sidecar `resolveAuth` only checks header shape (`services/server/scripts/mcp/auth.ts`).
- Streamable sessions are stored until `DELETE` (`services/server/scripts/mcp/index.ts`); no idle reap.

## How to reproduce

Against production (or any relayer with MCP + OAuth enabled):

```bash
KEY=$(python3 -c 'print("11"*32)')
ACCT="0x$(python3 -c 'print("22"*32)')'

# 1) initialize — expect HTTP 200 and an mcp-session-id
curl -sS -D - -X POST "https://relayer.memory.walrus.xyz/api/mcp" \
  -H "Authorization: Bearer $KEY" \
  -H "X-MemWal-Account-Id: $ACCT" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"repro","version":"0"}}}'
```

Copy `mcp-session-id` from the response headers, then:

```bash
SID='<mcp-session-id from step 1>'

# 2) tools/list — expect all six memwal_* tools, including writes
curl -sS -X POST "https://relayer.memory.walrus.xyz/api/mcp" \
  -H "Authorization: Bearer $KEY" \
  -H "X-MemWal-Account-Id: $ACCT" \
  -H "mcp-session-id: $SID" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'

# 3) memwal_health — expect a successful tool result (relayer /health is public)
curl -sS -X POST "https://relayer.memory.walrus.xyz/api/mcp" \
  -H "Authorization: Bearer $KEY" \
  -H "X-MemWal-Account-Id: $ACCT" \
  -H "mcp-session-id: $SID" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"memwal_health","arguments":{}}}'
```

To see the cap:

```bash
# 4) Repeat step 1 ~16 times without DELETE.
# Later initialize calls return:
#   HTTP 429
#   {"jsonrpc":"2.0","error":{"code":-32000,"message":"MCP rate limit: ip_active_cap. Try again in 30s."},"id":null}
```

Cleanup for any sessions you opened:

```bash
curl -sS -X DELETE "https://relayer.memory.walrus.xyz/api/mcp" \
  -H "Authorization: Bearer $KEY" \
  -H "X-MemWal-Account-Id: $ACCT" \
  -H "mcp-session-id: $SID"
```

Observed on `https://relayer.memory.walrus.xyz/api/mcp` (relayer `version` `0.1.0`, `serverInfo.version` `0.0.1`).

### LINKED ISSUE #710: MCP proxy grants full memwal:read memwal:write scope to requests with no Authorization header when mcp_oauth is unconfigured
## What kind of bug

Fail-open authorization / missing authentication on a privilege-granting code path, reachable through a single missing environment variable. The relayer vouches for a caller who presented **no credentials at all** by stamping the sidecar's internal trust headers with full `memwal:read memwal:write` scope.

## Not a duplicate of #707

#707 covers the path where a caller **does** present a bearer: `is_legacy_delegate_bearer(token)` classifies any 64-hex string as a legacy delegate and returns `Passthrough`. It notes in passing that this path stamps full scope.

This report is a **different branch of the same function**, with a different trigger and a strictly weaker precondition on the attacker: when `mcp_oauth` is `None`, `classify_and_resolve` returns `Passthrough` **before the `Authorization` header is read at all**. No bearer — valid, fake, or absent — is required. #707's finding still requires the caller to supply a 64-hex string and an account id header; this one requires nothing.

## What happens

`services/server/src/mcp_proxy.rs`, `classify_and_resolve`:

```rust
async fn classify_and_resolve(state: &AppState, headers: &HeaderMap) -> McpAuthOutcome {
    if state.config.mcp_oauth.is_none() {
        return McpAuthOutcome::Passthrough;      // ← returns before any header is inspected
    }
    let Some(auth_value) = headers
        .get(axum::http::header::AUTHORIZATION)
        .and_then(|v| v.to_str().ok())
    else {
        return McpAuthOutcome::Unauthorized(None);   // ← unreachable when mcp_oauth is None
    };
    ...
}
```

The "no Authorization header → Unauthorized" branch is dead code whenever `mcp_oauth` is `None`.

`sse_proxy`, `messages_proxy`, and `streamable_proxy` all map `Passthrough` to `identity: None`, which `apply_internal_headers` treats as fully privileged:

```rust
let scope = match identity {
    Some(identity) => identity.scope.clone(),
    // Legacy delegate-key callers have no OAuth grant, so the relayer says
    // outright that they hold every tool scope.
    None => format!("{} {}", crate::oauth::SCOPE_READ, crate::oauth::SCOPE_WRITE),
};
```

The comment states the assumption that makes this wrong: `None` is taken to mean "legacy delegate-key caller." With `mcp_oauth` unset, `None` *also* means "sent nothing." The two are indistinguishable by the time this runs.

The forwarded request then carries `x-memwal-internal-sidecar-token` — the header whose entire purpose is to prove to the sidecar that the request genuinely came from the relayer — alongside `x-memwal-internal-oauth-scope: memwal:read memwal:write`.

This inverts the invariant #685 was closed to establish:

> The sidecar registers no tools when it is absent, so silence means "no access" rather than "unrestricted" (GH #685).

At the sidecar, silence means no access. At the relayer, silence now means full scope.

## Why a deployment lands here by accident

`McpOAuthConfig::from_env()` in `services/server/src/oauth.rs` handles two misconfigurations of the same class inconsistently:

```rust
let encryption_key = load_encryption_key(&encryption_key_b64)
    .expect("MCP_OAUTH_DELEGATE_ENCRYPTION_KEY must decode to exactly 32 bytes (base64url, no padding)");
//   malformed key  → panic at boot (loud, fail-closed)

let issuer = std::env::var("MEMWAL_RELAYER_URL").ok()?;
//   missing URL    → silent None → OAuth off → every MCP request fails open
```

An operator who set `MCP_OAUTH_DELEGATE_ENCRYPTION_KEY` has unambiguously signalled they want OAuth on. If `MEMWAL_RELAYER_URL` is missing or renamed, OAuth is silently disabled instead.

Nothing surfaces this at runtime:

- `main.rs` has `if config.mcp_oauth.is_some() { ...mount oauth routes... }` with no `else` branch and no warning.
- The `/api/mcp`, `/api/mcp/sse`, and `/api/mcp/messages` routes are mounted **unconditionally**, and by design carry no signed-request auth: *"MCP proxy routes — reverse-proxy to the Node sidecar's `/mcp/*` routes. No signed-request auth here"*.
- Those routes also get no rate limiting, since the weighted limiter keys off the `AuthInfo` extension that only `verify_signature` populates.

So the same binary silently serves MCP with the proxy's authorization layer switched off, and the only external symptom is the absence of a `WWW-Authenticate` challenge.

## How to reproduce

1. Start the relayer with the MCP sidecar configured and `SIDECAR_AUTH_TOKEN` set, then:
   - set `MCP_OAUTH_DELEGATE_ENCRYPTION_KEY` to a valid 32-byte base64url value
   - leave `MEMWAL_RELAYER_URL` **unset**

2. Confirm OAuth silently disabled — this returns 404, and startup logs contain no warning:
   ```bash
   curl -i "$RELAYER/.well-known/oauth-protected-resource"
   ```

3. Send an MCP request with **no `Authorization` header and no `X-MemWal-Account-Id`**:
   ```bash
   curl -i -X POST "$RELAYER/api/mcp" \
     -H "Content-Type: application/json" \
     -H "Accept: application/json, text/event-stream" \
     -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"repro","version":"0"}}}'
   ```

   Expected (current OAuth-enabled behaviour): `401` with `WWW-Authenticate: Bearer resource_metadata=...`.
   Actual: no 401 from the proxy. The request is forwarded to the sidecar carrying `x-memwal-internal-sidecar-token` and `x-memwal-internal-oauth-scope: memwal:read memwal:write`.

4. Observe the forwarded headers on the sidecar (sidecar request log, or a proxy placed at `SIDECAR_URL`). Both internal headers are present on a request that carried no credentials.

Code-path confirmation without running anything: `classify_and_resolve` returns at the first statement, so the `Authorization` lookup below it never executes.

## Scope of impact

Stated plainly rather than overstated:

- The residual gate is the sidecar's own `resolveAuth()`, which still requires `Authorization: Bearer <64-hex>`. This is therefore not by itself a full unauthenticated compromise — it removes the relayer's authorization layer and leaves a single header-shape check (which #707 shows accepts arbitrary values).
- Hosted production is **not** currently in this state. Verified: `POST https://relayer.memory.walrus.xyz/api/mcp` returns `401` with `www-authenticate: Bearer resource_metadata="https://relayer.memory.walrus.xyz/.well-known/oauth-protected-resource"`, so `mcp_oauth` is configured there.
- The exposure is self-hosted deployments and any environment where `MEMWAL_RELAYER_URL` goes missing while the encryption key stays set — a state the process currently enters silently, and which the malformed-key branch three lines above treats as a boot-time panic.

### LINKED ISSUE #716: `ip_active_cap` rate limit (HTTP 429) can persist well past its stated 30s cooldown, and triggers a native crash on Windows instead of a clean exit
Summary

Rapid reconnects to the `memwal-mcp` server (e.g. repeated MCP client
window reloads) can trigger:

```
[memwal-mcp] fatal: Walrus Memory relayer SSE handshake failed: HTTP 429
{"jsonrpc":"2.0","error":{"code":-32000,"message":"MCP rate limit:
ip_active_cap. Try again in 30s."},"id":null}
Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 76
```

Two problems:

1. The stated 30s cooldown was not reliable in practice, repeated checks
   several minutes apart still hit the same 429, suggesting either the cap
   counts something other than a simple rolling window (e.g. it resets on
   every new connection attempt, including ones a client's own automatic
   reconnect logic makes), or the actual cooldown is significantly longer
   than the message states.
2. When the SSE handshake fails this way, the process does not exit
   cleanly, it crashes with a native libuv assertion failure on Windows
   (`UV_HANDLE_CLOSING`), rather than surfacing the 429 as a normal error
   exit. This makes the failure harder for an MCP client to detect and
   handle gracefully, and harder for a user to diagnose (the visible
   symptom is just "the server never registers any tools").

Suggestion: handle the 429 response by exiting cleanly with a
descriptive error (no native crash), and consider whether `ip_active_cap`
should be a longer/backoff-based limit rather than a flat "try again in
30s" that can be re-triggered by a client's own reconnect attempts.

ducnmm:
Linear: [WALM-402](https://linear.app/mysten-labs/issue/WALM-402/harden-mcp-session-auth-occupancy-caps-and-sse-429-handling)
Fixes https://github.com/MystenLabs/MemWal/issues/710
Fixes https://github.com/MystenLabs/MemWal/issues/707
Fixes https://github.com/MystenLabs/MemWal/issues/716
Fixes https://github.com/MystenLabs/MemWal/issues/624

## What this does

- MCP proxy always requires `Authorization`. A 64-hex bearer is forwarded only after on-chain delegate verification against `X-MemWal-Account-Id`. Unregistered keys get 401. Sui RPC failures get 503, not 401.
- Streamable sessions that sit idle for 15 minutes are reaped so they stop pinning `ip_active_cap`.
- Occupancy caps (`ip_active_cap`, `global_cap`) no longer advertise a 30s cooldown.
- The stdio bridge drains handshake error bodies instead of aborting the socket (Windows libuv crash).

## Package

`@mysten-incubation/memwal-mcp` **0.0.10 → 0.0.11**. Changelog updated. No changeset.

### DIFF
--- a/packages/mcp/CHANGELOG.md
+++ b/packages/mcp/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Do not abort the SSE handshake socket on HTTP 429, 401, or a bad content-type (that crash path fired on Windows).
 - Report MCP session occupancy caps as concurrent, not a 30-second cooldown.
+- Exit with code 1 when `login` is run without a TTY, instead of booting the auth-required stub and reporting success.
+- Concurrent `memwal_login` calls reuse the in-flight listener and URL instead of starting a second flow that can hang later remember/recall.
 
 ## 0.0.10
 
--- a/packages/mcp/CHANGELOG.md
+++ b/packages/mcp/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @mysten-incubation/memwal-mcp
 
+## 0.0.11
+
+### Fixed
+
+- Do not abort the SSE handshake socket on HTTP 429, 401, or a bad content-type (that crash path fired on Windows).
+- Report MCP session occupancy caps as concurrent, not a 30-second cooldown.
+
 ## 0.0.10
 
 ### Fixed
--- a/packages/mcp/CHANGELOG.md
+++ b/packages/mcp/CHANGELOG.md
@@ -8,6 +8,8 @@
 - Report MCP session occupancy caps as concurrent, not a 30-second cooldown.
 - Exit with code 1 when `login` is run without a TTY, instead of booting the auth-required stub and reporting success.
 - Concurrent `memwal_login` calls reuse the in-flight listener and URL instead of starting a second flow that can hang later remember/recall.
+- Serialize outbound MCP POSTs and reconnect when the SSE handle is gone, so a burst of `memwal_recall` calls cannot poison the bridge until restart.
+- Login timeout now warns that an on-chain delegate key may already exist if the browser approved after the local listener expired.
 
 ## 0.0.10
 
--- a/packages/mcp/package.json
+++ b/packages/mcp/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@mysten-incubation/memwal-mcp",
-    "version": "0.0.10",
+    "version": "0.0.11",
     "description": "Walrus Memory MCP client — single-binary stdio MCP server that bridges Cursor / Claude Desktop / Antigravity / Claude Code to the Walrus Memory relayer. Handles browser-based wallet login on first run.",
     "type": "module",
     "engines": {
--- a/packages/mcp/plugin/plugin.json
+++ b/packages/mcp/plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "id": "memwal",
   "name": "memwal",
-  "version": "0.0.10",
+  "version": "0.0.11",
   "description": "Automatic Walrus Memory for Antigravity — proactive recall and durable-fact saving via the MemWal MCP + lifecycle hooks.",
   "author": { "name": "Mysten Labs" },
   "homepage": "https://memory.walrus.xyz",
--- a/packages/mcp/src/auth-required.ts
+++ b/packages/mcp/src/auth-required.ts
@@ -23,7 +23,7 @@
  */
 import { loadCreds, type MemWalCredentials } from "./auth.js";
 import { log } from "./logger.js";
-import { loginFlow } from "./login.js";
+import { startOrReuseLoginFlow } from "./login.js";
 import { AUTH_REQUIRED_INSTRUCTIONS } from "./instructions.js";
 import { MEMWAL_MCP_VERSION } from "./version.js";
 
@@ -255,42 +255,33 @@ async function handleLoginToolCall(
     config: AuthRequiredConfig,
     _progressToken: unknown,
 ): Promise<{ text: string; isError: boolean }> {
-    let connectUrl: string | null = null;
-
-    // Promise that resolves with the URL as soon as the listener is bound.
-    const urlReady = new Promise<string>((resolve) => {
-        // Fire loginFlow but DO NOT await — it runs in the background.
-        // openBrowser: false because (a) child-process spawning a browser is
-        // unreliable across MCP clients, and (b) macOS `open <url>` often
-        // foregrounds an existing memory.walrus.xyz tab instead of navigating to
-        // the full /connect/mcp?... URL — so user lands on the homepage,
-        // not the consent screen. The agent surfaces the clickable URL
-        // from the tool result instead.
-        loginFlow({
+    // Fire login but DO NOT await the wallet callback — it runs in the
+    // background. openBrowser: false because (a) child-process spawning a
+    // browser is unreliable across MCP clients, and (b) macOS `open <url>`
+    // often foregrounds an existing memory.walrus.xyz tab instead of
+    // navigating to the full /connect/mcp?... URL. The agent surfaces the
+    // clickable URL from the tool result instead.
+    //
+    // Concurrent memwal_login calls join this in-flight flow instead of
+    // opening a second listener (that race hung later recall/remember).
+    const session = startOrReuseLoginFlow(
+        {
             relayerUrl: config.relayerUrl,
             webUrl: config.webUrl,
             label: config.label,
             timeoutMs: LOGIN_BG_TIMEOUT_MS,
             openBrowser: false,
             onUrl: (url) => {
-                connectUrl = url;
-                resolve(url);
-                // Also push to log notification — clients that surface these
-                // (Cursor) get a second visible copy of the URL.
                 sendLogMessage("info", `Walrus Memory MCP login URL: ${url}`);
             },
-        })
-            .then((creds) => {
-                log.info("memwal_login.bg.success", {
-                    accountId: creds.accountId,
-                    delegateAddress: creds.delegateAddress,
-                });
-            })
-            .catch((err) => {
-                const msg = err instanceof Error ? err.message : String(err);
-                log.warn("memwal_login.bg.failed", { msg });
+        },
+        (creds) => {
+            log.info("memwal_login.bg.success", {
+                accountId: creds.accountId,
+                delegateAddress: creds.delegateAddress,
             });
-    });
+        },
+    );
 
     // Race the URL-ready against a short timeout. The listener bind is
     // synchronous-ish (single port allocation); 5s is a hard cap for a
@@ -304,7 +295,7 @@ async function handleLoginToolCall(
 
     let url: string;
     try {
-        url = await Promise.race([urlReady, timeoutPromise]);
+        url = await Promise.race([session.url, timeoutPromise]);
     } catch (err) {
         const msg = err instanceof Error ? err.message : String(err);
         log.error("memwal_login.tool.url_not_ready", { msg });
--- a/packages/mcp/src/bin/memwal-mcp.ts
+++ b/packages/mcp/src/bin/memwal-mcp.ts
@@ -6,5 +6,7 @@ main().catch((err) => {
     if (err?.stack && process.env.MEMWAL_MCP_DEBUG) {
         process.stderr.write(err.stack + "\n");
     }
-    process.exit(1);
+    // Let the event loop drain so undici/libuv can close the handshake
+    // socket. process.exit(1) here asserts on Windows (UV_HANDLE_CLOSING).
+    process.exitCode = 1;
 });
--- a/packages/mcp/src/bridge.ts
+++ b/packages/mcp/src/bridge.ts
@@ -19,7 +19,7 @@ import { clearCreds, credsPath } from "./auth.js";
 import { TOOL_DEFINITIONS } from "./auth-required.js";
 import { ensureCompatibleRelayer, resolveConnectTimeoutMs } from "./compatibility.js";
 import { PROACTIVE_INSTRUCTIONS } from "./instructions.js";
-import { loginFlow } from "./login.js";
+import { startOrReuseLoginFlow } from "./login.js";
 import { log, note } from "./logger.js";
 import { MEMWAL_MCP_VERSION } from "./version.js";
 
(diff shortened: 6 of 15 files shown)


HoangDucBach (approved):
Approved.

This PR now carries the full stack (#738 and #745 merged in). Re-verified
after the rebase: `mcp_proxy.rs` diff is unchanged from the earlier review
(94+/20-), and the full MCP test suite passes on this exact head (46/46,
tsc clean), run in an isolated worktree against `c0c75bef`.

Issue #655, opened 2026-08-15
Hosted Playground remember() fails with 503 insufficient SUI, retry hits delegate-key 429


shmalik06:
## Summary

Using the official Walrus Memory Developer Playground with the hosted relayer, `remember()` is accepted successfully but the durable Walrus upload later fails because of insufficient SUI balance.

A subsequent retry can then hit the delegate-key rate limit (`429`).

## Environment

* Walrus Memory Developer Playground
* SDK: `@mysten-incubation/memwal`
* Server: `https://relayer.memory.walrus.xyz`
* Namespace: `default`
* Delegate key: configured
* Health check: successful

## Steps to reproduce

1. Create/register a delegate key through the Walrus Memory Dashboard.
2. Open the Developer Playground.
3. Verify that the health check succeeds.
4. Run `remember()` with a normal text memory.
5. The request is initially accepted with HTTP 202.
6. Allow the Playground to poll the remember job.
7. The job eventually fails during the durable Walrus upload.
8. Retry the operation after the failure.

## Actual behavior

The initial request is accepted:

```text
accepted (HTTP 202)
status: running
```

After polling, the job fails with an error similar to:

```text
Internal Error: durable Walrus upload failed (503 Service Unavailable):
Unable to perform gas selection due to insufficient SUI balance
```

A subsequent attempt can return:

```text
Walrus Memory server error (429):
Rate limit exceeded
layer: delegate_key
limit: 30 weighted-requests/min
retry_after_seconds: 60
```

I also observed that the address referenced by the insufficient-balance error was not stable across attempts, which made it unclear whether the user was expected to fund an address or whether this was relayer-side infrastructure.

## Expected behavior

When using the official hosted relayer, a valid authenticated `remember()` request should either:

1. complete the durable Walrus upload successfully, or
2. fail early with an actionable error explaining what the user needs to do.

If hosted-relayer funding is managed server-side, an insufficient relayer balance should ideally not surface as an ambiguous user-facing gas-selection error after the job has already been accepted.

## Developer experience impact

This creates a confusing sequence:

`remember accepted → long polling → 503 insufficient SUI → retry → 429 rate limit`

From the Playground UI it is difficult to determine whether:

* the user's wallet needs funding;
* the MemWalAccount needs funding;
* the delegate-derived address needs funding; or
* the hosted relayer itself needs funding.

This can lead users to send SUI to an address that may not actually be theirs to fund.

## Suggested improvement

A few possibilities:

[...]

PR #736, opened 2026-08-22
fix(relayer): restore timeouts, oauth authorize cap, analyze idempotency

### LINKED ISSUE #414: memwal_restore returns HTTP 500 (Internal server error) intermittently, and can hang >400s past the 30s tool timeout
**Environment:** macOS 26.5.2 (Darwin 25.5.0) · Node v22.23.1 · `@mysten-incubation/memwal-mcp` 0.0.5 · Claude Code CLI (MCP stdio transport) · mainnet relayer `https://relayer.memory.walrus.xyz` · relayer version 0.1.0 (`memwal_health` → `status=ok`)

## Summary

`memwal_restore` intermittently fails server-side. Two distinct failure modes were observed in MCP logs, on a namespace that restores fine on other days — so this is server-side flakiness, not a malformed request.

## Failure mode A — repeated HTTP 500s

On 2026-07-12, six consecutive `memwal_restore` calls (same namespace, ~30s apart) each failed after ~30s with:

```
MemWal server error (500): {"error":"Internal server error (traceId: <id>)"}
```

Relayer trace IDs for log correlation:

- `2e68f3e7-ccf5-486f-b974-f1a6339b20ac`
- `4074ac5e-58f9-4d1d-a0e6-b5640a8c7ba4`
- `c860318b-90ce-41a0-aa8f-4fe87d021f42`
- `9f70eaab-cddb-4dde-9e58-a9b7aa25396f`
- `b2e76329-4c9a-4b4e-82b9-73d9e18bdf68`
- `2771e4ab-dad7-4f0e-9434-49fc881f2027`

## Failure mode B — hang past timeout

On 2026-07-13, one `memwal_restore` ran **428s** then failed with `MCP error -32001: AbortError: remote-cancel`; the stdio connection then dropped after 905s uptime. The "still running (30s elapsed)" keepalive fired 14 times before the client-side abort.

## Expected vs actual

**Expected:** restore completes, or returns a fast typed error the agent can act on.
**Actual:** either an HTTP 500 after ~30s, or an indefinite hang requiring client-side cancel.

## Same namespace, same day, succeeds

```
Restore complete for namespace "<redacted>":
  total=15  restored=0  skipped=15
```

Completed in <10s — confirming intermittency rather than a client-side error.

## Impact

`restore` is the documented recovery path when `recall` returns empty. A flaky or hanging restore blocks the exact scenario it exists for.

**Related but distinct:** #399 (restore doesn't *report* per-blob drops) and #382 (count mismatch) concern the *success response shape*; this issue is the call *failing or hanging* outright.

### LINKED ISSUE #620: [Bug][Security] /oauth/authorize has no auth and no rate limit — unbounded growth of the delegates table (which permanently stores encrypted Ed25519 private keys)
### Summary

`GET /oauth/authorize` is mounted with no auth middleware and no rate limiting. Every call generates a delegate key, inserts a **permanent** row into `mcp_oauth_delegates` (containing an AES-256-GCM envelope of a live Ed25519 private key), plus an authorize-session row. The housekeeping job only ever deletes expired sessions and authorization codes — it never deletes delegates. An attacker with one self-registered client (redirect `http://localhost/callback` is on the allowlist) can fire unlimited authorize GETs → unbounded DB growth → disk exhaustion / degraded queries, plus unbounded accumulation of sensitive key material at rest.

### Files

- `services/server/src/routes/oauth.rs:388-450` (delegate + session insert on every authorize)
- `services/server/src/main.rs:1406-1409` (mounted without auth middleware)
- `services/server/src/storage/db.rs:1450-1475` (`evict_expired_oauth_state` — only sessions and codes; explicitly does NOT touch delegates)

### Root cause

```rust
// routes/oauth.rs:388-418 — runs on EVERY /oauth/authorize call
let (private_key_hex, public_key_hex, delegate_address) = oauth::generate_delegate_key();
let encrypted = oauth::encrypt_delegate_private_key(&cfg.encryption_key, &private_key_hex)?;
let delegate_row = OAuthDelegateRow { ..., encrypted_private_key: encrypted, status: "pending", ... };
if state.db.insert_oauth_delegate(&delegate_row).await.is_err() { ... }
```

There is **no `DELETE FROM mcp_oauth_delegates` anywhere** in the codebase (verified by grep). The register throttle (20/hr/IP) does not bound this: one `client_id` enables unlimited authorize calls.

### Reproduction

```
loop:
  GET /oauth/authorize?client_id=<registered>&redirect_uri=http://localhost/callback
```

Each iteration inserts a new row into `mcp_oauth_delegates` and `mcp_oauth_authorize_sessions`; pending delegates from abandoned authorizes are never pruned.

### Impact

- Unbounded PostgreSQL growth → disk exhaustion / degraded queries for everyone (DoS).
- Permanent accumulation of AES-256-GCM envelopes of live Ed25519 private keys at rest.

### Suggested fix

Rate-limit `/oauth/authorize` per IP; add a TTL / cleanup for pending delegates that are never activated; cap pending delegates per client.

### LINKED ISSUE #729: memwal_analyze reports N facts with N blob_ids but the namespace ends up with 2N blobs — every successfully extracted fact is stored twice
## Summary

A single `memwal_analyze` call (MCP tool, mainnet relayer) that reports **3 facts extracted,
3 succeeded, 3 blob_ids** leaves the target namespace with **6 blobs**: `memwal_restore` reports
`total=6`, and `memwal_recall` returns each fact **twice** (two hits with identical text at the
same score). The tool's own response under-reports its writes by half, so the duplication is
invisible unless you audit with `restore` — which most agents never do, especially since there is
no list operation (#626).

## Environment

- MCP tools via `memwal-mcp` (Claude Code client), production relayer, `memwal_health` → `status=ok version=0.1.0`
- Walrus **Mainnet**
- Date: 2026-08-21
- Fresh namespaces created for the reproduction (no prior contents)

## Reproduction

1. Pick a fresh namespace (mine: `s7probe3`). Confirm empty: `memwal_restore(namespace="s7probe3", limit=100)` → `total=0`.
2. Call `memwal_analyze` with a short multi-fact passage:

   > The Sunblade was forged in Ravengard. Kane carries the Sunblade after Chapter 7. The Fold exacts a memory cost from anyone who crosses it.

3. Tool response — 3 facts, all `done`, one blob_id each:

   ```
   Extracted 3 fact(s) — succeeded=3 failed=0
   1. [done] blob_id=1jS-BdqyiUA1FIIQJ7CB1_7q2IMypF2VCwcVPAjxzm8 The Sunblade was forged in Ravengard
   2. [done] blob_id=16cbYawnxiDXrg7w7j56eeOByNd6EbLFdzHb6KCA2-o Kane carries the Sunblade after Chapter 7
   3. [done] blob_id=gymbAWFe45b0h8ee51WKHLGOsR0XMhOZgY39LyG5m7E The Fold exacts a memory cost from anyone who crosses it
   ```

4. Audit the namespace:

   ```
   memwal_restore(namespace="s7probe3", limit=100)
   → total=6  restored=0  skipped=6  truncated=false
   ```

5. Recall corroborates — each fact appears twice, pairwise identical text at identical scores:

   ```
   memwal_recall(query="The Sunblade was forged in Ravengard", namespace="s7probe3", limit=5)
   1. [score=1.000] The Sunblade was forged in Ravengard
   2. [score=1.000] The Sunblade was forged in Ravengard
   3. [score=0.554] Kane carries the Sunblade after Chapter 7
   4. [score=0.554] Kane carries the Sunblade after Chapter 7
   5. [score=0.175] The Fold exacts a memory cost from anyone who crosses it
   ```

3 facts reported → 6 blobs on Walrus, 2 per fact.

## A second run that did NOT duplicate — narrows the cause

Earlier the same day, during visible relayer degradation, `memwal_analyze` on a different fresh
namespace (`s7probe2`, 6-fact passage) returned:

```
Extracted 6 fact(s) — succeeded=1 failed=5
5. [done] blob_id=-_OyaI6QPf3S7mBAMDL8tKtaBYqGKM8LEJq150ulajI ...
(other 5: [timeout])
```

Audit: `memwal_restore(namespace="s7probe2")` → `total=1`. The one successful fact was stored
**once**, and re-checking ~15 minutes later it was still `total=1`.

So the duplication is tied to the **successful completion path**, not to timeout/retry recovery:
a run where facts complete cleanly writes each twice; a run where most facts time out writes the
lone survivor once. Consistent with a pipeline that persists each fact at extraction time and
then persists the completed batch again (or an internal job that reports one write while the
worker commits two) — rather than with client-side retry.

## Impact

- **2× storage cost** — every analyze spends double the WAL/quota its response admits to.
- **Dedup logic downstream is poisoned**: any prompt that recalls before writing (recall-then-
  remember dedup gates are the standard pattern in the current Prompt Jam corpus) sees a phantom
  "existing duplicate" for facts analyze created, and skips legitimate writes.
- **Blob-count proofs are inflated** — sessions events require agents to report mainnet blob
  counts; an agent using analyze honestly reports N from tool responses while the chain holds 2N.
- **Undetectable in normal operation**: the response shows N blob_ids, there is no list op
  (#626), and only a `restore` count or an exact-text recall reveals the extra writes.

## Expected

One extracted fact → one blob, and the response's blob_id list matches what was written. If the
server intentionally writes twice (e.g. raw + normalized), the response should say so and return
both ids.

## Related — and why this is not a duplicate of them

- **#562 (merged 2026-08-12)** — made `remember` writes idempotent precisely to prevent "a second
  paid blob" from client or internal (Apalis) retries. This reproduction is from **2026-08-21,
  nine days after that merge**, on the **analyze** path: either the analyze fan-out's per-fact
  writes don't go through the guarded path, or the guard doesn't cover them. Note the shape
  differs from a retry race: in the clean run ALL 3 facts were duplicated uniformly, which looks
  more like two persistence passes (per-fact at extraction, then the completed batch) than a
  stochastic retry — while in the degraded run the job presumably died before the second pass.
- **#87 (merged April)** — "analyze amplification" is a different amplification: unbounded fact
  *extraction* cost. This report is about each successfully extracted fact being *stored* twice.
- **#626** — no list operation, which is why this is invisible in normal operation.
- **#656** — read-your-writes lag; different failure (reads missing writes, not writes doubling).

ducnmm:
Linear: [WALM-322](https://linear.app/mysten-labs/issue/WALM-322/bug-high-memwal-restore-maps-transient-errors-to-http-500-and-can-hang), [WALM-357](https://linear.app/mysten-labs/issue/WALM-357/security-high-rate-limit-oauth-authorize-and-prune-abandoned-delegates), [WALM-375](https://linear.app/mysten-labs/issue/WALM-375/gh-655-hosted-relayer-gas-exhaustion-surfaces-as-an-ambiguous-503), [WALM-403](https://linear.app/mysten-labs/issue/WALM-403/analyze-retries-mint-duplicate-blobs-add-per-fact-idempotency)
Fixes https://github.com/MystenLabs/MemWal/issues/414
Fixes https://github.com/MystenLabs/MemWal/issues/620
Fixes https://github.com/MystenLabs/MemWal/issues/729
Related: https://github.com/MystenLabs/MemWal/issues/655

## What this does

- `/api/restore`: sidecar query failures are 503; 55s deadline; embed fan-out capped at 3.
- `/oauth/authorize`: per-IP Redis rate limit; abandoned pending delegates older than 1 hour are pruned.
- `/api/analyze`: per-fact idempotency so a retry does not mint a second blob; failed no-blob rows can restart; enqueue failure marks the row failed.
- SUI gas-selection / insufficient SUI errors classify as pool budget failures (WAL excluded). Preflight-before-202 and rate-limit refund are not in this PR.

## Package

Relayer only. No SDK/MCP version bump.

### DIFF
--- a/services/server/src/routes/analyze.rs
+++ b/services/server/src/routes/analyze.rs
@@ -17,6 +17,7 @@ use axum::extract::State;
 use axum::http::StatusCode;
 use axum::{Extension, Json};
 use base64::Engine as _;
+use sha2::{Digest, Sha256};
 use std::sync::Arc;
 
 use crate::jobs::WalletOperation;
@@ -27,6 +28,11 @@ use crate::types::*;
 use super::{collect_bounded_results, enqueue_wallet_job};
 
 const ANALYZE_CONCURRENCY: usize = 5;
+
+fn analyze_fact_idempotency_key(owner: &str, namespace: &str, fact_text: &str) -> String {
+    let digest = Sha256::digest(format!("{owner}\n{namespace}\n{fact_text}").as_bytes());
+    format!("analyze:{}", hex::encode(digest))
+}
 
 // /api/analyze does not benefit from larger inputs — it sends the
 // full text to gpt-4o-mini for fact extraction in a single LLM call (no
@@ -548,21 +554,92 @@ pub async fn analyze(
     for (idx, (job_id, (fact_text, importance, vector, encrypted))) in
         pending.into_iter().enumerate()
     {
-        // Insert status row
-        if let Err(e) = sqlx::query(
-            "INSERT INTO remember_jobs (id, owner, namespace, status) VALUES ($1, $2, $3, 'pending')",
+        let idem_key = analyze_fact_idempotency_key(owner, namespace, &fact_text);
+        let inserted = match sqlx::query(
+            "INSERT INTO remember_jobs (id, owner, namespace, status, idempotency_key, request_fingerprint)
+             VALUES ($1, $2, $3, 'pending', $4, $4)
+             ON CONFLICT (owner, idempotency_key) WHERE idempotency_key IS NOT NULL DO NOTHING",
         )
         .bind(&job_id)
         .bind(owner)
         .bind(namespace)
+        .bind(&idem_key)
         .execute(state.db.pool())
         .await
         {
-            rate_limit::release_storage_quota(&state, &all_ids[idx..]).await;
-            return Err(AppError::Internal(format!(
-                "Failed to create analyze job row: {}",
-                e
-            )));
+            Ok(result) => result,
+            Err(e) => {
+                rate_limit::release_storage_quota(&state, &all_ids[idx..]).await;
+                return Err(AppError::Internal(format!(
+                    "Failed to create analyze job row: {}",
+                    e
+                )));
+            }
+        };
+        let mut upload_job_id = job_id.clone();
+        if inserted.rows_affected() == 0 {
+            rate_limit::release_storage_quota(&state, std::slice::from_ref(&job_id)).await;
+            let existing = sqlx::query_as::<_, (String, String, Option<String>)>(
+                "SELECT id, status, blob_id FROM remember_jobs WHERE owner = $1 AND idempotency_key = $2",
+            )
+            .bind(owner)
+            .bind(&idem_key)
+            .fetch_optional(state.db.pool())
+            .await;
+            match existing {
+                Err(e) => {
+                    rate_limit::release_storage_quota(&state, &all_ids[idx + 1..]).await;
+                    return Err(AppError::Internal(format!(
+                        "Failed to look up analyze job: {}",
+                        e
+                    )));
+                }
+                Ok(None) => {
+                    rate_limit::release_storage_quota(&state, &all_ids[idx + 1..]).await;
+                    return Err(AppError::Internal(
+                        "Failed to create analyze job row: insert affected no rows".into(),
+                    ));
+                }
+                Ok(Some((existing_id, status, blob_id))) => {
+                    if status != "failed" || blob_id.is_some() {
+                        accepted_facts.push(AnalyzeAcceptedFact {
+                            text: fact_text,
+                            id: existing_id.clone(),
+                            job_id: existing_id.clone(),
+                        });
+                        job_ids.push(existing_id);
+                        continue;
+                    }
+                    if let Err(e) = sqlx::query(
+                        "UPDATE remember_jobs SET status = 'pending', error_msg = NULL, updated_at = NOW()
+                         WHERE id = $1 AND status = 'failed' AND blob_id IS NULL",
+                    )
+                    .bind(&existing_id)
+                    .execute(state.db.pool())
+                    .await
+                    {
+                        rate_limit::release_storage_quota(&state, &all_ids[idx + 1..]).await;
+                        return Err(AppError::Internal(format!(
+                            "Failed to reset analyze job: {}",
+                            e
+                        )));
+                    }
+                    if let Err(e) = rate_limit::reserve_storage_quota(
+                        &state,
+                        owner,
+                        &[crate::storage::db::StorageReservationRequest {
+                            id: existing_id.clone(),
+                            bytes: encrypted.len() as i64,
+                        }],
+                    )
+                    .await
+                    {
+                        rate_limit::release_storage_quota(&state, &all_ids[idx + 1..]).await;
+                        return Err(e);
+                    }
+                    upload_job_id = existing_id;
+                }
+            }
         }
 
         // Pick next wallet slot (round-robin) and enqueue UploadAndTransfer
@@ -584,14 +661,23 @@ pub async fn analyze(
                 package_id: state.config.package_id.clone(),
                 account_id: account_id.clone(),
                 agent_public_key: Some(auth_pubkey_base.clone()),
-                remember_job_id: Some(job_id.clone()),
+                remember_job_id: Some(upload_job_id.clone()),
                 prepare_claim_token: None,
                 epochs: state.config.walrus_storage_epochs,
             },
         )
         .await
         {
-            rate_limit::release_storage_quota(&state, &all_ids[idx..]).await;
+            let _ = sqlx::query(
+                "UPDATE remember_jobs SET status = 'failed', error_msg = $2, updated_at = NOW()
+                 WHERE id = $1 AND blob_id IS NULL",
+            )
+            .bind(&upload_job_id)
+            .bind(e.to_string())
+            .execute(state.db.pool())
+            .await;
+            rate_limit::release_storage_quota(&state, std::slice::from_ref(&upload_job_id)).await;
+            rate_limit::release_storage_quota(&state, &all_ids[idx + 1..]).await;
             return Err(AppError::Internal(format!(
                 "Failed to enqueue analyze job: {}",
                 e
@@ -601,17 +687,17 @@ pub async fn analyze(
         // Queued: jobs.rs owns this reservation from here on.
 
         tracing::info!(
-            job_id = %job_id,
+            job_id = %upload_job_id,
             wallet_index,
             fact_len = fact_text.len(),
             "analyze fact enqueued"
         );
         accepted_facts.push(AnalyzeAcceptedFact {
             text: fact_text,
-            id: job_id.clone(),
-            job_id: job_id.clone(),
+            id: upload_job_id.clone(),
+            job_id: upload_job_id.clone(),
         });
-        job_ids.push(job_id);
+        job_ids.push(upload_job_id);
     }
 
     let fact_count = job_ids.len();
@@ -636,7 +722,7 @@ pub async fn analyze(
 
 #[cfg(test)]
 mod tests {
-    use super::{ANALYZE_CONCURRENCY, MAX_ANALYZE_TEXT_BYTES};
+    use super::{analyze_fact_idempotency_key, ANALYZE_CONCURRENCY, MAX_ANALYZE_TEXT_BYTES};
     use crate::routes::remember::MAX_REMEMBER_TEXT_BYTES;
     use crate::services::extractor::MAX_ANALYZE_FACTS;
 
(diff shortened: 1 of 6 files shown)


services/server/src/routes/analyze.rs
```diff
@@ -28,6 +29,11 @@ use super::{collect_bounded_results, enqueue_wallet_job};
 
 const ANALYZE_CONCURRENCY: usize = 5;
 
+fn analyze_fact_idempotency_key(owner: &str, namespace: &str, fact_text: &str) -> String {
```
HoangDucBach, line 40: ## Major: idempotency key survives `forget`, causing silent data loss

This key is built from `owner + namespace + fact_text` and never expires. `/api/forget` only deletes from `vector_entries` (`storage/db.rs:1619`, `delete_by_namespace`) — it never touches `remember_jobs`, so the row and its `idempotency_key` survive a forget.

Repro:

1. `analyze("I drink green tea every morning")` — creates a row, uploads to Walrus, `status = 'done'`, `blob_id` set.
2. `forget(namespace)` — removes the vector entry, but the `remember_jobs` row above is untouched.
3. `analyze("I drink green tea every morning")` again — same key, insert hits the unique constraint, `ON CONFLICT DO NOTHING`, reuse branch at line 604 sees `blob_id.is_some()` and treats it as already stored.

Result: the API returns success, but nothing is written. The fact the user just tried to re-save is gone, silently, with no error anywhere.

The idempotency key needs either a short TTL, or `forget` needs to clear `idempotency_key` on the rows it affects.

services/server/src/routes/analyze.rs
```diff
@@ -548,21 +554,92 @@ pub async fn analyze(
...
+                    )));
+                }
+                Ok(None) => {
+                    rate_limit::release_storage_quota(&state, &all_ids[idx + 1..]).await;
+                    return Err(AppError::Internal(
+                        "Failed to create analyze job row: insert affected no rows".into(),
+                    ));
+                }
+                Ok(Some((existing_id, status, blob_id))) => {
+                    if status != "failed" || blob_id.is_some() {
```
HoangDucBach: ## Major: idempotency key ignores `importance`, so a second call can be dropped

`analyze_fact_idempotency_key` (line 32) hashes `owner + namespace + fact_text` only. `importance` is not part of the key, but it does affect ranking in `recall`.

Example:

- `analyze("likes rust", importance = LOW)` — stored with importance LOW.
- `analyze("likes rust", importance = HIGH)` — same key as above, hits the same conflict path, gets treated as a duplicate.

The second call's `importance` is dropped with no error. The fact stays at LOW even though the caller explicitly asked for HIGH.

Either fold `importance` into the key, or treat a changed `importance` on an existing row as an update rather than a no-op.

HoangDucBach (changes requested):
Two major issues in the analyze idempotency change, both flagged inline.

- The idempotency key never expires, and forget does not clear it, so re-analyzing text after forgetting it returns success while writing nothing.
- The key does not include importance, so a repeat call with a different importance silently drops that field.

Everything else, the /api/restore 503 mapping and 55s deadline, the embed fan-out cap, the OAuth authorize rate limit, and the gas-selection classification, looks correct.

Happy to switch this to approve once these are fixed, or if I am missing context that makes either of them a non-issue.

ducnmm:
@HoangDucBach both majors should be fixed on `3ca3d118`.

**Forget vs idempotency key.** `delete_by_namespace` now NULLs `remember_jobs.idempotency_key` for that owner+namespace in the same transaction as the vector delete + tombstone. Re-analyzing the same text after forget inserts a new job instead of hitting `ON CONFLICT` and returning success with no write. Keys in other namespaces are left alone.

**Importance.** The key stays `owner + namespace + fact_text` so HTTP retries still collapse onto one paid upload. A later extract of the same text with a different vital/standard/trivial bucket updates `vector_entries.importance` in place instead of dropping the new score or minting a second blob.

If the job is `done` but the live row is already gone (forget raced, or another delete path), the old key is released and the fact is written again. Failed-with-blob still collapses so we do not pay twice.

services/server/src/routes/analyze.rs
ducnmm: Fixed on `3ca3d118`. `forget` (`delete_by_namespace`) now NULLs `remember_jobs.idempotency_key` for that owner+namespace in the same transaction as the vector delete. Re-analyzing the same text after forget inserts a new job.

If a `done` job is still keyed but the live row is already gone, analyze also releases that key and writes again (covers expiry / a raced forget).

services/server/src/routes/analyze.rs
ducnmm: Fixed on `3ca3d118`. Importance stays out of the idempotency key so HTTP retries still collapse onto one paid upload. A later extract of the same text with a different bucket updates `vector_entries.importance` in place instead of dropping the new score or minting a second blob.

HoangDucBach (approved):
Thanks for the quick turnaround. Verified both on 3ca3d118, not just from the description.

Forget vs idempotency key: delete_by_namespace now clears idempotency_key for that owner+namespace inside the same transaction as the vector delete and tombstone insert, so a re-analyze after forget inserts fresh instead of silently colliding.

Importance: classify_analyze_job_reuse gives four clean outcomes (collapse, update in place, rewrite after forget, retry an unpaid failure), each covered by its own test, and the live-row-gone race is handled by falling back to a rewrite when the importance update affects zero rows.


Approved.

PR #737, opened 2026-08-22
fix(sdk): align remember-manual, embed, hex, and env=dev

### LINKED ISSUE #608: [SDK/Bug] rememberManual sends blob_id but /api/remember/manual requires encrypted_data
# SDK/Bug: `rememberManual` / `remember_manual` send `blob_id`, but `/api/remember/manual` requires `encrypted_data`

Reporter contact: Discord/Telegram `masterwattson`

## Summary

The lightweight manual APIs on both SDKs document a "register a pre-uploaded blob ID + vector" flow:

- TypeScript: `memwal.rememberManual({ blobId, vector, namespace? })`
- Python: `memwal.remember_manual(RememberManualOptions(blob_id=..., vector=...))`

But the current relayer contract for `POST /api/remember/manual` no longer accepts a pre-uploaded `blob_id`. It requires `encrypted_data` and uploads the blob itself.

As a result, the public SDK methods send a request body that the current server cannot deserialize/handle.

## Affected client code

TypeScript SDK, `packages/sdk/src/memwal.ts`:

```ts
async rememberManual(opts: RememberManualOptions): Promise<RememberManualResult> {
    return this.signedRequest<RememberManualResult>(
        "POST",
        "/api/remember/manual",
        {
            blob_id: opts.blobId,
            vector: opts.vector,
            namespace: opts.namespace ?? this.namespace,
        },
        { includeDelegateKey: false },
    );
}
```

Python SDK, `packages/python-sdk-memwal/memwal/client.py`:

```py
data = await self._signed_request(
    "POST",
    "/api/remember/manual",
    {
        "blob_id": opts.blob_id,
        "vector": opts.vector,
        "namespace": opts.namespace or self._namespace,
    },
    include_seal_session=False,
)
```

The Python option type also documents the old shape:

```py
@dataclass
class RememberManualOptions:
    blob_id: str
    vector: List[float]
    namespace: Optional[str] = None
```

## Current server contract

`services/server/src/types.rs`:

```rs
pub struct RememberManualRequest {
    pub encrypted_data: String, // base64-encoded SEAL-encrypted bytes
    pub vector: Vec<f32>,
    #[serde(default = "default_namespace")]
    pub namespace: String,
}
```

`services/server/src/routes/remember.rs`:

```rs
/// - Client sends {encrypted_data (base64), vector}
/// - Server uploads encrypted bytes to Walrus via upload-relay sidecar -> gets blob_id
/// - Server stores {blob_id, vector} in Vector DB
pub async fn remember_manual(
    ...
    Json(body): Json<RememberManualRequest>,
) -> Result<Json<RememberManualResponse>, AppError> {
    if body.encrypted_data.is_empty() {
        return Err(AppError::BadRequest(
            "encrypted_data cannot be empty".into(),
        ));
    }
    ...
}
```

Because `encrypted_data` is a required field, a request containing `blob_id` instead of `encrypted_data` cannot use the current route.

## Documentation also points users at the broken shape

`docs/sdk/api-reference.md`:

```md
| `rememberManual({ blobId, vector, namespace? })` | Register a pre-uploaded blob ID with a pre-computed vector |
```

`docs/python-sdk/usage/memwal-manual.md`:

```md
Register a pre-uploaded Walrus blob with a pre-computed vector. The relayer stores the `{blob_id, vector, owner, namespace}` mapping; it does not upload for you here.
```

## Expected behavior

The SDK, server, and docs should agree on one manual-write contract.

Either:

1. restore the pre-uploaded-blob registration route, so `blob_id + vector` is valid again; or
2. change the lightweight SDK methods/types/docs to accept/send `encrypted_data` instead of `blob_id`; or
3. remove/deprecate the lightweight `MemWal.rememberManual` / `MemWal.remember_manual` APIs and direct users to the full `MemWalManual.rememberManual(text)` flow, which already sends `{ encrypted_data, vector }`.

## Why this matters

Manual mode is documented as the escape hatch for advanced callers who manage Walrus upload, SEAL encryption, or embeddings themselves. Today those callers can follow the public SDK API exactly and still hit a server contract mismatch. This is especially confusing because:

- the TypeScript and Python SDKs both expose the old `blob_id + vector` shape;
- the docs still describe registering a pre-uploaded blob;
- the server route comments and request type now require encrypted bytes and perform the upload itself.

This is a public API break/friction point, not just an internal refactor mismatch.

## Suggested fix

The least surprising fix is to split the two use cases explicitly:

- `rememberManual({ encryptedData, vector, namespace? })` for "server uploads encrypted bytes";
- `registerManualBlob({ blobId, vector, namespace? })` or similar for "client already uploaded to Walrus".

If only the server-upload flow is intended going forward, update both SDKs and docs so no public method claims that a pre-uploaded `blob_id` is accepted.

### LINKED ISSUE #618: [SDK/Bug] embed() calls POST /api/embed, which the relayer server does not register — 100% 404
### Summary

Both the TypeScript and Python SDKs expose an `embed()` method that POSTs to `/api/embed`. The Rust relayer server never registers that route — `services/server/src/main.rs` only mounts remember/recall/analyze/ask/restore/forget/stats/bulk. Every `embed()` call fails with 404.

### Files

- `packages/python-sdk-memwal/memwal/client.py:760-773`
- `packages/sdk/src/memwal.ts:750`
- Server router: `services/server/src/main.rs` (no `/api/embed`)

### Evidence

Python SDK:

```python
# client.py:770
"/api/embed",
```

TS SDK:

```ts
// memwal.ts:750
return this.signedRequest<EmbedResult>("POST", "/api/embed", { text });
```

The server registers no such route (grep for `api/embed` in `services/server/` returns zero hits). The SDK's own example acknowledges the drift:

```python
# examples/async_remember_demo.py:172
print(f"  skipped (server may not expose /api/embed yet): {e}")
```

### Reproduction

```python
await memwal.embed("hello")
# -> _HttpStatusError: 404
```

### Impact

A documented public API method that can never work against the shipped server. Friction point for every SDK user; the example code soft-skips the error, which hides the problem.

### Suggested fix

Either register `/api/embed` on the server or remove the method (and the example) from both SDKs.

### LINKED ISSUE #726: [SDK/Bug] Python hex_to_bytes silently accepts whitespace -> mistyped/pasted key decodes to a different valid key (Python analog of TS LOW-25, unfixed)
**[SDK/Bug] Python SDK `hex_to_bytes` silently accepts whitespace-laced input -> a mistyped/pasted private key decodes to a *different* valid key with no error (Python analog of the TS LOW-25 hardening, still unfixed)**

### Summary
`hex_to_bytes()` in the Python SDK delegates straight to `bytes.fromhex()`, which **silently strips ASCII whitespace between byte pairs**. A private key copy-pasted with a stray space, tab, or trailing newline therefore decodes to a valid 32-byte seed, builds a working `SigningKey`, and derives a Sui address -- with **no error surfaced**. Several visually different input strings collapse to the *same* key, and an empty string returns `b''` instead of raising.

This is the exact bug class the TypeScript SDK was hardened against under **LOW-25** (`packages/sdk/src/utils.ts:41`, whose comment states: *"Strict validation -- rejects non-hex characters, odd-length input, and empty strings. Previously, `parseInt("zz", 16)` silently produced `NaN` which was coerced to `0`, yielding a wrong-but-valid-looking key."*). The Python port never received the same guard.

### Location
`packages/python-sdk-memwal/memwal/utils.py:32-33` (`hex_to_bytes`)
```python
clean = hex_str[2:] if hex_str.startswith("0x") else hex_str
return bytes.fromhex(clean)   # <-- silently strips interior ASCII whitespace; no charset/length/empty guard
```

### Reproduction
```python
# python3, against packages/python-sdk-memwal/memwal/utils.py
from memwal.utils import hex_to_bytes, build_signing_key, bytes_to_hex

good    = "ab" * 32                       # canonical 64-hex-char seed
spaced  = "  ".join(("ab" * 32)[i:i+2] for i in range(0, 64, 2))  # same bytes, spaces between pairs
newline = ("ab" * 32) + "\n"              # trailing newline

for name, s in [("canonical", good), ("space-grouped", spaced), ("trailing \\n", newline)]:
    seed = hex_to_bytes(s)
    print(f"{name:16} len={len(seed)} equal_to_canonical={seed == hex_to_bytes(good)}")

print("distinct strings?", good != spaced, "| identical seed?", hex_to_bytes(good) == hex_to_bytes(spaced))
print("empty ->", repr(hex_to_bytes("")))   # b''  (TS raises "empty hex string")
```

Output (Python 3.13.14):
```
canonical        len=32 equal_to_canonical=True
space-grouped    len=32 equal_to_canonical=True
trailing \n      len=32 equal_to_canonical=True
distinct strings? True | identical seed? True
empty -> b''
```

### Impact
- A user who pastes a key with a stray space/newline/tab silently authenticates as an **unexpected identity** instead of getting a clear error -- the precise failure mode LOW-25 was created to prevent.
- Empty input returns `b''` and fails later with an opaque length error rather than the intended `"empty hex string"`.
- **Input-normalization asymmetry between the two SDKs**: the same malformed string is rejected by the TS SDK but accepted by the Python SDK, so behavior depends on which SDK a caller happens to use.

### Fix
Mirror the TS LOW-25 guard (codebase style):
```python
def hex_to_bytes(hex_str: str) -> bytes:
    if not isinstance(hex_str, str):
        raise TypeError("hex_to_bytes: expected string input")
    clean = hex_str[2:] if hex_str[:2].lower() == "0x" else hex_str
    if clean == "":
        raise ValueError("hex_to_bytes: empty hex string")
    if len(clean) % 2 != 0:
        raise ValueError(f"hex_to_bytes: odd-length hex string (length={len(clean)})")
    if any(c not in "0123456789abcdefABCDEF" for c in clean):
        raise ValueError("hex_to_bytes: input contains non-hex characters")
    return bytes.fromhex(clean)
```

### Related (same file, same normalization family)
`normalize_private_key` at `utils.py:284` only strips a lowercase `0x` prefix:
```python
return candidate[2:] if candidate.startswith("0x") else candidate
```
The TS `normalizePrivateKey` strips **both** `0x` and `0X` (`utils.ts:241`). A key entered as `0XABAB...` (a form the TS SDK explicitly accepts) is passed through with the `0X` still attached and then crashes inside `bytes.fromhex` with `non-hexadecimal number found in fromhex() arg at position 1` -- an error that names nothing the user can act on. The docstring says *"Matches the TypeScript SDK's `normalizePrivateKey`"* -- it currently doesn't for the `0X` case. Fix: `candidate[2:] if candidate[:2].lower() == "0x" else candidate`.

### Notes
- `bytes.fromhex()` on Python 3.13 confirmed to skip interior whitespace (`'ab cd' -> abcd`, `'ab\ncd' -> abcd`).
- Distinct from #593 (base64 argument-spread crash) and from the TS LOW-25 fix (which only covered the TS `hexToBytes`). This is the unfixed Python side.
- Environment: `@mysten-incubation/memwal` Python SDK `packages/python-sdk-memwal/memwal/utils.py`, Python 3.13.14.

### LINKED ISSUE #727: [SDK/Bug] MemWalManual.recallManual decodes SEAL id with unguarded inline parseInt/.match! instead of LOW-25-hardened hexToBytes (silent NaN->0 + null-deref)
**[SDK/Bug] `MemWalManual.recallManual` decodes the SEAL id with an unguarded inline hex parse (`parseInt`/`.match!`) instead of the LOW-25-hardened `hexToBytes` — silent NaN->0 and a null-deref crash path**

### Summary
The manual (client-side-encryption) recall path decodes the SEAL object id from hex to bytes with a hand-rolled expression that reproduces exactly the failure modes `hexToBytes` was hardened against under **LOW-25**: a non-hex pair yields `parseInt(..,16) === NaN`, which `Uint8Array.from` silently coerces to `0` (wrong id, no error), and an id that doesn't match the `/.{1,2}/g` regex (e.g. empty string) makes the non-null assertion `!` dereference `null` and throw `TypeError: Cannot read properties of null (reading 'map')`. The repo already ships a safe decoder (`hexToBytes`, `utils.ts:41`) created specifically to stop this, but this call-site doesn't use it.

This is a robustness / cross-module-consistency bug, not a demonstrated exploit: `fullId` here comes from `EncryptedObject.parse(blob.data).id` (SEAL-library output), so in the normal path it is well-formed hex. It is filed because it silently diverges from the hardening the team already committed to (LOW-25) and turns a malformed/truncated blob id into either a wrong-but-silent key id or an opaque crash instead of a clear error.

### Location
`packages/sdk/src/manual.ts:502-504`
```ts
const idBytes = Array.from(
    Uint8Array.from(fullId.match(/.{1,2}/g)!.map((b: string) => parseInt(b, 16)))
);
```
Compare `utils.ts:41` (`hexToBytes`), whose comment: *"Strict validation — rejects non-hex characters, odd-length input, and empty strings. Previously, `parseInt(\"zz\", 16)` silently produced `NaN` which was coerced to `0`, yielding a wrong-but-valid-looking key."*

### Reproduction
```js
// node >= 18, standalone — mirrors the exact expression at manual.ts:503
function idBytesLikeMemwal(fullId) {
  return Array.from(Uint8Array.from(fullId.match(/.{1,2}/g).map((b) => parseInt(b, 16))));
}
const cases = {
  "canonical (64 hex)": "ab".repeat(32),
  "non-hex chars":      "zz".repeat(32),
  "empty string":       "",
};
for (const [name, id] of Object.entries(cases)) {
  try {
    const out = idBytesLikeMemwal(id);
    console.log(`${name.padEnd(20)} -> len=${out.length} first8=[${out.slice(0,8)}]`);
  } catch (e) {
    console.log(`${name.padEnd(20)} -> ${e.constructor.name}: ${e.message}`);
  }
}
```
Output (Node v22.23.1):
```
canonical (64 hex)   -> len=32 first8=[171,171,171,171,171,171,171,171]
non-hex chars        -> len=32 first8=[0,0,0,0,0,0,0,0]     (every byte silently 0 — wrong SEAL id)
empty string         -> TypeError: Cannot read properties of null (reading 'map')
```

### Impact
- A malformed / truncated `fullId` produces an all-zero (or partly-zero) SEAL id **with no error** — `seal_approve` is then invoked with the wrong id and decryption fails in a way that looks like a SEAL/network problem rather than a decode bug.
- An id that fails the regex throws a raw `TypeError` from the `!` assertion — no actionable message.
- Silent inconsistency with the LOW-25 hardening: the SDK validates hex in one module and hand-rolls the unsafe version in another.

### Fix
Use the existing hardened decoder (already exported from the same package):
```ts
import { hexToBytes } from "./utils.js";
// ...
const idBytes = Array.from(hexToBytes(fullId));
```
`hexToBytes` rejects odd-length, non-hex, and empty input with a clear message, removing both the NaN->0 and the null-deref paths.

### Notes
- Distinct from #593 (base64 argument-spread encode crash) — this is the hex **decode** path on recall.
- Environment: `@mysten-incubation/memwal` SDK `packages/sdk/src/manual.ts`, Node v22.23.1.

### LINKED ISSUE #728: [SDK/Bug] Python env="dev" documented in 3 docstrings but missing from ENV_PRESETS -> ValueError on a documented value
**[SDK/Bug] Python SDK: `env="dev"` is documented in 3 public docstrings but missing from `ENV_PRESETS` — passing the documented value crashes with `ValueError`**

### Summary
`MemWal.create()` and both middleware wrappers document `env` as accepting `"prod"`, `"dev"`, `"staging"`, or `"local"`. But `ENV_PRESETS` only defines `prod`, `staging`, `local`. Passing the documented `env="dev"` raises `ValueError` at client construction. The docs and the implementation ship a broken contract simultaneously — and the docs even disagree with each other (the `MemWalConfig` dataclass docstring lists only prod/staging/local).

### Location
`packages/python-sdk-memwal/memwal/types.py:24-28` (`ENV_PRESETS`) vs docstrings at `client.py:252` and the two middleware wrappers.
```python
ENV_PRESETS = {
    "prod": "https://relayer.memory.walrus.xyz",
    "staging": "https://relayer-staging.memory.walrus.xyz",
    "local": "http://127.0.0.1:8000",
}   # <-- no "dev", but client.py:252 docstring promises: '"prod", "dev", "staging", or "local"'
```

### Reproduction
```python
# python3, against packages/python-sdk-memwal/
from memwal.types import ENV_PRESETS
print("presets in code:", sorted(ENV_PRESETS))     # ['local', 'prod', 'staging']
# docstring at client.py:252 promises: prod / dev / staging / local
from memwal import MemWal
try:
    MemWal.create(key="ab"*32, account_id="0x1", env="dev")
except ValueError as e:
    print("env='dev' ->", type(e).__name__ + ":", e)
```
Output (Python 3.13.14):
```
presets in code: ['local', 'prod', 'staging']
env='dev' -> ValueError: Unknown env preset 'dev'. Valid presets: local, prod, staging
```

### Impact
A user following the public docstring (`env="dev"`) gets a hard `ValueError` at construction — a documentation/implementation contract break. Either the preset is missing or the docstrings are wrong; both are shipped at once.

### Fix
Reconcile the two. If a dev relayer exists, add it to `ENV_PRESETS`:
```python
ENV_PRESETS = {
    "prod": "https://relayer.memory.walrus.xyz",
    "dev": "https://relayer-dev.memory.walrus.xyz",
    "staging": "https://relayer-staging.memory.walrus.xyz",
    "local": "http://127.0.0.1:8000",
}
```
Otherwise drop `"dev"` from the `client.py:252` and both middleware docstrings (and align the `MemWalConfig` docstring, which currently lists only prod/staging/local).

### Notes
- Three docstrings advertise `dev`; the `MemWalConfig` dataclass docstring lists only prod/staging/local — the docs are internally inconsistent even before the code mismatch.
- Distinct from #593. Environment: `packages/python-sdk-memwal/memwal/{types,client,middleware}.py`, Python 3.13.14.

ducnmm:
Linear: [WALM-404](https://linear.app/mysten-labs/issue/WALM-404/align-sdk-remember-manual-embed-hex-and-envdev-contracts)
Fixes https://github.com/MystenLabs/MemWal/issues/608
Fixes https://github.com/MystenLabs/MemWal/issues/618
Fixes https://github.com/MystenLabs/MemWal/issues/726
Fixes https://github.com/MystenLabs/MemWal/issues/727
Fixes https://github.com/MystenLabs/MemWal/issues/728

## What this does

- `MemWal.rememberManual` / `remember_manual` send `encrypted_data` (base64 SEAL ciphertext) instead of a pre-uploaded `blob_id`.
- Relayer registers signed `POST /api/embed` so SDK `embed()` is no longer a 100% 404.
- Manual recall decodes SEAL ids with `hexToBytes`.
- Python `hex_to_bytes` rejects empty, odd-length, whitespace, and non-hex input.
- Adds the documented `env="dev"` preset (`https://relayer.dev.memwal.ai`).

## Package

- `@mysten-incubation/memwal` `0.1.4` → `0.1.5`
- `memwal` (Python) `0.1.7` → `0.1.8`

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -163,7 +163,7 @@ const stored = await memwal.waitForRememberJob(accepted.job_id, {
 
 | Method | Description |
 |---|---|
-| `rememberManual({ blobId, vector, namespace? })` | Register pre-uploaded blob with pre-computed vector |
+| `rememberManual({ encryptedData, vector, namespace? })` | Send SEAL-encrypted bytes + pre-computed vector; relayer uploads |
 | `recallManual({ vector, limit?, namespace? })` | Search with pre-computed vector (returns blob IDs only) |
 | `embed(text)` | Generate embedding vector (no storage) |
 
--- a/docs/python-sdk/api-reference.md
+++ b/docs/python-sdk/api-reference.md
@@ -211,7 +211,7 @@ Hex-encoded public key for the current delegate key.
 
 | Method | Description |
 | --- | --- |
-| `remember_manual(RememberManualOptions)` | Register a pre-uploaded blob with a pre-computed vector → `RememberManualResult` |
+| `remember_manual(RememberManualOptions)` | Send SEAL-encrypted bytes + a pre-computed vector; the relayer uploads to Walrus → `RememberManualResult` |
 | `recall_manual(RecallManualOptions)` | Search with a pre-computed query vector → `RecallManualResult` (blob_id + distance only) |
 | `embed(text)` | Embedding vector for text, no storage → `EmbedResult(vector)` |
 
--- a/docs/python-sdk/usage/memwal-manual.md
+++ b/docs/python-sdk/usage/memwal-manual.md
@@ -2,7 +2,7 @@
 title: "Manual Methods"
 description: >-
   Lower-level remember_manual, recall_manual, and embed methods for callers that manage
-  their own vectors or have pre-uploaded Walrus blobs in the Python SDK.
+  their own vectors or already have SEAL-encrypted payloads in the Python SDK.
 keywords:
   - Walrus Memory
   - MemWal
@@ -12,7 +12,7 @@ keywords:
   - remember_manual
   - recall_manual
 goal:
-  description: Call the manual Python SDK methods to register a pre-uploaded blob, search with a pre-computed embedding vector, and compute embeddings independently from storage.
+  description: Call the manual Python SDK methods to register SEAL-encrypted bytes with a pre-computed embedding vector, search with a pre-computed vector, and compute embeddings independently from storage.
   requires:
     - has_frontmatter:
         - title
@@ -31,13 +31,14 @@ questions:
   - When should I use manual methods vs standard remember/recall in MemWal?
 answer: >-
   The Python SDK provides lower-level manual methods on the same MemWal/MemWalSync client:
-  `embed` computes a vector without storing anything, `remember_manual` registers a
-  pre-uploaded blob with a pre-computed vector, and `recall_manual` searches with a
-  pre-computed query vector returning raw blob IDs and distances.
+  `embed` computes a vector without storing anything, `remember_manual` sends
+  base64 SEAL-encrypted bytes plus a pre-computed vector (the relayer uploads to
+  Walrus), and `recall_manual` searches with a pre-computed query vector returning
+  raw blob IDs and distances.
 ---
 
 <Note>
-Unlike the TypeScript SDK there is **no separate `MemWalManual` class** in Python. The Python SDK is relayer-backed: the relayer always handles embedding, SEAL encryption, and Walrus storage. The "manual" methods are lower-level entry points on the same `MemWal` / `MemWalSync` client for callers that already have a vector or a pre-uploaded blob.
+Unlike the TypeScript SDK there is **no separate `MemWalManual` class** in Python. The Python SDK is relayer-backed for the standard `remember` / `recall` flow. The "manual" methods are lower-level entry points on the same `MemWal` / `MemWalSync` client for callers that already have a vector and SEAL-encrypted bytes.
 </Note>
 
 Use these when you want to control indexing or do your own vector math. For the standard flow, prefer [`remember` / `recall`](/python-sdk/usage/memwal).
@@ -57,14 +58,14 @@ print(len(vec.vector))
 
 ## `remember_manual`
 
-Register a pre-uploaded Walrus blob with a pre-computed vector. The relayer stores the `{blob_id, vector, owner, namespace}` mapping; it does not upload for you here.
+Send base64 SEAL-encrypted bytes plus a pre-computed vector. The relayer uploads the ciphertext to Walrus and stores the `{blob_id, vector, owner, namespace}` mapping.
 
 ```python
 from memwal import RememberManualOptions
 
 result = await memwal.remember_manual(
     RememberManualOptions(
-        blob_id="<walrus-blob-id>",
+        encrypted_data="<base64-seal-ciphertext>",
         vector=vec.vector,
         namespace="chatbot-prod",   # optional; falls back to client default
     )
@@ -93,7 +94,7 @@ for hit in hits.results:
 | --- | --- |
 | Plain text, want it stored | `remember` / `remember_and_wait` |
 | Plain text, want only the vector | `embed` |
-| A vector + an already-uploaded blob | `remember_manual` |
+| A vector + SEAL-encrypted bytes | `remember_manual` |
 | A query vector, want raw hits | `recall_manual` |
 | Plain query text, want decrypted matches | `recall` |
 
--- a/docs/python-sdk/usage/memwal.md
+++ b/docs/python-sdk/usage/memwal.md
@@ -98,9 +98,9 @@ print(result.restored, result.skipped, result.total)
 
 ## Lower-Level Methods
 
-Use these when you already have a vector or a pre-uploaded blob — see [Manual methods](/python-sdk/usage/memwal-manual):
+Use these when you already have a vector or SEAL-encrypted bytes — see [Manual methods](/python-sdk/usage/memwal-manual):
 
-- `remember_manual(RememberManualOptions(blob_id=..., vector=..., namespace=...))`
+- `remember_manual(RememberManualOptions(encrypted_data=..., vector=..., namespace=...))`
 - `recall_manual(RecallManualOptions(vector=..., limit=..., namespace=...))`
 - `embed(text)` — embedding vector only, no storage
 - `get_public_key_hex()` — the delegate public key
--- a/docs/python-sdk/usage/memwal.md
+++ b/docs/python-sdk/usage/memwal.md
@@ -98,7 +98,7 @@ print(result.restored, result.skipped, result.total)
 
 ## Lower-Level Methods
 
-Use these when you already have a vector or SEAL-encrypted bytes — see [Manual methods](/python-sdk/usage/memwal-manual):
+Use these when you already have a vector or SEAL-encrypted bytes. See [Manual methods](/python-sdk/usage/memwal-manual):
 
 - `remember_manual(RememberManualOptions(encrypted_data=..., vector=..., namespace=...))`
 - `recall_manual(RecallManualOptions(vector=..., limit=..., namespace=...))`
--- a/docs/relayer/api-reference.md
+++ b/docs/relayer/api-reference.md
@@ -1,7 +1,7 @@
 ---
 title: API Reference
 description: >-
-  Complete HTTP API reference for the Walrus Memory relayer, including authentication headers, public routes, the protected endpoints for remember, recall, analyze, ask, restore, forget, and stats, and the MCP transport endpoints.
+  Complete HTTP API reference for the Walrus Memory relayer, including authentication headers, public routes, the protected endpoints for remember, recall, analyze, embed, ask, restore, forget, and stats, and the MCP transport endpoints.
 keywords:
   - Walrus Memory
   - MemWal
@@ -28,7 +28,7 @@ questions:
   - "How do I authenticate requests to the MemWal relayer?"
   - "What is the request and response format for the remember and recall APIs?"
 answer: >-
-  The Walrus Memory relayer exposes public routes (health, version, config, metrics, sponsor) and protected routes (remember, recall, analyze, ask, restore, forget, stats) that require Ed25519 signed headers, plus MCP transport endpoints that use bearer authentication. Signed authentication uses x-public-key, x-signature, x-timestamp, and x-nonce headers, with the signature covering the timestamp, method, path, body hash, nonce, and account ID.
+  The Walrus Memory relayer exposes public routes (health, version, config, metrics, sponsor) and protected routes (remember, recall, analyze, embed, ask, restore, forget, stats) that require Ed25519 signed headers, plus MCP transport endpoints that use bearer authentication. Signed authentication uses x-public-key, x-signature, x-timestamp, and x-nonce headers, with the signature covering the timestamp, method, path, body hash, nonce, and account ID.
 ---
 
 The Rust relayer exposes these routes. The route table lives in `services/server/src/main.rs`.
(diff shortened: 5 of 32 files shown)


jessiemongeon1:
## Style Guide Audit

All 6 file(s) pass the style guide audit.

HoangDucBach (approved):
Approved.

**Verified**
- `remember_manual`/`rememberManual` now send `encrypted_data`, matching the server's actual `/api/remember/manual` contract. The old `blob_id` field was a 100% failure mode per the linked issues.
- `hex_to_bytes`/`hexToBytes` reject whitespace, odd-length, and non-hex input instead of silently decoding a different key via `bytes.fromhex`'s auto-strip behavior.
- `/api/embed` is wired with `includeDelegateKey: false` / `include_seal_session=False` on both SDKs, consistent with the other manual-mode routes.

Issue #714, opened 2026-08-21
memwal_recall / memwal_remember hang for the full 30-minute MCP idle timeout after a login race, while the relayer stays healthy


akaniooh:
Summary

Triggering `memwal_login` twice in quick succession (a second call issued
before confirming whether the first had completed) leaves the local
`memwal-mcp` server process in a state where subsequent `memwal_recall`
and `memwal_remember` calls hang indefinitely — no response, no error —
until the MCP client's own idle timeout kills the call (Claude Code's
default is 30 minutes).

Steps to reproduce

1. Configure `memwal-mcp` in an MCP client (e.g. Claude Code `.mcp.json`).
2. Call `memwal_login`, get a sign-in URL, but before completing it in the
   browser, call `memwal_login` again (or otherwise trigger a second
   concurrent login attempt).
3. Complete a wallet sign-in from either link.
4. Call `memwal_recall` or `memwal_remember`.

Expected: the call either succeeds, or fails fast with a clear error
(e.g. "stale session, please retry login").

Actual: the call hangs with zero response. During the hang,
`curl https://relayer.memory.walrus.xyz/health` from the same machine
responds in under 2 seconds the entire time, the relayer itself is
healthy, so the stuck component is the local server process (likely
holding a stale/conflicting delegate-key session from the login race).

Worround: fully restart the MCP client so it respawns a fresh
`memwal-mcp` process. The fresh process then works immediately,
`memwal_health` and `memwal_remember` both succeed on the first call.

Suggestion: either reject/cancel an in-flight login when a second one
starts, or have recall/remember fail fast with a clear "credentials stale,
re-run memwal_login" error instead of hanging indefinitely.

nikola0x0:
Resolved in https://github.com/MystenLabs/MemWal/pull/738 and https://github.com/MystenLabs/MemWal/pull/745.

Issue #629, opened 2026-08-13
memwal-mcp login exits 0 without logging in when stdin is not a TTY


Olalekan2345:
### Summary

`memwal-mcp login` reports success and does nothing when run without an interactive terminal.

### Reproduction

Run the login command from any non-interactive context — a script, a CI step, an editor task runner,
or any tool that spawns a child process without a TTY:

    npx -y @mysten-incubation/memwal-mcp login --label my-project

Output:

    warn  creds.missing_at_spawn.serving_auth_required
    info  auth_required_server.started
    info  auth_required_server.closed

Exit code **0**. No browser opens, no URL is printed, no credential file is written.

### Cause

`dist/index.js` parses `login` correctly and sets `forceLogin`, but a TTY check at line 138 fires
before `loginFlow()` is reached at line 176:

```js
if (!process.stdin.isTTY) {
    // serve the auth-required stub instead
}
```

The environment check silently overrides an explicit user command.

### Impact

Scripted or automated setup appears to succeed while leaving the user unauthenticated. The
consequence surfaces much later as unexplained auth failures, with nothing pointing back to the login
that never happened.

### Expected

When `login` is passed explicitly and no TTY is available, fail loudly:

    error: `login` requires an interactive terminal (stdin is not a TTY).
           Run it directly in a terminal, or set credentials via <documented alternative>.

Exit non-zero. The implicit case — a server spawned without credentials — should keep its current
stub behaviour; only the explicit command should error.

HoangDucBach:
[WALM-389](https://linear.app/mysten-labs/issue/WALM-389/memwal-mcp-login-exits-0-without-logging-in-when-stdin-is-not-a-tty)

DannyWebDesigns:
Still reproduces on 2026-08-21 with memwal-mcp 0.0.10 via npx on macOS 15 / Node 22.22, Mainnet relayer: run from a non-interactive context (an agent's shell tool), `memwal-mcp login` printed the URL, waited out the 5-minute window, and exited without a non-zero status. Checked the dashboard afterwards: the attempt's public key (e9c671bd…) was never registered on-chain, so unlike #384 this run did not silently succeed — it silently did nothing. A TTY check up front, or a non-zero exit with "no key registered", would have saved a debugging session. Found while building a Prompt Evolution entry (BuildMEM v2).

nikola0x0:
Resolved in https://github.com/MystenLabs/MemWal/pull/738.

PR #738, opened 2026-08-22
fix(mcp): fail non-TTY login and single-flight memwal_login

ducnmm:
Linear: [WALM-405](https://linear.app/mysten-labs/issue/WALM-405/fix-mcp-login-non-tty-exit-and-concurrent-memwal-login-race)
Fixes https://github.com/MystenLabs/MemWal/issues/629
Fixes https://github.com/MystenLabs/MemWal/issues/714

Stacked on #735 so `@mysten-incubation/memwal-mcp` stays at `0.0.11`.

## What this does

- `memwal-mcp login` with no TTY exits 1 instead of booting the auth-required stub and reporting success. Implicit MCP spawn without credentials still uses the stub.
- Concurrent `memwal_login` calls reuse the in-flight listener and URL so a second flow cannot hang later remember/recall.


HoangDucBach (approved):
LGTM!

PR #744, opened 2026-08-23
fix(relayer): align analyze charge windows and report write_ready

### LINKED ISSUE #609: charge_explicit_weight appears to prune current rate-limit history by using window_start = now
Reporter contact: Discord/Telegram @masterwattson

## Severity

Medium

Rate-limit accounting error. Extra requests may be accepted within the configured sliding window because valid in-window usage can be removed from Redis.

## Summary

In `services/server/src/rate_limit.rs`, the normal rate-limit middleware records usage with real sliding-window lower bounds, for example:

```rust
let dk_window_start = now - 60_000.0;
let burst_window_start = now - 60_000.0;
let hourly_window_start = now - 3_600_000.0;
```

However, `charge_explicit_weight`, which is called after `/api/analyze` finishes and the final fact count is known, appears to call `check_and_record_window` with `window_start = now`:

```rust
let _ = check_and_record_window(&mut redis, &dk_key, now, now, i64::MAX, weight, 120).await;

let _ = check_and_record_window(
    &mut redis,
    &burst_key,
    now,
    now + 0.1,
    i64::MAX,
    weight,
    120,
)
.await;

let _ =
    check_and_record_window(&mut redis, &hr_key, now, now + 0.2, i64::MAX, weight, 3700).await;
```

The shared Lua path always prunes before inserting the new charge:

```lua
redis.call('ZREMRANGEBYSCORE', key, '-inf', window_start)
```

Because Redis removes scores inclusively through `window_start`, passing `window_start = now` removes existing members scored `<= now`. On the same rate-limit keys used by the live limiter, this can clear valid current-window history before inserting the post-hoc charge.

Affected keys include:

```text
rate:...
rate:dk:...
rate:hr:...
```

The apparent intent of `charge_explicit_weight` is to add extra post-hoc weight after the final `/api/analyze` fact count is known. The current window bounds do not preserve existing in-window history.

## Expected Behavior

Post-hoc charging should add the discovered extra weight while preserving existing usage inside the active sliding window.

For minute windows, the lower bound should match the middleware behavior:

```rust
let minute_window_start = now - 60_000.0;
```

For hourly windows, the lower bound should use the hourly equivalent:

```rust
let hourly_window_start = now - 3_600_000.0;
```

If post-hoc charging should never deny because the response has already completed, the effective limit can remain permissive, for example `i64::MAX`, while still using the correct sliding-window lower bound.

## Actual Behavior

`charge_explicit_weight` passes `window_start = now`.

Since the Lua script runs:

```lua
redis.call('ZREMRANGEBYSCORE', key, '-inf', window_start)
```

this removes existing entries with scores at or before `now`, including entries that are still inside the intended 60-second or 1-hour window.

After the new charge is inserted, Redis may contain only the new charge instead of the previous in-window usage plus the new charge.

## Local Redis Verification

I verified this against a real local Redis 7.2.5 instance only. No MemWal/relayer endpoints or hosted services were called.

Redis was run locally on `127.0.0.1:6380`, using disposable DB index `15` and a throwaway key.

Setup used:

```bash
curl -fsSLO https://download.redis.io/releases/redis-7.2.5.tar.gz
tar xzf redis-7.2.5.tar.gz
make -C redis-7.2.5 -j2 MALLOC=libc BUILD_TLS=no

/tmp/redis-7.2.5/src/redis-server \
  --bind 127.0.0.1 \
  --port 6380 \
  --daemonize yes \
  --save '' \
  --appendonly no
```

Verification commands and output:

```bash
$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 FLUSHDB
OK

# Case 1: charge_explicit_weight-style bound: window_start = now
$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 ZADD memwal:rate-limit-window-proof 950000 m1 970000 m2 990000 m3
3

$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 ZCARD memwal:rate-limit-window-proof
3

$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 EVAL "redis.call('ZREMRANGEBYSCORE', KEYS[1], '-inf', ARGV[1]); return redis.call('ZCARD', KEYS[1])" 1 memwal:rate-limit-window-proof 1000000
0

$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 ZADD memwal:rate-limit-window-proof 1000000 charge
1

$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 ZCARD memwal:rate-limit-window-proof
1

# Case 2: middleware-style bound: window_start = now - 60000
$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 DEL memwal:rate-limit-window-proof
1

$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 ZADD memwal:rate-limit-window-proof 950000 m1 970000 m2 990000 m3
3

$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 ZCARD memwal:rate-limit-window-proof
3

$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 EVAL "redis.call('ZREMRANGEBYSCORE', KEYS[1], '-inf', ARGV[1]); return redis.call('ZCARD', KEYS[1])" 1 memwal:rate-limit-window-proof 940000
3

$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 ZADD memwal:rate-limit-window-proof 1000000 charge
1

$ /tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 ZCARD memwal:rate-limit-window-proof
4
```

Result:

- With `window_start = now`, Redis prunes the 3 valid in-window members, then the new charge leaves `ZCARD = 1`.
- With `window_start = now - 60000`, Redis preserves the 3 valid in-window members, then the new charge leaves `ZCARD = 4`.

Cleanup:

```bash
/tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 -n 15 FLUSHDB
/tmp/redis-7.2.5/src/redis-cli -h 127.0.0.1 -p 6380 SHUTDOWN NOSAVE
```

## Suggested Fix

Use the same sliding-window lower bounds in `charge_explicit_weight` that the middleware uses.

For example:

```rust
let dk_window_start = now - 60_000.0;
let burst_window_start = now - 60_000.0;
let hourly_window_start = now - 3_600_000.0;
```

Then call `check_and_record_window` with those lower bounds while keeping `i64::MAX` if post-hoc charging should record usage without denying after the request has completed.

Conceptually:

```rust
let _ = check_and_record_window(
    &mut redis,
    &dk_key,
    dk_window_start,
    now,
    i64::MAX,
    weight,
    120,
)
.await;

let _ = check_and_record_window(
    &mut redis,
    &burst_key,
    burst_window_start,
    now + 0.1,
    i64::MAX,
    weight,
    120,
)
.await;

let _ = check_and_record_window(
    &mut redis,
    &hr_key,
    hourly_window_start,
    now + 0.2,
    i64::MAX,
    weight,
    3700,
)
.await;
```

## Impact

This appears to be an incorrect sliding-window parameter rather than an intentional redesign. The post-hoc charge path can unintentionally prune current-window rate-limit history from the same Redis keys used by the normal limiter, causing subsequent checks to observe less recorded usage than expected.

### LINKED ISSUE #709: `memwal_health` reports `ok` during a total write-path outage, and the write error leaks an internal `localhost:9000` address that misleads users into debugging their own machine
# MemWal bug report — ready to paste

File at: https://github.com/MystenLabs/MemWal/issues

Checked against open issues #575–#593 — not a duplicate. The nearest is #580 (post-login
credential refresh → 401), which is an **auth** failure. This is not auth: login was valid
and reads kept working throughout.

Remove the account identifiers at the bottom if you'd rather not publish them. They only
help the engineers find server logs.

---

## Title

`memwal_health` reports `ok` during a total write-path outage, and the write error leaks an internal `localhost:9000` address that misleads users into debugging their own machine

---

### Summary

During a study session on 2026-08-16 the write path failed almost completely — **10 of 12
`memwal_remember` calls were lost, only 2 persisted.** Reads were unaffected: `memwal_recall`
and decryption worked normally the entire time.

Two things made this much worse than a plain outage:

1. **`memwal_health` returned `ok` throughout**, while writes were failing 100%. The health
   check does not cover the write path, so a green result is not evidence that memory is
   working.
2. **The write error surfaced `http://localhost:9000/seal/encrypt`.** That reads as a service
   on the user's own machine. It is not one — the published npm client contains no reference
   to port 9000 or to `seal` anywhere. The agent assisting me believed it and instructed me to
   restart a local process that does not exist on my computer.

Individually each is a papercut. Together they mean a user can lose data silently, be told by
the health check that everything is fine, and then waste an hour debugging their own setup.

### Environment

- `@mysten-incubation/memwal-mcp` **0.0.6** (stdio MCP server, Claude Code)
- Relayer `https://relayer.memory.walrus.xyz`, reported version **0.1.0**
- Node **v22.17.1**, Windows 11
- Date/time: **2026-08-16, approx. 18:00–19:00 UTC+1**

### What happened

1. Ran a study session that generated 12 memories to write across namespaces
   `move-core`, `sui-obj`, `move-sec`, `exam-intel`.
2. `memwal_remember` calls began timing out. The agent backed off rather than retrying hard.
3. `memwal_health` was called and returned `ok` (relayer up, v0.1.0).
4. `memwal_recall` and `memwal_restore` continued to work correctly against existing data,
   including decryption.
5. Final state, verified by `recall` + `restore` per namespace:

   | namespace | stored |
   |---|---|
   | `move-core` | 0 |
   | `sui-obj` | 1 |
   | `move-sec` | 1 |
   | `exam-intel` | 0 |

   **2 of 12 written. 10 lost.** There is no local queue, so the unsent entries existed only
   in the chat session.
6. The error text referenced `http://localhost:9000/seal/encrypt`.

### Why the error message is a bug in its own right

`localhost` in a user-facing error means "your machine" to every reader. Encryption in this
architecture happens server-side — the client's own README describes the package as a bridge
to the hosted relayer, which handles Seal encryption and Walrus storage.

I verified the published `0.0.6` bundle: it contains **no** occurrence of `9000` and **no**
occurrence of `seal`. Its only local addresses are the loopback listener used for the browser
login callback and a `localhost:5173` dev URL.

So the address cannot be acted on by the user, and it actively points them in the wrong
direction. A capable agent read it, concluded a local sidecar had died, and told me to restart
it. There was nothing to restart.

### Impact

- **Silent data loss.** Writes fail, nothing is queued locally, and the memories are gone once
  the session ends.
- **The obvious diagnostic lies.** `health` is the tool a user reaches for, and it said `ok`.
- **The error sends users the wrong way.** Time lost debugging a non-existent local process.

For a memory product this is the worst failure shape: the user believes their data is safe
when it is not.

### Suggested fixes

1. **Make `memwal_health` cover the write path end to end** — or return a per-component
   status (`relayer`, `encryption`, `storage`) rather than a single `ok`. A health check that
   passes during a total write outage is worse than none, because it is trusted.
2. **Stop leaking internal hostnames into client-facing errors.** Map them to something the
   user can act on: *"Encryption service unavailable on the Walrus relayer — this is
   server-side, retry shortly."* Explicitly say it is not the user's machine.
3. **Consider a local retry queue** for failed writes, so a transient outage does not destroy
   captured memories when the session closes.

### Reproducibility

I cannot trigger the outage on demand — it was an incident window, and writes recovered later
the same evening. Server logs for the account below should show it.

Items 1 and 2 do not depend on reproducing the outage. Both are constant properties of the
current design: `health` never checks the write path, and the error string is emitted whenever
that failure occurs.

- Delegate public key: `7aa33f567bb3c3d64d6f97f6b5fa1a38d92f0193a7920dc94d5f1fc0d1d3779f`
- MemWal account ID: `0xe9455a5452488deb949d015c68aadfb333e1617ccb9446ec3c17cc42d77a67d5`

ducnmm:
Linear: [WALM-406](https://linear.app/mysten-labs/issue/WALM-406/fix-analyze-rate-limit-window-prune-and-health-write-path-signal)
Fixes https://github.com/MystenLabs/MemWal/issues/609
Fixes https://github.com/MystenLabs/MemWal/issues/709

## What this does

- `charge_explicit_weight` used `window_start = now`, so Redis pruned in-window history after `/api/analyze`. Charge now shares the live limiter windows (`now-60s` / `now-60s` / `now-1h`).
- `GET /health` stays HTTP 200. New `write_ready` reports whether the encryption sidecar process answered `/health` (2s cache, 300ms probe).
- Seal transport errors map to `UpstreamUnavailable` with a generic client message. The reqwest URL is logged server-side only.
- Sidecar MCP `health` / `recall` tools surface `write_ready` and `dropped_count`.

### DIFF
--- a/services/server/scripts/mcp/tools/recall.ts
+++ b/services/server/scripts/mcp/tools/recall.ts
@@ -83,12 +83,20 @@ export function registerRecallTool(
         },
         wrapTool<{ query: string; limit: number; namespace?: string }>(async ({ query, limit, namespace }) => {
             const result = await session.memwal.recall(query, limit, namespace);
+            const dropped =
+                typeof (result as { dropped_count?: number }).dropped_count === "number"
+                    ? (result as { dropped_count: number }).dropped_count
+                    : 0;
             if (result.results.length === 0) {
+                const empty =
+                    dropped > 0
+                        ? `No matching memories could be returned (${dropped} matched but failed to download or decrypt). This is not an empty namespace.`
+                        : "No matching memories found.";
                 return {
                     content: [
                         {
                             type: "text",
-                            text: "No matching memories found.",
+                            text: empty,
                         },
                     ],
                 };
@@ -104,6 +112,11 @@ export function registerRecallTool(
             if (collapsed > 0) {
                 lines.push(
                     `\n(${collapsed} duplicate ${collapsed === 1 ? "copy" : "copies"} of the above collapsed; the same fact is stored more than once.)`
+                );
+            }
+            if (dropped > 0) {
+                lines.push(
+                    `\n(${dropped} additional matches could not be decrypted and were omitted.)`,
                 );
             }
             return {
--- a/services/server/scripts/mcp/tools/recall.ts
+++ b/services/server/scripts/mcp/tools/recall.ts
@@ -83,10 +83,8 @@ export function registerRecallTool(
         },
         wrapTool<{ query: string; limit: number; namespace?: string }>(async ({ query, limit, namespace }) => {
             const result = await session.memwal.recall(query, limit, namespace);
-            const dropped =
-                typeof (result as { dropped_count?: number }).dropped_count === "number"
-                    ? (result as { dropped_count: number }).dropped_count
-                    : 0;
+            const droppedRaw = (result as { dropped_count?: unknown }).dropped_count;
+            const dropped = typeof droppedRaw === "number" ? droppedRaw : 0;
             if (result.results.length === 0) {
                 const empty =
                     dropped > 0
--- a/docs/relayer/api-reference.md
+++ b/docs/relayer/api-reference.md
@@ -80,7 +80,7 @@ These routes require no authentication.
 
 ### `GET /health`
 
-Service liveness check. `status` is `"ok"` when the relayer process is up. `write_ready` is `true` when the encryption sidecar process answered its own `/health` (cached a few seconds). A write outage can still return HTTP 200 with `write_ready: false`; `write_ready: true` is sidecar liveness, not a guarantee that remember/analyze will succeed.
+Service liveness check. `status` is `"ok"` when the relayer process is up. `write_ready` is `true` when the encryption sidecar process answered its own `/health` (cached a few seconds). A write outage can still return HTTP 200 with `write_ready: false`. `write_ready: true` is sidecar liveness, not a guarantee that remember or analyze succeed.
 
 **Response:**
 
--- a/docs/relayer/api-reference.md
+++ b/docs/relayer/api-reference.md
@@ -80,7 +80,7 @@ These routes require no authentication.
 
 ### `GET /health`
 
-Service health check.
+Service liveness check. `status` is `"ok"` when the relayer process is up. `write_ready` is `true` when the encryption sidecar process answered its own `/health` (cached a few seconds). A write outage can still return HTTP 200 with `write_ready: false`; `write_ready: true` is sidecar liveness, not a guarantee that remember/analyze will succeed.
 
 **Response:**
 
@@ -106,7 +106,8 @@ Service health check.
   "prompt_versions": {
     "extract": "extract.v1",
     "ask": "ask.v1"
-  }
+  },
+  "write_ready": true
 }
 ```
 
--- a/services/server/scripts/mcp/tools/health.ts
+++ b/services/server/scripts/mcp/tools/health.ts
@@ -23,11 +23,18 @@ export function registerHealthTool(
         },
         wrapTool<Record<string, never>>(async () => {
             const result = await session.memwal.health();
+            const extra = result as { write_ready?: boolean };
+            const writeNote =
+                extra.write_ready === false
+                    ? " write_ready=false (relayer is up; encryption sidecar did not answer health)"
+                    : extra.write_ready === true
+                      ? " write_ready=true"
+                      : "";
             return {
                 content: [
                     {
                         type: "text",
-                        text: `Walrus Memory is reachable. status=${result.status} version=${result.version}`,
+                        text: `Walrus Memory is reachable. status=${result.status} version=${result.version}${writeNote}`,
                     },
                 ],
             };
--- a/services/server/src/rate_limit.rs
+++ b/services/server/src/rate_limit.rs
@@ -118,6 +118,13 @@ impl RateLimitConfig {
     }
 }
 
+/// Sliding-window lower bounds shared by the live limiter and post-hoc
+/// `charge_explicit_weight`. Passing `now` as `window_start` would prune
+/// every in-window Redis member before inserting the extra charge.
+fn sliding_window_starts(now: f64) -> (f64, f64, f64) {
+    (now - 60_000.0, now - 60_000.0, now - 3_600_000.0)
+}
+
 // ============================================================
 // Cost Weights — per endpoint
 // ============================================================
@@ -447,9 +454,7 @@ pub async fn rate_limit_middleware(
     let burst_key = format!("rate:{}", auth.owner);
     let hourly_key = format!("rate:hr:{}", auth.owner);
 
-    let dk_window_start = now - 60_000.0; // 1-min window (ms)
-    let burst_window_start = now - 60_000.0; // 1-min window (ms)
-    let hourly_window_start = now - 3_600_000.0; // 1-hr  window (ms)
+    let (dk_window_start, burst_window_start, hourly_window_start) = sliding_window_starts(now);
 
     // --- Atomic check-and-record via Lua script for all 3 layers ---
     // Each layer is checked+recorded atomically. If Redis is unavailable,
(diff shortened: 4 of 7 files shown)


jessiemongeon1:
## Style Guide Audit

services/server/scripts/mcp/tools/recall.ts
```diff
@@ -83,12 +83,20 @@ export function registerRecallTool(
         },
         wrapTool<{ query: string; limit: number; namespace?: string }>(async ({ query, limit, namespace }) => {
             const result = await session.memwal.recall(query, limit, namespace);
+            const dropped =
+                typeof (result as { dropped_count?: number }).dropped_count === "number"
+                    ? (result as { dropped_count: number }).dropped_count
+                    : 0;
```
HoangDucBach: Blocking: this fails the build. TS2352 — the second `as` cast to `{ dropped_count: number }` doesn't sufficiently overlap with `RecallResult`.

```suggestion
            const droppedRaw = (result as { dropped_count?: unknown }).dropped_count;
            const dropped = typeof droppedRaw === "number" ? droppedRaw : 0;
```

HoangDucBach (changes requested):
The rate-limit fix and the seal-error-URL-leak fix are both correct and verified against their described behavior. One thing blocks the build.

HoangDucBach (approved):
Approved. A suggested fix is already on the recall.ts thread for the build error — please check and apply.

HoangDucBach (approved):
Approved. 
A suggested fix is already on the recall.ts thread for the build error — please check and apply.

ducnmm:
Applied the suggested TS2352 fix on `recall.ts`: read `dropped_count` as `unknown` and only use it when it is a number. That is the line that failed Compile & CLI Smoke.

Once checks go green I will merge.

Issue #739, opened 2026-08-23
memwal-mcp stops responding after concurrent memwal_recall calls — even memwal_health fails until restart, while the relayer/SDK stay healthy


DannyWebDesigns:
**Summary.** Issuing several `memwal_recall` calls in the same agent turn makes every call fail, and the MCP server does not recover afterwards — even `memwal_health` fails — while the relayer itself is healthy (the TypeScript SDK against the same relayer keeps answering).

**Environment.** `@mysten-incubation/memwal-mcp` 0.0.10 via `npx -y … --prod --namespace <ns>`, Claude Code (VS Code extension), Node 22.22, macOS 15 arm64, hosted Mainnet relayer. Reproduced 2026-08-22 in two separate Claude Code sessions on the same machine.

**Steps to reproduce.**
1. Configure the MCP server and log in; confirm `memwal_health` → `status=ok`.
2. Have the agent issue 4 `memwal_recall` calls in one turn (different queries, same namespace, limit 3–5).
3. All four return: `Walrus Memory did not answer this call. The connection to the relayer dropped before the result came back. Please retry.`
4. Call `memwal_health` → same error. Call a single `memwal_recall` → same error. This persists for the life of the process.
5. In parallel, `MemWal.create({...}).health()` / `.recall()` through `@mysten-incubation/memwal` 0.1.3 with the same credentials → `status: ok`, full results in < 2 s.

**Expected.** Concurrent calls are queued or serialised; one failed call does not poison the bridge; `memwal_health` reflects the relayer rather than the bridge's own broken state.

**Actual.** Every subsequent tool call fails until the MCP server process is restarted (reload the editor window).

**Impact.** Prompts that recall proactively (session-start handoff + pre-action check + dedup-before-write) naturally batch calls. The only workaround is strictly one memwal call per turn, which costs several extra round-trips per task.

Related but distinct: #632 (timeout with no retry/diagnostic), #714 (hang after a login race), #709 (`memwal_health` ok during a write outage). This one is triggered purely by concurrency, with no login involved, and the bridge never recovers on its own.

ducnmm:
[WALM-408](https://linear.app/mysten-labs/issue/WALM-408/serialize-mcp-sse-posts-so-concurrent-recall-cannot-hang-the-bridge)
Fix in https://github.com/MystenLabs/MemWal/pull/745

PR #745, opened 2026-08-23
fix(mcp): serialize SSE POSTs so concurrent recall cannot hang the bridge

ducnmm:
Linear: [WALM-408](https://linear.app/mysten-labs/issue/WALM-408/serialize-mcp-sse-posts-so-concurrent-recall-cannot-hang-the-bridge)
Fixes https://github.com/MystenLabs/MemWal/issues/739

Stacked on #738 so `@mysten-incubation/memwal-mcp` stays at `0.0.11`.

## What this does

- Serialize outbound MCP POSTs on the SSE session so a burst of `memwal_recall` cannot drop the stream.
- Skip queued posts against a stale session URL after reconnect (reconnect already replays in-flight calls).
- Login timeout now warns that an on-chain delegate key may already exist if the browser approved after the local listener expired.


HoangDucBach (approved):
LGTM!

PR #746, opened 2026-08-23
fix(app): show namespace counts and persist playground namespace

### LINKED ISSUE #723: Dashboard shows no memories and no blob count for an account holding 185 blobs; the Playground's namespace field resets to default on navigation
## Summary

The dashboard at https://memory.walrus.xyz reports **no memories and no blob count** for an account
that holds 185 blobs on mainnet. The only control that could fix it — the namespace field — lives in
the Developer Playground rather than on the dashboard, and it **does not persist**: navigating to the
dashboard and back resets it to `default`.

The data is fine. The same account, same credentials, returns records immediately from the CLI. Only
the UI cannot reach them.

## Reproduction

1. Sign in to https://memory.walrus.xyz with an account whose memories are all in explicit
   namespaces — nothing written to `default`.
2. Dashboard shows no blob count and no memories. Its reload control changes nothing.
3. Open the Developer Playground, set the namespace to a real one, run a recall — records come back
   normally, decrypted, with `distance` values.
4. Navigate back to the dashboard, then return to the Playground. The namespace field has reset to
   `default`.

Confirmed against the same account at the same time:

```
memwal_recall namespace="markov.state.markov-v2" query="current task checkpoint goal status" limit=100
  -> 6 records, newest 2026-08-20T14:08:00Z
memwal_health
  -> status=ok version=0.1.0
```

Namespaces on the account: `markov.index` (1), `markov.facts.global` (11), `markov.capsules` (0),
`markov.state.markov-v2` (6), `markov.facts.markov-v2` (7), plus four `mk.exp*` namespaces (160).
None of them is `default`, and the dashboard shows none of them.

## Likely cause

This looks like the first-party consequence of #626 / #698 — there is no operation to enumerate
namespaces. With nothing to iterate, the dashboard can only show one namespace, falls back to
`default`, and finds it empty. Any agent following the documented per-topic namespace pattern is
therefore invisible to its own account UI by construction.

Worth noting how strongly the docs push users into this state: the MCP setup guidance advises against
setting a global default namespace so that per-call routing is explicit. Follow that advice and your
dashboard is guaranteed to be empty.

## Why it matters beyond cosmetics

The dashboard is where a user goes to confirm their memory exists, check a blob count, and manage
delegate keys. **Showing zero for a populated account is indistinguishable from total data loss**,
which is an alarming thing to display about encrypted storage the user cannot otherwise inspect. My
first reaction was that something had gone wrong with the account, and the only reason I knew
otherwise is that I had a CLI open in another window.

It also leaves the blob count genuinely unobtainable through any first-party surface — see #626, and
#385 on the Prompt Evolution session's own proof fields, which asks entrants to report exactly this
number. I ended up writing a script that recalls every namespace I could name and sums the results,
which only works because I wrote them and therefore remember what they are called.

## Suggested fix, cheapest first

1. **Persist the Playground's namespace across navigation.** One-line fix; makes the existing
   workaround usable today.
2. **Put a namespace input on the dashboard itself**, even with the caveat that you must know the
   name. Honest and immediately useful.
3. **Label the empty state with its scope.** "No memories in namespace `default`" is accurate and
   reassuring; a bare empty state reads as data loss.
4. **Best — add namespace enumeration (#626, #698)** and have the dashboard list namespaces with
   per-namespace counts. That fixes this properly, removes the need for every agent prompt to
   maintain its own registry namespace, and makes the blob count self-serve.

## Environment

- https://memory.walrus.xyz, observed 2026-08-20 / 2026-08-21
- Relayer `https://relayer.memory.walrus.xyz`, relayer 0.1.0, API 1.0.0, build `2162d261`
- Signing in from a browser that had never held a delegate key prompted creation of a new one. That
  worked correctly and was additive — the existing CLI delegate kept working, verified by recall
  immediately afterwards — so this is not a key-scoping problem. Possibly related to #719.

ducnmm:
Linear: [WALM-409](https://linear.app/mysten-labs/issue/WALM-409/show-dashboard-namespace-counts-and-persist-playground-namespace)
Fixes https://github.com/MystenLabs/MemWal/issues/723

## What this does

- Dashboard lists owner namespaces and memory counts from `GET /v1/owners/{owner}/namespaces` (signed GET). Each row links into the playground.
- Playground persists namespace per account in `localStorage`. A dashboard `?namespace=` link is applied once, then dropped so it cannot overwrite another account.
- Connect MCP callback-fail copy tells the user a delegate key was still added on-chain and to remove unused keys from the dashboard.

### DIFF
--- a/apps/app/src/pages/Playground.tsx
+++ b/apps/app/src/pages/Playground.tsx
@@ -5,7 +5,7 @@
  * that executes the call against a live server using the real SDK.
  */
 
-import { useState, useCallback, useMemo, type ReactNode } from 'react'
+import { useState, useCallback, useMemo, useEffect, useRef, type ReactNode } from 'react'
 import { Link } from 'react-router-dom'
 import { LayoutDashboard, LogOut } from 'lucide-react'
 import { Light as SyntaxHighlighter } from 'react-syntax-highlighter'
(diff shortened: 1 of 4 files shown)


apps/app/src/pages/Playground.tsx
```diff
@@ -188,7 +188,46 @@ export default function Playground() {
...
 
-    const [namespace, setNamespace] = useState('default')
+    const initialQueryNamespace = useMemo(() => {
+        if (typeof window === 'undefined') return null
+        const value = new URLSearchParams(window.location.search).get('namespace')?.trim()
+        return value || null
+    }, [])
+    const consumedQueryNamespace = useRef(false)
+    const [hydratedAccountId, setHydratedAccountId] = useState<string | null>(null)
+    const [namespace, setNamespace] = useState(() => {
```
HoangDucBach, line 198: **Cosmetic**: if `accountObjectId` isn't ready yet at first render and there's no
query param, this falls back to `default` before the hydration effect
corrects it to the saved value. A brief flash, not a data bug. Could init to
null instead and let the effect be the single source of truth.

HoangDucBach (approved):
Approved.

Traced the query-namespace-consume-once logic by hand across an account switch (A applies and persists `?namespace=foo`, then switching to B without reload correctly reads B's own saved namespace, not A's) - the ref guard is what makes this safe, not just component state. The signed GET correctly includes the query string in what gets signed.

**_One non-blocking note_**: when `accountObjectId` isn't available on first render
and there's no query param, namespace briefly initializes to `default` before
the hydration effect corrects it - cosmetic flash, not a data bug.

Issue #636, opened 2026-08-13
memwal_recall drops corrupt/inaccessible blobs silently — client sees fewer results with no indication of data loss


dexarxbt:
## Description

When the relayer encounters blobs that cannot be decrypted or embedded during recall, they are silently dropped. The client receives fewer results than exist in storage with no indication that some memories were lost. The `dropped_count` field computed by the server is stripped by the SDK and MCP output (covered in #622), making the problem invisible.

This compounds with `restore()` semantics where `restored + skipped` is a lower bound, not an exact count — healthy entries can also be silently lost.

## Steps to Reproduce

1. Store memories in a namespace
2. Rotate delegate key or corrupt one blob
3. Recall from that namespace
4. Receive N results where N < total stored, with no `dropped_count` or warning

## Expected Behavior

- Client should receive `dropped_count` (server already computes it — #622)
- SDK should expose a `recallWithDiagnostics()` or similar that returns both results and loss metadata
- MCP tool should include dropped_count in output
- Consider: `memwal.diagnostics(namespace)` to check storage health

## Actual Behavior

- Results silently truncated
- No way to know memories are missing without manually counting stored vs. returned
- #622 tracks the dropped_count being stripped; this issue tracks the broader silent-loss UX

## Related

- #622 — dropped_count computed by server is stripped by SDK/MCP
- #626 — no operation to list stored blobs (makes auditing impossible)

HoangDucBach:
[WALM-397](https://linear.app/mysten-labs/issue/WALM-397/memwal-recall-drops-corruptinaccessible-blobs-silently-client-sees)

ducnmm:
[WALM-407](https://linear.app/mysten-labs/issue/WALM-407/surface-recall-dropped-count-and-redact-loopback-urls-in-the-sdks)
Fix in https://github.com/MystenLabs/MemWal/pull/747

harrymove-ctrl:
Closing as resolved by #747 (already merged to `dev` and `main`). 

The core issue is fixed: `dropped_count` is now surfaced across both TS/Python SDKs and the MCP recall tool, clearly distinguishing decryption drops from truly empty namespaces. 

Remaining OpenClaw plugin adoption will be tracked separately.

Issue #622, opened 2026-08-13
recall: dropped_count computed by server is dropped by SDK type and MCP output, so an all-dropped recall is indistinguishable from an empty namespace


kiyeps:
## Summary

The relayer deliberately computes and returns `dropped_count` on `POST /api/recall`, which is the number of matched memories that were silently dropped due to blob download / SEAL decrypt / UTF-8 failures. But the field dies at the client layer: the TypeScript SDK's `RecallResult` type omits it, and the MCP `memwal_recall` tool never surfaces it. The server's own code comment states the field's purpose ("so clients can distinguish 'no matches' from 'matches we couldn't return'"), and no client can fulfill that purpose today.

## Reproduction (code path, commit 5010cd0, 2026-08-13)

1. Server side (works as designed):
   - `RecallResponse.dropped_count` is defined and serialized when non-zero (services/server/src/types.rs:1056-1063).
   - `routes/recall.rs` increments it via the storage engine and warns when > 0 ("recall: {} of {} matches dropped due to download/decrypt errors").
2. SDK (drops the signal):
   - `RecallResult` in packages/sdk/src/types.ts is `{ results: RecallMemory[]; total: number }`, with no `dropped_count` field. `MemWalManual` SDK users have no typed access to it even though the HTTP response carries it.
3. MCP tool (drops the signal again):
   - `memwal_recall` (services/server/scripts/mcp/tools/recall.ts:56-59) formats output as `N. [score=...] <text>` lines and returns `results.length`, ignoring `result.dropped_count` entirely. On a recall where every match failed decryption, the agent sees "No matching memories found.", which is indistinguishable from an empty namespace.

## Deterministic PoC

A two-part PoC proves the field dies at the type/format layers without needing a live deployment:

1. **Part A (wire shape):** a Rust program using the exact serde `RecallResponse` types copied from `services/server/src/types.rs` serializes a recall where both matches failed decryption. Output wire JSON includes `"dropped_count": 2`, confirming the server contract carries the signal.
2. **Part B (MCP formatter):** a Node script feeds that wire JSON through the `memwal_recall` output formatter. Results:
   - `CASE2_ALL_DROPPED -> "No matching memories found."`
   - `CASE3_EMPTY       -> "No matching memories found."`

[...]

PR #747, opened 2026-08-23
fix(sdk): surface dropped_count and redact loopback URLs

ducnmm:
Linear: [WALM-407](https://linear.app/mysten-labs/issue/WALM-407/surface-recall-dropped-count-and-redact-loopback-urls-in-the-sdks)
Fixes https://github.com/MystenLabs/MemWal/issues/622
Fixes https://github.com/MystenLabs/MemWal/issues/636

Stacked on #737 so `@mysten-incubation/memwal` stays at `0.1.5` and Python `memwal` stays at `0.1.8`.

## What this does

- `recall()` results include `dropped_count` when the relayer omitted matches that failed to download or decrypt.
- `health()` surfaces optional `write_ready`.
- HTTP error messages and remember-job error strings redact `localhost` / `127.0.0.1` URLs.

### DIFF
--- a/packages/python-sdk-memwal/CHANGELOG.md
+++ b/packages/python-sdk-memwal/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- `recall()` results include `dropped_count` when the relayer omitted matches that failed to download or decrypt.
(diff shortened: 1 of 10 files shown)


HoangDucBach (approved):
Approved.

PR #751, opened 2026-08-24
Promote dev to staging

ducnmm:
## Summary

Promote the latest `dev` changes to `staging` (since [#712](https://github.com/MystenLabs/MemWal/pull/712)).

Includes:

- **#554 / WALM-295/296/297** — owner-scoped memory read API (`GET /v1/owners/{owner}/{namespaces,memories,agents}`) plus short-lived bearer token auth for Console.
- **#701 / WALM-361** — MCP credentials resolve per project (local `.memwal/credentials.json` over the global file); stop `login` deleting credentials before the new ones exist.
- **#702** — empty-body 401s point at `memwal_login` instead of `<no message>`.
- **#720 / WALM-359** — storage quota admission is atomic with insertion.
- **#721** — SDK accepts a bech32 `suiprivkey` delegate key, not just hex.
- **#719 / WALM-326** — dashboard “no delegate key” notice is scoped to this browser.
- **#684** — Noter gRPC migration, auth race fixes, Playwright e2e suite + CI.
- **#679** — JS SDK live e2e suite and CI against the relayer.
- **#641** — sidecar crash-recovery reconcile looked up the wrong on-chain tag.
- **#733 / GH #619** — persist the OAuth client loopback redirect URI.
- **#732 / WALM-401** — GitHub bug and feature issue forms (blank issues disabled).

`staging` is 10 commits ahead of `dev`; those are previous promote merge commits only.

## Test plan

- [x] Listed PRs merged to `dev`
- [ ] After merge: staging smoke — `memwal_login`, remember, recall
- [ ] Console owner-scoped read API against staging (bearer token + Ed25519 fallback)
- [ ] MCP project-local credentials: local `.memwal/` vs global `~/.memwal/`
- [ ] Empty 401 from a signed-out `memwal_*` tool shows the login hint

PR #756, opened 2026-08-24
fix(bench): retry Cloudflare HTML in the live latency check

ducnmm:
## Summary

Unblocks the style-guide comment and the Memory API Latency failure on #751 (promote `dev` → `staging`).

**Style guide (#751 bot audit).** `docs/api/memory-read-api.md`, `docs/api/owner-token-auth.md`, and `docs/sdk/changelog.mdx`: `on-chain` → `onchain`, `via` → `through`, italic `*only*` → `**only**`, passive voice on the owner-token path, unstacked `## 0.1.4` heading.

**Latency check.** The job failed because 5/10 warm `/api/recall` calls got a Cloudflare HTML error page (`<!DOCTYPE html> … ie6 oldie`) in ~365ms, then later runs succeeded. The bench treated that as a failed recall and exited 1. The 500ms warm-p50 line in the log is informational only; it does not fail the job, and this PR does not try to make live Walrus+SEAL recall hit 500ms.

The bench now retries Cloudflare HTML, 429, and 5xx (up to 4 attempts, 400ms backoff). JSON 401/403/400 stay fatal.

## Test plan

- [x] Style-guide nits match the #751 audit list
- [x] `cd services/server/scripts && npm test` (includes new `live-bench-retry` cases)
- [ ] After merge to `dev`, #751's Memory API Latency re-run should no longer fail on a one-off Cloudflare HTML burst

### DIFF
--- a/docs/api/memory-read-api.md
+++ b/docs/api/memory-read-api.md
@@ -110,9 +110,9 @@ sends the signature in `x-signature`:
 Server-side verification flow (`auth.rs::verify_signature`): validate the
 Ed25519 signature against the canonical string → check the nonce hasn't
 been seen before (Redis, fail-closed on Redis outage) → resolve the account
-(cache, then the `x-account-id` hint, then a bounded on-chain registry scan
+(cache, then the `x-account-id` hint, then a bounded onchain registry scan
 as a last resort) → verify the public key is a registered delegate key on
-that account's on-chain `MemWalAccount.delegate_keys` (cached in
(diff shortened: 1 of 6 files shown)


HoangDucBach (approved):
Approved. Verified the root cause matches what I saw in #751's failing log
(HTML on recall-warm runs 4-9), and the retry classification is scoped
correctly - JSON 401/403/400 stay fatal per the new test, only HTML/edge
statuses retry. CI 15/15.

PR #790, opened 2026-08-25
fix(recall): reject whitespace-only queries (WALM-300)

### LINKED ISSUE #425: [SDK/Validation] Whitespace only query bypasses input validation and returns embedding based noise
**Summary**
memwal_recall (SDK + relayer) rejects query: "" with a proper 400 ("String must contain at least 1 character(s)"), but accepts query: " " (whitespace-only) and returns embedding-based results as if the query were meaningful. Inconsistent validation — and the failure is invisible to the caller.

This is a small standalone bug, but it fits a broader pattern of silent-failure gaps at the SDK/relayer boundary I hit during a two-week build:

[#309](https://github.com/MystenLabs/MemWal/issues/309) — recall param name mismatch (limit vs topK), the wrong-name param is silently ignored
[#400](https://github.com/MystenLabs/MemWal/issues/400) — namespace whitespace drift silently returns empty results
All three (this one + #309 + #400) share the same fix surface: tighter Zod validation at the SDK/relayer boundary. Filing this one separately since it isn't covered by either existing issue.

Reproduction environment: @mysten-incubation/memwal 0.0.7, @mysten-incubation/memwal-mcp 0.0.7, relayer https://relayer.memory.walrus.xyz, tested 2026-07-13 against a namespace with 26 real memories.

**Reproduction**
```
const mw = MemWal.create({ key, accountId, serverUrl: "https://relayer.memory.walrus.xyz", namespace: "debugmem" });

await mw.recall({ query: "",     limit: 3 }); // → 400 error: "String must contain at least 1 character(s)"
await mw.recall({ query: "   ",  limit: 3 }); // → 3 results with valid scores (~0.19–0.22), all unrelated to whitespace
await mw.recall({ query: "\t\n ", limit: 3 }); // → same: accepted, results returned
```
Same behavior when called via memwal-mcp memwal_recall tool.

**Why it matters**
Agent workflows often construct queries from user input or upstream context. If that string ends up as whitespace-only (or was meant to be trimmed but wasn't), the caller gets back real-looking results instead of an error, leading the agent to act on irrelevant memories.
The empty-string check already in place is doing the right thing — extending it to trim().length === 0 closes the gap at zero cost.
Because the caller sees results, they have no signal that anything is wrong. Same silent-failure shape as #309 and #400.

**Suggested fix**
Prefer option 1:
1. In the Zod schema for recall() query, reject queries whose .trim() is empty with the same 400 error used for empty strings:

`z.string().min(1).refine(s => s.trim().length > 0, "Query cannot be whitespace-only")`

2. Alternatively, apply .trim() to the query before embedding — but that hides the caller's bug instead of surfacing it, so option 1 is preferred.
One-line change, closes an invisible-failure class for agent authors. Happy to PR if the direction is confirmed.

Additional friction noted during build (not filing separately)
restore() returned HTTP 500 intermittently on the mainnet relayer — transient, resolves on ~5s retry.
remember() occasionally times out but still succeeds in the background; a naive retry then creates a duplicate on-chain blob.
MEMWAL_AGENT_ID labeling in the dashboard — covered by [#364](https://github.com/MystenLabs/MemWal/issues/364).

harrymove-ctrl:
Fixes WALM-300 / closes #425.

## Problem

`body.query.is_empty()` is a byte-length check, so `"   ".is_empty()` is `false`. A whitespace-only query passed validation, reached `generate_recall_embedding_cached` → the embeddings API, and returned a real but semantically meaningless vector. Callers got a plausible-looking 200 instead of the 400 they'd get from `query: ""`.

Severity is low: results stay scoped to `owner + namespace`, so there is no cross-tenant exposure. The cost is a wasted embedding call and a misleading success response.

## Fix

| File | Change |
|---|---|
| `services/server/src/routes/recall.rs:157` | `is_empty()` → `trim().is_empty()` |
| `services/server/scripts/mcp/tools/recall.ts` | added Zod `.refine((v) => v.trim().length > 0, ...)` |
| `packages/sdk/src/manual.ts:389` | `if (!query)` → `if (!query.trim())` |

The server gate is the one that matters: every client path (SDK, MCP tool, raw HTTP) funnels through it, so that single line closes the gap everywhere. The other two are fail-fast client-side mirrors. `recallManual()` had the same truthiness pattern and is fixed alongside.

## Verification

`cargo check --all-targets` passes in `services/server` (exit 0; remaining warnings are pre-existing and unrelated).

The two TypeScript edits are one-liners with no new imports, but they were **not** typechecked locally: `node_modules` is not installed in this worktree, so `npx tsc` fails with `MODULE_NOT_FOUND`. CI covers them.

### DIFF
--- a/packages/sdk/src/manual.ts
+++ b/packages/sdk/src/manual.ts
@@ -386,7 +386,7 @@ export class MemWalManual {
         limitOrOptions: number | MemWalManualRecallOptions = 10,
         namespace?: string
     ): Promise<RecallManualResult> {
-        if (!query) throw new Error("Query cannot be empty");
+        if (!query.trim()) throw new Error("Query cannot be empty");
 
         const options = typeof limitOrOptions === "number" ? { limit: limitOrOptions, namespace } : limitOrOptions;
         const limit = options.limit ?? 10;
--- a/services/server/scripts/mcp/tools/recall.ts
+++ b/services/server/scripts/mcp/tools/recall.ts
@@ -8,6 +8,7 @@ const RECALL_INPUT = {
     query: z
         .string()
         .min(1)
+        .refine((v) => v.trim().length > 0, "Query cannot be empty.")
         .describe("Natural-language search query to match against stored memories."),
     limit: z
         .number()
--- a/services/server/src/routes/recall.rs
+++ b/services/server/src/routes/recall.rs
@@ -154,7 +154,7 @@ pub async fn recall(
     Extension(auth): Extension<AuthInfo>,
     Json(body): Json<RecallRequest>,
 ) -> Result<Json<RecallResponse>, AppError> {
-    if body.query.is_empty() {
+    if body.query.trim().is_empty() {
         return Err(AppError::BadRequest("Query cannot be empty".into()));
     }
 


HoangDucBach (approved):
LGTM!

PR #793, opened 2026-08-25
fix(mcp,server): stop losing the delegate key when a login is interrupted [WALM-332]

nikola0x0:
Closes WALM-332.

## The bug

Signing in creates a delegate keypair, the browser registers the **public** half on-chain — paid, irreversible — and only *afterwards* POSTs a callback that makes us save the **private** half.

Until now that private half existed only in memory, and the callback listener lived in the same process. Any death in that window destroyed the only copy of a key that had already been paid for and committed on-chain, leaving an orphaned registration nobody can use.

Nothing reported it. The browser's POST hit a closed port; the process was gone so it logged nothing; and `handleLocalLogin` had already returned the URL and told the client the call succeeded.

## This is not the bug the ticket described

WALM-332 originally attributed this to a **state-nonce mismatch**. That path is unreachable, and fixing it would have accomplished nothing.

The callback handler gates `preflightVerified` → payload shape → *then* state (`login.ts`). Reaching `login.callback_bad_state` therefore requires the *same live process* to have already accepted a preflight carrying its own nonce. A stale tab dies at `/preflight` first.

Measured with the real `memwal-mcp` binary over stdio against a mock relayer:

| Step | Observed |
| --- | --- |
| Preflight while process alive | `200 {"ok":true}` |
| Process killed, dashboard POSTs `/callback` | **`ECONNREFUSED`** |
| Stale tab preflights at respawned process | `403 preflight mismatch` |
| Stale tab POSTs callback at respawned process | `428 verification required` |

Ports are never reused across restarts (51858 → 51862; 51820 → 51825). There are zero `callback_bad_state` or `preflight_mismatch` events across any local `mcp-server-memwal.log*`. The ticket has been rewritten with this evidence; the original report is preserved there verbatim.

## The fix: write-ahead

Persist the keypair **before** anything can hand the public key to a browser, and clear it once the key is safely in `credentials.json`. On the next start a stranded record is reclaimed automatically.

Reclaiming needs account metadata the lost callback would have carried, and nothing exposed it — `find_account_by_delegate_key` is internal to the auth middleware, `account_exists` takes an owner address and returns only `{exists}`, `StatsResponse` returns `owner`. Hence the new **authenticated** `GET /api/whoami`, which returns what the middleware already resolved while authenticating.

Returning `account_id` there is safe *because* the route is authenticated: the caller proved possession of a key registered against that account, so it only ever learns about itself. The public existence-check route still withholds it.

## Two deliberate constraints

**A rejected key is never deleted.** A 401 means "not registered" on mainnet — but on testnet the registry scan is disabled outright and a *genuinely registered* key is rejected for want of an `x-account-id` hint (`services/server/src/auth.rs`). Deleting on 401 would destroy a paid key in exactly that environment. The record waits out its 24h TTL instead.

**Recovery never rolls back a newer sign-in.** If the user gave up on a stuck login and signed in again, adopting the older stranded key would silently downgrade them. Covered by a test.

## Overlap with WALM-333, resolved

This branch also fixed the swallowed failure in `startOrReuseLoginFlow` — a login failing while the process is still alive (timeout, listener error) was eaten into a `warn` and never reached the client. **WALM-333 (#725) landed that fix on `dev` first**, independently, with the same `onFailure` shape. Credit for it belongs there, not here.

Merging produced a working-but-wrong result: `onFailure` invoked from two catches, firing twice per failure. Dev's own `onFailure runs once when a concurrent call joins the in-flight flow` catches it.

Resolved to one report per failure, with dev's version as the base:

- **`login.ts`** — dev's `(err: unknown)` signature and its single call site, in the catch that rejects the URL promise. That placement also covers a caller passing only `onFailure`, which the `onSuccess` chain does not. This branch's `log.error` + stderr `note` moved there with it, and an `onSuccess` that throws is now its own event instead of being folded into `login.inflight.failed`.
- **`bridge.ts`** — dev's handler, whose level and wording its tests pin. Kept one clause from this branch: *"if a delegate key was already registered on-chain, the next start reclaims it"* is new information, and it is only true because of the write-ahead record.

Two things the merge surfaced that were not textual conflicts:

- Dev added a guard test rejecting any test that sandboxes `HOME` without `USERPROFILE`. Both new test files here tripped it. Switched them to `MEMWAL_CREDS_DIR`, which dev added for exactly this — and which also closes the project-local creds walk they were still exposed to (`credsPath()` looks for a `.memwal` above the working directory, i.e. inside the real checkout, before it ever consults `HOME`).
- **Docs / Freshness** fails on a registered route no docs page mentions, so `GET /api/whoami` is now documented in `docs/relayer/api-reference.md`. CHANGELOG entry added alongside it.

## Testing

`packages/mcp` **69/69 pass**, 12 of them new here. `services/server` compiles clean and `cargo test` gives **645 passed, 1 failed** against the project's own `docker-compose.yml`, no overrides. That one failure — `routes::memory_read::query_owner_namespaces_surfaces_namespace_created_mid_walk_on_next_walk` — is in a file neither side of the merge touches, and reproduces identically on a clean `origin/dev` checkout. Server CI is green.

The signature format is duplicated across Rust and TypeScript because they can't share code. Silent drift there would fail only in production, as an opaque 401 — so the exact canonical string is pinned by a test on **each** side, with a comment pointing at the other. Change one and that one fails; change the format in `auth.rs` and both fail.

New coverage: key on disk before the URL is published · cleared on success · reclaim rebuilds credentials · never rolls back a newer sign-in · ambiguous rejection never deletes · unreachable relayer retries later · expired records discarded · `whoami` field mapping cannot transpose `account_id`/`owner`.

## Deferred, not forgotten

The better long-term fix is deleting the loopback listener entirely for a device-code/polling flow (RFC 8628-shaped) — nothing to bind, nothing to die, and it would work over SSH and in containers where the current design cannot work at all. It needs relayer work and overlaps WALM-382, so it is written up in WALM-332 as a follow-up. **This change stays correct under that rewrite.**

## Reviewer notes

- Cross-package: TypeScript (`packages/mcp`) + Rust (`services/server`). Two commits split by package, then a merge from `dev` — `--first-parent` gives the original two; the merge commit on its own is the conflict resolution described above.
- `/api/whoami` **must** stay inside the authenticated router group.
- Recovery is a no-op with no network call when nothing is pending — the common case.


jessiemongeon1:
## Style Guide Audit

**3** violation(s) found. All must be fixed before merge.

### `docs/mcp/changelog.mdx` (3 violation(s))

_3 violation(s) (1 regex, 2 claude)_

- **Line 40** — "on-chain" is one word → "onchain"
  - Current: `on-chain`
  - Fix: `onchain`
- **Line 40** — passive_voice
  - Current: `before the sign-in URL is handed to the browser`
  - Fix: `before sign-in hands the URL to the browser`
- **Line 40** — passive_voice
  - Current: `The browser's on-chain `add_delegate_key` is paid and irreversible, and it happens before the callback that saved the private half`
  - Fix: `The browser's on-chain `add_delegate_key` costs gas and is irreversible, and it happens before the callback that saves the private half`

---

packages/mcp/src/recovery.ts
```diff
@@ -0,0 +1,252 @@
...
+/** sha256 of an empty body. A GET sends none; the server hashes it anyway. */
+export const EMPTY_BODY_SHA256 = createHash("sha256").update("").digest("hex");
+
+async function whoami(
+    relayerUrl: string,
+    privateKeyHex: string,
+    publicKeyHex: string,
+): Promise<{ status: number; body: unknown } | null> {
+    const path = "/api/whoami";
+    const timestamp = String(Date.now());
```
ducnmm: **[bug]** Recovery authenticates with `const timestamp = String(Date.now())` (milliseconds, ~1.7e12). Relayer auth documents `x-timestamp` as Unix seconds and accepts only `|age| <= drift` (default 300s). A millisecond timestamp is ~1.7e12 seconds in the future, so `whoami` always 401s with `ERR_TIMESTAMP_OUT_OF_BOUNDS` and `recoverPendingLogin` reports `rejected`. Signing tests pin `"1700000000"` (seconds) and the HTTP mock never checks the header.

**Suggestion:** Use `String(Math.floor(Date.now() / 1000))`. Extend the recovery HTTP test to assert `x-timestamp` is 10-digit seconds.

packages/mcp/src/recovery.ts
```diff
@@ -0,0 +1,252 @@
...
+    );
+
+    if (res === null) {
+        log.warn("login.pending.relayer_unreachable", {
+            publicKey: pending.delegatePublicKeyHex,
+        });
+        return { outcome: "unavailable", strandedPublicKey: pending.delegatePublicKeyHex };
+    }
+
+    if (res.status !== 200 || !isWhoami(res.body)) {
```
ducnmm: **[bug]** Any non-200 `whoami` response, including 503 `AUTH_UPSTREAM_UNAVAILABLE`, is classified as `rejected`. Auth unavailability is not a revoke (WALM-429). The stranded notice for `rejected` tells the user to run `memwal_login` again and revoke the key — wrong for a transient outage, and also what they see today because of the timestamp 401.

**Suggestion:** Treat 503 / `x-auth-error: AUTH_UPSTREAM_UNAVAILABLE` (and transport failure) as `unavailable` so the record is kept. Only use the "sign in again, then revoke" copy when identity is actually refused (401 without timestamp/unavailable reasons).

Also: on testnet, auth fails closed without `x-account-id` and never runs the registry scan, so reclaim cannot return 200 on `--dev`/`--staging`. Document that automatic reclaim is mainnet-only, or give `/api/whoami` a pubkey path that does not require the hint.

packages/mcp/src/login.ts
```diff
@@ -211,6 +217,21 @@ export async function loginFlow(opts: LoginOptions = {}): Promise<MemWalCredenti
         timeoutMs: opts.timeoutMs ?? resolveLoginTimeoutMs(),
     };
     const keypair = await generateKeypair();
```
ducnmm: **[bug]** Every `loginFlow` call generates a new keypair and `savePendingLogin` overwrites `login-pending.json`. Recovery only runs at process start and is skipped on `--login`. The failure copy tells the agent to `call memwal_login again to retry`. Following that in the same session (timeout after the user already approved `add_delegate_key`) replaces the stranded private key with an unrelated one — the original WALM-332 failure without a crash.

**Suggestion:** If an unexpired pending record exists for the same relayer, reuse that keypair instead of `generateKeypair()`. At minimum, run recovery before `--login` so a stranded key is saved or backed up first.

packages/mcp/src/index.ts
```diff
@@ -157,6 +158,25 @@ export async function main(argv: string[] = process.argv.slice(2)): Promise<void
     // sign-in belongs to a different account (GH #628). The old file is now
     // replaced only on success, and backed up when the account changes.
     let creds = args.forceLogin ? null : loadCreds();
+    // A previous sign-in may have died after the browser registered our
+    // delegate key on-chain but before the callback saved it (WALM-332). The
+    // key was write-ahead-persisted, so try to reclaim it rather than making
+    // the user register — and pay for — a replacement. Cheap no-op when there
+    // is no pending record, which is the overwhelmingly common case: it hits
+    // the network only when there is genuinely something stranded.
+    if (!args.forceLogin) {
```
ducnmm: **[bug]** `recoverPendingLogin` now runs on every non-`--login` start and will rebuild `credentials.json` from `login-pending.json`. `--logout` and `handleLocalLogout` still only call `clearCreds()`. Signing out therefore leaves the write-ahead secret on disk, and the next spawn logs the user back in.

**Suggestion:** Call `clearPendingLogin()` from the CLI `--logout` path and from `handleLocalLogout`. Do not fold it into `clearCreds()` itself — a 401/wipe of current credentials should still be able to reclaim a newer pending key.

packages/mcp/src/auth.ts
```diff
@@ -286,3 +286,119 @@ function isValid(obj: unknown): obj is MemWalCredentials {
...
+/** Sits beside whichever credentials file `credsPath()` resolves to, so a
+ * project-local sign-in recovers into that same project. */
+export function pendingLoginPath(): string {
+    return join(dirname(credsPath()), PENDING_FILE);
+}
+
+/** Persist the pending keypair. Best-effort: a login that cannot write its
+ * write-ahead record is still better than no login at all, so this never
+ * throws — it degrades to today's behaviour. */
+export function savePendingLogin(pending: PendingLogin): void {
```
ducnmm: **[suggestion]** `savePendingLogin` swallows every write error and does not log. A permissions or disk failure degrades to the pre-fix in-memory-only key with no signal, and the browser is still given the connect URL.

**Suggestion:** Log a warning in the catch (no private key material). Optionally refuse to publish the connect URL if the pending record could not be written.

packages/mcp/test/login-recovery.test.mjs
```diff
@@ -0,0 +1,214 @@
...
+import assert from "node:assert/strict";
+import http from "node:http";
+import { existsSync, mkdtempSync, mkdirSync, writeFileSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+const ACCOUNT = `0x${"a".repeat(64)}`;
+const OWNER = `0x${"b".repeat(64)}`;
+const PACKAGE = `0x${"c".repeat(64)}`;
+
```
ducnmm: **[suggestion]** `freshHome()` assigns `HOME`, `USERPROFILE`, and `MEMWAL_CREDS_DIR` and never restores them. `node --test` can run files in one process; a leftover `MEMWAL_CREDS_DIR` makes `credsPath()` ignore HOME.

**Suggestion:** Capture previous env values and restore them in `t.after`, matching `login-preflight.test.mjs`.

packages/mcp/src/recovery.ts
```diff
@@ -0,0 +1,252 @@
...
+    const w = o as Record<string, unknown>;
+    return (
+        typeof w.account_id === "string" &&
+        /^0x[0-9a-fA-F]{64}$/.test(w.account_id) &&
+        typeof w.owner === "string" &&
+        typeof w.package_id === "string"
+    );
+}
+
+/**
```
ducnmm: **[nit]** Two consecutive JSDoc blocks sit on `canonicalRequestMessage`. The first ("Ask the relayer who this delegate key belongs to") is leftover from `whoami`.

**Suggestion:** Delete the orphaned block and keep the canonical-message comment.

packages/mcp/test/login-write-ahead.test.mjs
```diff
@@ -0,0 +1,133 @@
...
+        pending.delegatePrivateKey ?? "",
+        /^(0x)?[0-9a-f]{64}$/i,
+        "the private key must be recoverable from the record",
+    );
+    assert.equal(pending.relayerUrl, RELAYER);
+    assert.ok(pending.createdAt, "record needs a timestamp so it can expire");
+
+    // Same handling as credentials.json — owner-only.
+    assert.equal(
+        statSync(pendingPath(home)).mode & 0o777,
```
ducnmm: **[nit]** The test asserts `statSync(...).mode & 0o777 === 0o600`. `savePendingLogin` already treats `chmod` as best-effort on Windows, so this assertion fails there.

**Suggestion:** Skip the mode check on `win32`, or assert only that the file exists and contains the key.

packages/mcp/src/recovery.ts
ducnmm: **[bug]** Recovery builds `x-timestamp` with `String(Date.now())` (Unix milliseconds). Relayer auth documents `x-timestamp` as Unix seconds and freshness-checks it against `chrono::Utc::now().timestamp()` within typically ±300s. A ms value is ~10^12 and always lands outside that window (`ERR_TIMESTAMP_OUT_OF_BOUNDS` 401), so `GET /api/whoami` never succeeds.

**Suggestion:** Use `Math.floor(Date.now() / 1000).toString()` and add a test that the live `whoami` timestamp is 10-digit seconds.

packages/mcp/src/recovery.ts
ducnmm, line 194: **[bug]** Any non-200 or unparseable body is `outcome: "rejected"`. A 503 with `x-auth-error: AUTH_UPSTREAM_UNAVAILABLE` is lumped in with 401. `formatStrandedLoginNotice` then tells the user to run `memwal_login` and revoke the key — the wrong action for a transient upstream failure, and the action that destroys the stranded key if login overwrites pending.

**Suggestion:** Treat 5xx / 429 and `x-auth-error: AUTH_UPSTREAM_UNAVAILABLE` as `unavailable`. Reserve `rejected` for identity failures (401/403).

packages/mcp/src/login.ts
ducnmm: **[bug]** Every `loginFlow` call `generateKeypair()`s and `savePendingLogin()` overwrites `login-pending.json`. Recovery only runs at process start and is skipped for `--login` / `forceLogin`. A timeout then `memwal_login` in the same process replaces the write-ahead record with a new key. If the browser already paid for `add_delegate_key` on the old public key, that private half is gone.

**Suggestion:** If `loadPendingLogin()` returns a still-valid record for the same `relayerUrl`, reuse that keypair instead of minting and overwriting.

packages/mcp/src/login.ts
```diff
@@ -431,6 +452,9 @@ export async function loginFlow(opts: LoginOptions = {}): Promise<MemWalCredenti
                     version: 1,
                 };
                 const saved = saveCreds(creds);
+                // The key is durable in credentials.json now, so the
+                // write-ahead record has done its job.
+                clearPendingLogin();
```
ducnmm, line 483: **[bug]** `clearPendingLogin()` runs after a successful callback, but CLI `--logout` (`clearCreds()` only) and `handleLocalLogout` never clear `login-pending.json`. After an interrupted re-login, logout wipes `credentials.json` and leaves the pending key; the next start hits `recoverPendingLogin()` and signs the user back in.

**Suggestion:** Call `clearPendingLogin()` from both logout paths. Do not fold it into `clearCreds()`, which also runs on 401 session teardown and should not discard a newer stranded key.

packages/mcp/src/recovery.ts
```diff
@@ -0,0 +1,252 @@
...
+    });
+    const signature = await signMessage(privateKeyHex, message);
+
+    const controller = new AbortController();
+    const timer = setTimeout(() => controller.abort(), WHOAMI_TIMEOUT_MS);
+    timer.unref?.();
+    try {
+        const resp = await fetch(`${relayerUrl.replace(/\/+$/, "")}${path}`, {
+            method: "GET",
+            headers: {
```
ducnmm, line 130: **[suggestion]** Recovery omits `x-account-id` because the client does not know the account. On testnet, auth refuses delegate-key auth without that hint and never scans the registry, so whoami cannot succeed even with a correct timestamp.

**Suggestion:** Exempt `GET /api/whoami` from the testnet `x-account-id` requirement, or document mainnet-only recovery.

packages/mcp/src/auth.ts
```diff
@@ -286,3 +286,119 @@ function isValid(obj: unknown): obj is MemWalCredentials {
...
+        writeFileSync(path, JSON.stringify(pending, null, 2), {
+            encoding: "utf8",
+            mode: 0o600,
+        });
+        try {
+            chmodSync(path, 0o600);
+        } catch {
+            /* Windows etc. — best effort */
+        }
+    } catch {
```
ducnmm, line 368: **[suggestion]** `savePendingLogin` swallows all write errors. The PR’s invariant is that the private key is on disk before the connect URL is handed to the browser; a silent failure restores the original WALM-332 loss with no log line.

**Suggestion:** Log the failure. Prefer failing the login (or refusing to publish the URL) over continuing as if the key were durable.

packages/mcp/src/recovery.ts
ducnmm: **[nit]** Orphaned JSDoc for `whoami()` sits immediately above the `canonicalRequestMessage` JSDoc after the helper was extracted.

**Suggestion:** Attach that block to `whoami`, or drop it.

packages/mcp/test/login-write-ahead.test.mjs
ducnmm: **[nit]** `statSync(...).mode & 0o777 === 0o600` is not meaningful on Windows. `savePendingLogin` already treats `chmodSync` as best-effort there.

**Suggestion:** Assert the mode only when `process.platform !== "win32"`.

ducnmm (dismissed):
## Summary

The `origin/dev` merge did not change the WALM-332 recovery path. Write-ahead of `login-pending.json` is in place, but reclaiming a stranded key still cannot work: `whoami` signs a millisecond timestamp the relayer will reject, non-200 responses (including 503 `AUTH_UPSTREAM_UNAVAILABLE`) are treated as a hard rejection, and the next `loginFlow` overwrites the only copy of the paid key. Logout still leaves the pending record, so the following start can sign the user back in.

## Issue counts by severity

- bugs: 4
- suggestions: 2
- nits: 2

packages/mcp/src/recovery.ts
nikola0x0: Fixed in f5b4ccd. `Math.floor(Date.now() / 1000).toString()`. Added a test that captures the header the relayer actually receives and asserts it is 10-digit seconds and within 300s of now.

packages/mcp/src/recovery.ts
nikola0x0: Fixed. `rejected` is now only 401/403 without `x-auth-error: AUTH_UPSTREAM_UNAVAILABLE`. Everything else is `unavailable` and retried: 429 and 5xx, that header on any status, a 404 from a relayer too old to serve the route, and a 200 whose body is not a whoami (which means we are not talking to the endpoint we think we are).

Tests cover each case, and assert the `unavailable` notice never says "revoke" while the `rejected` one still does.

packages/mcp/src/login.ts
nikola0x0: Fixed. `reusablePendingLogin(relayerUrl)` adopts a still-valid record instead of minting over it, and the record is written back unchanged, `createdAt` included, so the TTL keeps measuring from the attempt that may have registered the key and the `superseded` comparison in recovery stays meaningful.

A different relayer still mints fresh, since a key registered against one relayer's account proves nothing to another. That case logs a warning now rather than being silent.

packages/mcp/src/login.ts
nikola0x0: Fixed in both paths, and kept out of `clearCreds` for exactly the reason you gave. Tests pin both directions: an explicit logout removes the record, and `clearCreds` on its own leaves it.

packages/mcp/src/recovery.ts
nikola0x0: I don't think the exemption can work, so I took the second option instead.

The testnet gate is not policy. The registry scan behind it runs over Sui JSON-RPC, and `auth.rs` Strategy 3 fails closed on testnet *because* testnet no longer serves that endpoint:

> The legacy registry scan uses JSON-RPC. Testnet no longer serves JSON-RPC, so fail closed when a modern signed `x-account-id` hint is absent instead of silently contacting a retired endpoint.

Exempting `whoami` would send it to the retired endpoint and fail anyway, just less legibly than the current 401. And recovery is the one caller that has no account id to send, so there is no hint path either: the pending record is written before the browser acts, so it cannot know the account id.

Documented as mainnet-only in `docs/relayer/api-reference.md`. Testnet recovery would need the scan ported to gRPC, which seemed well outside this PR. Happy to file that if you think it is worth having.

packages/mcp/src/auth.ts
nikola0x0: It now logs and throws, so the login fails rather than publishing a URL it cannot back.

I went past "log it" deliberately, and it is worth a second opinion. This file sits beside `credentials.json`, so a directory that cannot take it cannot take the credentials either. That login was going to fail at the callback regardless, one paid `add_delegate_key` later, so failing early costs the user nothing and failing late costs them gas. Test covers it with a read-only credentials directory.

Say the word if you would rather it stay best-effort with just the log line.

packages/mcp/src/recovery.ts
nikola0x0: Fixed, reattached to `whoami`.

packages/mcp/test/login-write-ahead.test.mjs
nikola0x0: Fixed, guarded on `process.platform !== "win32"`.

packages/mcp/src/bridge.ts
```diff
@@ -660,6 +671,14 @@ async function handleLocalLogin(
 function handleLocalLogout(): { text: string; isError: boolean } {
     try {
         const cleared = clearCreds();
+        // Explicit sign-out discards the write-ahead record too. Without this
+        // an interrupted re-login leaves `login-pending.json` behind, and the
+        // next start's `recoverPendingLogin` signs the user straight back in.
+        //
+        // NOT folded into `clearCreds()`: that also runs on 401 session
```
ducnmm: **[suggestion]** The new comment says pending is kept out of `clearCreds()` because that helper also runs on 401 session teardown. In this tree a 401 does not wipe credentials. Shorten to logout-only, or drop the 401 claim.

packages/mcp/src/bridge.ts
```diff
@@ -606,7 +611,13 @@ async function handleLocalLogin(
                 params: {
                     level: "warning",
                     logger: "memwal-mcp",
-                    data: `Walrus Memory sign-in did not complete: ${msg}. Existing credentials are unchanged; call memwal_login again to retry.`,
+                    // The last clause is only true because of the write-ahead
+                    // record (WALM-332): a key the browser already paid to
+                    // register is no longer lost with the process.
+                    data:
+                        `Walrus Memory sign-in did not complete: ${msg}. Existing credentials are ` +
```
ducnmm, line 680: **[suggestion]** Timeout copy still says to remove unused keys from the dashboard, then the suffix says the next start will reclaim the key. Revoking first destroys the registration write-ahead is meant to recover. Point a failed flow at retry / next-start reclaim, not revoke-then-relogin.

ducnmm (commented):
## Summary

The four recovery-breaking bugs from the earlier review are fixed: `whoami` signs a Unix-seconds timestamp, pending-key reuse, logout clearing, fail-closed write-ahead. Remaining notes are comment/wording, not another broken reclaim path.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

packages/mcp/src/bridge.ts
nikola0x0: You're right, and I should have checked rather than carrying that reason over from the earlier review. `bridge.ts` explicitly refuses to wipe credentials on a relayer 401 ("DO NOT wipe creds here", the creds-wipe DoS note), and the two logout paths are `clearCreds`'s only callers in this tree.

Reworded to the reason that does hold: discarding a key that may still be reclaimable is a decision only a deliberate sign-out gets to make, and `clearCreds` is exported so its callers are not fixed. Renamed the test that repeated the same false claim.

packages/mcp/src/bridge.ts
nikola0x0: Fixed. The timeout message now points at the two paths that actually work: run login again and the same key is reused, or restart and it is reclaimed. Revoking is mentioned only as the deliberate abandon-it option.

The same defect was in `formatStrandedLoginNotice`'s `rejected` branch in the other order — sign in again, *then* revoke — which keypair reuse makes self-defeating, since that sign-in adopts the very key it tells you to revoke. Reworded that too. Left `superseded` alone: that record is cleared and the key really is dangling, so revoke is the right advice there.

packages/mcp/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ### Fixed
 
+- Persist the delegate keypair before the sign-in URL is handed to the browser, and reclaim it on the next start. The browser's on-chain `add_delegate_key` is paid and irreversible, and it happens before the callback that saved the private half, so a client that died in that window destroyed the only copy of a key the user had already paid for and left an orphaned registration nobody could use. (#793) Signing out discards the pending record along with the credentials, a second sign-in against the same relayer reuses the stranded key rather than minting over it, and a sign-in that cannot write the record fails instead of publishing a URL it cannot back. Reclaiming works on mainnet; testnet requires an account-id hint the recovering client does not have.
```
ducnmm: **[bug]** origin/dev == origin/main at MCP 0.0.12, so this change must bump 0.0.12 → 0.0.13 in `packages/mcp/package.json`, `scripts/verify-manual-sdk-release.mjs`, and the six plugin/marketplace JSON files the verify script tracks. Move the WALM-332 / #793 note out of released `## 0.0.12` into new `## 0.0.13` in both `packages/mcp/CHANGELOG.md` and `docs/mcp/changelog.mdx` (mdx `answer:` too). Leave TS 0.1.6, Python 0.1.9, and OpenClaw 0.0.6 alone.

ducnmm (dismissed):
## Summary

The failed-sign-in copy no longer tells the user to revoke a reclaimable key. Recovery is still fail-closed.

Dump is the remaining blocker: origin/dev and origin/main both pin MCP at 0.0.12, this PR changes MCP, and the WALM-332 note was folded into already-released `## 0.0.12`. `docs/mcp/changelog.mdx` was not updated.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

packages/mcp/CHANGELOG.md
nikola0x0: `origin/dev` moved after you wrote this and did the 0.0.13 bump itself, for WALM-480 — all seven manifests, `verify-manual-sdk-release.mjs`, and both changelogs. And 0.0.13 is still open: npm `latest` is `0.0.12`, with only `0.0.13-dev.0` published as a prerelease. So this belongs in 0.0.13 rather than a bump on top of it.

Merged `dev` and moved the #793 note out of shipped 0.0.12 into `## 0.0.13` in `packages/mcp/CHANGELOG.md` and `docs/mcp/changelog.mdx` — the mdx had no mention of it at all, so it also gets the release-summary line and the `answer` frontmatter. TS 0.1.7, Python 0.1.10 and OpenClaw 0.0.6 untouched (dev bumped the first two).

Also capitalized `Testnet` in the whoami note — the style-guide violation from the earlier audit was never fixed, only the em dashes were.

71981f4a. `verify-manual-sdk-release` reports MCP 0.0.13 synchronized; docs freshness and code-sync green; `packages/mcp` 87/87.

ducnmm (commented):
## Summary

71981f4a after the `origin/dev` merge clears the dump blocker from ab1636bd. origin/dev is MCP 0.0.13, origin/main is 0.0.12, this PR stays on 0.0.13 (no 0.0.14 bump). The WALM-332 / #793 note sits under unreleased `## 0.0.13` in both `packages/mcp/CHANGELOG.md` and `docs/mcp/changelog.mdx`; the mdx `answer:` and release summary mention the write-ahead reclaim. Manifests and `scripts/verify-manual-sdk-release.mjs` remain 0.0.13. Recovery is still fail-closed: seconds timestamp, pending-key reuse, logout clears `login-pending.json` outside `clearCreds`, write-ahead throws, 401/403 never delete, 5xx/429/`AUTH_UPSTREAM_UNAVAILABLE` stay `unavailable`. User-facing timeout/`rejected` copy no longer tells the user to revoke a reclaimable key. Remaining notes are comments and the docs style-guide audit on the new mdx bullet.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 2

packages/mcp/src/recovery.ts
```diff
@@ -0,0 +1,283 @@
...
+        log.warn("login.pending.relayer_unreachable", {
+            publicKey: pending.delegatePublicKeyHex,
+        });
+        return { outcome: "unavailable", strandedPublicKey: pending.delegatePublicKeyHex };
+    }
+
+    if (res.status !== 200 || !isWhoami(res.body)) {
+        // `rejected` is reserved for the relayer actually denying this
+        // identity, because that is the only outcome whose advice — sign in
+        // again, then revoke the key — is safe to give. Telling a user to do
```
ducnmm: **[nit]** User-facing copy is updated (`login.ts:329` timeout; `recovery.ts:277` `rejected` branch: reuse / next-start reclaim; revoke only to abandon). Two comments still describe the old advice: the `denied` split at line 197 says rejected's advice is "sign in again, then revoke the key", and `formatStrandedLoginNotice`'s JSDoc at line 249 still says "the actionable step is revoking it from the dashboard". `superseded` keeping revoke is correct (the record is cleared). The `unavailable` tests still pin no `revoke`. In `bridge.ts:681` the logout comment still points at "the relayer-401 handling below"; that handling is above `handleLocalLogout` (module doc at line 10 and the SSE 401 path at line 329).

**Suggestion:** Reword the `denied` comment and the `formatStrandedLoginNotice` JSDoc to match the new copy (reuse / next-start reclaim; revoke only to abandon). Point the logout comment at the 401 handling above, or drop the direction.

docs/mcp/changelog.mdx
```diff
@@ -28,15 +28,16 @@ questions:
...
 ---
 
 ## 0.0.13
 
-This release reports restore `failed` counts and retries the same page when truncation is a transient download or embed blip.
+This release saves the delegate keypair before sign-in hands a URL to the browser and reclaims it on the next start, and it reports restore `failed` counts and retries the same page when truncation is a transient download or embed blip.
 
 ### Fixed
 
+- Persist the delegate keypair before the sign-in URL is handed to the browser, and reclaim it on the next start. The browser's on-chain `add_delegate_key` is paid and irreversible, and it happens before the callback that saved the private half, so a client that died in that window destroyed the only copy of a key the user had already paid for and left an orphaned registration nobody could use. (#793) Signing out discards the pending record along with the credentials, a second sign-in against the same relayer reuses the stranded key rather than minting over it, and a sign-in that cannot write the record fails instead of publishing a URL it cannot back. Reclaiming works on Mainnet; Testnet requires an account-id hint the recovering client does not have.
```
ducnmm: **[nit]** The style-guide audit re-ran on 71981f4a and still reports three violations on the new 0.0.13 Fixed bullet: "on-chain" → "onchain"; two passive-voice hits (`before the sign-in URL is handed to the browser`; `The browser's on-chain add_delegate_key is paid and irreversible, and it happens before the callback that saved the private half`). The mdx release summary at line 36 was already rewritten in active voice; this bullet was not. `packages/mcp/CHANGELOG.md:7` is the same sentence, so a mdx-only edit would desync the two changelogs.

**Suggestion:** Apply the audit's wording to the Fixed bullet in both `docs/mcp/changelog.mdx` and `packages/mcp/CHANGELOG.md` so they stay identical. Leave MCP at 0.0.13.

packages/mcp/src/recovery.ts
nikola0x0: Fixed in `68d10917`. The denied split and `formatStrandedLoginNotice`'s JSDoc now both say that signing in again reuses the key and that revoking is only to abandon it. For the logout comment I dropped the direction rather than repointing it.

docs/mcp/changelog.mdx
nikola0x0: Fixed in `68d10917`, in both changelogs. The 0.0.13 bullet now uses the audit's wording: `onchain`, "before sign-in hands the URL to the browser", "costs gas and is irreversible", "the callback that saves the private half". Re-checked after `02a3bf76`: both 0.0.13 Fixed lists are identical, 9 bullets each. MCP stays 0.0.13.

packages/mcp/src/login.ts
```diff
@@ -279,7 +326,7 @@ export async function loginFlow(opts: LoginOptions = {}): Promise<MemWalCredenti
         const timer = setTimeout(() => {
             if (!creds) {
                 error = new Error(
-                    `Login timed out after ${cfg.timeoutMs}ms. If you already approved the wallet transaction, a delegate key may exist on-chain without local credentials. Remove unused keys from the dashboard, then run login again.`,
+                    `Login timed out after ${cfg.timeoutMs}ms. If you already approved the wallet transaction, that delegate key is on disk and still usable: run login again and it is reused, or restart and it is reclaimed automatically. Only remove it from the dashboard if you mean to abandon it.`,
```
ducnmm: **[bug]** This timeout copy is right (reuse / reclaim; revoke only to abandon). The signed-out stub still wraps it with `loginFailureNotice` (`packages/mcp/src/messages.ts:128`, not in this diff), which says the unused key “may already be registered” and to “Remove it from the dashboard if you are not using it. Sign in again and open the new link.” Same leftover in `docs/troubleshooting/overview.md` (“should be revoked”, “Signing in again is a full new attempt”). In this session the pending record is on disk and a second `memwal_login` reuses it; following that banner revokes the only copy restart would reclaim. The `Reason:` line then contradicts the wrapper.

**Suggestion:** Match this timeout / stranded-rejected copy in `loginFailureNotice` and the troubleshooting page: call `memwal_login` again (same key is reused), or restart to reclaim; revoke only to abandon. Pin the notice in a test the way `login-recovery.test.mjs` already forbids `/revoke/i` on `unavailable`.

ducnmm (changes requested):
## Summary

Recovery is still fail-closed on `8e5a0949`: Unix-seconds whoami timestamp, pending-key reuse, logout clears `login-pending.json`, write-ahead throws, 401/403 never delete, 5xx/429/`AUTH_UPSTREAM_UNAVAILABLE` stay `unavailable`. Dump stays under unpublished `## 0.0.13` in both changelogs (mdx `answer:` included).

One leftover: `loginFlow`'s timeout copy is correct, but the in-session `loginFailureNotice` wrapper (and the troubleshooting page) still tell the user to revoke the key this PR exists to keep.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

packages/mcp/src/login.ts
nikola0x0: Fixed in 4b6d5d86. `loginFailureNotice`, the timeout reason, the bridge warning, the rejected-recovery notice and the troubleshooting page no longer tell anyone to revoke the key.

One change from your suggestion: they do not say "call `memwal_login` again, the same key is reused" for an approved key. `ConnectMcp.tsx` always sends `add_delegate_key`, and the contract aborts with `EDelegateKeyAlreadyExists` on a key that is already registered, so signing in again only works if the wallet step was never approved. The copy now splits on that: approved → restart to reclaim; not approved → sign in again; dashboard only to abandon (Testnet, where reclaim cannot confirm the key: remove it first, then sign in).

Pinned in `login-failure-notice.test.mjs`: the notice must say "Restart the MCP client" and "reclaim", must not match /revoke/i, and any sentence mentioning the dashboard must say "abandon". MCP tests 137/137.

packages/mcp/src/messages.ts
```diff
@@ -125,10 +130,12 @@ export function loginFailureNotice(reason: string | null): string {
         "",
         `Reason: ${reason}`,
         "",
-        "The unused key from this attempt may already be registered on your account. Remove it",
-        "from the dashboard if you are not using it. Sign in again and open the new link",
-        "straight away. A retry only helps once the MCP client is left running through the",
-        "wallet prompt.",
+        "If you approved the wallet step, that key is registered and saved on this machine.",
+        "Restart the MCP client within 24 hours and it is reclaimed. Signing in again cannot",
```
ducnmm, line 134: **[bug]** This notice now correctly says restart to reclaim and that signing in again cannot register the same key twice. After a timed-out `memwal_login`, `auth-required.ts` still concatenates `LOGIN_INSTRUCTION` ("easiest fix: `memwal_login`", "No terminal command, no client restart"). The unique commit's test pins `/memwal_login/` on that blob, so the contradiction ships. An agent following the last CTA retries the key instead of restarting.

**Suggestion:** When `lastLoginFailure` is set, do not append the generic no-restart instruction. Point at the same split as this notice. Pin the concatenated blob so it cannot say both reclaim and no client restart.

ducnmm (changes requested):
## Summary

Head `cbe48766`. Previous 8e5a0949 finding is fixed on `loginFailureNotice` and troubleshooting (approved stranded key → restart; dashboard removal is abandon-only). Two leftover surfaces still tell the user to retry/revoke that key.

## Issue counts by severity

- bugs: 2
- suggestions: 0
- nits: 0

## Issues outside the diff

- **[bug]** `apps/app/src/pages/ConnectMcp.tsx` callback-failed card still says sign in again and remove from the dashboard. The troubleshooting section this card links to now says the opposite. Match the restart/reclaim split; dashboard removal only to abandon.
- **[bug]** `packages/mcp/src/auth-required.ts` concatenates `loginFailureNotice` with `LOGIN_INSTRUCTION` ("easiest fix: memwal_login", "no client restart"). When `lastLoginFailure` is set, do not append the generic no-restart instruction.

packages/mcp/src/messages.ts
nikola0x0: Fixed in 898df373. A failed sign-in is now followed by a retry instruction scoped to the case it fixes (the wallet step was never approved); the generic `LOGIN_INSTRUCTION`, with "no terminal command, no client restart", is only used when there is no failure to report.

`login-failure-notice.test.mjs` pins the concatenated blob: on top of the existing restart / reclaim / no-revoke checks it must not match /no client restart/i. MCP tests 137/137.

nikola0x0:
@ducnmm the other one, `ConnectMcp.tsx` callback-failed card: fixed in 898df373. It now says restart the MCP client within 24 hours, do not sign in again first (the key is already registered and cannot be registered twice), and remove it from the dashboard only to abandon it — matching the troubleshooting section the card links to.

Pinned in `ConnectMcp.test.tsx`: the restart copy and the abandon-only dashboard sentence must be present, and the old "sign in again / remove it if you are not using it" copy must be gone. App tests 91/91, MCP 137/137.

ducnmm (approved):
## Summary

Head `898df373`. Both leftover copy bugs from `cbe48766` are fixed.

- `ConnectMcp.tsx` callback-failed card now says restart within 24 hours, do not sign in again first, dashboard removal only to abandon. Matches the troubleshooting section it links to. The old "Sign in again" / "Remove it from the dashboard if you are not using it" copy is pinned gone.
- `auth-required.ts` no longer concatenates `loginFailureNotice` with the generic `LOGIN_INSTRUCTION` ("no client restart"). After a failed `memwal_login` it appends `LOGIN_RETRY_INSTRUCTION`, scoped to the unapproved-wallet case. The concatenated blob is pinned not to say both reclaim and no-restart.

Recovery is still fail-closed. Approving.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #794, opened 2026-08-25
docs(mcp): correct login help text and README for non-destructive re-auth (WALM-308)

harrymove-ctrl:
## WALM-308 — `login` no longer wipes credentials, but the docs still said it did

Review of WALM-308 against current `dev`: the code half of the ticket is already fixed and covered.

1. `clearCreds()` before re-auth is gone (`83f1e913`). `main()` now does `args.forceLogin ? null : loadCreds()`, so `login` ignores the saved file instead of deleting it, and `saveCreds()` overwrites only after the browser flow succeeds.
2. Non-TTY `login` fails loudly with an actionable message instead of wiping and stalling in the auth-required stub (`01481eed`), covered by `test/login-non-tty.test.mjs`.

What was still outstanding is item 4 of the ticket: the user-facing text never caught up.

### Changes

1. `src/index.ts` help text: replaces "Force re-authentication (wipes existing credentials and opens browser)" with the actual contract — needs an interactive terminal, existing credentials kept until the new login succeeds.
2. `README.md` "## Login": same two guarantees, plus a pointer to the `memwal_login` tool for MCP-client (non-TTY) spawns and to `--logout` when removal is what the user actually wants.
3. `test/login-non-tty.test.mjs`: new regression test — a non-TTY `login` against a populated `~/.memwal/credentials.json` exits 1 and leaves the file byte-identical. This is the assertion that would have caught the original bug.

### Verification

`npm test` in `packages/mcp`: 52 passing, 0 failing (includes `tsc`).

No runtime behaviour change; documentation and test coverage only.


harrymove-ctrl:
Closing unmerged. The underlying WALM-308 bug is already fixed on dev (83f1e913 removed the pre-emptive clearCreds, 01481eed made non-TTY login fail loudly). This PR was help-text/README wording plus a test that only re-asserts an early-exit path, so it does not change behaviour or add real coverage.

PR #803, opened 2026-08-27
fix(relayer): stop leaking pool-wallet funding errors to remember callers (WALM-399)

harrymove-ctrl:
Addresses WALM-399 and #655.

Does not fully close WALM-399: the ticket's rate-limit item (do not charge infra-failed jobs against the delegate-key budget) is tracked separately, so the ticket stays open after this merges.

## Problem

A failed `remember` job returned the sidecar error verbatim to the caller. From the hosted Playground that surfaced as:

```
Internal Error: durable Walrus upload failed (503 Service Unavailable):
Unable to perform gas selection due to insufficient SUI balance
```

alongside a relayer-owned pool-wallet address. That address is infrastructure, not the caller's, and it rotates between attempts. Users read it as a funding instruction, which is the ticket's worst outcome: sending SUI to an address that was never theirs to fund.

PR #736 already classified this error string for retry and ops alerting, so the pool rotates onto a healthy wallet and ops gets a `GasPoolExhausted` alert. That fixed the operational half. The client-facing half was untouched: `remember_status` at `remember.rs:1168` passed `error_msg` straight through, and the Playground prints `terminal.error` verbatim, so it inherited the leak.

## Change

1. `WalletJobError::is_infrastructure_funding_error` in `jobs.rs`, composed from the existing #736 classifiers: gas-pool budget exhaustion, low WAL balance, and upload-slot congestion. No new matching heuristics.
2. `sanitize_job_error_for_client` in `remember.rs`, applied at both client read points (`remember_status`, `remember_bulk_status`):
   * infrastructure failures collapse to a single message stating it is our side, the memory was not stored, retry shortly, and no address shown is theirs to fund;
   * everything else keeps its diagnostic text with 32+ hex-digit addresses redacted to `0xdeadbeef…[redacted]`, so short hex like `0x2::coin` and Move abort codes stay readable.

The stored `error_msg` is unchanged, so `admin_dashboard` and ops alerts still see the raw text. Fixing it at the API boundary rather than in the Playground means the SDK, MCP, and Playground all benefit with no client changes.

## Verification

```bash
cd services/server && cargo check --all-targets && cargo test --bin memwal-server -- routes::remember jobs::
```

* `cargo check --all-targets` compiles clean. Worth calling out: CI does not build all targets, so a non-compiling test would otherwise land as green.
* 5 new tests, covering the exact error string from the issue, the low-WAL variant, address redaction on non-infra errors, the `None` case, and non-ASCII/short-hex preservation.
* Full `routes::remember` + `jobs::` suites pass 86/86, no regressions.
* `cargo fmt --check` is clean on both touched files. The `types.rs` and `memory_read.rs` diffs it reports are pre-existing on `dev`.

## The 429 half: the ticket's stated cause is wrong

The ticket attributes the `429` to failed infrastructure jobs being charged against the delegate-key budget, and suggests refunding them. I verified that against the code, and refunding would not have fixed it.

The delegate-key window is **30 weighted requests / 60s sliding** (`sliding_window_starts`, `max_requests_per_delegate_key`). `POST /api/remember` costs 5. `GET /api/remember/{job_id}` falls through `endpoint_weight` to 1, and it sits inside `protected_routes`, so **every poll is charged**. The Playground drove its own loop at a flat `POLL_MS = 1500` for up to 90s:

```
40 polls/min + 5 for the accepting POST = 45   vs   limit 30
```

The budget is exhausted at poll ~25, about **T+37s**, on a single `remember()` with no retry and nothing wrong with the job. That is the reported 429.

And the accepting POST's 5 units age out of the 60s window long before a slow job fails, so by the time the infra failure lands there is no charge left to refund.

The SDK already solved this: `waitForRememberJob` polls through `pollingDelayMs` (`1500 * 1.5^min(attempt,6)`, capped at 10s, jittered), roughly 12 polls over 90s. The Playground hand-rolls its loop so each intermediate status reaches the UI, and reimplemented the cadence without the backoff.

This PR applies the same curve to the Playground loop, and fixes the displayed code sample, which taught the flat 1.5s poll and reproduced the bug for anyone who copied it.

## Still out of scope

A genuine rate-limit refund path (worker credits weight back into the Redis window after an infra failure) is still not here. Given the above it is a much smaller concern than the ticket implies, since the charge has usually already expired. Tracked separately.

"Preflight hosted-relayer funding before accepting the job" is likewise a larger architectural change that #736's pool rotation and escalation partly supersedes.

### DIFF
--- a/services/server/src/jobs.rs
+++ b/services/server/src/jobs.rs
@@ -2518,6 +2518,19 @@ impl WalletJobError {
         lower.contains("timed out waiting for") && lower.contains("upload slot")
     }
 
+    /// True if `msg` describes a failure of the hosted relayer's own
+    /// infrastructure rather than anything the caller controls: pool-wallet SUI
+    /// gas exhaustion, pool-wallet WAL balance depletion, or upload-slot
+    /// congestion. None of these are actionable by the caller, and their raw
+    /// text names relayer-owned addresses and balances — surfacing it verbatim
+    /// on a job status has led users to send SUI to an address that was never
+    /// theirs to fund. `sanitize_job_error_for_client` gates on this.
+    pub fn is_infrastructure_funding_error(msg: &str) -> bool {
+        Self::is_gas_pool_budget_error(msg)
+            || Self::is_upload_slot_congestion_error(msg)
+            || parse_wal_balance_alert_info(msg).is_some()
+    }
+
     pub fn classify_sidecar_error(msg: &str) -> Self {
         let lower = msg.to_ascii_lowercase();
         // PostgreSQL B-tree tuple-size failures are deterministic for the
--- a/services/server/src/jobs.rs
+++ b/services/server/src/jobs.rs
@@ -2518,17 +2518,24 @@ impl WalletJobError {
         lower.contains("timed out waiting for") && lower.contains("upload slot")
     }
 
-    /// True if `msg` describes a failure of the hosted relayer's own
-    /// infrastructure rather than anything the caller controls: pool-wallet SUI
-    /// gas exhaustion, pool-wallet WAL balance depletion, or upload-slot
-    /// congestion. None of these are actionable by the caller, and their raw
-    /// text names relayer-owned addresses and balances — surfacing it verbatim
-    /// on a job status has led users to send SUI to an address that was never
-    /// theirs to fund. `sanitize_job_error_for_client` gates on this.
+    /// True if `msg` is a pool-wallet WAL shortfall. Deliberately the
+    /// substring half of `parse_wal_balance_alert_info` without its
+    /// `available < WAL_BALANCE_LOW_THRESHOLD_MIST` gate: that threshold
+    /// decides whether ops gets paged, and a shortfall on a large blob can
+    /// hold well over 2 WAL and still fail. Both are pool-wallet funding.
+    pub fn is_walrus_wal_shortfall(msg: &str) -> bool {
+        let lower = msg.to_ascii_lowercase();
+        lower.contains("insufficient balance") && lower.contains("::wal::wal")
+    }
+
+    /// True if `msg` is a failure of the relayer's own pool wallets or upload
+    /// pipeline rather than anything the caller controls. Gates client-facing
+    /// error replacement; each arm matches on substrings only, with no
+    /// alert-threshold coupling.
     pub fn is_infrastructure_funding_error(msg: &str) -> bool {
         Self::is_gas_pool_budget_error(msg)
             || Self::is_upload_slot_congestion_error(msg)
-            || parse_wal_balance_alert_info(msg).is_some()
+            || Self::is_walrus_wal_shortfall(msg)
     }
 
     pub fn classify_sidecar_error(msg: &str) -> Self {
--- a/services/server/src/routes/remember.rs
+++ b/services/server/src/routes/remember.rs
@@ -67,25 +67,24 @@ struct PendingBulkRememberItem {
 // Client-facing job-error sanitization
 // ============================================================
 
-/// Replacement text for any relayer-infrastructure failure surfaced on a job
-/// status. The raw sidecar text ("Unable to perform gas selection due to
-/// insufficient SUI balance", low-WAL alerts) names relayer-owned pool wallets
-/// and reads like a funding instruction, so callers have tried to fund an
-/// address that was never theirs. Say plainly that it is our side, that the
-/// job is retriable, and that no funding action is expected of them.
+/// Client-facing stand-in for a pool-wallet / upload-congestion failure on a
+/// job that has stopped retrying.
 const INFRA_JOB_ERROR_MESSAGE: &str = "The hosted relayer could not complete the durable Walrus upload: temporary capacity limit on Walrus Memory infrastructure. This is not caused by your account, wallet, or delegate key, and there is nothing for you to fund — do not send SUI or WAL to any address in response to this error. The memory was not stored. Retry in a few minutes, or contact support if it persists.";
 
-/// Redact 0x-prefixed hex blobs (Sui addresses, object IDs, package IDs) from a
-/// job error before it reaches the caller. Even on non-infrastructure errors,
-/// a bare address in a failure message reads as "send funds here". Keeps a
-/// short prefix so the message stays diagnosable when a user reports it.
+/// Same, for a job still in flight. Retryable infrastructure errors are
+/// recorded on rows that stay `running` (see `update_remember_job_after_wallet_error`),
+/// so this must not read as terminal: telling a caller to retry while the
+/// original attempt is live is what produces the resubmit-then-429 loop.
+const INFRA_JOB_RETRYING_MESSAGE: &str = "Temporary capacity limit on Walrus Memory infrastructure; the hosted relayer is still retrying this upload. This is not caused by your account, wallet, or delegate key, and there is nothing for you to fund — do not send SUI or WAL to any address in response to this error. Keep polling; do not resubmit unless the job reaches status `failed`.";
+
+/// Redact `0x` runs of 32+ hex digits (Sui addresses, object ids), keeping a
+/// short prefix. The length floor is what leaves `0x2::coin` and Move abort
+/// codes readable.
 fn redact_hex_addresses(msg: &str) -> String {
     let bytes = msg.as_bytes();
     let mut out = String::with_capacity(msg.len());
     let mut i = 0;
     while i < bytes.len() {
-        // Only 0x-prefixed runs long enough to be an address/object id, so
-        // short hex like an abort code or "0x2::coin" module path is kept.
         let is_start = bytes[i] == b'0'
             && i + 1 < bytes.len()
             && (bytes[i + 1] == b'x' || bytes[i + 1] == b'X')
@@ -101,7 +100,6 @@ fn redact_hex_addresses(msg: &str) -> String {
         while end < bytes.len() && bytes[end].is_ascii_hexdigit() {
             end += 1;
         }
-        // 32+ hex digits: a Sui address or object id, never an abort code.
         if end - (i + 2) >= 32 {
             out.push_str(&msg[i..i + 2 + 8]);
             out.push_str("…[redacted]");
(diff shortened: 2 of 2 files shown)


services/server/src/jobs.rs
```diff
@@ -2518,6 +2518,19 @@ impl WalletJobError {
...
+    /// infrastructure rather than anything the caller controls: pool-wallet SUI
+    /// gas exhaustion, pool-wallet WAL balance depletion, or upload-slot
+    /// congestion. None of these are actionable by the caller, and their raw
+    /// text names relayer-owned addresses and balances — surfacing it verbatim
+    /// on a job status has led users to send SUI to an address that was never
+    /// theirs to fund. `sanitize_job_error_for_client` gates on this.
+    pub fn is_infrastructure_funding_error(msg: &str) -> bool {
+        Self::is_gas_pool_budget_error(msg)
+            || Self::is_upload_slot_congestion_error(msg)
+            || parse_wal_balance_alert_info(msg).is_some()
```
ducnmm: **[suggestion]** `is_infrastructure_funding_error` treats WAL failures as infra only when `parse_wal_balance_alert_info` returns `Some`. That helper is an alerting predicate: it requires `insufficient balance` + `::wal::wal` **and** a parseable `available:` **and** `available < WAL_BALANCE_LOW_THRESHOLD_MIST` (2e9 / 2 WAL). A real pool-wallet WAL shortfall that omits `available:` or has `available >= 2 WAL` (large blob / long epochs, or a slightly different sidecar wording) therefore falls through to `redact_hex_addresses`. Callers still see “Insufficient balance of 0x…::wal::WAL for owner 0x… Required/Available,” which is the same funding-instruction class WALM-399 is trying to stop. Redaction removes a pasteable address, but the instruction remains. Gas-pool and upload-slot arms match on substrings with no extra threshold; WAL should too.

**Suggestion:** Gate WAL sanitization on the same first-pass check as the parser (`insufficient balance` && `::wal::wal`), or on `classify_sidecar_error` == `WalrusBalanceLow` plus that substring pair — not on `parse_wal_balance_alert_info(...).is_some()`. Keep the thresholded parser for Slack alerts only. Add a test whose `Available:` is `>= 2_000_000_000` (and one with no `available:` at all) and assert they still become `INFRA_JOB_ERROR_MESSAGE`.

services/server/src/routes/remember.rs
```diff
@@ -63,6 +63,68 @@ struct PendingBulkRememberItem {
...
+// Client-facing job-error sanitization
+// ============================================================
+
+/// Replacement text for any relayer-infrastructure failure surfaced on a job
+/// status. The raw sidecar text ("Unable to perform gas selection due to
+/// insufficient SUI balance", low-WAL alerts) names relayer-owned pool wallets
+/// and reads like a funding instruction, so callers have tried to fund an
+/// address that was never theirs. Say plainly that it is our side, that the
+/// job is retriable, and that no funding action is expected of them.
+const INFRA_JOB_ERROR_MESSAGE: &str = "The hosted relayer could not complete the durable Walrus upload: temporary capacity limit on Walrus Memory infrastructure. This is not caused by your account, wallet, or delegate key, and there is nothing for you to fund — do not send SUI or WAL to any address in response to this error. The memory was not stored. Retry in a few minutes, or contact support if it persists.";
```
ducnmm, line 72: **[suggestion]** `INFRA_JOB_ERROR_MESSAGE` states a terminal outcome (“The memory was not stored. Retry in a few minutes”) and `sanitize_job_error_for_client` applies it whenever `error_msg` is present, with no look at job status. `update_remember_job_after_wallet_error` writes the raw sidecar text onto still-`running` rows for retryable Transient gas rotation, `WalrusBalanceLow`, and congestion requeues (`status = "running"` when `!aborts_retries()`). Playground dumps the full status JSON while polling (`JSON.stringify(current)`), so a user can see `status: "running"` next to “memory was not stored / retry,” and submit a duplicate remember while the original is still in flight — the same retry-then-429 sequence as #655, minus the funding misread.

**Suggestion:** Phrase the infra text as non-terminal (“hosted relayer is retrying; do not fund any address; do not resubmit until status is failed”), or pick copy from status (`failed` → current text; `running`/`uploaded` → shorter “temporary infrastructure issue, still in progress”). Sanitizing only on `failed` would also work if you are willing to show nothing (or a milder string) during retries.

services/server/src/routes/remember.rs
```diff
@@ -63,6 +63,68 @@ struct PendingBulkRememberItem {
     namespace: String,
 }
 
+// ============================================================
+// Client-facing job-error sanitization
+// ============================================================
+
+/// Replacement text for any relayer-infrastructure failure surfaced on a job
```
ducnmm: **[suggestion]** The new comments restate the ticket and the surrounding code rather than a non-obvious constraint: the `INFRA_JOB_ERROR_MESSAGE` block (who funded what, why users sent SUI), `sanitize_job_error_for_client` (admin still sees raw text — already obvious from not writing the DB), and `WalletJobError::is_infrastructure_funding_error` in `jobs.rs` (same history). Inline notes in `redact_hex_addresses` (UTF-8 boundaries, keep `0x2::coin`) are the useful ones.

**Suggestion:** Drop the ticket narration. Keep a one-liner on the const (client-facing stand-in for pool-wallet / congestion failures) and a one-liner on the hex helper (32+ hex digits only, so abort codes and `0x2::coin` survive).

services/server/src/routes/remember.rs
```diff
@@ -2110,4 +2173,60 @@ mod tests {
         assert!(seen.iter().all(|len| *len <= SUMMARIZE_CHUNK_BYTES + 1024));
         assert!(seen.iter().all(|len| *len < MAX_REMEMBER_TEXT_BYTES / 4));
     }
+
+    // ---- client-facing job-error sanitization (WALM-399) ----
+
+    const GAS_SELECTION_ERR: &str = "Internal Error: durable Walrus upload failed (503 Service Unavailable): Unable to perform gas selection due to insufficient SUI balance";
+    const LOW_WAL_ERR: &str = "walrus upload failed: Insufficient balance of 0x356a26eb9e012a68958082340d4c4116e7f55615cf27affcff209cf0ae544f59::wal::WAL for owner 0x8d3c1f0a9b2e4d6c7a5f8e1b0d4c9a2f3e6b7d8c1a0f9e2b3c4d5a6f7e8b9c0d. Required: 64367730, Available: 10708877";
+
+    #[test]
```
ducnmm, line 2186: **[suggestion]** New tests cover the exact #655 gas-selection string, the happy-path WAL string (available well below 2 WAL), address redaction, `None`, and short/non-ASCII hex. They never exercise the third disjunct, `is_upload_slot_congestion_error`, even though `is_infrastructure_funding_error` advertises it and jobs.rs already has the production string (`walrus upload failed: Internal Error: walrus upload failed: timed out waiting for wallet 3 upload slot`).

**Suggestion:** Add one `sanitize_job_error_for_client` case with that congestion string (and the global-slot variant) asserting equality with `INFRA_JOB_ERROR_MESSAGE` and that `wallet 3` / `upload slot` do not leak.

ducnmm (commented):
## Summary

The change sanitizes client-facing remember job errors at the two poll endpoints (`remember_status`, `remember_bulk_status`) and leaves the stored `error_msg` intact, so `/api/admin/upload-errors` and ops alerts still see the raw sidecar text. The reported GH #655 / WALM-399 string (`gas selection` + `insufficient SUI`) is correctly collapsed via the existing `is_gas_pool_budget_error` matcher, and 32+ digit `0x` blobs on non-infra errors are truncated so a pool-wallet address is not a pasteable funding destination. Dominant residual risk is that WAL coverage reuses the *ops-alert* parser (requires parseable `available:` and `available < 2 WAL`) rather than “this is a pool-wallet WAL shortfall,” and the replacement copy claims a terminal failure even though retryable rows stay `running` with `error_msg` set.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 0

services/server/src/jobs.rs
harrymove-ctrl: Verified and fixed in c9bd4073. You are right: `parse_wal_balance_alert_info` bails on an unparseable `Available:` and on `available >= WAL_BALANCE_LOW_THRESHOLD_MIST` (2e9), so a large-blob shortfall holding well over 2 WAL fell through to plain redaction and the caller still read the funding instruction with only the address truncated.

Split out `is_walrus_wal_shortfall`, matching `insufficient balance` + `::wal::wal` and nothing else, so all three arms are substring-only with no alert-threshold coupling. The thresholded parser is untouched and still gates the Slack alert.

Added `wal_shortfall_is_infra_regardless_of_the_ops_alert_threshold` covering both cases you named: `Available: 40000000000` (20x the threshold) and a message with no `available:` field at all. Both assert `INFRA_JOB_ERROR_MESSAGE` and that neither `Insufficient balance` nor `::wal::WAL` survives.

services/server/src/routes/remember.rs
harrymove-ctrl: Good catch, this was the most consequential one. Confirmed at jobs.rs:215: `update_remember_job_after_wallet_error` sets `status = if error.aborts_retries() { "failed" } else { "running" }` and writes `error_msg` either way, so retryable gas rotation, WAL shortfall and congestion all land terminal-sounding copy on a live row. Combined with the Playground `JSON.stringify(current)` dump that is a direct invitation to resubmit over an in-flight attempt, which is the resubmit-then-429 half of #655.

Fixed in c9bd4073 by taking your second option: `sanitize_job_error_for_client` now takes the job status and returns `INFRA_JOB_RETRYING_MESSAGE` unless the job reached `failed`. The retrying variant says the relayer is still retrying and to keep polling rather than resubmit unless the job reaches `failed`. Both variants keep the do-not-fund sentence, since that is the point of the change.

`retryable_infra_error_on_a_running_job_does_not_read_as_terminal` asserts the split across running/pending/uploaded, that "was not stored" never appears on those, and that the funding guidance survives in both.

services/server/src/routes/remember.rs
harrymove-ctrl: Fair, that was narration. Trimmed in c9bd4073 to a one-liner on each constant and on `redact_hex_addresses`, keeping the UTF-8 boundary note and the 32-hex-digit floor rationale, which are the parts not readable off the code. The one comment I did keep at length is on `INFRA_JOB_RETRYING_MESSAGE`, since why it must not read as terminal depends on `update_remember_job_after_wallet_error` writing errors onto still-running rows, which is not visible from this file.

services/server/src/routes/remember.rs
harrymove-ctrl: Correct, the third disjunct was advertised but never exercised. Added `upload_slot_congestion_is_infra_and_hides_the_wallet_index` in c9bd4073 using both production strings, the wallet-scoped one you quoted and the global-slot variant, asserting equality with `INFRA_JOB_ERROR_MESSAGE` and that neither `wallet 3` nor `upload slot` leaks.

ducnmm (approved):
Re-reviewed c9bd4073. All four review notes are addressed and the new tests pin them.

- WAL infra match is now `insufficient balance` + `::wal::wal`, not the 2 WAL alert gate. The above-threshold `Available:` fixture and the missing-`Available:` case both become `INFRA_JOB_ERROR_MESSAGE`.
- `sanitize_job_error_for_client` now keys off job status: terminal copy only on `failed`; `running` / `pending` / `uploaded` get `INFRA_JOB_RETRYING_MESSAGE` so Playground cannot look like a dead job while Apalis is still rotating.
- Congestion strings (wallet-scoped and global-slot) are covered and do not leak `wallet 3` / `upload slot`.
- Comment pass is down to the constraints that are not readable off the code.

Non-blocking, not merge-blocking: `classify_sidecar_error` still promotes `WalrusBalanceLow` only through `parse_wal_balance_alert_info`, so a large-blob shortfall sitting above 2 WAL is sanitized for the client but will not take the `WalrusBalanceLow` ops arm. Client leak is closed; paging that class of failure is a separate ops question.

CI is green on Clippy + server unit/e2e. Approving.

PR #804, opened 2026-08-27
ci: split live JS SDK e2e out of test.yml (WALM-426)

ducnmm:
## Summary
- Move live JS SDK e2e into `test-sdk.yml` so `test.yml` is no longer carrying schedule guards, `sdk-changed`, or duplicated Node/Playwright setup.
- Slim `live.e2e.mjs` to serial auth-rejection tests plus one remember → recall → namespace flow. Method-level coverage stays in unit tests.
- Push to `dev`/`staging`/`main` waits until Railway `/health.build.commit` equals this SHA before writing.
- `workflow_dispatch` / schedule wait for `/health` `write_ready` only (Railway does not deploy a feature-branch SHA).
- Live e2e does **not** run on PRs. Preview: Actions → Test JS SDK / Test Python SDK → Run workflow.

## Test plan
- [x] Cold JS e2e on `dev` (dispatch): https://github.com/MystenLabs/MemWal/actions/runs/33040805124 — 11/11
- [x] Cold Python e2e on `dev` (dispatch): https://github.com/MystenLabs/MemWal/actions/runs/33046091162 — 23/23
- [x] Preview with waiter + User-Agent fix (dispatch): JS https://github.com/MystenLabs/MemWal/actions/runs/33048699261 (11/11), Python https://github.com/MystenLabs/MemWal/actions/runs/33048702734 (23/23)
- [ ] Full run on merge to `dev` (wait for Railway SHA, then JS + Python live e2e)

Linear: WALM-426

### DIFF
--- a/.github/actions/setup-playwright/action.yml
+++ b/.github/actions/setup-playwright/action.yml
@@ -1,0 +1,26 @@
+name: Setup Playwright
+description: Cache and install Chromium for a workspace package
+
+inputs:
+  package:
+    description: pnpm filter name (e.g. @memwal/chatbot)
+    required: true
+  cache-id:
+    description: Cache key suffix so parallel jobs do not race one entry
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Cache Playwright browsers
+      uses: actions/cache@v4
+      with:
+        path: ~/.cache/ms-playwright
+        key: pw-${{ inputs.cache-id }}-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
+        restore-keys: |
+          pw-${{ inputs.cache-id }}-${{ runner.os }}-
+          pw-${{ runner.os }}-
+
+    - name: Install Playwright (Chromium + OS deps)
+      shell: bash
+      run: pnpm --filter ${{ inputs.package }} playwright:install
--- a/.github/actions/wait-for-relayer/action.yml
+++ b/.github/actions/wait-for-relayer/action.yml
@@ -59,7 +59,11 @@ runs:
         last_error = "no successful /health response"
         while time.monotonic() < deadline:
             try:
-                with urllib.request.urlopen(url, timeout=10) as response:
+                request = urllib.request.Request(
+                    url,
+                    headers={"User-Agent": "MemWal-ci-wait-for-relayer"},
+                )
+                with urllib.request.urlopen(request, timeout=10) as response:
                     data = json.load(response)
                 status = str(data.get("status") or "")
                 ready = data.get("write_ready") is True
--- a/.github/actions/wait-for-relayer/action.yml
+++ b/.github/actions/wait-for-relayer/action.yml
@@ -40,27 +40,24 @@ runs:
           exit 1
         fi
 
-        expected="${EXPECTED_SHA:-}"
-        case "$(printf '%s' "$expected" | tr '[:upper:]' '[:lower:]')" in
-          false|null|none) expected="" ;;
-        esac
-        since="${SINCE_SHA:-}"
+        expected="$(printf '%s' "${EXPECTED_SHA:-}" | tr '[:upper:]' '[:lower:]' | tr -d '[:space:]')"
+        since="$(printf '%s' "${SINCE_SHA:-}" | tr '[:upper:]' '[:lower:]' | tr -d '[:space:]')"
+        case "$expected" in false|null|none) expected="" ;; esac
+        case "$since" in false|null|none) since="" ;; esac
         zeros="0000000000000000000000000000000000000000"
 
         # Railway relayer watchPatterns is /services/server/**. A push that
-        # does not touch that tree is recorded SKIPPED — no build, no new SHA
-        # — so waiting for github.sha never clears (PR 804 / nikola0x0).
+        # does not touch that tree is SKIPPED — no new SHA is served — so
+        # waiting on expected-commit would hang until timeout.
         if [ -n "$expected" ] && [ -n "$since" ] && [ "$since" != "$zeros" ]; then
-          if git fetch --no-tags --depth=1 origin "$since" >/tmp/wait-relayer-fetch.log 2>&1; then
-            if ! git diff --name-only "$since" "$expected" | grep -qE '^services/server/'; then
-              echo "No services/server/** changes in this push; Railway will SKIP the relayer deploy. Waiting on write_ready only."
+          if git fetch --no-tags --depth=1 origin "$since" \
+            && changed="$(git diff --name-only "$since" "$expected")"; then
+            if ! printf '%s\n' "$changed" | grep -Eq '^services/server/'; then
+              echo "Railway will SKIP this push, so SHA wait is skipped."
               expected=""
-            else
-              echo "services/server/** changed; waiting for Railway to serve ${expected:0:12}."
             fi
           else
-            echo "Could not fetch since-commit ${since:0:12}; keeping SHA wait."
-            cat /tmp/wait-relayer-fetch.log >&2 || true
+            echo "Could not fetch/diff since-commit; keeping SHA wait."
           fi
         fi
         export EXPECTED_SHA="$expected"
--- a/.github/actions/wait-for-relayer/action.yml
+++ b/.github/actions/wait-for-relayer/action.yml
@@ -40,6 +40,9 @@ runs:
 
         base = os.environ["SERVER_URL"].rstrip("/")
         expected = os.environ.get("EXPECTED_SHA", "").strip().lower()
+        # GitHub `false && github.sha || ''` can stringify as "false".
+        if expected in {"", "false", "null", "none"}:
+            expected = ""
         timeout = int(os.environ.get("TIMEOUT_SECS", "480"))
         poll = int(os.environ.get("POLL_SECS", "10"))
         url = f"{base}/health"
--- a/.github/actions/wait-for-relayer/action.yml
+++ b/.github/actions/wait-for-relayer/action.yml
@@ -1,0 +1,91 @@
+name: Wait for relayer
+description: Poll GET /health until Railway is serving this git SHA
+
+inputs:
+  server-url:
+    description: Relayer base URL
+    required: true
+  expected-commit:
+    description: Full git SHA to wait for. Empty waits for health only.
+    required: false
+    default: ""
+  timeout-seconds:
+    description: Give up after this many seconds (Railway deploys are ~2m, cold image longer)
+    required: false
+    default: "480"
+  poll-seconds:
+    required: false
+    default: "10"
+
+runs:
+  using: composite
+  steps:
+    - name: Wait until /health serves the expected commit
+      shell: bash
+      env:
+        SERVER_URL: ${{ inputs.server-url }}
+        EXPECTED_SHA: ${{ inputs.expected-commit }}
+        TIMEOUT_SECS: ${{ inputs.timeout-seconds }}
+        POLL_SECS: ${{ inputs.poll-seconds }}
+      run: |
+        set -euo pipefail
+
+        if [ -z "${SERVER_URL:-}" ]; then
+          echo "::error::server-url is empty"
+          exit 1
+        fi
+
+        python3 - <<'PY'
+        import json, os, sys, time, urllib.error, urllib.request
+
+        base = os.environ["SERVER_URL"].rstrip("/")
+        expected = os.environ.get("EXPECTED_SHA", "").strip().lower()
+        timeout = int(os.environ.get("TIMEOUT_SECS", "480"))
+        poll = int(os.environ.get("POLL_SECS", "10"))
+        url = f"{base}/health"
+        deadline = time.monotonic() + timeout
+
+        print(
+            f"Waiting on {url} (timeout {timeout}s, expected commit {expected or '<any>'})",
+            flush=True,
+        )
+
+        def commit_matches(commit: str) -> bool:
+            if not expected:
+                return True
+            got = commit.lower()
+            return bool(got) and (got == expected or expected.startswith(got) or got.startswith(expected))
+
+        last_error = "no successful /health response"
+        while time.monotonic() < deadline:
+            try:
+                with urllib.request.urlopen(url, timeout=10) as response:
+                    data = json.load(response)
+                status = str(data.get("status") or "")
+                ready = data.get("write_ready") is True
+                commit = str((data.get("build") or {}).get("commit") or "")
+                print(
+                    f"waiting: status={status or '?'} write_ready={str(ready).lower()} commit={commit or 'none'}",
+                    flush=True,
+                )
+                if status == "ok" and ready and commit_matches(commit):
+                    print(f"Relayer ready (commit={commit or 'none'})", flush=True)
+                    raise SystemExit(0)
+                last_error = (
+                    f"status={status!r} write_ready={ready} commit={commit or 'none'}"
+                )
+            except urllib.error.URLError as exc:
+                last_error = f"health unreachable: {exc}"
+                print(f"waiting: {last_error}", flush=True)
+            except json.JSONDecodeError as exc:
+                last_error = f"health returned non-JSON: {exc}"
+                print(f"waiting: {last_error}", flush=True)
+            time.sleep(poll)
+
+        print(
+            f"::error::Timed out after {timeout}s waiting for {url} to serve "
+            f"commit {expected or '<any>'} ({last_error}). Railway deploys take ~2m.",
+            file=sys.stderr,
+        )
+        raise SystemExit(1)
+        PY
--- a/.github/actions/wait-for-relayer/action.yml
+++ b/.github/actions/wait-for-relayer/action.yml
@@ -1,5 +1,5 @@
 name: Wait for relayer
-description: Poll GET /health until Railway is serving this git SHA
+description: Poll GET /health until the relayer is write-ready (and serving this SHA if Railway built it)
 
 inputs:
   server-url:
@@ -7,6 +7,10 @@ inputs:
     required: true
   expected-commit:
     description: Full git SHA to wait for. Empty waits for health only.
+    required: false
+    default: ""
+  since-commit:
+    description: Push `github.event.before`. When set, SHA wait is skipped unless services/server/** changed (Railway watchPatterns SKIP those deploys).
     required: false
     default: ""
   timeout-seconds:
@@ -25,6 +29,7 @@ runs:
       env:
         SERVER_URL: ${{ inputs.server-url }}
         EXPECTED_SHA: ${{ inputs.expected-commit }}
+        SINCE_SHA: ${{ inputs.since-commit }}
         TIMEOUT_SECS: ${{ inputs.timeout-seconds }}
         POLL_SECS: ${{ inputs.poll-seconds }}
       run: |
@@ -35,8 +40,33 @@ runs:
           exit 1
         fi
 
+        expected="${EXPECTED_SHA:-}"
+        case "$(printf '%s' "$expected" | tr '[:upper:]' '[:lower:]')" in
+          false|null|none) expected="" ;;
+        esac
+        since="${SINCE_SHA:-}"
+        zeros="0000000000000000000000000000000000000000"
+
+        # Railway relayer watchPatterns is /services/server/**. A push that
+        # does not touch that tree is recorded SKIPPED — no build, no new SHA
+        # — so waiting for github.sha never clears (PR 804 / nikola0x0).
+        if [ -n "$expected" ] && [ -n "$since" ] && [ "$since" != "$zeros" ]; then
+          if git fetch --no-tags --depth=1 origin "$since" >/tmp/wait-relayer-fetch.log 2>&1; then
+            if ! git diff --name-only "$since" "$expected" | grep -qE '^services/server/'; then
+              echo "No services/server/** changes in this push; Railway will SKIP the relayer deploy. Waiting on write_ready only."
+              expected=""
+            else
+              echo "services/server/** changed; waiting for Railway to serve ${expected:0:12}."
+            fi
+          else
+            echo "Could not fetch since-commit ${since:0:12}; keeping SHA wait."
+            cat /tmp/wait-relayer-fetch.log >&2 || true
+          fi
+        fi
+        export EXPECTED_SHA="$expected"
+
         python3 - <<'PY'
-        import json, os, sys, time, urllib.error, urllib.request
+        import http.client, json, os, sys, time, urllib.error, urllib.request
 
         base = os.environ["SERVER_URL"].rstrip("/")
         expected = os.environ.get("EXPECTED_SHA", "").strip().lower()
@@ -81,11 +111,9 @@ runs:
                 last_error = (
                     f"status={status!r} write_ready={ready} commit={commit or 'none'}"
                 )
-            except urllib.error.URLError as exc:
-                last_error = f"health unreachable: {exc}"
-                print(f"waiting: {last_error}", flush=True)
-            except json.JSONDecodeError as exc:
-                last_error = f"health returned non-JSON: {exc}"
+            except (urllib.error.URLError, OSError, http.client.HTTPException,
+                    json.JSONDecodeError) as exc:
+                last_error = f"health check failed: {exc}"
                 print(f"waiting: {last_error}", flush=True)
             time.sleep(poll)
 
--- a/.github/workflows/benchmark-live.yml
+++ b/.github/workflows/benchmark-live.yml
@@ -104,11 +104,10 @@ jobs:
         run: npm ci
 
       - name: Wait for Railway deploy
-        if: github.event_name == 'push'
         uses: ./.github/actions/wait-for-relayer
         with:
           server-url: ${{ env.SERVER_URL }}
-          expected-commit: ${{ github.sha }}
+          expected-commit: ${{ github.event_name == 'push' && github.sha || '' }}
 
       - name: Run memory API latency benchmark
         shell: bash
--- a/.github/workflows/benchmark-live.yml
+++ b/.github/workflows/benchmark-live.yml
@@ -64,7 +64,7 @@ jobs:
       name: benchmark-${{ github.event_name == 'workflow_dispatch' && inputs.target_environment || (github.ref_name == 'staging' && 'staging' || 'dev') }}
     timeout-minutes: 30
 
-    # Serializes against the sdk-e2e job in test.yml and the e2e job in
+    # Serializes against the sdk-e2e job in test-sdk.yml and the e2e job in
     # test-python-sdk.yml, which sign with this same BENCH_DELEGATE_KEY. The
     # relayer rate-limits per delegate key and per account, so two suites
     # running concurrently 429 each other. Group name must stay identical in
@@ -102,6 +102,13 @@ jobs:
       - name: Install sidecar script dependencies
         working-directory: services/server/scripts
         run: npm ci
+
+      - name: Wait for Railway deploy
+        if: github.event_name == 'push'
+        uses: ./.github/actions/wait-for-relayer
+        with:
+          server-url: ${{ env.SERVER_URL }}
+          expected-commit: ${{ github.sha }}
 
       - name: Run memory API latency benchmark
         shell: bash
--- a/.github/workflows/benchmark-live.yml
+++ b/.github/workflows/benchmark-live.yml
@@ -108,6 +108,7 @@ jobs:
         with:
           server-url: ${{ env.SERVER_URL }}
           expected-commit: ${{ github.event_name == 'push' && github.sha || '' }}
+          since-commit: ${{ github.event_name == 'push' && github.event.before || '' }}
 
       - name: Run memory API latency benchmark
         shell: bash
--- a/.github/workflows/test-sdk.yml
+++ b/.github/workflows/test-sdk.yml
@@ -1,10 +1,13 @@
 name: Test JS SDK
 
 # Live JS SDK e2e against the bench relayer.
-# Unit tests stay in test.yml so they run on every PR.
+# Unit tests stay in test.yml so they run on every PR; this job also runs
+# them before live writes so a typecheck-only green cannot mint memories.
 #
-# push to main/staging/dev: wait until /health.build.commit == github.sha
-#   (Railway takes ~2m to serve the new replica).
+# push that touched services/server/**: wait until /health.build.commit == github.sha
+#   (Railway takes ~2m to serve the new replica). Other pushes are SKIPPED by
+#   Railway watchPatterns, so wait-for-relayer drops the SHA and waits on
+#   write_ready only.
 # workflow_dispatch / schedule: wait until /health is ok — Railway does not
 #   deploy a feature-branch SHA, so this is a preview against the live replica.
 
@@ -68,7 +71,7 @@ jobs:
         uses: ./.github/actions/setup-js
 
       - name: Build
-        run: pnpm build:sdk
+        run: pnpm build:sdk && pnpm --filter @mysten-incubation/memwal test
 
       - name: Check credentials
         id: config
@@ -99,6 +102,7 @@ jobs:
         with:
           server-url: ${{ env.MEMWAL_SERVER_URL }}
           expected-commit: ${{ github.event_name == 'push' && github.sha || '' }}
+          since-commit: ${{ github.event_name == 'push' && github.event.before || '' }}
 
       - name: E2E tests
         id: e2e
--- a/.github/workflows/test-sdk.yml
+++ b/.github/workflows/test-sdk.yml
@@ -1,0 +1,128 @@
+name: Test JS SDK
+
+# Live JS SDK e2e against the bench relayer.
+# Unit tests stay in test.yml so they run on every PR.
+
+on:
+  push:
+    branches: [main, staging, dev]
+    paths:
+      - "packages/sdk/**"
+      - ".github/workflows/test-sdk.yml"
+      - ".github/actions/setup-js/**"
+      - ".github/actions/wait-for-relayer/**"
+  workflow_dispatch:
+    inputs:
+      target_environment:
+        description: Benchmark environment to run the authenticated SDK e2e suite against
+        required: false
+        type: choice
+        default: dev
+        options:
+          - dev
+          - staging
+          - production
+  schedule:
+    # Offset from benchmark-live.yml (0 9 * * 1) and test-python-sdk.yml
+    # (17 9 * * 1) so the three suites do not write through the shared bench
+    # account at once. The concurrency group below enforces that too.
+    - cron: "47 9 * * 1"
+
+concurrency:
+  group: test-js-sdk-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
+  sdk-e2e:
+    name: SDK / E2E (live relayer)
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    # Byte-identical to benchmark-live.yml and test-python-sdk.yml so the three
+    # suites serialize on the shared BENCH_DELEGATE_KEY. Do not edit one copy.
+    concurrency:
+      group: bench-account-${{ github.event_name == 'workflow_dispatch' && inputs.target_environment || (github.ref_name == 'main' && 'production' || github.ref_name == 'staging' && 'staging' || 'dev') }}
+      cancel-in-progress: false
+
+    environment:
+      name: benchmark-${{ github.event_name == 'workflow_dispatch' && inputs.target_environment || (github.ref_name == 'main' && 'production' || github.ref_name == 'staging' && 'staging' || 'dev') }}
+
+    env:
+      ENVIRONMENT_NAME: benchmark-${{ github.event_name == 'workflow_dispatch' && inputs.target_environment || (github.ref_name == 'main' && 'production' || github.ref_name == 'staging' && 'staging' || 'dev') }}
+      MEMWAL_PRIVATE_KEY: ${{ secrets.BENCH_DELEGATE_KEY }}
+      MEMWAL_ACCOUNT_ID: ${{ secrets.BENCH_ACCOUNT_ID }}
+      MEMWAL_SERVER_URL: ${{ vars.BENCH_SERVER_URL }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup JS
+        uses: ./.github/actions/setup-js
+
+      - name: Build
+        run: pnpm build:sdk
+
+      - name: Check credentials
+        id: config
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          # Without credentials the suite skips every authenticated test and
+          # the job would report success having never exercised the relayer
+          # beyond /health. Without the URL there is no safe fallback.
+          missing=0
+          for name in MEMWAL_PRIVATE_KEY MEMWAL_ACCOUNT_ID MEMWAL_SERVER_URL; do
+            if [ -z "${!name:-}" ]; then
+              echo "::error::${name} is empty — set BENCH_DELEGATE_KEY / BENCH_ACCOUNT_ID / BENCH_SERVER_URL on the ${ENVIRONMENT_NAME} environment."
+              missing=1
+            fi
+          done
+          if [ "$missing" -ne 0 ]; then
+            exit 1
+          fi
+          echo "authenticated=true" >> "$GITHUB_OUTPUT"
+
+      # Push races Railway: the relayer image takes ~2m to rebuild, so /health
+      # still serves the previous SHA until the new replica is live.
+      - name: Wait for Railway deploy
+        if: github.event_name == 'push'
+        uses: ./.github/actions/wait-for-relayer
+        with:
+          server-url: ${{ env.MEMWAL_SERVER_URL }}
+          expected-commit: ${{ github.sha }}
+
+      - name: E2E tests
+        id: e2e
+        working-directory: packages/sdk
+        run: |
+          node --test --test-concurrency=1 \
+            --test-reporter=spec --test-reporter-destination=stdout \
+            --test-reporter=junit --test-reporter-destination=e2e-results.xml \
+            "test/e2e/live.e2e.mjs"
+
+      - name: Write run summary
+        if: always()
+        run: |
+          {
+            echo "## JS SDK e2e"
+            echo
+            echo "| Field | Value |"
+            echo "| --- | --- |"
+            echo "| Environment | \`${ENVIRONMENT_NAME}\` |"
+            echo "| Relayer | \`${MEMWAL_SERVER_URL:-<unset>}\` |"
+            echo "| Authenticated | ${{ steps.config.outputs.authenticated }} |"
+            echo "| Result | ${{ steps.e2e.outcome }} |"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload e2e results
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: js-sdk-e2e-${{ env.ENVIRONMENT_NAME }}-${{ github.run_id }}
+          path: packages/sdk/e2e-results.xml
+          if-no-files-found: warn
+          retention-days: 14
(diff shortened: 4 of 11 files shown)


.github/workflows/test-sdk.yml
```diff
@@ -0,0 +1,133 @@
...
+          - production
+  schedule:
+    # Offset from benchmark-live.yml (0 9 * * 1) and test-python-sdk.yml
+    # (17 9 * * 1) so the three suites do not write through the shared bench
+    # account at once. The concurrency group below enforces that too.
+    - cron: "47 9 * * 1"
+
+concurrency:
+  group: test-js-sdk-${{ github.ref }}
+  cancel-in-progress: false
```
ducnmm: **[suggestion]** Workflow concurrency is `test-js-sdk-${{ github.ref }}` with `cancel-in-progress: false`, while push runs wait until `/health.build.commit` equals this run's `github.sha`. Relayer `railway.json` uses a single replica and has no `watchPatterns`, so a later push to the same branch cancels/replaces the in-progress Railway deploy. The older wait then polls for a SHA that will never become live and burns the full 480s timeout before the queued successor can start. Job-level `bench-account-...` should stay `cancel-in-progress: false` (that group is shared with Python e2e and the live benchmark); this workflow-level group only serializes Test JS SDK runs on one ref.

**Suggestion:** Set workflow-level `cancel-in-progress: true` (or true on `push` only). Optionally also fail-fast in `wait-for-relayer` when `/health` reports a different full SHA than `expected-commit`, so a superseded run does not sit on the timeout even if it is not cancelled.

packages/sdk/test/e2e/live.e2e.mjs
```diff
@@ -126,243 +88,159 @@ async function rawSignedRequest(method, path, body, privateKey, overrides = {})
...
-    const mw = client();
-    const result = await mw.restore(E2E_NAMESPACE);
-    assert.equal(result.namespace, E2E_NAMESPACE);
-    assert.ok(Number.isInteger(result.restored) && result.restored >= 0);
-    assert.ok(Number.isInteger(result.skipped) && result.skipped >= 0);
-    assert.ok(Number.isInteger(result.total) && result.total >= 0);
-    assert.equal(typeof result.truncated, "boolean");
-    assert.ok(result.owner.startsWith("0x"));
-});
+    test("remember accepts a job", async () => {
```
ducnmm: **[suggestion]** The file header says the authenticated path is one remember→recall→namespace flow, but `remember accepts a job` and `analyze extracts facts` still run first. `remember()` accepts and returns immediately while the relayer continues embed→SEAL→Walrus, and `analyze()` extracts then starts one background write per fact. Those pipelines are still in flight when `full flow` calls `rememberAndWait` on the same bench account, which is the rate-limit contention this suite was slimmed to avoid. `--test-concurrency=1` only serializes test functions, not the server-side jobs they already kicked off.

**Suggestion:** Drop the fire-and-forget `remember` / `analyze` cases from live e2e (keep them in unit tests), or wait for those job ids before starting the full flow so only one Walrus write pipeline runs at a time.

ducnmm:
@nikola0x0 mind reviewing this when you have a minute?

Live JS e2e moved out of `Tests` into `test-sdk.yml` (same shape as Python). Dispatch preview against relayer `dev` already passed (JS 11/11, Python 23/23). Full SHA-wait still needs merge to `dev`.

.github/workflows/test-sdk.yml
```diff
@@ -0,0 +1,133 @@
...
+          echo "authenticated=true" >> "$GITHUB_OUTPUT"
+
+      # Push: wait until this SHA is live (Railway ~2m). Dispatch/schedule:
+      # Railway never deploys a feature-branch SHA, so only wait until the
+      # already-deployed replica is healthy.
+      - name: Wait for Railway deploy
+        uses: ./.github/actions/wait-for-relayer
+        with:
+          server-url: ${{ env.MEMWAL_SERVER_URL }}
+          expected-commit: ${{ github.event_name == 'push' && github.sha || '' }}
```
nikola0x0: **Blocking — this gate can never clear on push.**

Railway only redeploys the relayer when `services/server/**` changes, but this workflow's push trigger fires only on `packages/sdk/**` and the three CI paths above. So `github.sha` is never a SHA Railway deploys: `commit_matches()` stays false for the full 480s and the step exits 1. Every merge to `dev` that runs this job fails.

Verified live — `relayer.dev.memwal.ai/health` serves `28e8ba307064c78f`, which is 27 commits behind `origin/dev`, and `git log 28e8ba30..origin/dev -- services/server` is empty.

The dispatch previews in the test plan pass because dispatch resolves this to `''` and only checks `write_ready`. The merge-to-dev row is unchecked, and it's the one that breaks.

Options: drop the SHA gate and wait on `write_ready` only (same guarantee the dispatch path already accepts), or only pass `expected-commit` when the push actually touched `services/server/**`.

.github/workflows/benchmark-live.yml
```diff
@@ -103,6 +103,12 @@ jobs:
         working-directory: services/server/scripts
         run: npm ci
 
+      - name: Wait for Railway deploy
```
nikola0x0: Same unsatisfiable gate as `test-sdk.yml:101`, and worse here: `benchmark-live.yml` has no `paths:` filter, so it runs on **every** push to `dev`/`staging`. A docs-only or app-only merge — 27 in a row on current `dev` — now burns 8 minutes polling and then fails a job that used to pass green.

.github/actions/wait-for-relayer/action.yml
```diff
@@ -0,0 +1,98 @@
...
+            return bool(got) and (got == expected or expected.startswith(got) or got.startswith(expected))
+
+        last_error = "no successful /health response"
+        while time.monotonic() < deadline:
+            try:
+                request = urllib.request.Request(
+                    url,
+                    headers={"User-Agent": "MemWal-ci-wait-for-relayer"},
+                )
+                with urllib.request.urlopen(request, timeout=10) as response:
```
nikola0x0: CPython's `AbstractHTTPHandler.do_open` only wraps `h.request()` in `URLError`; `getresponse()` and the body read in `json.load()` re-raise bare. So a `RemoteDisconnected`/`ConnectionResetError` from a Railway replica swap, or a `TimeoutError` on the 10s socket, escapes both `except` clauses and kills the loop with a traceback — during exactly the event this action exists to wait through, and with ~7 minutes of budget left unused.

```diff
-            except urllib.error.URLError as exc:
-                last_error = f"health unreachable: {exc}"
-                print(f"waiting: {last_error}", flush=True)
-            except json.JSONDecodeError as exc:
-                last_error = f"health returned non-JSON: {exc}"
+            except (urllib.error.URLError, OSError, json.JSONDecodeError) as exc:
+                last_error = f"health check failed: {exc}"
                 print(f"waiting: {last_error}", flush=True)
```

(`SystemExit` is a `BaseException`, so the success path at line 80 still escapes cleanly.)

.github/actions/wait-for-relayer/action.yml
```diff
@@ -0,0 +1,98 @@
...
+        print(
+            f"Waiting on {url} (timeout {timeout}s, expected commit {expected or '<any>'})",
+            flush=True,
+        )
+
+        def commit_matches(commit: str) -> bool:
+            if not expected:
+                return True
+            got = commit.lower()
+            return bool(got) and (got == expected or expected.startswith(got) or got.startswith(expected))
```
nikola0x0: `expected.startswith(got)` matches in both directions, so any truncated `got` — `"2"`, `"28"`, a half-populated `GIT_SHA` — satisfies the gate against an expected SHA that starts with it. That's the stale-replica run this action exists to prevent.

```diff
-            return bool(got) and (got == expected or expected.startswith(got) or got.startswith(expected))
+            return len(got) >= 7 and expected.startswith(got)
```

Still accepts a legitimately shortened SHA, rejects a stub.

.github/workflows/test-sdk.yml
```diff
@@ -0,0 +1,133 @@
...
+
+concurrency:
+  group: test-js-sdk-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
+  sdk-e2e:
```
nikola0x0: `needs: [sdk-checks, sdk-changed]` didn't survive the move (it's at `test.yml:220` on `dev`). `sdk-checks` now lives in a workflow this one can't declare a `needs:` on, so the Node 22/24 unit matrix no longer gates the live run.

Concretely: a commit that breaks remember/recall semantics but still typechecks passes `pnpm build:sdk`, and this job writes real memories through the full SEAL/Walrus pipeline on the shared bench account — `benchmark-production` on a push to `main` — before a single unit test has reported.

.github/workflows/test.yml
```diff
@@ -545,72 +226,40 @@ jobs:
...
-          # Separate from chatbot's `pw-` key so the two jobs don't race the
-          # same cache entry. Fall back to the shared prefix on a cold start.
-          key: pw-noter-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
-          restore-keys: |
-            pw-noter-${{ runner.os }}-
-            pw-${{ runner.os }}-
-
-      - name: Install Playwright (Chromium + OS deps)
+      - name: Setup Playwright
         timeout-minutes: 8
```
nikola0x0: `timeout-minutes` isn't honored on a step that `uses:` a composite action ([actions/runner#1979](https://github.com/actions/runner/issues/1979)). It worked on the old `run:` step; here it's silently ignored, so a hung Chromium/OS-deps download runs to the job's 25 minutes instead of failing fast at 8.

Same dead flag at `test.yml:125` and `test.yml:189`. Composite steps don't support it either, so it has to become a `timeout 480 pnpm --filter …` inside the action — or be dropped from all three, since right now it reads as a guard that doesn't exist.

.github/workflows/test-sdk.yml
```diff
@@ -0,0 +1,133 @@
...
+          done
+          if [ "$missing" -ne 0 ]; then
+            exit 1
+          fi
+          echo "authenticated=true" >> "$GITHUB_OUTPUT"
+
+      # Push: wait until this SHA is live (Railway ~2m). Dispatch/schedule:
+      # Railway never deploys a feature-branch SHA, so only wait until the
+      # already-deployed replica is healthy.
+      - name: Wait for Railway deploy
```
nikola0x0: This up-to-8-minute wait runs inside the `bench-account-*` critical section declared at line 51, so each holder occupies the shared group doing nothing but polling.

With `cancel-in-progress: false`, GitHub cancels a *pending* run when a newer one queues. Two dev merges close together plus the Monday cron and one of the three suites gets dropped rather than run — losing the drift signal the schedule exists for. The wait belongs in a preceding job, outside the group. Same placement in `test-python-sdk.yml:141` and `benchmark-live.yml:106`.

packages/sdk/test/e2e/live.e2e.mjs
```diff
@@ -126,243 +88,159 @@ async function rawSignedRequest(method, path, body, privateKey, overrides = {})
...
+        for (const entry of recalled.results) {
+            assert.equal(typeof entry.text, "string");
+            assert.equal(typeof entry.blob_id, "string");
+            assert.equal(typeof entry.distance, "number");
+        }
 
-    // Store a distinctive memory in an isolated namespace.
-    const memory = await mw.rememberAndWait(text, namespace, { timeoutMs: REMEMBER_TIMEOUT_MS });
-    assert.ok(memory.id.length > 0);
+        const isolated = await mw.recall({
```
nikola0x0: Nothing writes to `E2E_NAMESPACE_ALT` any more — the deleted `remember honors a per-call namespace override` test was the only writer (`live.e2e.mjs:257` on `dev`). So this passes trivially whether isolation works or the namespace simply doesn't exist, which makes it a weak assertion rather than a wrong one.

Related: line 206 passes `undefined` for the namespace, so no live test exercises a per-call override at all now. An SDK regression that silently fell back to `this.namespace` would ship green.

packages/sdk/test/e2e/live.e2e.mjs
```diff
@@ -126,243 +88,159 @@ async function rawSignedRequest(method, path, body, privateKey, overrides = {})
...
-});
-
-test("embed returns a vector", { skip: requiresKey }, async () => {
-    const mw = client();
-    const result = await mw.embed("hello world");
-    assert.ok(Array.isArray(result.vector));
-    assert.ok(result.vector.length > 0);
-    assert.equal(typeof result.vector[0], "number");
-});
+describe("authenticated", { skip: requiresKey, concurrency: 1 }, () => {
```
nikola0x0: `rememberBulkAndWait` and `analyzeAndWait` are mock-only after this. Bulk hits a distinct endpoint (`/api/remember/bulk`), and the deleted test asserted the SDK's client-side namespace bookkeeping precisely because the bulk status response carries no namespace — only a live relayer can falsify that. `mock.test.mjs` checks the SDK against a mock the SDK itself defines.

Reasonable trade for runtime, but worth calling out in the PR body so it's a decision rather than an accident.

packages/sdk/package.json
```diff
@@ -28,7 +28,7 @@
         "dev": "tsc --watch",
         "typecheck": "tsc --noEmit",
         "test": "tsc && node --test \"test/**/*.test.mjs\"",
-        "test:e2e": "tsc && node --test \"test/e2e/*.e2e.mjs\""
+        "test:e2e": "tsc && node --test --test-concurrency=1 \"test/e2e/*.e2e.mjs\""
```
nikola0x0: `--test-concurrency` caps how many test *files* run in parallel, and `test/e2e/` has exactly one. Within a file, `node:test` already runs top-level tests and describe subtests serially, so the `concurrency: 1` at `live.e2e.mjs:168` is also the default.

Both are inert. Harmless in themselves, but they read as a guarantee against parallel live writes through the shared `BENCH_DELEGATE_KEY` that nothing actually enforces — add a second `*.e2e.mjs` and you get exactly that.

.github/workflows/test-sdk.yml
```diff
@@ -0,0 +1,133 @@
...
+        uses: ./.github/actions/wait-for-relayer
+        with:
+          server-url: ${{ env.MEMWAL_SERVER_URL }}
+          expected-commit: ${{ github.event_name == 'push' && github.sha || '' }}
+
+      - name: E2E tests
+        id: e2e
+        working-directory: packages/sdk
+        run: |
+          node --test --test-concurrency=1 \
```
nikola0x0: Second copy of `packages/sdk`'s `test:e2e` script, and the two have already diverged — the script uses the glob `test/e2e/*.e2e.mjs`, this hardcodes `live.e2e.mjs`. Add a second e2e file and the workflow silently keeps running only the one.

`pnpm --filter @mysten-incubation/memwal test:e2e`, with the reporter flags moved into a `test:e2e:ci` script, keeps them in one place.

.github/workflows/test-sdk.yml
```diff
@@ -0,0 +1,133 @@
...
+
+# Live JS SDK e2e against the bench relayer.
+# Unit tests stay in test.yml so they run on every PR.
+#
+# push to main/staging/dev: wait until /health.build.commit == github.sha
+#   (Railway takes ~2m to serve the new replica).
+# workflow_dispatch / schedule: wait until /health is ok — Railway does not
+#   deploy a feature-branch SHA, so this is a preview against the live replica.
+
+on:
```
nikola0x0: No `pull_request` trigger, so this workflow gets zero validation on the PR that adds it, or on any PR that edits it. While `sdk-e2e` lived in `test.yml` a YAML or expression error at least surfaced when the file was parsed for the other jobs.

A typo in the `environment:` expression or a bad `uses:` path now first surfaces after merge to `dev`, against a live production-capable bench account.

.github/actions/setup-playwright/action.yml
```diff
@@ -0,0 +1,26 @@
+name: Setup Playwright
+description: Cache and install Chromium for a workspace package
+
+inputs:
+  package:
+    description: pnpm filter name (e.g. @memwal/chatbot)
+    required: true
+  cache-id:
```
nikola0x0: `cache-id` is the basename of `package` at all three call sites (`@memwal/chatbot`/`chatbot`, `researcher`/`researcher`, `@memwal/noter`/`noter`). Two required inputs that must agree can drift — and two callers passing the same `cache-id` with different packages silently share one cache entry, which is the race the doc comment warns about.

`cache_id=${package##*/}` inside the action drops the input.

.github/actions/setup-playwright/action.yml
```diff
@@ -0,0 +1,26 @@
...
+  using: composite
+  steps:
+    - name: Cache Playwright browsers
+      uses: actions/cache@v4
+      with:
+        path: ~/.cache/ms-playwright
+        key: pw-${{ inputs.cache-id }}-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
+        restore-keys: |
+          pw-${{ inputs.cache-id }}-${{ runner.os }}-
+          pw-${{ runner.os }}-
```
nikola0x0: After this PR nothing writes an unprefixed `pw-<os>-<hash>` entry — chatbot was the only job doing so and it now writes `pw-chatbot-…`. So this fallback can only hit pre-PR caches, which GitHub evicts after 7 days without access.

From then on every cold start (a `pnpm-lock.yaml` bump) falls through both restore-keys to a full Chromium + OS-deps download in all three jobs, with no shared warm path — the opposite of what the fallback is for.

.github/workflows/test.yml
```diff
@@ -36,320 +18,67 @@ permissions:
...
-      # packages/*/package.json, and MCP config blocks from the packages/mcp
-      # README. Failing the job blocks the merge, so drift is fixed in the
-      # same pull request.
-      - name: Check docs code references against the packages
-        run: node scripts/check-docs-code-sync.mjs
-
-  docs-freshness:
-    name: Docs / Freshness
-    if: github.event_name != 'schedule'
+  meta:
```
nikola0x0: `check-compatibility-contract.mjs`, `check-docs-code-sync.mjs` and `check-docs-freshness.mjs` are dependency-free scripts that finish in milliseconds. This provisions three runners with a full checkout + `setup-node` each, on every push and PR.

One job looping over the three — recording failures and exiting non-zero at the end — preserves the `fail-fast: false` behaviour at a third of the cost. The only loss is three check names collapsing into one.

.github/workflows/test-sdk.yml
nikola0x0: **Blocking — this gate hangs whenever the push doesn't also touch `services/server/**`.**

Railway's `relayer` service is configured with `watchPatterns: ["/services/server/**"]` and `rootDirectory: services/server/` in all three environments (confirmed against the Railway API). A push changing nothing under `services/server/` produces a deployment record marked **SKIPPED** — no build, no new SHA. The dev relayer's recent history is 11 SKIPPED to 1 SUCCESS.

This workflow's `paths:` filter fires on `packages/sdk/**` and the three CI paths, so an SDK-only change never triggers a Railway build. `/health.build.commit` keeps reporting the old SHA, `commit_matches()` stays false for the full 480s, and the step exits 1.

Observed live today: the dev relayer sat on `28e8ba30` (merge of #790, the last `services/server` change, Aug 25) through 27 pushes, then #803 merged at 15:20 — which does touch `services/server` — and Railway deployed `6102c76c` within seconds.

So the gate passes only when a PR happens to touch both the SDK and the server. That's the minority case, and it makes the failure look flaky rather than broken.

The dispatch previews pass because dispatch resolves this to `''` and only checks `write_ready` — which is why the merge-to-dev row is the one still unchecked.

Options: drop the SHA gate and wait on `write_ready` only (the guarantee the dispatch path already accepts), or pass `expected-commit` only when the push actually touched `services/server/**`.

.github/workflows/benchmark-live.yml
nikola0x0: Same gate as `test-sdk.yml:101`, and it misfires more often here: `benchmark-live.yml` has no `paths:` filter, so it runs on **every** push to `dev`/`staging` — while Railway only rebuilds the relayer on `/services/server/**`.

So every docs-only, app-only or SDK-only merge — 27 in a row on `dev` between Aug 25 and today — burns 8 minutes polling for a SHA Railway marked SKIPPED, then fails a job that used to pass green.

.github/actions/wait-for-relayer/action.yml
nikola0x0: CPython's `AbstractHTTPHandler.do_open` only wraps `h.request()` in `URLError`; `getresponse()` and the body read in `json.load()` re-raise bare. So a `RemoteDisconnected`/`ConnectionResetError` from a Railway replica swap, or a `TimeoutError` on the 10s socket, escapes both `except` clauses and kills the loop with a traceback — during exactly the event this action exists to wait through, and with ~7 minutes of budget left unused.

```diff
-            except urllib.error.URLError as exc:
-                last_error = f"health unreachable: {exc}"
-                print(f"waiting: {last_error}", flush=True)
-            except json.JSONDecodeError as exc:
-                last_error = f"health returned non-JSON: {exc}"
+            except (urllib.error.URLError, OSError, json.JSONDecodeError) as exc:
+                last_error = f"health check failed: {exc}"
                 print(f"waiting: {last_error}", flush=True)
```

(`SystemExit` is a `BaseException`, so the success path at line 80 still escapes cleanly.)

.github/actions/wait-for-relayer/action.yml
nikola0x0: `expected.startswith(got)` matches in both directions, so any truncated `got` — `"2"`, `"28"`, a half-populated `GIT_SHA` — satisfies the gate against an expected SHA that starts with it. That's the stale-replica run this action exists to prevent.

```diff
-            return bool(got) and (got == expected or expected.startswith(got) or got.startswith(expected))
+            return len(got) >= 7 and expected.startswith(got)
```

Still accepts a legitimately shortened SHA, rejects a stub.

.github/workflows/test-sdk.yml
nikola0x0: `needs: [sdk-checks, sdk-changed]` didn't survive the move (it's at `test.yml:220` on `dev`). `sdk-checks` now lives in a workflow this one can't declare a `needs:` on, so the Node 22/24 unit matrix no longer gates the live run.

Concretely: a commit that breaks remember/recall semantics but still typechecks passes `pnpm build:sdk`, and this job writes real memories through the full SEAL/Walrus pipeline on the shared bench account — `benchmark-production` on a push to `main` — before a single unit test has reported.

.github/workflows/test.yml
nikola0x0: `timeout-minutes` isn't honored on a step that `uses:` a composite action ([actions/runner#1979](https://github.com/actions/runner/issues/1979)). It worked on the old `run:` step; here it's silently ignored, so a hung Chromium/OS-deps download runs to the job's 25 minutes instead of failing fast at 8.

Same dead flag at `test.yml:125` and `test.yml:189`. Composite steps don't support it either, so it has to become a `timeout 480 pnpm --filter …` inside the action — or be dropped from all three, since right now it reads as a guard that doesn't exist.

.github/workflows/test-sdk.yml
nikola0x0: This up-to-8-minute wait runs inside the `bench-account-*` critical section declared at line 51, so each holder occupies the shared group doing nothing but polling.

With `cancel-in-progress: false`, GitHub cancels a *pending* run when a newer one queues. Two dev merges close together plus the Monday cron and one of the three suites gets dropped rather than run — losing the drift signal the schedule exists for. The wait belongs in a preceding job, outside the group. Same placement in `test-python-sdk.yml:141` and `benchmark-live.yml:106`.

packages/sdk/test/e2e/live.e2e.mjs
nikola0x0: Nothing writes to `E2E_NAMESPACE_ALT` any more — the deleted `remember honors a per-call namespace override` test was the only writer (`live.e2e.mjs:257` on `dev`). So this passes trivially whether isolation works or the namespace simply doesn't exist, which makes it a weak assertion rather than a wrong one.

Related: line 206 passes `undefined` for the namespace, so no live test exercises a per-call override at all now. An SDK regression that silently fell back to `this.namespace` would ship green.

packages/sdk/test/e2e/live.e2e.mjs
nikola0x0: `rememberBulkAndWait` and `analyzeAndWait` are mock-only after this. Bulk hits a distinct endpoint (`/api/remember/bulk`), and the deleted test asserted the SDK's client-side namespace bookkeeping precisely because the bulk status response carries no namespace — only a live relayer can falsify that. `mock.test.mjs` checks the SDK against a mock the SDK itself defines.

Reasonable trade for runtime, but worth calling out in the PR body so it's a decision rather than an accident.

packages/sdk/package.json
nikola0x0: `--test-concurrency` caps how many test *files* run in parallel, and `test/e2e/` has exactly one. Within a file, `node:test` already runs top-level tests and describe subtests serially, so the `concurrency: 1` at `live.e2e.mjs:168` is also the default.

Both are inert. Harmless in themselves, but they read as a guarantee against parallel live writes through the shared `BENCH_DELEGATE_KEY` that nothing actually enforces — add a second `*.e2e.mjs` and you get exactly that.

.github/workflows/test-sdk.yml
nikola0x0: Second copy of `packages/sdk`'s `test:e2e` script, and the two have already diverged — the script uses the glob `test/e2e/*.e2e.mjs`, this hardcodes `live.e2e.mjs`. Add a second e2e file and the workflow silently keeps running only the one.

`pnpm --filter @mysten-incubation/memwal test:e2e`, with the reporter flags moved into a `test:e2e:ci` script, keeps them in one place.

.github/workflows/test-sdk.yml
nikola0x0: No `pull_request` trigger, so this workflow gets zero validation on the PR that adds it, or on any PR that edits it. While `sdk-e2e` lived in `test.yml` a YAML or expression error at least surfaced when the file was parsed for the other jobs.

A typo in the `environment:` expression or a bad `uses:` path now first surfaces after merge to `dev`, against a live production-capable bench account.

.github/actions/setup-playwright/action.yml
nikola0x0: `cache-id` is the basename of `package` at all three call sites (`@memwal/chatbot`/`chatbot`, `researcher`/`researcher`, `@memwal/noter`/`noter`). Two required inputs that must agree can drift — and two callers passing the same `cache-id` with different packages silently share one cache entry, which is the race the doc comment warns about.

`cache_id=${package##*/}` inside the action drops the input.

.github/actions/setup-playwright/action.yml
nikola0x0: After this PR nothing writes an unprefixed `pw-<os>-<hash>` entry — chatbot was the only job doing so and it now writes `pw-chatbot-…`. So this fallback can only hit pre-PR caches, which GitHub evicts after 7 days without access.

From then on every cold start (a `pnpm-lock.yaml` bump) falls through both restore-keys to a full Chromium + OS-deps download in all three jobs, with no shared warm path — the opposite of what the fallback is for.

.github/workflows/test.yml
nikola0x0: `check-compatibility-contract.mjs`, `check-docs-code-sync.mjs` and `check-docs-freshness.mjs` are dependency-free scripts that finish in milliseconds. This provisions three runners with a full checkout + `setup-node` each, on every push and PR.

One job looping over the three — recording failures and exiting non-zero at the end — preserves the `fail-fast: false` behaviour at a third of the cost. The only loss is three check names collapsing into one.

.github/workflows/test-sdk.yml
nikola0x0: You flagged this as the one unverified path — *"Full SHA-wait still needs merge to dev."* Here is the answer without merging: it won't clear.

Your commit messages scope the problem to branches — "Feature branches cannot match github.sha on Railway" (48671a5), "Dispatch/schedule cannot match github.sha" (3a79faf). That's right, but there is a second filter underneath it. The Railway `relayer` service is configured with `watchPatterns: ["/services/server/**"]` and `rootDirectory: services/server/` in dev, staging and production. Even on a tracked branch, a push that changes nothing under `services/server/` is recorded as a deployment and marked **SKIPPED** — no build, no new SHA.

This workflow's `paths:` fires on `packages/sdk/**` and the CI files, so an SDK-only merge to `dev` never triggers a Railway build. `/health.build.commit` keeps reporting the previous SHA, `commit_matches()` stays false for the full 480s, and the step exits 1.

Observed today: the dev relayer sat on `28e8ba30` (merge of #790, the last `services/server` change, Aug 25) across 27 pushes. #803 merged at 15:20, touched `services/server`, and Railway served `6102c76c` seconds later. Recent dev deploy history is 11 SKIPPED to 1 SUCCESS.

So the gate clears only when a PR happens to touch both the SDK and the server — which will read as flaky rather than broken.

Worth noting why this couldn't surface in review either: `wait-for-relayer` is used only by the two e2e jobs, and both are excluded from PRs by design. Nothing executes this action until a merge.

Options: wait on `write_ready` only (the guarantee the dispatch path already accepts), or pass `expected-commit` only when the push actually touched `services/server/**`.

.github/workflows/benchmark-live.yml
nikola0x0: Same gate as `test-sdk.yml:101`, and it misfires more often here: `benchmark-live.yml` has no `paths:` filter, so it runs on **every** push to `dev`/`staging` — while Railway only rebuilds the relayer on `/services/server/**`.

So every docs-only, app-only or SDK-only merge — 27 in a row on `dev` between Aug 25 and today — burns 8 minutes polling for a SHA Railway marked SKIPPED, then fails a job that used to pass green.

.github/actions/wait-for-relayer/action.yml
nikola0x0: CPython only wraps `h.request()` in `URLError` — `getresponse()` and the body read in `json.load()` re-raise bare. I ran the four failure modes a replica swap actually produces against this exact code:

```
RST mid-body                 -> ConnectionResetError   ESCAPES
clean close mid-body         -> IncompleteRead         ESCAPES
read timeout after headers   -> TimeoutError           ESCAPES
close before response        -> RemoteDisconnected     ESCAPES
non-JSON body                -> JSONDecodeError        caught
```

Only the last one is handled. Any of the other four kills the poll loop with a traceback and fails the step immediately, with most of the 480s budget unused — during exactly the event this action exists to wait through.

Note `IncompleteRead` is an `http.client.HTTPException`, *not* an `OSError`, so catching `OSError` alone still misses it:

```diff
-        import json, os, sys, time, urllib.error, urllib.request
+        import http.client, json, os, sys, time, urllib.error, urllib.request
```
```diff
-            except urllib.error.URLError as exc:
-                last_error = f"health unreachable: {exc}"
-                print(f"waiting: {last_error}", flush=True)
-            except json.JSONDecodeError as exc:
-                last_error = f"health returned non-JSON: {exc}"
+            except (urllib.error.URLError, OSError, http.client.HTTPException,
+                    json.JSONDecodeError) as exc:
+                last_error = f"health check failed: {exc}"
                 print(f"waiting: {last_error}", flush=True)
```

Verified that tuple catches all five. `SystemExit` is a `BaseException`, so the success path at line 80 still escapes cleanly.

.github/workflows/test-sdk.yml
nikola0x0: `needs: [sdk-checks, sdk-changed]` didn't survive the move (it's at `test.yml:220` on `dev`), so the Node 22/24 unit matrix no longer gates the live run.

`test-python-sdk.yml` — the workflow this one is modelled on — keeps exactly that gate: its `e2e` job declares `needs: unit` (line 69). The JS workflow diverges from the sibling precisely here.

Concretely: a commit that breaks remember/recall semantics but still typechecks passes `pnpm build:sdk`, and this job writes real memories through the full SEAL/Walrus pipeline on the shared bench account — `benchmark-production` on a push to `main` — before a single unit test has reported.

Simplest fix is to move `sdk-checks` into this workflow, the way the Python one is self-contained.

.github/workflows/test-sdk.yml
nikola0x0: You flagged this as unverified — *"Full SHA-wait still needs merge to dev."* It won't clear.

Railway's `relayer` has `watchPatterns: ["/services/server/**"]` in all three envs. A push touching nothing there is recorded as a deployment and marked **SKIPPED** — no build, no new SHA. This workflow's `paths:` only fires on `packages/sdk/**` and the CI files, so an SDK-only merge never triggers a build: `commit_matches()` stays false for 480s, step exits 1.

Dev relayer sat on `28e8ba30` (#790, the last `services/server` change) across 27 pushes; #803 touched the server at 15:20 today and it served `6102c76c` seconds later. Recent history: 11 SKIPPED, 1 SUCCESS.

Your commits scope this to branches — "Feature branches cannot match github.sha" (48671a5). True, but the path filter applies on tracked branches too. Net effect: the gate clears only when a PR touches both SDK and server, so it'll read as flaky rather than broken.

Fix: wait on `write_ready` only, or pass `expected-commit` only when the push touched `services/server/**`.

.github/workflows/benchmark-live.yml
nikola0x0: Same gate, worse trigger: no `paths:` filter here, so this runs on every push to `dev`/`staging` — including the 27 consecutive non-server merges between Aug 25 and today. Each would poll 8 minutes and then fail a job that currently passes.

.github/actions/wait-for-relayer/action.yml
nikola0x0: Only `h.request()` is wrapped in `URLError`; `getresponse()` and the `json.load()` read re-raise bare. Ran the five ways this poll fails:

```
RST mid-body                ConnectionResetError   ESCAPES
clean close mid-body        IncompleteRead         ESCAPES
read timeout after headers  TimeoutError           ESCAPES
close before response       RemoteDisconnected     ESCAPES
non-JSON body               JSONDecodeError        caught
```

Any of the four kills the loop with a traceback and fails the step with most of the 480s unused. `IncompleteRead` is an `http.client.HTTPException`, not an `OSError`, so catching `OSError` alone still misses it.

```diff
-        import json, os, sys, time, urllib.error, urllib.request
+        import http.client, json, os, sys, time, urllib.error, urllib.request
```
```diff
-            except urllib.error.URLError as exc:
-                last_error = f"health unreachable: {exc}"
-                print(f"waiting: {last_error}", flush=True)
-            except json.JSONDecodeError as exc:
-                last_error = f"health returned non-JSON: {exc}"
+            except (urllib.error.URLError, OSError, http.client.HTTPException,
+                    json.JSONDecodeError) as exc:
+                last_error = f"health check failed: {exc}"
                 print(f"waiting: {last_error}", flush=True)
```

Verified that tuple catches all five; `SystemExit` still escapes cleanly for the success path.

.github/workflows/test-sdk.yml
nikola0x0: Dropped `needs: [sdk-checks, sdk-changed]` (was `test.yml:220`). `test-python-sdk.yml:69` keeps the equivalent `needs: unit` — this is the one place the two shapes diverge.

Effect: a commit that typechecks but breaks recall passes `pnpm build:sdk` and writes real memories through SEAL/Walrus on the shared bench account — `benchmark-production` on a push to `main` — before any unit test reports.

Fix: move `sdk-checks` into this workflow so it's self-contained like the Python one.

.github/workflows/test-sdk.yml
nikola0x0, line 105: This won't clear on an SDK-only push.

Railway's `relayer` has `watchPatterns: ["/services/server/**"]` in all three envs. A push touching nothing there is recorded as a deployment and marked **SKIPPED** — no build, no new SHA. This workflow's `paths:` only fires on `packages/sdk/**` and the CI files, so an SDK-only merge never triggers a build: `commit_matches()` stays false for 480s, step exits 1.

Dev relayer sat on `28e8ba30` (#790, the last `services/server` change) across 27 pushes; #803 touched the server at 15:20 today and it served `6102c76c` seconds later. Recent history: 11 SKIPPED, 1 SUCCESS.

This is separate from the feature-branch case already handled in 48671a5 — the path filter applies on tracked branches too. Net effect: the gate clears only when a push happens to touch both SDK and server, so it'll present as flaky rather than broken.

Fix: wait on `write_ready` only, or pass `expected-commit` only when the push touched `services/server/**`.

.github/workflows/benchmark-live.yml
nikola0x0, line 106: Same gate, worse trigger: no `paths:` filter here, so this runs on every push to `dev`/`staging` — including the 27 consecutive non-server merges between Aug 25 and today. Each would poll 8 minutes and then fail a job that currently passes.

.github/actions/wait-for-relayer/action.yml
nikola0x0, line 96: Only `h.request()` is wrapped in `URLError`; `getresponse()` and the `json.load()` read re-raise bare. Ran the five ways this poll fails:

```
RST mid-body                ConnectionResetError   ESCAPES
clean close mid-body        IncompleteRead         ESCAPES
read timeout after headers  TimeoutError           ESCAPES
close before response       RemoteDisconnected     ESCAPES
non-JSON body               JSONDecodeError        caught
```

Any of the four kills the loop with a traceback and fails the step with most of the 480s unused. `IncompleteRead` is an `http.client.HTTPException`, not an `OSError`, so catching `OSError` alone still misses it.

```diff
-        import json, os, sys, time, urllib.error, urllib.request
+        import http.client, json, os, sys, time, urllib.error, urllib.request
```
```diff
-            except urllib.error.URLError as exc:
-                last_error = f"health unreachable: {exc}"
-                print(f"waiting: {last_error}", flush=True)
-            except json.JSONDecodeError as exc:
-                last_error = f"health returned non-JSON: {exc}"
+            except (urllib.error.URLError, OSError, http.client.HTTPException,
+                    json.JSONDecodeError) as exc:
+                last_error = f"health check failed: {exc}"
                 print(f"waiting: {last_error}", flush=True)
```

Verified that tuple catches all five; `SystemExit` still escapes cleanly for the success path.

.github/workflows/test-sdk.yml
nikola0x0, line 47: Dropped `needs: [sdk-checks, sdk-changed]` (was `test.yml:220`). `test-python-sdk.yml:69` keeps the equivalent `needs: unit` — this is the one place the two shapes diverge.

Effect: a commit that typechecks but breaks recall passes `pnpm build:sdk` and writes real memories through SEAL/Walrus on the shared bench account — `benchmark-production` on a push to `main` — before any unit test reports.

Not suggesting you move `sdk-checks` here — "unit tests stay in test.yml so they run on every PR" is right, and this file's `paths:` would narrow them. Cheapest restore is to make the build step run them too:

```diff
-        run: pnpm build:sdk
+        run: pnpm build:sdk && pnpm --filter @mysten-incubation/memwal test
```

nikola0x0 (commented):
The SHA wait won't clear on an SDK-only push, so both live suites fail on merge. Nothing else blocks.

**Blocking**

Railway's `relayer` is watch-path gated on `/services/server/**`. A push touching nothing there is marked SKIPPED — no build, no new SHA — so `commit_matches()` never clears. Separate from the feature-branch case already handled — the path filter applies on tracked branches too. Hits `test-sdk.yml:101` on any SDK-only merge, and `benchmark-live.yml:106` on every push to `dev`/`staging`.

**Majors (details inline)**

* `wait-for-relayer/action.yml:69` — 4 of the 5 ways a `/health` poll fails escape both `except` clauses. The loop dies on the first blip of the replica swap it's waiting for.
* `test-sdk.yml:44` — dropped `needs: [sdk-checks]`; `test-python-sdk.yml` keeps the equivalent. Live writes no longer wait on unit tests.

<details>
<summary><b>What I verified</b></summary>

* Railway API: `watchPatterns: ["/services/server/**"]`, `rootDirectory: services/server/`, all three envs. Recent dev deploys: 11 SKIPPED, 1 SUCCESS.
* Dev relayer sat on `28e8ba30` (#790, last `services/server` change) across 27 pushes; #803 touched the server at 15:20 today and it moved to `6102c76c` in seconds.
* `test-python-sdk.yml` keeps `needs: unit` (line 69) and a `pull_request:` trigger with `e2e` excluded by `if:` (lines 4-8, 41-44).
* Ran the relayer poller against four real socket failures — results inline on `action.yml:69`.
* Checked `timeout-minutes` on the new composite `uses:` steps before flagging it: the runner *does* honor it (`ActionRunner.Timeout`, then `CreateEmbeddedChild(timeout: GetRemainingTimeout())`). Dropped that finding — the unsupported case is `timeout-minutes` *inside* a composite action.
* Split itself is good: the three composite actions remove real duplication and `test.yml` is far easier to read at −488.

</details>

<details>
<summary><b>Minor / info (9)</b></summary>

* `test-sdk.yml:11` — no `pull_request` trigger, so neither this file nor `wait-for-relayer` runs until a merge. Mirroring Python's `if:`-gated shape buys YAML validation only — it wouldn't have caught the gate above. Real coverage needs actionlint, which the repo lacks.
* `wait-for-relayer/action.yml:60` — `expected.startswith(got)` matches both ways, so a truncated commit passes. `len(got) >= 7 and expected.startswith(got)`.
* `test-sdk.yml:97` — the wait sits inside the `bench-account-*` group, holding the shared lock for up to 8 min of polling. Move it to a preceding job.
* `test-sdk.yml:107` — duplicates `packages/sdk`'s `test:e2e` and has already diverged (hardcoded filename vs. glob). Call the script; put reporter flags in `test:e2e:ci`.
* `live.e2e.mjs:228` — nothing writes to `E2E_NAMESPACE_ALT` now, so the isolation assert passes either way. Line 206 also dropped the last per-call namespace override.
* `live.e2e.mjs:168` — `rememberBulkAndWait` / `analyzeAndWait` are mock-only. Fine per "method-level coverage stays in unit tests" — worth stating, since bulk has its own endpoint.
* `packages/sdk/package.json:31` — `--test-concurrency=1` caps parallel *files* and there's one; the describe's `concurrency: 1` is already default. Both inert.
* `setup-playwright/action.yml:8` — `cache-id` is the `package` basename everywhere; `cache_id=${package##*/}` drops an input that can drift.
* `setup-playwright/action.yml:22` — nothing writes an unprefixed `pw-<os>-` entry after this PR, so that restore-key dies with the old caches. Then every lockfile bump is a full Chromium download ×3.

</details>

.github/workflows/test-sdk.yml
ducnmm: Addressed in 2c3f3d3d.

`wait-for-relayer` now takes `since-commit` (`github.event.before`). If that range has no `services/server/**` paths, the SHA gate is dropped and the step waits on `status=ok` + `write_ready` only — Railway SKIPPED those deploys, so `commit_matches()` can never clear.

Wired on this job and on the Python live suite (same action, same trap).

.github/workflows/benchmark-live.yml
ducnmm: Addressed in 2c3f3d3d.

Same `since-commit` path in `wait-for-relayer`: a `dev`/`staging` push that does not touch `services/server/**` no longer polls 8 minutes for a SHA Railway will never serve.

.github/actions/wait-for-relayer/action.yml
ducnmm: Addressed in 2c3f3d3d.

Poller now catches `(URLError, OSError, HTTPException, JSONDecodeError)` as you listed, so RST / IncompleteRead / TimeoutError / RemoteDisconnected keep looping instead of killing the step. `SystemExit` still escapes for the success path.

.github/workflows/test-sdk.yml
ducnmm: Addressed in 2c3f3d3d.

Build step is now `pnpm build:sdk && pnpm --filter @mysten-incubation/memwal test`. Unit tests stay in `test.yml` for every PR; this job runs them too so a typecheck-only green cannot write through the bench account.

nikola0x0 (approved):
LGTM!

PR #805, opened 2026-08-27
feat(mcp): identify coding agent on MCP sessions and log it

ducnmm:
Closes WALM-373.

## What

The relayer sidecar can tell which coding agent an MCP session belongs to, and logs it.

1. Stdio `memwal-mcp` reads `initialize.clientInfo` and forwards `x-memwal-client` / `x-memwal-client-version` (the proxy already forwards `x-memwal-*`).
2. Sidecar maps the name to a stable `agentClient` (`claude-code`, `codex`, `cursor`, …; unknown stays `other` with the raw `clientName`).
3. JSON stderr logs: `session.identified`, `tool.call`, `session.closed`.

View in **Railway → Relayer** after this sidecar is deployed. Search `session.identified` or `agentClient":"claude-code"`.

## Not in this PR

No install CLI / marketplace / per-client plugin pack. Local `memwal-mcp` stderr is only on the user's machine.

## Test

- `packages/mcp`: 60/60
- sidecar MCP scripts: 233/233 (includes handshake stamp for `claude-code`)

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -125,10 +125,7 @@ export async function main(argv: string[] = process.argv.slice(2)): Promise<void
         args.relayerUrl ?? process.env.MEMWAL_SERVER_URL ?? "https://relayer.memory.walrus.xyz";
     const webUrl =
         args.webUrl ?? process.env.MEMWAL_WEB_URL ?? "https://memory.walrus.xyz";
-    // Label is the on-chain delegate-key name shown in the dashboard.
-    // Default is generic; the user can rename anytime. Which coding agent
-    // opened the MCP session is detected from `initialize.clientInfo` and
-    // logged on the sidecar — it is not this login-time label.
+    // On-chain delegate-key display name; users can rename it in the dashboard.
     const label = args.label ?? process.env.MEMWAL_CLIENT_LABEL ?? "MCP Client";
     // Default memory namespace applied to memory tool calls when the agent
     // omits one. CLI > env, then UNSET — we deliberately do NOT hardcode a
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -125,11 +125,10 @@ export async function main(argv: string[] = process.argv.slice(2)): Promise<void
         args.relayerUrl ?? process.env.MEMWAL_SERVER_URL ?? "https://relayer.memory.walrus.xyz";
     const webUrl =
         args.webUrl ?? process.env.MEMWAL_WEB_URL ?? "https://memory.walrus.xyz";
-    // Label is the on-chain delegate-key name shown in the dashboard. We
-    // default to a generic value at registration time — the bridge updates
-    // it to the actual client's `clientInfo.name` ("Cursor", "Claude",
-    // "Antigravity", ...) after the first MCP `initialize` request. User
-    // can rename anytime from the dashboard.
+    // Label is the on-chain delegate-key name shown in the dashboard.
+    // Default is generic; the user can rename anytime. Which coding agent
+    // opened the MCP session is detected from `initialize.clientInfo` and
+    // logged on the sidecar — it is not this login-time label.
     const label = args.label ?? process.env.MEMWAL_CLIENT_LABEL ?? "MCP Client";
     // Default memory namespace applied to memory tool calls when the agent
     // omits one. CLI > env, then UNSET — we deliberately do NOT hardcode a
--- a/services/server/scripts/mcp/index.ts
+++ b/services/server/scripts/mcp/index.ts
@@ -21,11 +21,18 @@ import { randomUUID } from "node:crypto";
 import { SSEServerTransport } from "@modelcontextprotocol/sdk/server/sse.js";
 import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
 
-import type { AuthResolution } from "./auth.js";
+import type { AuthResolution, MemWalSession } from "./auth.js";
 import { McpAuthError, resolveAuth } from "./auth.js";
+import {
+    applyAgentClient,
+    applyAgentClientFromServer,
+    CLIENT_NAME_HEADER,
+    CLIENT_VERSION_HEADER,
+} from "./agent-client.js";
 import { createLogger } from "./logger.js";
 import { createMcpServer } from "./server.js";
 import { McpRateLimiter, clientIpFromRequest } from "./rateLimit.js";
+import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 
 const log = createLogger("mcp");
 
@@ -70,6 +77,8 @@ function rateLimitDeny(
 
 interface McpConnection {
     sessionKey: string;
+    session: MemWalSession;
+    server: McpServer;
     transport: SSEServerTransport;
     cleanup: () => void;
 }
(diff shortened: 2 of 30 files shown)


services/server/scripts/mcp/index.ts
```diff
@@ -415,6 +475,7 @@ async function handleStreamableHttp(
                 accountId: auth.session.accountId,
                 delegatePubKey: auth.session.delegatePubKeyHex,
                 transportId: newId,
+                agentClient: auth.session.agentClient ?? null,
```
ducnmm: **[suggestion]** Streamable `session.opened` always logs `agentClient: null`. `onsessioninitialized` runs when the SDK assigns a session id, which is before `_oninitialize` stores `clientInfo` and before `identifyConnection` runs after `handleRequest`. Operators grepping `session.opened` will think identification failed even when the same POST carries `clientInfo` and `session.identified` fires moments later. SSE is inconsistent with this: it stamps from headers *before* `session.opened`.

**Suggestion:** Call `applyAgentClient(auth.session, agentClientFromHeaders(req), …)` inside `onsessioninitialized` before logging (mirrors the SSE GET path), and/or omit `agentClient` from this `session.opened` line so `session.identified` is the event that means identity is known.

packages/mcp/src/index.ts
```diff
@@ -125,11 +125,10 @@ export async function main(argv: string[] = process.argv.slice(2)): Promise<void
...
     const webUrl =
         args.webUrl ?? process.env.MEMWAL_WEB_URL ?? "https://memory.walrus.xyz";
-    // Label is the on-chain delegate-key name shown in the dashboard. We
-    // default to a generic value at registration time — the bridge updates
-    // it to the actual client's `clientInfo.name` ("Cursor", "Claude",
-    // "Antigravity", ...) after the first MCP `initialize` request. User
-    // can rename anytime from the dashboard.
+    // Label is the on-chain delegate-key name shown in the dashboard.
+    // Default is generic; the user can rename anytime. Which coding agent
+    // opened the MCP session is detected from `initialize.clientInfo` and
```
ducnmm: **[suggestion]** The replacement comment narrates the WALM-373 design choice (agent identity is detected from `initialize.clientInfo` and logged on the sidecar) instead of explaining the label itself. That is architecture history sitting on a one-line default.

**Suggestion:** Keep a short why for the label, e.g. `// On-chain delegate-key display name; users can rename it in the dashboard.`

ducnmm:
@harrymove-ctrl please review this PR (and the Linear ticket [WALM-373](https://linear.app/mysten-labs/issue/WALM-373/add-agent-client-claude-code-codex-to-mcp)).

Sidecar identifies the coding agent from MCP `initialize.clientInfo` and logs `session.identified` / `tool.call` / `session.closed`. Manual package bump to 0.0.12 (no changeset).

harrymove-ctrl (changes requested):
@ducnmm Reviewed. Design is sound and CI is fully green (`scripts/verify-manual-sdk-release.mjs` confirms `MCP package 0.0.12: manifests and changelogs synchronized`, so the 11-file manual bump is complete). Three independent identification paths funnelling through one idempotent `applyAgentClient` is the right shape, client strings are length-capped and control-char-stripped before they reach a header or a log line, and no tool arguments are logged. `mcp_proxy.rs` correctly needed only a test, since `x-memwal-` is already a forward prefix and `x-memwal-internal-` is denied first.

Two changes requested before merge.

---

### 1. `x-memwal-client` almost never reaches the first SSE connect, so `session.opened` will log `agentClient: null`

`packages/mcp/src/bridge.ts`

`connectInBackground` opens the SSE stream *before* stdin is wired. That is deliberate and correct (it is the cold-start fix), but it means `extraHeaders`, seeded at line 751 from `lastClientInfoHeaders()`, is still empty at connect time in the ordinary signed-in path. `lastClientInfo` is only populated when the auth-required handoff ran, so on a normal already-signed-in start the GET carries no `x-memwal-client`. The `Object.assign(extraHeaders, ...)` on `initialize` therefore only takes effect on reconnects.

The feature still works end to end, because the sidecar recovers the name from `server.server.oninitialized` and the POST-side `identifyConnection`. But the PR description reads as though the header is the mechanism, and in practice the SDK handshake is. Concretely this means the `agentClient` field newly added to the `session.opened` log lines in `services/server/scripts/mcp/index.ts` (SSE at ~line 279, streamable at ~line 478) will be `null` for the large majority of sessions, which is misleading for anyone grepping Railway.

Please either:

* drop `agentClient` from `session.opened` and rely on `session.identified` as the single place the agent name appears, or
* keep it and add a comment at the `extraHeaders` declaration stating that the header is a best-effort hint that only lands after the auth-required handoff or on reconnect, and that the handshake is the authoritative source.

Either is fine, but the current state advertises a signal that is usually empty.

### 2. `clientVersion` can be permanently lost for a session

`services/server/scripts/mcp/agent-client.ts`, `applyAgentClient`

```ts
if (session.agentClient === id && session.clientName === name) return false;
```

The guard ignores `version`. The realistic sequence:

1. SSE GET carries `x-memwal-client: claude-code` with no `x-memwal-client-version` (per finding 1 this is the common case once the header does land), so `clientVersion` is set to `undefined`.
2. The SDK handshake then reports `claude-code` **with** a real version.
3. Name and id match, the guard returns `false`, and `clientVersion` stays `undefined` for the life of the session.

Your own test asserts the idempotent case with a matching version, so it does not catch this. Please either include `version` in the equality check, or assign `session.clientVersion` unconditionally before the guard and keep the guard only for suppressing the duplicate log line. A test covering "header without version, then handshake with version" would pin it.

---

### Non-blocking

3. **Two divergent `sanitizeClientToken` copies.** `packages/mcp/src/client-info.ts` keeps only `\x20-\x7e`, which is correct because it produces an HTTP header value. The sidecar copy in `agent-client.ts` strips only C0 and DEL, so non-ASCII survives into logs. Functionally fine (JSON encodes it), but these two near-identical files will drift. A short comment in each pointing at the other would be enough; sharing code across the package boundary is probably not worth it.

4. **`let server!: McpServer` read from a closure** (`services/server/scripts/mcp/index.ts:444`). Correct today, and the comment explains why, but the definite-assignment assertion means a future reordering becomes a runtime TDZ `ReferenceError` inside `onsessioninitialized` instead of a compile error. A one-field holder object would make it structurally safe.

5. **`identifyConnection` runs after `await transport.handleRequest(req, res)`.** The response is already flushed so nothing blocks, but on a request that both initializes and calls a tool, `tool.call` is logged before `session.identified`. Ordering only, mentioning it so it is not a surprise in Railway.

6. **`identifyAgentClient` returns a bare `string`.** A union type (`AgentClientId`) would keep the returned values, the branch list, and the test table honest against each other.

7. **`claude.ai` mapping to `claude-desktop`** is a stretch, though harmless since unknown names fall through to `other` while retaining the raw `clientName`.

ducnmm:
@harrymove-ctrl addressed in c7e20af9:

1. **`session.opened` / `x-memwal-client` timing.** Dropped `agentClient` from both `session.opened` lines so `session.identified` is the event that means identity is known. Also documented on `extraHeaders` that the header is a best-effort hint (auth-required handoff or reconnect) and the handshake is authoritative.

2. **`clientVersion` lost after a header stamp without one.** `applyAgentClient` now fills in version when a later call supplies it, and a later call that omits version does not wipe one we already stored. Tests cover header-then-handshake and the wipe guard.

Non-blocking:

3. Cross-comments on the two `sanitizeClientToken` copies (header ASCII vs log C0/DEL).
4. Replaced `let server!: McpServer` with a holder object so a reorder logs instead of throwing TDZ.
5. Left handshake identification after `handleRequest` (SDK only exposes `clientInfo` once initialize has been processed); headers are stamped first, and a comment notes that a combined initialize+tool request can still emit `tool.call` before `session.identified`.
6. `identifyAgentClient` now returns `AgentClientId`; `AGENT_CLIENT_IDS` and the mapping table are pinned to each other.
7. Left `claude.ai` → `claude-desktop` with a comment that this is Claude Desktop's web `clientInfo`; unknown names still fall through to `other`.

harrymove-ctrl (approved):

PR #806, opened 2026-08-27
fix(mcp): write credentials through a fresh 0600 inode instead of chmod-after-write

### LINKED ISSUE #520: security(mcp): existing credential file retains broad permissions while new secret content is written before `chmodSync` [Security / Low]
`saveCreds` writes new credential content directly to the final path and only then calls `chmodSync(0600)`. Node's `writeFileSync(..., { mode: 0o600 })` applies the mode when creating a file; it does not replace the permissions of an existing file. If `credentials.json` already exists with broader permissions, the new delegate private key is visible under those old permissions for the interval between the write and the chmod.

## Affected code

`packages/mcp/src/auth.ts:58-67`:

```ts
/** Write credentials with secure (`0600`) permission. */
export function saveCreds(creds: MemWalCredentials): void {
    mkdirSync(dirname(CREDS_PATH), { recursive: true, mode: 0o700 });
    writeFileSync(CREDS_PATH, JSON.stringify(creds, null, 2), {
        encoding: "utf8",
        mode: 0o600,
    });
    // writeFileSync's `mode` argument is only honored on file creation; ensure
    // the permission on an existing file matches.
    try {
        chmodSync(CREDS_PATH, 0o600);
    } catch {
        /* Windows etc. — best effort */
    }
}
```

The file contains `delegatePrivateKey` in plaintext.

## Preconditions and practical scope

The destination file must already exist with permissions broader than `0600`, and a local reader must win the short write-to-chmod race. The containing `~/.memwal` directory is created with `0700`, which materially limits exposure: a different Unix user cannot normally traverse that directory unless the directory permissions were also weakened or the reader already has relevant local privileges. The confirmed primitive is therefore narrow and local; it is not a remote vulnerability.

## Reproduction

```bash
node evidence/poc_credential_file_toctou.mjs
```

The harness creates a pre-existing file with mode `0644`, reproduces the two operations from `saveCreds`, and calls `statSync` between them:

```text
Precondition: pre-existing file mode = 644
Mode immediately after writeFileSync (new secret content is now on disk): 644
Mode after the subsequent chmodSync: 600
```

The positive control is the final `0600` result, which confirms the harness observes real mode changes. The target observation is the intermediate `0644`, after the new secret has already been written.

## Impact

A local process able to read the path during the race can obtain the freshly written delegate private key. Exploitability requires a pre-existing permissive file and local access sufficient to observe the path despite the directory protection. Severity is Low because of those prerequisites and the small race window.

## Suggested fix

Write to a newly created temp file with mode `0600`, then atomically rename it over the destination. The new secret is never present at the final path under the old inode's permissions:

```diff
-import { mkdirSync, readFileSync, writeFileSync, chmodSync, unlinkSync, existsSync } from "node:fs";
+import { mkdirSync, readFileSync, writeFileSync, chmodSync, unlinkSync, existsSync, renameSync } from "node:fs";

 export function saveCreds(creds: MemWalCredentials): void {
-    mkdirSync(dirname(CREDS_PATH), { recursive: true, mode: 0o700 });
-    writeFileSync(CREDS_PATH, JSON.stringify(creds, null, 2), {
+    const dir = dirname(CREDS_PATH);
+    mkdirSync(dir, { recursive: true, mode: 0o700 });
+    const tmpPath = `${CREDS_PATH}.${process.pid}.${Date.now()}.tmp`;
+    writeFileSync(tmpPath, JSON.stringify(creds, null, 2), {
         encoding: "utf8",
         mode: 0o600,
     });
     try {
-        chmodSync(CREDS_PATH, 0o600);
+        chmodSync(tmpPath, 0o600);
     } catch {
         /* Windows etc. — best effort */
     }
+    renameSync(tmpPath, CREDS_PATH);
 }
```

Production hardening should also clean up the temp path if writing, chmod, or rename fails and should ensure concurrent writers cannot leave stale temp files.

## Regression test

```bash
node evidence/fix_test_credential_file_toctou.mjs
```

Observed result:

```text
Precondition: pre-existing file mode = 644
Final mode after fixed saveCreds(): 600
```

The file that becomes visible at `CREDS_PATH` is the fresh temp inode, already set to `0600`; the old broad-permission inode is replaced atomically.

This is a revalidation of legacy finding F-004 against the current target, with a new dynamic PoC and a dynamically checked fix.

**Environment:** `MystenLabs/MemWal`, PoC baseline `dev` at `a67e9f4a633a9431352aaeb7eb606b30c95c70ef`; revalidated unchanged on current `dev` `e7fd5c5a0d1849a65a93786f05baa98aed34542a`.

nikola0x0:
Closes #520. Linear: WALM-312.

## The bug

`saveCreds` wrote the plaintext delegate private key to the final path and only then called `chmodSync(0600)`:

```ts
writeFileSync(path, JSON.stringify(creds, null, 2), { encoding: "utf8", mode: 0o600 });
// writeFileSync's `mode` argument is only honored on file creation; ensure
// the permission on an existing file matches.
try { chmodSync(path, 0o600); } catch { /* Windows etc. — best effort */ }
```

`writeFileSync`'s `mode` follows POSIX `open()` — the kernel applies it when it creates the inode and ignores it for one that already exists. A `credentials.json` left world-readable by anything outside this code (a manual `chmod`, a restored backup, another tool) therefore received the new key under the *old* permission, with a second, non-atomic syscall to tighten it afterwards. Anyone reading the path in between gets the key.

## The fix

Both writes go through a new `writeSecretFile()`, which creates a randomly named temp file in the target directory with `{ mode: 0o600, flag: "wx" }` and renames it into place:

- `O_EXCL` (`wx`) means the mode is always honored on creation, and a temp path that already exists — an interrupted earlier run, or a file planted by someone else — fails hard rather than being written through.
- `rename(2)` repoints the name atomically, so a reader sees either the whole old file or the whole new one, never a permissive inode holding a fresh secret. This removes the window rather than shortening it.
- The temp file is unlinked on any failure, so a failed write never leaves the secret behind.

**Also fixed: the account-switch backup.** `backupIfReplacingAnotherAccount` did `copyFileSync` + `chmodSync` — the same bug class, and it needs no precondition at all. Every account switch created a second plaintext copy of the same delegate key under the process umask (typically `0644`) before tightening it. It now uses the same writer. This is beyond the literal scope of #520, but it is the same defect in the same function, three lines away.

## Test

`packages/mcp/test/credential-file-permissions.test.mjs`. The race cannot be asserted by watching for it, so the test asserts the property that closes it: a reader holding the pre-existing `0644` inode open across the save must never see the new key — the attacker in the report is what makes the observation deterministic.

Against the previous code it fails on exactly that assertion:

```
✖ saveCreds never writes the new secret through a pre-existing permissive file
  AssertionError: the new delegate private key must never be readable through the pre-existing 0644 inode
```

Full package suite: **61/61 pass**.

One caveat worth recording: the backup assertion (final mode `0600`, correct content) passes both before and after. A fresh timestamped inode that nobody holds a handle to is not deterministically observable, so it is a guard, not a regression proof — the atomicity property is proven by the `saveCreds` test, and the backup shares the code path.

## Note on the issue text

The issue's line references (`auth.ts:59-70`, `CREDS_PATH`) predate the project-local credential resolution. The target is now `credsPath()`, which can resolve to a `.memwal` inside a project directory — so the issue's "`~/.memwal` is `0700` and limits exposure" caveat is weaker than stated for project-local files.


jessiemongeon1:
## Style Guide Audit

packages/mcp/src/auth.ts
```diff
@@ -142,16 +142,49 @@ export function loadCreds(): MemWalCredentials | null {
...
+    // Named off the target and randomised, so concurrent saves cannot collide
+    // on it and no one can guess it ahead of time. Dot-prefixed to keep a
+    // crashed run's leftovers out of the way of directory listings.
+    const tmp = join(dir, `.${basename(path)}.${process.pid}.${randomUUID()}.tmp`);
     try {
-        chmodSync(path, 0o600);
-    } catch {
-        /* Windows etc. — best effort */
+        writeFileSync(tmp, contents, { encoding: "utf8", mode: 0o600, flag: "wx" });
+        renameSync(tmp, path);
```
ducnmm: **[suggestion]** `renameSync(tmp, path)` is the POSIX-correct replace, but on Windows Node implements it with `MoveFileEx(MOVEFILE_REPLACE_EXISTING)`, which returns `EPERM`/`EACCES` when another handle holds the destination. The previous `writeFileSync` to the final path succeeded while the file was open for read; `login.ts` treats a thrown `saveCreds` as HTTP 500, so a transient lock becomes a failed sign-in. The new tests also assert Unix `mode` bits that Windows does not honor.

**Suggestion:** On `win32`, retry `renameSync` on `EPERM`/`EACCES`/`EBUSY`, or fall back to in-place write. Skip or relax the mode and open-fd assertions when `process.platform === "win32"`.

packages/mcp/test/credential-file-permissions.test.mjs
```diff
@@ -0,0 +1,168 @@
...
+    const { auth, path } = await sandbox(t);
+
+    auth.saveCreds(makeCreds(ACCOUNT, NEW_KEY));
+
+    assert.equal(modeOf(path), 0o600);
+    assert.equal(modeOf(dirname(path)), 0o700, "the containing directory stays owner-only");
+});
+
+test("the backup of a displaced account is written at 0600", async (t) => {
+    const { auth, home } = await sandbox(t, { existingFileMode: 0o600 });
```
ducnmm: **[nit]** `const { auth, home }` destructures `home` but never uses it.

**Suggestion:** Change the binding to `const { auth } = await sandbox(...)`.

ducnmm (commented):
## Summary

This correctly closes GH #520 on POSIX: `writeSecretFile` creates a fresh `0600` inode with `O_EXCL` (`wx`) and `rename`s it over the destination, so a new delegate key is never written through a pre-existing permissive file. Routing the account-switch backup through the same helper is the right extra fix. Residual gaps are Windows: `renameSync` cannot replace a locked destination, and the new tests assume Unix `mode` bits.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

packages/mcp/src/auth.ts
nikola0x0: Fixed in bf51210. On `win32`, `renameSync` retries on EPERM/EACCES/EBUSY (5 attempts, linear backoff) and then falls back to an in-place write rather than failing the sign-in. POSIX keeps the plain `renameSync` with no retry and no fallback, since there the mode *is* the protection and `rename(2)` replaces a destination regardless of who holds it open.

The fallback gives up atomicity but not the guarantee this function exists for: Windows does not enforce POSIX mode bits at all, so `0600` was never doing that work there, and NTFS ACLs are inherited from the directory either way.

Tests whose premise is the mode bit now skip on win32; the backup test keeps its content assertions on every platform. Worth flagging that CI has no Windows runner, so the retry path is reasoned rather than exercised.

packages/mcp/test/credential-file-permissions.test.mjs
nikola0x0: Fixed.

packages/mcp/src/auth.ts
```diff
@@ -142,16 +142,103 @@ export function loadCreds(): MemWalCredentials | null {
...
+            return;
+        } catch (err) {
+            const code = (err as NodeJS.ErrnoException).code ?? "";
+            if (!WIN32_LOCKED_CODES.has(code)) throw err;
+            if (attempt < WIN32_RENAME_ATTEMPTS) {
+                sleepSync(WIN32_RENAME_BACKOFF_MS * attempt);
+                continue;
+            }
+            // Still locked. Write through the existing handle's inode instead
+            // of failing the sign-in. `writeSecretFile`'s catch removes `tmp`.
```
ducnmm: **[bug]** After five locked `renameSync` attempts, the Windows fallback writes in place and `return`s. That is a successful `replaceWithTemp`, so `writeSecretFile`'s `catch` does not run. The comment (`writeSecretFile`'s catch removes `tmp`) is false: `tmp` still holds the plaintext delegate key.

This is the path the commit is for. Old `writeFileSync` to the final path never created a sibling file. Each locked save can accumulate another `.credentials.json.<pid>.<uuid>.tmp`.

**Suggestion:** After the in-place `writeFileSync` succeeds, best-effort `unlinkSync(tmp)` (swallow unlink errors so a leftover temp cannot fail a completed save). Do not rely on the outer `catch` — that path rethrows.

ducnmm (dismissed):
## Summary

POSIX still creates a `wx`+`0600` temp and `renameSync`s it over the destination. Windows retries `renameSync` on lock then writes in place so a transient lock does not 500 login. The locked-destination success path does not unlink the temp, so a second plaintext copy of the key remains as `.credentials.json.<pid>.<uuid>.tmp`.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

packages/mcp/src/auth.ts
nikola0x0: Confirmed and fixed in ce4f66f. The fallback returns successfully, so the outer catch never runs, and my comment claiming it did was simply wrong. Unlinked best-effort right after the in-place write lands, so a temp that cannot be removed does not fail a save that already succeeded.

Also took the point that this branch had no test behind it. `replaceWithTemp` now takes an injectable platform / rename / sleep and is exported, so the Windows path is exercisable without a Windows runner — the same pattern `recovery.ts` uses for `canonicalRequestMessage`. Seven cases: each of EPERM/EACCES/EBUSY lands and leaves no temp, three consecutive locked saves leave none either, a lock that clears mid-retry renames instead of falling back, a non-lock code still propagates, and POSIX neither retries nor falls back.

Verified they fail against the leaking version — the four temp-file assertions go red, the other three stay green.

packages/mcp/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
...
+- Write the credentials file by creating a new `0600` file and renaming it into place, instead of writing the delegate private key to the existing path and tightening the permission afterwards. `writeFileSync`'s `mode` only applies when it creates the file, so a `credentials.json` left world-readable by anything outside the CLI (a manual `chmod`, a restored backup, another tool) received the new key under the old permission until the following `chmod` landed. The account backup taken when signing in as a different account is written the same way; it holds the same plaintext key and was previously created under the process umask. On Windows, where `rename` cannot replace a destination another process holds open and POSIX mode bits are not enforced at all, the CLI retries briefly and then writes in place rather than failing the sign-in. (#520)
```
ducnmm: **[bug]** origin/dev and origin/main both publish MCP 0.0.12. Equal versions mean this PR must patch-bump MCP to 0.0.13: `packages/mcp/package.json`, `scripts/verify-manual-sdk-release.mjs`, and the six plugin/marketplace JSON files the verify script tracks. Move this #520 note (only) under `## 0.0.13` in both `packages/mcp/CHANGELOG.md` and `docs/mcp/changelog.mdx`, including the mdx `answer:`. Rebase onto origin/dev first so 0.0.12 is not lost. Do not keep `## Unreleased`. Do not re-list #705. Leave TS 0.1.6, Python 0.1.9, and OpenClaw 0.0.6 alone.

ducnmm (dismissed):
## Summary

Windows leftover-temp on the locked-destination fallback is unlinked. POSIX `wx`+`0600`+rename still closes GH #520.

Dump is still wrong: origin/dev and origin/main both publish MCP 0.0.12, so this bugfix must ship as 0.0.13. Head is a stale 0.0.11 branch with `## Unreleased` (and still lists #705, which already shipped in 0.0.12).

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

packages/mcp/CHANGELOG.md
nikola0x0: Rebased onto origin/dev. The bump is already there: dev bumped MCP to 0.0.13 itself in 7bca3703 (COMG-719/WALM-480), about 30 minutes after this comment, so `packages/mcp/package.json`, the six plugin/marketplace JSON files and `verify-manual-sdk-release.mjs` all arrive at 0.0.13 with the rebase, and the script passes. npm `latest` is 0.0.12 and `dev` is 0.0.13-dev.0, so 0.0.13 is open rather than published; bumping to 0.0.14 would have opened a second unreleased version alongside it.

The rebase dropped `## Unreleased` and the duplicate #705 entry on its own, but it landed the #520 note at the end of the published 0.0.12 list. 6af5d2d5 moves it to 0.0.13 in both `packages/mcp/CHANGELOG.md` and `docs/mcp/changelog.mdx`, and updates the mdx release line and `answer:`. TS 0.1.7, Python 0.1.10 and OpenClaw 0.0.6 untouched. Package suite 72/72.

ducnmm (approved):
## Summary

Final look on `e8dc7dd4`. POSIX still creates a `wx`+`0600` temp and `renameSync`s it over the destination, so a new delegate key is never written through a pre-existing permissive file (GH #520). The Windows locked-destination fallback unlinks the leftover temp. Dump is under unpublished `## 0.0.13` in both changelogs (mdx `answer:` included); no 0.0.14 bump. No remaining issues on this head.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

Issue #412, opened 2026-07-13
##  Missing feature: can't easily confirm a save succeeded or get its blob ID


triumphsta:
**What happens**
When you save something, it doesn't finish right away — it starts as "pending" and finishes later in the background. The only way to get the final result (a real blob ID) is to use a separate function called `wait_for_remember_jobs()`. But if you're using the easy setup (`with_memwal_openai`), this function isn't available to you at all — it's hidden inside the wrapper.

**What would help**
Let developers using `with_memwal_openai` also get access to the job ID or a way to check "did this save finish yet?" — right now the only way to do this is to skip the easy setup entirely and use the lower-level tools directly.

**Why it matters**
If you need proof that something was saved (for example, for a hackathon submission, or just for debugging), there's currently no way to get that proof using the main recommended setup.

Issue #411, opened 2026-07-13
## Bug: auto_save doesn't store code (or anything that isn't a plain fact) — and doesn't warn you


triumphsta:
**What happens**
`with_memwal_openai`'s auto-save feature uses a function called `analyze()` to decide what to save. `analyze()` is built to pull out simple facts from normal sentences — like "my favorite hobby is chess." But if you send something that isn't a plain sentence — like a code snippet — `analyze()` finds no "fact" in it and saves nothing. No error. No warning. It just silently skips it.

Meanwhile, there's a *different* function, `remember()`, that saves things exactly as written — including code. But auto_save never uses this one.

**Proof**
We tested both with the exact same code snippet:
- `analyze(code)` → saved nothing (empty result)
- `remember(code)` → saved it correctly (returned a real job)

**What would help**
- Let developers choose which method auto_save uses — `analyze()` for extracting facts from chat, or `remember()` for saving exact content like code
- At the very least, explain clearly in the docs that `analyze()` only works for natural sentences, not code or other raw content

**Why it matters**
Any app built to remember code (or anything that isn't a simple spoken fact) will look like it's working — the AI will say "I've saved this" — while nothing is actually being stored. This is a serious, invisible gap between what the app claims and what's really happening.

harrymove-ctrl:
Hi @triumphsta, thanks for the clear writeup and repro. 

We traced through `with_memwal_openai`'s implementation (`packages/python-sdk-memwal/memwal/middleware.py`):
* `_post_analyze()` (the `auto_save` hook run after every completion) calls `memwal.analyze(user_text, namespace)` exclusively. There is no code path in `auto_save` that calls `remember()`, confirming your finding exactly.
* It runs through `_fire_and_forget()`, which wraps the call in a `try/except` that only does `logger.debug(...)` on failure. So if `analyze()` comes back with "no facts found" (not an exception—just an empty result), there's no signal at all. It's silent by design, not a missed edge case.
* There is currently no parameter to pick the save method. `auto_save` is a plain bool (default `True`); it's not a choice between `analyze()` and `remember()`.

This also perfectly explains #413's report: when there's no running event loop, `_fire_and_forget` spins up a plain background thread via `asyncio.run(...)` with no join. Nothing keeps it alive past process exit, so a short script can genuinely lose the save mid-flight.

### Next Steps
[...]

Issue #410, opened 2026-07-13
##  Developer experience: confirming a save works is way harder than it should be


triumphsta:
**What happens**
To check if a save actually worked, you have to:
1. Know that saving doesn't finish immediately
2. Find a function called `wait_for_remember_jobs()` — which isn't shown in the main examples, we only found it by printing out all available functions ourselves
3. Skip the recommended easy setup and use the lower-level client directly

None of this is explained in the getting-started docs.

**What would help**
- Add a simple example in the docs titled something like "How to confirm your save worked"
- Also clarify: does this same waiting/polling problem happen in other languages too (like the TypeScript SDK), or is it Python-only? We couldn't tell from the docs.

**Why it matters**
Confirming "did my data actually save?" should be one of the easiest things to do, not something that requires reading the SDK's source code to figure out.

PR #807, opened 2026-08-27
fix(python-sdk): with_memwal auto_save save_mode + confirmable saves (WALM-307)

harrymove-ctrl:
Fixes the two compounding gaps in the Python SDK `with_memwal_*` auto-save path reported in GH #411, #410, #412 (WALM-307).

## Problem

1. **Non-fact content was silently dropped (#411).** `_post_analyze()` called `memwal.analyze()` exclusively. `analyze()` only extracts spoken-fact-style statements, so a code snippet or other non-sentence content came back with zero facts: no error, no warning, nothing stored, while the AI reported success.
2. **No way to confirm a save (#410, #412).** `wait_for_remember_jobs()` lives on the low-level client, and the wrapper never surfaced it or the job IDs its auto-saves enqueued, so anyone on the easy-setup path could not poll for completion or read a blob ID.

Verified against `origin/dev` before changing anything: both gaps were still present.

## Changes

`packages/python-sdk-memwal/memwal/middleware.py`

1. New `save_mode` option on `with_memwal_langchain` and `with_memwal_openai`: `"analyze"` (default, unchanged behavior) or `"remember"` to store the user message verbatim as a single memory. Invalid values raise `ValueError` at wrap time, not on the first save.
2. When `analyze()` extracts zero facts, the middleware now logs a warning that names `save_mode="remember"` instead of returning silently.
3. `_PendingSaves` records the remember job IDs each auto-save enqueues. Two new controls on the wrapped client drain pending saves and then poll those jobs to a terminal state, returning a `RememberBulkResult` with per-job status and blob IDs:
   * `await client.memwal_wait_for_saves(opts)`
   * `client.memwal_wait_for_saves_sync(opts)`
4. The underlying client is exposed as a public `client.memwal` attribute, so reaching `wait_for_remember_jobs()` and friends no longer means touching `_memwal`.

Shared `_run_auto_save()` now backs all three call sites (LangChain, async OpenAI, sync OpenAI) so the modes and warning behave identically across them.

## Docs

`docs/python-sdk/usage/with-memwal.md` gains a "Choosing a save mode" section explaining that `analyze()` is lossy by design, a "Confirming a Save" section covering flush/wait and the returned blob IDs, and a `save_mode` row in the options table. `api-reference.md` updated to match.

## Tests

Seven new tests in `tests/test_middleware.py`: the zero-fact warning, job-ID tracking on a successful analyze, `save_mode="remember"` storing verbatim without calling `/api/analyze`, wrap-time rejection of an unknown mode, one-shot job-ID draining, end-to-end `memwal_wait_for_saves()` returning a blob ID, and the public client surface.

Full suite: 143 passed. `ruff check memwal/ tests/` clean. Both are what `test-python-sdk.yml` runs.

## Not in scope

No version bump or changelog entry: this repo bumps 11 files together as a separate release step.

### DIFF
--- a/packages/python-sdk-memwal/memwal/middleware.py
+++ b/packages/python-sdk-memwal/memwal/middleware.py
@@ -55,7 +55,7 @@ from typing import (
 )
 
 from .client import MemWal, _with_fresh_http_client
-from .types import RecallMemory
+from .types import RecallMemory, RememberBulkOptions, RememberBulkResult
 
 if TYPE_CHECKING:
     from langchain_core.language_models.chat_models import BaseChatModel
@@ -130,6 +130,25 @@ def _format_memories(
     ])
 
 
+SAVE_MODES = ("analyze", "remember")
+"""Valid ``save_mode`` values for the auto-save path.
+
+``"analyze"`` extracts spoken-fact-style statements via the server LLM and
+stores one memory per fact — it is lossy by design, and non-sentence content
+(code snippets, structured data) yields zero facts. ``"remember"`` stores the
+user message verbatim as a single memory, which is what callers saving code or
+other non-conversational content want.
+"""
+
+
+def _validate_save_mode(save_mode: str) -> str:
+    if save_mode not in SAVE_MODES:
+        raise ValueError(
+            f"save_mode must be one of {SAVE_MODES}, got {save_mode!r}"
+        )
+    return save_mode
+
+
 class _PendingSaves:
     """Tracks in-flight fire-and-forget auto-save work (asyncio Tasks and
     background Threads) so a caller can drain it deterministically via
@@ -143,7 +162,23 @@ class _PendingSaves:
     def __init__(self) -> None:
         self._tasks: "set[asyncio.Task[Any]]" = set()
         self._threads: List[threading.Thread] = []
+        self._job_ids: List[str] = []
         self._lock = threading.Lock()
+
+    def track_job_ids(self, job_ids: List[str]) -> None:
+        """Record the background remember jobs an auto-save enqueued, so a
+        caller can later confirm they finished and read their blob IDs."""
+        if not job_ids:
+            return
+        with self._lock:
+            self._job_ids.extend(job_ids)
+
+    def drain_job_ids(self) -> List[str]:
+        """Return every job ID recorded so far and forget them, so a second
+        wait does not re-poll jobs that already settled."""
+        with self._lock:
+            job_ids, self._job_ids = self._job_ids, []
+        return job_ids
 
     def track_task(self, task: "asyncio.Task[Any]") -> None:
         self._tasks.add(task)
@@ -205,6 +240,41 @@ class _PendingSaves:
             thread.join()
 
 
+async def _run_auto_save(
+    memwal: MemWal,
+    text: str,
+    namespace: str,
+    save_mode: str,
+    pending: _PendingSaves,
+    log: Callable[..., Any],
+) -> None:
+    """Persist `text` via the configured save mode and record its job IDs.
+
+    ``analyze()`` returns cleanly with zero facts for content it cannot read
+    as spoken facts (code, structured data), so a plain call there looks
+    identical to a successful save. Warn in that case and point at
+    ``save_mode="remember"`` rather than dropping the write silently.
+    """
+    try:
+        if save_mode == "remember":
+            accepted = await memwal.remember(text, namespace)
+            pending.track_job_ids([accepted.job_id])
+            return
+
+        result = await memwal.analyze(text, namespace)
+        pending.track_job_ids(list(result.job_ids))
+        if not result.facts and not result.job_ids:
+            logger.warning(
+                "Walrus Memory auto-save extracted 0 facts from a %d-character "
+                "message and stored nothing. analyze() only extracts "
+                "spoken-fact-style statements — pass save_mode=\"remember\" to "
+                "store the message verbatim instead.",
+                len(text),
+            )
+    except Exception as e:  # noqa: BLE001 -- auto-save must never break the LLM call
+        log(f"[Walrus Memory] Auto-save failed: {e}")
+
+
 def _expose_memwal_controls(obj: Any, memwal: MemWal, pending: _PendingSaves) -> None:
     """Attach the underlying client and a way to drain pending auto-saves —
     without this, a short-lived process has no way to avoid silently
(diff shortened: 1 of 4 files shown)


ducnmm (changes requested):
Reviewed `7996b3ad`. Direction matches WALM-307 (`save_mode` + confirmable saves), CI is green, and the new tests cover the happy paths. Two confirmability bugs should land before merge — they undercut the #410/#412 fix.

### 1. `memwal_wait_for_saves` drains job IDs before the poll, so a failed wait looks like success on retry

`packages/python-sdk-memwal/memwal/middleware.py` — `memwal_wait_for_saves` / `_sync`:

```python
await pending.flush()
return await memwal.wait_for_remember_jobs(pending.drain_job_ids(), opts)
```

`drain_job_ids()` runs as the argument, so the list is emptied before `wait_for_remember_jobs` returns. That helper re-raises non-transient `_HttpStatusError` and any auth/network error from `_signed_request`. After that raise, a retry drains `[]` and returns `total=0, succeeded=0`, which reads as "nothing pending, all good" — the opposite of confirmability.

Please drain after a successful poll, or restore the IDs on exception:

```python
await pending.flush()
job_ids = pending.drain_job_ids()
try:
    return await memwal.wait_for_remember_jobs(job_ids, opts)
except Exception:
    pending.track_job_ids(job_ids)
    raise
```

Add a test that a raising poll still lets the next `memwal_wait_for_saves` see those job IDs.

### 2. Zero-fact warning is `logger.warning` on every analyze miss, including ordinary questions

`_run_auto_save` logs at WARNING for any `analyze()` that returns no facts. That includes ordinary recall questions, not just code. A 20-turn Q&A then emits 20 warnings suggesting `save_mode="remember"`, which is the wrong advice for a question and buries the code-snippet case this ticket is about.

Route this through the wrapper's `log` (debug unless `debug=True`), or warn once per wrapped client. The current test asserts `logger.warning.called`; please update it to match the quieter signal.

### Non-blocking

3. `_PendingSaves._job_ids` grows without bound if the caller never waits (the documented default). Same class already untracks completed threads for this reason. Cap the list, or only start tracking after the first `memwal_wait_for_saves` call.

The `save_mode` wrap-time validation, shared `_run_auto_save` across LangChain/OpenAI, public `client.memwal`, and the docs sections are the right shape once 1 and 2 are in.

jessiemongeon1:
## Style Guide Audit

harrymove-ctrl:
Thanks for the review @ducnmm — sorry this sat. All three points are addressed in `5ec03d04`.

### 1. Job IDs drained before the poll — fixed

You were right, and the failure mode was exactly as you described. Both controls now drain into a local and restore on exception:

```python
await pending.flush()
job_ids = pending.drain_job_ids()
try:
    return await memwal.wait_for_remember_jobs(job_ids, opts)
except Exception:
    pending.track_job_ids(job_ids)
    raise
```

Same shape in `memwal_wait_for_saves_sync`, which had the same bug via `_run_blocking`.

New test `TestConfirmableSaves::test_failed_wait_leaves_job_ids_for_the_next_wait`: the bulk-status route returns 403 (non-transient, so `wait_for_remember_jobs` re-raises), the wait raises, then the route is switched to a `done` response and the next `memwal_wait_for_saves` still sees `job-1` and returns its blob ID. Against the old code that test fails with `assert 0 == 1` on `RememberBulkResult(results=[], total=0, ...)` — the exact "nothing pending, all good" reading you flagged.

### 2. Zero-fact warning too loud — fixed

Agreed, and it was also inconsistent with the same function's own error path, which already used `log`. The zero-fact branch now goes through the wrapper's `log` (debug unless `debug=True`) instead of `logger.warning`, so a 20-turn Q&A is silent by default.

`test_analyze_mode_warns_when_zero_facts_extracted` is renamed to `test_analyze_mode_reports_zero_facts_through_wrapper_log` and now asserts the message reaches the wrapper's `log` and that `logger.warning` is *not* called. One doc line in `usage/with-memwal.md` that said "logs a warning" is corrected to name the level.

### 3. Unbounded `_job_ids` — capped

Took the cap option. `_PendingSaves.track_job_ids` keeps at most `_MAX_TRACKED_JOB_IDS = 1000`, dropping the oldest, with `test_tracked_job_ids_are_capped_keeping_the_newest` pinning it. This also keeps a later wait from posting thousands of long-settled job IDs in one `/api/remember/bulk/status` request.

I did not take the "only start tracking after the first `memwal_wait_for_saves` call" variant: the first wait would then always return `total=0`, and a single save-then-confirm — the main thing #410/#412 asked for — would never work.

### Branch state

`fix/python-autosave-confirmable` is now 201 commits behind `dev`, but it still merges clean: `git merge-tree --write-tree origin/dev <head>` produces a tree with no conflicts, and nothing in `dev` has touched `middleware.py` or `test_middleware.py` since this branch was cut. I merged `dev` in locally to check for semantic drift rather than just textual: `184 passed, 17 skipped`, `ruff check memwal/ tests/` clean. I left the branch un-rebased so the review diff stays readable — say the word if you'd rather it were rebased before merge.

### Tests

On the branch as pushed: `152 passed, 16 skipped` (was `150 passed, 16 skipped` before this commit — the three new/changed tests net +2 since one existing test was rewritten rather than added). `ruff check memwal/ tests/` clean. The 16/17 skips are the live-relayer integration tests, which need credentials this environment doesn't have.

packages/python-sdk-memwal/memwal/middleware.py
```diff
@@ -221,9 +295,44 @@ def _expose_memwal_controls(obj: Any, memwal: MemWal, pending: _PendingSaves) ->
...
     """
+    async def memwal_wait_for_saves(
+        opts: Optional[RememberBulkOptions] = None,
+    ) -> RememberBulkResult:
+        """Drain pending auto-saves, then poll their remember jobs to a
+        terminal state and return per-job status + blob IDs."""
+        await pending.flush()
+        job_ids = pending.drain_job_ids()
+        try:
+            return await memwal.wait_for_remember_jobs(job_ids, opts)
```
ducnmm, line 306: **[bug]** `memwal_wait_for_saves` (and the sync twin) passes the full drained list to `MemWal.wait_for_remember_jobs`, which POSTs every id in one `/api/remember/bulk/status` call. The relayer rejects `job_ids.len() > MAX_BULK_ITEMS` (20) with 400. 400 is not transient, so the poll raises; after `5ec03d04` the ids are restored, so a retry hits the same 400 forever.

`_MAX_TRACKED_JOB_IDS = 1000` and the docs ("poll every enqueued remember job") advertise waiting far past that cap. Easy to hit: `save_mode="remember"` is one job per turn (21 turns), and default `analyze` can emit up to 20 facts per message, so two fat extracts already overflow. The e2e test only waits on a single remember job.

**Suggestion:** Chunk status polls into batches of 20. The durable fix is in `wait_for_remember_jobs` so every caller gets it; a local chunk in `memwal_wait_for_saves` / `_sync` is enough for this PR. Add a test that 21 tracked ids confirm instead of 400-looping.

ducnmm (changes requested):
## Summary

`5ec03d04` fixes the previous CHANGES_REQUESTED on `7996b3ad`: `memwal_wait_for_saves` now restores job IDs if the poll raises, the zero-fact path no longer uses `logger.warning`, and `_job_ids` is capped. Remaining blocker is confirmability past the relayer's 20-id bulk-status cap.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

## Issues outside the diff

- **[suggestion]** packages/python-sdk-memwal/CHANGELOG.md — `save_mode`, `memwal_wait_for_saves`, and the failed-wait restore have no bullets under unpublished `## 0.1.11` in `packages/python-sdk-memwal/CHANGELOG.md` or `docs/python-sdk/changelog.mdx`. `verify-manual-sdk-release.mjs` will not fail (the heading already exists on origin/dev); still dump them when releasing.

PR #808, opened 2026-08-27
docs(guides): add system prompt template library (WALM-199)

harrymove-ctrl:
Closes WALM-199.

## Why

Connecting the MCP server exposes the memory tools but does not change agent behaviour. Without a standing instruction, most agents write only when a user says "remember this", which is why an untuned setup stalls at a handful of lifetime memories instead of writing every session.

## What

**New docs page: `docs/guides/system-prompt-templates.md`**

Six copy-paste system prompts, one per agent type:

| Template | Suggested namespace | Expected writes |
| --- | --- | --- |
| Universal starter | `personal` | 3 to 8 / session |
| Personal assistant | `personal` | 5 to 15 / session |
| Coding agent | `repo-<name>` | 4 to 10 / session |
| CRM and customer support | `customer-<id>` | 2 to 6 / conversation |
| Research agent | `research-<topic>` | 8 to 20 / session |
| Partner backend (SDK) | `user-<id>`, per request | 1 to 5 / user turn |

All six follow the same four-part shape (recall first, write proactively, what to skip, which namespace), and the page documents that shape so users can write their own. It also covers how to verify the prompt took effect and what to check when the write count does not move.

Wired into the nav under both Guides tabs and the MCP tab. Tool names and semantics match `packages/mcp/src/auth-required.ts`; the CRM and partner-backend templates carry explicit privacy rules against writing credentials or another user's data.

**Wizard success state: `apps/app/src/pages/ConnectMcp.tsx`**

The MCP connect success screen now shows the universal starter prompt with a copy button and a link to the full library, placed right after the account details and before the dashboard CTA. It renders only when the credential hand-off actually landed, so the two failure states stay focused on recovery rather than adding a next step to a flow that has not finished.

## Testing

- `apps/app` suite: 91 passed (12 files), including 2 new cases asserting the card appears on a delivered hand-off and is withheld when it was not.
- `npm run lint` and `npm run build` in `apps/app`: clean.
- `docs.json` edited through a JSON round-trip, so the diff is 3 added lines with no reformatting.

## Notes

The ticket lists the docs URL as `docs.memwal.ai/prompt-templates`; the page ships at `/guides/system-prompt-templates` to sit with the other guides. Say the word if you want a redirect from the shorter path.

### DIFF
--- a/apps/app/src/pages/ConnectMcp.test.tsx
+++ b/apps/app/src/pages/ConnectMcp.test.tsx
@@ -85,6 +85,26 @@ describe('MCP sign-in hand-off', () => {
         expect(screen.getByText(/handed off to your MCP client/i)).toBeInTheDocument()
     })
 
+    it('offers the starter system prompt once the client is connected', async () => {
+        stubListener(async () => new Response('{}', { status: 200 }))
+        await signIn()
+
+        expect(await screen.findByText(/MCP client connected/i)).toBeInTheDocument()
+        expect(screen.getByText(/most agents only write when told to/i)).toBeInTheDocument()
+        expect(screen.getByText(/call memwal_remember in that same turn/i)).toBeInTheDocument()
+        expect(screen.getByRole('button', { name: /copy prompt/i })).toBeInTheDocument()
+    })
+
+    it('withholds the starter prompt when the hand-off never landed', async () => {
+        stubListener(async () => {
+            throw new TypeError('Failed to fetch')
+        })
+        await signIn()
+
+        expect(await screen.findByText(/Almost there/i)).toBeInTheDocument()
+        expect(screen.queryByRole('button', { name: /copy prompt/i })).not.toBeInTheDocument()
+    })
+
     it('does not claim success when nothing answers on localhost', async () => {
         stubListener(async () => {
             throw new TypeError('Failed to fetch')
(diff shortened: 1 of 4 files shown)


ducnmm (changes requested):
Reviewed `7d39444`. Ticket coverage is right: six templates (universal + the five named agent types), namespace + expected-writes on each, four-part shape, wizard success card only when `callbackOutcome === 'delivered'`, tests for both the shown and withheld cases, CI green. Tool names match `packages/mcp/src/auth-required.ts`. CRM and partner templates carry the privacy rules.

One change before merge:

### Wrong "where to paste" link

`docs/guides/system-prompt-templates.md` step 2 (and the Related list) points at `/guides/agent-runtimes` for "the exact file per client: `CLAUDE.md` / `AGENTS.md` / `.cursor/rules`". That page is the EVM/cross-chain runtime guide — it does not mention those files.

Please point at the MCP client pages instead (`/mcp/claude-code` for `CLAUDE.md`, plus Codex/Cursor if they exist) or drop the link and keep the file names inline. As written, the first instruction on this page sends people to the wrong doc.

Non-blocking:

- Universal starter (also the wizard copy) does not tell the agent that `memwal_recall` results are data, not instructions. Personal assistant and partner backend already do. Worth one line on the starter, since that is what we paste at connect time.
- Ticket listed `docs.memwal.ai/prompt-templates`; `/guides/system-prompt-templates` is the better home. No redirect needed unless we already advertised the short path.

ducnmm (approved):
Re-reviewed `0fa9af14`. Both notes are in, CI is green.

1. Step 2 keeps the file names inline and points Claude Code at `/mcp/claude-code`. Related lists `/mcp/claude-code`, `/mcp/codex`, `/mcp/cursor`, and keeps `/guides/agent-runtimes` only for the partner-backend SDK path.
2. Universal starter (docs + wizard copy) now treats recall as data, not instructions. The two copies are byte-identical.

Approving.

PR #809, opened 2026-08-27
fix(mcp): confirm a completed sign-in, and keep serving stdin after it

### LINKED ISSUE #633: memwal-mcp sign-in flow gives no confirmation that credentials persisted or bridge reconnected
## Description

The `memwal-mcp login` flow opens a browser URL for Sui wallet connection. After the user approves the transaction in the browser, there is no confirmation back to the agent or terminal that:

1. The credential was actually written to `~/.memwal/credentials.json`
2. The MCP bridge picked up the new credential
3. Subsequent calls will use the authenticated session

The agent received a 401, followed the sign-in URL, the user approved — then the next call timed out (#1 related issue). It was unclear whether the sign-in succeeded, failed, or succeeded but the bridge didn't reload.

## Steps to Reproduce

1. Start memwal-mcp without valid credentials
2. Attempt a recall → 401
3. Agent follows sign-in URL, user approves in browser
4. No output in terminal/agent confirming success
5. Next call may still fail (timeout or 401)

## Expected Behavior

- After wallet approval, the sign-in endpoint should return a clear success/failure response
- The MCP bridge should reload credentials without requiring process restart
- Agent should receive a structured success message it can act on

## Actual Behavior

- Sign-in URL opens, user approves, but no confirmation flows back to the calling context
- Credential persistence is assumed but not verified
- #580 (in-session login not updating bridge) was marked closed/fixed, but the UX gap remains

## Related

- #580 (closed) — in-session login not updating bridge credentials
- #628 — global credentials.json scope issue
- #629 — login exits 0 without TTY

nikola0x0:
Fixes [WALM-394](https://linear.app/mysten-labs/issue/WALM-394). Closes #633.

## Two faults, found together

**Success was never reported.** The failure path says so twice — a `notifications/message` and a notice prefixed onto the next tool call. Success only wrote to the log file, so a user who approved in the browser had no way to tell whether credentials landed, whether the bridge adopted them, or whether retrying was worth it.

**The bridge stopped reading stdin after signing in.** Writing the test for the confirmation surfaced this. The auth-required stub hands off by calling `process.stdin.pause()`, and a stream paused that way does not resume when a new `data` listener attaches — so `readStdinLines` attached and read nothing. The bridge served only the request replayed from `pendingLines`, then went silent. Sign-in worked, the retry worked, and **every call after it hung until the client timed it out**.

That second one is the reported "next call timed out", and it made the first much worse: with no confirmation either way, a dead connection was indistinguishable from a failed sign-in. Confirmed against unmodified `dev` before fixing.

## Changes

- `readStdinLines` resumes stdin, with a regression test that signs in mid-session and then makes **two** calls rather than one.
- A one-shot banner is queued on credential adoption and consumed by the first `tools/call` result — naming the account, delegate, and resolved credentials path once. Unlike the failure notice, which repeats while the problem persists, success is an event, so the read clears it.
- The `notifications/message` twin of the existing failure warning, in both modes.
- Sign-in copy moved into `messages.ts`. The signed-out stub and signed-in bridge had drifted into different assistant instructions, step wording, and closing lines, and both claimed credentials land at `~/.memwal/credentials.json` even when the resolved path was project-local (#628).

## Verification

`packages/mcp` suite: **65/65**, up from 60. Three new files — `login-success-notice`, `login-handoff-stdin`, `login-prompt-unified`. The browser half is driven by the same preflight-then-callback handshake `login-preflight` already uses, so no wallet or on-chain transaction is involved.

Also smoke-tested against **dev with a real wallet**, twice:

| Check | Result |
|---|---|
| confirmation notification after approval | names the real account + resolved path |
| banner on first `tools/call` | present, on a genuine upstream success |
| second call answered | yes — the call that used to hang |
| second call banner | absent, one-shot held |

## Noticed while testing, NOT addressed here

Every tool call through the dev relayer's MCP session sits at a ~60s floor regardless of cost: `memwal_health` took 63.0s and 60.6s through the bridge, while the same endpoint answers in 0.24s over plain curl. Recall 504s because it is that floor plus real work. The bridge is not the cause — the same code against a mock relayer answers in ~130ms across all 65 tests. Filing separately.


packages/mcp/src/bridge.ts
```diff
@@ -631,27 +661,9 @@ async function handleLocalLogin(
...
-            `Steps:`,
-            `1. Open the URL in any browser`,
-            `2. Click **Connect Sui Wallet** and approve the on-chain \`add_delegate_key\` transaction`,
-            `3. Once "Connected" appears, retry the previous request — credentials at \`~/.memwal/credentials.json\` get overwritten with the new wallet's delegate key`,
-            ``,
-            `_The login link stays valid for 5 minutes._`,
-        ].join("\n"),
+        // Signed in already: this flow REPLACES the stored delegate key, which
+        // the shared prompt calls out.
+        text: loginPrompt({ url, credentialsPath: credsPath(), signedIn: true }),
```
ducnmm: **[bug]** `handleLocalLogin` always builds the prompt with `signedIn: true`, on the assumption that the bridge only runs while credentials exist. That is false after `memwal_logout`: login is intercepted before the `loggedOut` guard, the credentials file is gone, and this is a supported same-session path (`logout-invalidation.test.mjs` drives it). The new note then lies twice — “you are already signed in” and “replaces the stored delegate key” — which can look like logout did not take. `handleLoginToolCall` has the opposite hardcode (`signedIn: false` at `auth-required.ts:362`), so a second `memwal_login` after a successful callback but before the hand-off also skips the replacement warning even though `saveCreds` will overwrite the file. The flag’s own contract is “credentials already exist, so signing in REPLACES them,” which is a disk fact, not a mode fact.

**Suggestion:** At both call sites pass `signedIn: loadCreds() !== null` (import `loadCreds` in `bridge.ts`). That matches the `LoginPromptInfo` comment, is correct after logout, and is correct if `memwal_login` is called again while a file is already on disk. Add a case to `login-prompt-unified` or the logout-then-login integration test that the post-logout prompt does not match `/already signed in/i`.

packages/mcp/src/bridge.ts
```diff
@@ -1174,6 +1232,15 @@ export async function runBridge(
         releaseLogoutPark?.();
         releaseLogoutPark = null;
         logoutPark = null;
+
+        // Queue the confirmation only once the session is actually live, so
+        // the banner cannot claim an authenticated connection before there is
+        // one. It rides out on the next `tools/call` result.
+        notePendingLoginSuccess({
```
ducnmm, line 1284: **[suggestion]** Confirmation is wired twice — auth-required `sendLogMessage` + `notePendingLoginSuccess` in `index.ts` for the signed-out hand-off, and `handleLocalLogin`’s `notifications/message` + `adoptCredentials`’s `notePendingLoginSuccess` for re-login. `login-success-notice.test.mjs` only spawns signed-out and therefore never exercises the second pair. A regression that dropped the `adoptCredentials` queue (or the bridge-side notification) would still pass the new suite; the existing logout-then-login test only asserts `RECALL_OK` and would not notice a missing banner either.

**Suggestion:** Extend the success test (or the logout-invalidation flow) to start already signed in, call `memwal_login`, complete the callback, and assert both the notification and a one-shot banner on the next forwarded `tools/call`.

packages/mcp/src/bridge.ts
```diff
@@ -717,6 +729,52 @@ function handleLocalLogout(): { text: string; isError: boolean } {
  * file directly. They appear in `tools/list` by splicing them into the
  * relayer's response on the way back to the client.
  */
+/**
```
ducnmm, line 756: **[nit]** Inserting the `pendingLoginSuccess` JSDoc immediately after the existing `runBridge` JSDoc leaves two consecutive docblocks. The first now documents nothing (JSDoc binds to the next statement, which is the second `/**`), and `runBridge` itself lost its module-level documentation.

**Suggestion:** Keep one file-level or `runBridge` JSDoc on the function, and a short comment on the module-level pending slot — or pass the initial success info into `runBridge` next to `pendingLines` so the slot does not need to live outside the function.

packages/mcp/src/messages.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/** Sui ids are 66 characters; showing one whole swamps the message. */
+function shortId(id: string): string {
+    return id.length > 20 ? `${id.slice(0, 10)}…${id.slice(-6)}` : id;
+}
+
+/**
+ * Banner prefixed onto the first tool result after a sign-in completes.
+ *
+ * Deliberately a ONE-SHOT, unlike {@link loginFailureNotice}'s repeat-until-
```
ducnmm: **[nit]** `{@link loginFailureNotice}` does not resolve. That helper is a private function in `auth-required.ts` and is not exported from this module (or anywhere else).

**Suggestion:** Drop the `@link` and refer to the failure notice in prose, or move `loginFailureNotice` into `messages.ts` next to the success copy if the intent is one module for both voices.

ducnmm (dismissed):
## Summary

The stdin `resume()` after an auth-required hand-off is a real Node stream fix, and the new regression test (two post-login calls, not one) pins the “deaf after sign-in” failure. Queuing a one-shot banner plus a `notifications/message` twin of the existing failure path is the right shape for a completed login, and unifying the prompt in `messages.ts` stops the two modes from drifting. The main defect is that the bridge always passes `signedIn: true`, so `memwal_login` after `memwal_logout` in the same process claims the user is already signed in and that a stored key will be replaced. Re-login confirmation (`adoptCredentials` / `handleLocalLogin`) is also unwired from the new tests.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 2

packages/mcp/src/bridge.ts
nikola0x0: Fixed in ce2a1ee. Both call sites pass `signedIn: loadCreds() !== null` — it is a fact about the disk, as the `LoginPromptInfo` comment says, not about which mode answered.

The logout-then-login test in `logout-invalidation.test.mjs` now asserts the post-logout prompt matches neither `/already signed in/i` nor `/replaces the stored delegate key/i`. Confirmed it fails with the old `signedIn: true` restored.

packages/mcp/src/bridge.ts
nikola0x0: Added. `login-success-notice.test.mjs` has a case that seeds credentials so the bridge runs, calls `memwal_login`, completes the callback, and asserts the `notifications/message`, the banner on the next forwarded `tools/call`, and that a second call does not repeat it. It also pins that this is the one path where the replacement warning is correct.

The logout-then-login flow picked up the same notification + banner + one-shot assertions, so a dropped `adoptCredentials` queue no longer passes on `RECALL_OK` alone.

packages/mcp/src/bridge.ts
nikola0x0: Fixed — the `runBridge` block is back on `runBridge`, and the module-level pending slot keeps its own.

packages/mcp/src/messages.ts
nikola0x0: Took the second option: `loginFailureNotice` moved into `messages.ts` beside the success copy, as a pure function of the reason rather than reading `lastLoginFailure` directly. One module for both voices, which is what that module says it is for, and the link resolves.

ducnmm (commented):
## Summary

Follow-up on REQUEST_CHANGES at `1668d926`. Head `02e51fb6` addresses the prior items. Sign-in confirmation and stdin-after-login both hold (file-backed `signedIn`, `resume()` after attaching the stdin listener, one-shot success banner).

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

harrymove-ctrl (approved):
Approving. Second round is resolved: all four review threads are closed and @ducnmm's 08:52 follow-up reports 0 bugs / 0 suggestions / 0 nits, leaving only the earlier REQUEST_CHANGES standing. CI is green.

Read the diff rather than relying on the sign-off:

- `process.stdin.resume()` in `readStdinLines` is the right fix. A stream paused via `pause()` does not restart when a new `data` listener attaches, so the hand-off really did leave the bridge deaf after the replayed `pendingLines`. No-op on the cold path.
- `signedIn: loadCreds() !== null` at both call sites is correct — it is a fact about the disk, which is what `LoginPromptInfo` documents, and it fixes the post-logout prompt claiming a stored key would be replaced.
- The one-shot banner is consumed on read and only applied to `tools/call` replies, so it cannot be swallowed by a `tools/list` the user never reads.

One non-blocking note for whoever cuts the next release: the four entries here are added under `## 0.0.12`, which `origin/dev` and `origin/main` have already published. That is the same placement flagged on #793, and it wants sorting out once for the batch rather than per-PR — see my comment there.

packages/mcp/CHANGELOG.md
```diff
@@ -18,6 +18,10 @@
 
 ### Fixed
 
+- Keep reading stdin after an in-session `memwal_login`. The auth-required stub hands off to the bridge by pausing stdin, and a stream paused that way does not resume when a new `data` listener attaches, so the bridge served only the request replayed from the hand-off and then went deaf. Signing in appeared to work, the retry succeeded, and every call after it hung until the client timed it out. (#633)
```
ducnmm: **[bug]** `origin/dev` is already unpublished MCP `0.0.13` (`origin/main` is `0.0.12`). These four #633 / WALM-394 bullets sit under published `## 0.0.12`. Merging as-is rewrites a released section. `docs/mcp/changelog.mdx` does not have them at all (0.0.13 FAQ/body still only WALM-480 + WALM-390).

**Suggestion:** Move the four bullets under existing `## 0.0.13` in `packages/mcp/CHANGELOG.md` and `docs/mcp/changelog.mdx`. Extend the mdx `answer:` and 0.0.13 intro. Leave `package.json` at `0.0.13` (do not bump 0.0.14). Keep the two files in sync.

ducnmm (dismissed):
## Summary

Merge of `origin/dev` into WALM-394. Sign-in confirmation / stdin resume code still looks right. Dump is wrong after the merge: MCP is unpublished `0.0.13` on `dev` vs `0.0.12` on `main`, but the four #633 bullets landed under published `## 0.0.12` in `packages/mcp/CHANGELOG.md` and are missing from `docs/mcp/changelog.mdx`.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

Issue #621, opened 2026-08-13
memwal_recall has no date-aware ordering or write-time field — newest-wins protocols can't be correct by construction


kiyeps:
## Summary

`memwal_recall` returns records ordered only by semantic distance, with no date-aware ordering and no exposed write-time field on the results. This makes any "newest-wins" protocol built on top of recall impossible to implement correctly by construction: the most recent record can silently fall outside the top-N window whenever an older record happens to match the query string more closely.

## Why this matters

A very common pattern for agent memory is *current state = the newest checkpoint*. To implement that on top of `memwal_recall`, the caller has to:

1. recall the state namespace,
2. pick the record with the newest date.

But step 2 can only see whatever step 1 returned, and step 1 is ranked purely by embedding distance. If the newest record is worded naturally and an older record echoes the query keywords more literally, the older one ranks higher and the newest one is pushed out of the `limit N` window entirely. The "pick the newest" logic then never even sees the newest record, so it confidently resumes stale state.

This is not hypothetical — I hit it while using the Markov prompt (Walrus Session 7) for real cross-tool work: a fresh session resumed a task I had finished two days earlier, because the newest checkpoint was outside the recall window. Widening the window and re-ranking by date in the prompt mitigates it, but that is a workaround at the prompt layer for a gap in the retrieval API.

## Repro (conceptual)

- Store 5 checkpoints in one namespace, dates 07-01..07-06. Word the four oldest so they closely match a generic query like "current task checkpoint goal status"; word the newest (07-06) more naturally.
- `memwal_recall(query="current task checkpoint goal status", namespace=..., limit=3)`
- Observed: the 07-06 record is not in the results; the top-3 are all older.
- A caller doing "newest date wins" over the returned set therefore selects a stale record.

## Suggested improvements (any one would help)

1. **Expose each record's write-time** as a structured field in recall results (not just embedded in the text), so callers can order/verify by recency deterministically.
2. **Optional `sort` / recency-boost parameter** on `memwal_recall` (e.g. `sort=recency`, or a recency weight blended into the distance score), so time-sensitive retrieval doesn't depend on prompt-side re-ranking.
3. **Document the ordering guarantee** clearly: state explicitly that results are distance-ranked with no recency guarantee, so integrators don't assume date order.

## Environment

- Package: `@mysten-incubation/memwal` via the MCP `memwal_*` tools and the local `app/memwal.js` bridge.
[...]

PR #810, opened 2026-08-27
feat(server,sdk): date-aware recall — write-time on results + sort=recent

nikola0x0:
Refs WALM-383 (#621). The MCP half is tracked separately as WALM-428 — see **What's not here**.

`memwal_recall` returned records ordered only by semantic distance, with no write-time on the results. That makes a newest-wins protocol impossible to build correctly: ranking happens before `limit` truncates, so a newest record worded less literally than an older one is dropped before the caller ever sees it, and no client-side sort recovers it.

## What's here

**`created_at` on recall results** (ask #1) — threaded `SearchHit` → `HydratedMemory` → `RecallResult`, through the SDK's `RecallMemory`, and rendered as `written=YYYY-MM-DD` in the `memwal_recall` output. Optional on the wire, so default-weighted responses are byte-identical to today.

**`sort=recent`** (ask #2, server-side only) — over-fetches semantic candidates (`5×limit`, capped at 50, never below `limit`), orders by write-time descending, then truncates. Semantic similarity becomes the candidate generator and write-time decides the order. Selection runs on `SearchHit`s *before* hydration, so the over-fetch costs one wider SQL query rather than 5× the Walrus downloads and SEAL decrypts. `sort=relevance` is the default and issues the identical query it always has.

**A documented ordering guarantee** (ask #3) — new *Ordering* section in the SDK API reference, and the tool description now states outright that results are relevance-ranked with no recency guarantee.

Also extracts `recall_results_from_ranked`: `/api/recall` and `/api/ask` carried identical inline map closures, so the fields reaching a client are now decided in one place.

## What's not here

**The `sort` parameter on the MCP tool** — tracked as WALM-428.

`services/server/scripts` installs the *published* SDK via `npm ci` from its own `package-lock.json`, with `services/server/` as the Docker build context (`services/server/Dockerfile:48-49`), so it cannot resolve the workspace SDK. It is pinned at `0.0.3`, whose only signature is `recall(query, limit?, namespace?)`. `sort` therefore has to ship in a published release before the tool can expose it — this PR is what puts it in the SDK. Writing it anyway would advertise a knob that gets silently dropped, so it's backed out.

Worth flagging for whoever picks up WALM-428: the tool's input schema is duplicated, zod in `services/server/scripts/mcp/tools/recall.ts` and a hand-written mirror in `packages/mcp/src/auth-required.ts`, with nothing asserting they agree.

**So `#621` stays open.** It was reported from the MCP surface, and that surface can't ask for recency until WALM-428 lands. The HTTP API and the TS SDK can today.

## Design note

The `recent` preset originally proposed here — `semantic 0.3 / recency 0.7 / half-life 30d` — does not work, per @henry.nguyen on the ticket. Weights only re-rank rows the search already returned, and at a 30-day half-life two records days apart barely differ: on the ticket's own dates, newest (07-06, distance 0.45) scores `0.849` against older (07-01, distance 0.10) at `0.880`, so the closer-worded older row still wins even when the newest *is* in the window. Hence fixing selection rather than scoring. The docs that had recommended that preset are corrected in c2a1a95a.

## Testing

| Suite | Result |
| --- | --- |
| Rust `services/server` | 656 pass, 0 fail |
| `packages/sdk` | 97 pass, 0 fail |
| `services/server/scripts` | 232 pass, 0 fail |

15 new tests, each watched fail before implementing. Plus `tsc --noEmit` on the SDK and `pnpm check:compatibility`.

Also fixes a pre-existing type error this work surfaced: `mountMcpRoutes` declared `Express` but the integration test mounts an `express.Router()`. The scripts package had no `typecheck` script — its tests run `node --test --import tsx`, which strips types — so it never failed. Both are fixed here.

Two pre-existing issues noticed, neither addressed here:

- `routes::memory_read::tests` is flaky on `dev` — 4, then 3, then 0 failures across runs, `no entry found for key`. Confirmed against a clean checkout before any of this work, so a red CI run there is not from this PR.
- `services/server/scripts` pins the published SDK at `0.0.3` while the workspace is on `0.1.5`. That gap is what blocks the MCP `sort` parameter above. Possibly deliberate — the sidecar deploys independently — but nothing in the repo records the intent.

### DIFF
--- a/.changeset/recall-write-time-and-recency-weights.md
+++ b/.changeset/recall-write-time-and-recency-weights.md
@@ -2,12 +2,12 @@
 "@mysten-incubation/memwal": patch
 ---
 
-Expose each memory's write-time on `recall()`, and let callers ask for recency-weighted ranking (#621).
+Expose each memory's write-time on `recall()`, and let callers ask for a newest-first ordering (#621).
 
-`RecallMemory` now carries `created_at` (RFC3339, the time the fact was stored), and `RecallOptions` accepts `scoringWeights` — which the relayer's composite ranker has always supported but which only the manual bring-your-own-embedding paths could reach.
+`RecallMemory` now carries `created_at` (RFC3339, the time the fact was stored). `RecallOptions` accepts `sort: "relevance" | "recent"`, and `scoringWeights` — which the relayer's composite ranker has always supported but which only the manual bring-your-own-embedding paths could reach.
 
 `created_at` is the piece a "newest wins" protocol was missing: previously the only date available was whatever the memory text happened to mention, so callers could neither order nor verify by write-time.
 
-Note that `scoringWeights` re-ranks the candidates the vector search already returned — it does not widen the search. A record that fell outside the cosine top-`limit` cannot be recovered by any weighting, so raise `limit` when the record you need might not be in the window. A dedicated recency mode that over-fetches server-side is still open (WALM-383).
+`sort: "recent"` is the newest-wins mode: the relayer over-fetches semantic candidates (5x `limit`, capped at 50), orders them by write-time descending, then truncates to `limit`. `scoringWeights` does something narrower — it re-ranks the candidates the vector search already returned and does not widen the search, so a record that fell outside the cosine top-`limit` cannot be recovered by any weighting.
 
-Fully backward-compatible: omitting `scoringWeights` leaves the request byte-identical to a plain cosine sort, and `created_at` is optional so relayers that don't send it behave exactly as before.
+Fully backward-compatible: omitting `sort` and `scoringWeights` leaves the request byte-identical to a plain cosine sort, and `created_at` is optional so relayers that don't send it behave exactly as before.
--- a/.changeset/recall-write-time-and-recency-weights.md
+++ b/.changeset/recall-write-time-and-recency-weights.md
@@ -1,0 +1,11 @@
+---
+"@mysten-incubation/memwal": patch
+---
+
+Expose each memory's write-time on `recall()`, and let callers ask for recency-weighted ranking (#621).
+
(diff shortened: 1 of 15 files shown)


ducnmm (approved):
Reviewed `3e842cab` against the WALM-383 call (`sort: relevance | recent`, not weights).

- `created_at` is on the wire (`SearchHit` → SDK `RecallMemory`) and MCP renders `[written=YYYY-MM-DD]`.
- `sort=recent` is newest-among-matches: over-fetch `limit*5` capped at 50, never below `limit` (so `limit>50` does not under-fetch). Truncate happens on `SearchHit`s before hydration.
- Default `relevance` is the same cosine query as today.
- MCP `sort` is correctly left to WALM-428 until the sidecar can take a published SDK that has the param.

CI is green. Approving.

Non-blocking: WALM-428 still has to keep the zod schema in `tools/recall.ts` and the hand-written `auth-required.ts` mirror in lockstep — as you flagged.

jessiemongeon1:
## Style Guide Audit

**8** violation(s) found. All must be fixed before merge.

### `docs/sdk/api-reference.md` (8 violation(s))

_8 violation(s) (4 regex, 4 claude)_

- **Line 117** — No em dashes in prose
  - Current: `—`
- **Line 140** — No em dashes in prose
  - Current: `—`
- **Line 176** — No em dashes in prose
  - Current: `—`
- **Line 188** — No em dashes in prose
  - Current: `—`
- **Line 117** — no-em-dashes
  - Current: `blends recency and importance into the ranking — see`
  - Fix: `blends recency and importance into the ranking. See`
- **Line 140** — no-em-dashes
  - Current: `recency guarantee — not within the returned set`
  - Fix: `recency guarantee: not within the returned set`
- **Line 154** — passive-voice
  - Current: `a record that was never returned`
  - Fix: `a record the server never returned`
- **Line 187** — no-italic
  - Current: `Use `created_at` to *verify* or display an order.`
  - Fix: `Use `created_at` to verify or display an order.`

---

PR #811, opened 2026-08-27
fix(auth): don't treat Sui RPC 429 as a revoked delegate key (WALM-429)

ducnmm:
## Summary
- Signed HTTP auth treated **every** `verify_delegate_key_onchain` error as a revoke, including Sui gRPC `GetObject` 429s. That evicted a still-valid cache row, retried RPC, and returned an empty 401.
- Empty 401s made the SDK print `Walrus Memory isn't signed in. Call the memwal_login tool` even though the key was still on-chain. Recall looked intermittent.
- HTTP `/api/recall` now matches MCP SSE on classification: `RpcError` / `ScanCapExceeded` are unavailability, not a revoke.
- **Fail-closed on cache hit + RPC failure:** keep the cached mapping (do not evict) but return 503. We do **not** authenticate from a stale 24h cache while Sui is unreachable, so a revoked key cannot ride an outage. 503 carries `x-auth-error: AUTH_UPSTREAM_UNAVAILABLE` and `Retry-After: 5`.
- Cache miss + unavailability → 503. Evict only on `KeyNotFound` / `AccountDeactivated` / `WrongObjectType`.
- Registry-scan semaphore exhaustion (`try_acquire` fail) is also 503. That is not an RPC error — it is retryable saturation, previously a timing-normalized 401, and is now externally observable as unavailable. Called out because it is a behavior change beyond the PR title.
- TS SDK 0.1.6 and Python SDK 0.1.9 report **auth** 503 (header present) as retryable credential-verification unavailability, not a sign-in failure. Other relayer 503s (Redis, rate limiter, LLM) keep the generic sanitized body. Versions and changelogs dumped by hand (no changeset).

## Test plan
- [x] `cargo test --bin memwal-server` filters: `resolve_account_cache_hit_*`, `upstream_unavailable_is_503_not_401`, `test_error_variants_are_distinct`
- [x] SDK `sanitize-server-error` tests (auth 503 vs generic 503)
- [x] Python `test_auth_rejected_message` (3/3)
- [ ] After merge: watch prod `POST /api/recall` during Sui 429 bursts — should be 503 with `AUTH_UPSTREAM_UNAVAILABLE`, not empty 401 + `revoked on-chain`

Linear: WALM-429

### DIFF
--- a/packages/sdk/src/utils.ts
+++ b/packages/sdk/src/utils.ts
@@ -321,6 +321,18 @@ export function sanitizeServerError(
         };
     }
 
+    // Relayer 503 = Sui RPC/gRPC could not be consulted (WALM-429). This is
+    // retryable and is not a sign-in failure — do not send callers to
+    // memwal_login.
+    if (Number(status) === 503) {
+        return {
+            message:
+                "Walrus Memory temporarily cannot verify credentials (upstream unavailable). Retry; this is not a sign-in failure.",
+            raw: rawBody,
+            serverCode: "UPSTREAM_UNAVAILABLE",
+        };
+    }
+
     const MAX = 200;
     let serverCode: string | undefined;
     let text = rawBody;
--- a/packages/sdk/src/utils.ts
+++ b/packages/sdk/src/utils.ts
@@ -304,6 +304,7 @@ export function redactInternalUrls(text: string): string {
 export function sanitizeServerError(
     status: number,
     rawBody: string,
+    authError?: string | null,
 ): { message: string; raw: string; serverCode?: string } {
     // Number() so a string "401" (some MCP / HTTP paths) still hits this branch.
     if (Number(status) === 401) {
@@ -321,15 +322,15 @@ export function sanitizeServerError(
         };
     }
 
-    // Relayer 503 = Sui RPC/gRPC could not be consulted (WALM-429). This is
-    // retryable and is not a sign-in failure — do not send callers to
-    // memwal_login.
-    if (Number(status) === 503) {
+    // Auth-path 503 only: Sui could not be consulted (WALM-429). Other
+    // relayer 503s (Redis, rate limiter, LLM) keep the generic sanitizer
+    // so they are not mislabeled as a credential-verification failure.
+    if (Number(status) === 503 && authError === "AUTH_UPSTREAM_UNAVAILABLE") {
         return {
             message:
                 "Walrus Memory temporarily cannot verify credentials (upstream unavailable). Retry; this is not a sign-in failure.",
             raw: rawBody,
-            serverCode: "UPSTREAM_UNAVAILABLE",
+            serverCode: "AUTH_UPSTREAM_UNAVAILABLE",
         };
     }
 
--- a/services/server/src/auth.rs
+++ b/services/server/src/auth.rs
@@ -93,8 +93,8 @@ pub(crate) fn upstream_unavailable() -> Response {
     (
         StatusCode::SERVICE_UNAVAILABLE,
         [
-            ("retry-after", "5"),
-            ("x-auth-error", AUTH_UPSTREAM_UNAVAILABLE),
+            ("retry-after", AUTH_UPSTREAM_RETRY_AFTER_SECS.to_string()),
+            ("x-auth-error", AUTH_UPSTREAM_UNAVAILABLE.to_string()),
         ],
         "upstream unavailable",
     )
@@ -975,11 +975,12 @@ mod tests {
                 .and_then(|v| v.to_str().ok()),
             Some(AUTH_UPSTREAM_UNAVAILABLE),
         );
+        let retry_after = AUTH_UPSTREAM_RETRY_AFTER_SECS.to_string();
         assert_eq!(
             resp.headers()
                 .get("retry-after")
                 .and_then(|v| v.to_str().ok()),
-            Some("5"),
+            Some(retry_after.as_str()),
         );
         assert_eq!(AUTH_UPSTREAM_RETRY_AFTER_SECS, 5);
         let body = axum::body::to_bytes(resp.into_body(), 64).await.unwrap();
@@ -1003,6 +1004,15 @@ mod tests {
     #[test]
     fn resolve_account_cache_hit_key_not_found_evicts() {
         let action = cache_reverify_action(Err(OnchainVerifyError::KeyNotFound("gone".into())));
+        assert!(matches!(action, CacheReverifyAction::Evict { .. }));
+    }
+
+    #[test]
+    fn resolve_account_cache_hit_object_not_found_evicts() {
+        // Typo'd x-account-id / gRPC NOT_FOUND is a sign-in failure, not a 503.
+        let action = cache_reverify_action(Err(OnchainVerifyError::NotFound(
+            "gRPC GetObject failed: not found".into(),
+        )));
         assert!(matches!(action, CacheReverifyAction::Evict { .. }));
     }
 
--- a/services/server/src/auth.rs
+++ b/services/server/src/auth.rs
@@ -73,6 +73,28 @@ fn unsupported_legacy_sdk() -> StatusCode {
     StatusCode::UPGRADE_REQUIRED
 }
 
+/// Matches the MCP proxy: Sui could not be consulted, so this is not a login
+/// failure. Empty 401 here is what made the SDK print memwal_login (WALM-429).
+fn upstream_unavailable() -> Response {
+    (StatusCode::SERVICE_UNAVAILABLE, "upstream unavailable").into_response()
+}
+
+/// Outcome of resolving a signed delegate key to a MemWal account.
+enum AccountResolveError {
+    /// Identity could not be established. Maps to a timing-normalized 401.
+    Unauthorized(String),
+    /// On-chain lookup could not be completed. Maps to 503 — retryable.
+    Unavailable(String),
+}
+
+impl std::fmt::Display for AccountResolveError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Unauthorized(msg) | Self::Unavailable(msg) => write!(f, "{msg}"),
+        }
+    }
+}
+
 /// Whether a request whose signed timestamp is `age` seconds old (negative =
 /// future-dated) is fresh, given the accepted drift window. The window is
 /// inclusive and symmetric: `|age| <= drift`. `drift == 0` requires an exact
@@ -326,12 +348,17 @@ pub async fn verify_signature(
     }
 
     // Step 2: Resolve account — cache → signed header hint/config fallback → registry scan
-    // Always use constant_time_reject so that timing of the resolution error
-    // ("account not found" vs "key not in account") cannot be observed by callers.
+    // Identity failures stay on constant_time_reject (bare 401) so "account not
+    // found" vs "key not in account" cannot be timed. RPC/scan unavailability
+    // is 503: a Sui 429 is not a revoke (WALM-429).
     let (account_id, owner) =
         match resolve_account(&state, &public_key_hex, &pk_array, account_id_hint).await {
             Ok(pair) => pair,
-            Err(e) => {
+            Err(AccountResolveError::Unavailable(e)) => {
+                tracing::warn!("Account resolution unavailable: {}", e);
+                return Ok(upstream_unavailable());
+            }
+            Err(AccountResolveError::Unauthorized(e)) => {
                 tracing::warn!("Account resolution failed: {}", e);
                 return Err(constant_time_reject().await);
             }
@@ -366,12 +393,14 @@ async fn resolve_account(
     public_key_hex: &str,
     pk_bytes: &[u8; 32],
     account_id_hint: Option<String>,
-) -> Result<(String, String), String> {
+) -> Result<(String, String), AccountResolveError> {
     // Strategy 1: Check PostgreSQL cache
-    if let Ok(Some((cached_account_id, _cached_owner))) =
+    if let Ok(Some((cached_account_id, cached_owner))) =
         state.db.get_cached_account(public_key_hex).await
     {
-        // Verify the cached mapping is still valid onchain
+        // Re-verify the cached mapping is still valid onchain when Sui is
+        // reachable. A transient RPC failure is *not* a revoke: keep the row
+        // and serve it so a Sui 429 does not 401 a still-registered key.
         match verify_delegate_key_onchain(
             &state.http_client,
             &state.config.sui_rpc_url,
@@ -386,14 +415,24 @@ async fn resolve_account(
                 tracing::debug!("account resolved from cache: {}", cached_account_id);
                 return Ok((cached_account_id, owner));
             }
-            Err(_) => {
-                // Key was revoked on-chain. Delete the stale cache row
-                // immediately so subsequent requests don't loop: cache-hit → RPC fail →
-                // fall-through, burning RPC quota and generating log noise on every call.
+            Err(e) if e.is_unavailable() => {
                 tracing::warn!(
-                    "delegate key {} revoked on-chain for account {}; evicting from cache",
+                    "on-chain re-verify unavailable for key {} on account {} ({}); serving cached mapping",
                     public_key_hex,
-                    cached_account_id
+                    cached_account_id,
+                    e
+                );
+                return Ok((cached_account_id, cached_owner));
+            }
+            Err(e) => {
+                // Definitive miss: key gone, account deactivated, or the
+                // cached object is not a MemWalAccount. Evict so later
+                // requests don't loop cache-hit → RPC → fall-through.
+                tracing::warn!(
+                    "cached delegate key {} is stale for account {} ({}); evicting from cache",
+                    public_key_hex,
+                    cached_account_id,
+                    e
                 );
                 let _ = state.db.delete_cached_key(public_key_hex).await;
             }
@@ -410,7 +449,7 @@ async fn resolve_account(
         .as_deref()
         .or(state.config.memwal_account_id.as_deref())
     {
-        let owner = verify_delegate_key_onchain(
+        match verify_delegate_key_onchain(
             &state.http_client,
             &state.config.sui_rpc_url,
             state.sui_grpc_client.as_ref(),
@@ -419,32 +458,41 @@ async fn resolve_account(
             &state.config.package_id,
         )
         .await
-        .map_err(|e| {
-            format!(
-                "exact account {} verification failed: {}",
-                exact_account_id, e
-            )
-        })?;
+        {
+            Ok(owner) => {
+                let _ = state
+                    .db
+                    .cache_delegate_key(public_key_hex, exact_account_id, &owner)
+                    .await;
 
-        let _ = state
-            .db
-            .cache_delegate_key(public_key_hex, exact_account_id, &owner)
-            .await;
-
-        tracing::debug!(
-            "account resolved from exact account id: {}",
-            exact_account_id
-        );
-        return Ok((exact_account_id.to_string(), owner));
+                tracing::debug!(
+                    "account resolved from exact account id: {}",
+                    exact_account_id
+                );
+                return Ok((exact_account_id.to_string(), owner));
+            }
+            Err(e) if e.is_unavailable() => {
+                return Err(AccountResolveError::Unavailable(format!(
+                    "exact account {} verification unavailable: {}",
+                    exact_account_id, e
+                )));
+            }
+            Err(e) => {
+                return Err(AccountResolveError::Unauthorized(format!(
+                    "exact account {} verification failed: {}",
+                    exact_account_id, e
+                )));
+            }
+        }
     }
 
     // Strategy 3: The legacy registry scan uses JSON-RPC. Testnet no longer
     // serves JSON-RPC, so fail closed when a modern signed x-account-id hint
     // is absent instead of silently contacting a retired endpoint.
     if state.config.sui_network == "testnet" {
-        return Err(
+        return Err(AccountResolveError::Unauthorized(
             "x-account-id is required for delegate-key authentication on testnet".to_string(),
-        );
+        ));
     }
 
     // Non-testnet compatibility path: scan AccountRegistry only when no exact
@@ -453,19 +501,20 @@ async fn resolve_account(
     // unknown-key floods can't stack unbounded scans, and a per-scan page
     // cap (MEMWAL_REGISTRY_SCAN_MAX_PAGES) inside the scan itself. Both
     // rejection messages name the x-account-id remediation, but they surface
-    // only in server logs: the middleware collapses every auth failure to a
-    // bare 401 (no oracle). A key past the page cap therefore cannot
-    // self-resolve — operators must diagnose the lockout from the warn logs
-    // and either raise the cap or have the client send the header hint,
-    // which Strategy 2 verifies directly without any scan.
+    // only in server logs: the middleware collapses identity failures to a
+    // bare 401 (no oracle) and RPC/scan unavailability to 503. A key past the
+    // page cap therefore cannot self-resolve — operators must diagnose the
+    // lockout from the warn logs and either raise the cap or have the client
+    // send the header hint, which Strategy 2 verifies directly without any
+    // scan.
     let _scan_permit = match state.registry_scan_semaphore.try_acquire() {
         Ok(permit) => permit,
         Err(_) => {
-            return Err(
+            return Err(AccountResolveError::Unavailable(
                 "registry scan concurrency limit reached; retry, or send the x-account-id \
                  header hint to skip the registry scan"
                     .to_string(),
-            );
+            ));
         }
     };
     match find_account_by_delegate_key(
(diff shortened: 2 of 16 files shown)


services/server/src/auth.rs
```diff
@@ -386,14 +415,24 @@ async fn resolve_account(
...
-                // immediately so subsequent requests don't loop: cache-hit → RPC fail →
-                // fall-through, burning RPC quota and generating log noise on every call.
+            Err(e) if e.is_unavailable() => {
+                tracing::warn!(
+                    "on-chain re-verify unavailable for key {} on account {} ({}); serving cached mapping",
+                    public_key_hex,
+                    cached_account_id,
+                    e
+                );
+                return Ok((cached_account_id, cached_owner));
```
harrymove-ctrl: **Cached authorization can remain valid for up to 24 hours during a Sui outage.**

This branch authenticates with the cached mapping whenever the chain cannot be consulted. The row expires 24 hours after it was inserted or refreshed:

```sql
VALUES ($1, $2, $3, NOW() + INTERVAL '24 hours')
```

If the delegate is revoked or the account is deactivated while Sui is returning 429 or is otherwise unavailable, the old delegate can continue to authenticate until that cached row expires.

Before this PR, the request failed closed when Sui could not verify the key, although it also incorrectly evicted the cache and mislabeled the failure as 401. Keeping the row fixes the eviction bug; using the row as the authorization authority is a separate security posture change.

The safest behavior is to retain the cache row but return 503 on verification failure. If stale-cache authorization is intentional, it should have explicit security approval and a documented maximum staleness bound. A 5-15 minute bound reduces exposure compared with 24 hours, but it still permits revoked credentials during that window.

services/server/src/auth.rs
```diff
@@ -73,6 +73,28 @@ fn unsupported_legacy_sdk() -> StatusCode {
     StatusCode::UPGRADE_REQUIRED
 }
 
+/// Matches the MCP proxy: Sui could not be consulted, so this is not a login
+/// failure. Empty 401 here is what made the SDK print memwal_login (WALM-429).
+fn upstream_unavailable() -> Response {
```
harrymove-ctrl: **Please add an explicit retry/backoff signal to this response.**

Both SDKs now tell callers that this failure is retryable, but the response has no `Retry-After` header. During the Sui 429 window, immediate client retries can amplify pressure on the same dependency.

Please add a short `Retry-After` value and ensure callers use bounded exponential backoff. The 100 ms sleep used by 401 rejection exists for timing normalization; it should not be treated as load shedding or rate limiting for this 503 path.

services/server/src/storage/sui.rs
```diff
@@ -998,6 +998,19 @@ impl std::fmt::Display for OnchainVerifyError {
...
+impl OnchainVerifyError {
+    /// True when the chain could not be consulted, as opposed to a definitive
+    /// "this key is not registered / this account is dead" answer.
+    ///
+    /// HTTP signed auth and the MCP proxy must not treat these as a revoke:
+    /// a Sui gRPC 429 is `RpcError`, and logging it as "revoked on-chain"
+    /// produced intermittent empty 401s that the SDK mapped to memwal_login
+    /// (WALM-429).
+    pub fn is_unavailable(&self) -> bool {
+        matches!(self, Self::RpcError(_) | Self::ScanCapExceeded(_))
```
harrymove-ctrl: **Suggest an exhaustive `match` instead of `matches!`.**

This is correct for the five variants that exist today. But if someone later adds a `Timeout` or `Transport` variant, it silently defaults to `false`, becomes a 401, and reintroduces exactly this bug with no compiler warning.

Since this function is now the single source of truth for both the HTTP and MCP auth paths, an exhaustive `match` with one arm per variant is worth the extra four lines: it forces the next author to make the call deliberately.

services/server/src/auth.rs
```diff
@@ -453,19 +501,20 @@ async fn resolve_account(
...
+    // bare 401 (no oracle) and RPC/scan unavailability to 503. A key past the
+    // page cap therefore cannot self-resolve — operators must diagnose the
+    // lockout from the warn logs and either raise the cap or have the client
+    // send the header hint, which Strategy 2 verifies directly without any
+    // scan.
     let _scan_permit = match state.registry_scan_semaphore.try_acquire() {
         Ok(permit) => permit,
         Err(_) => {
-            return Err(
+            return Err(AccountResolveError::Unavailable(
```
harrymove-ctrl, line 558: **This reclassifies semaphore exhaustion from 401 to 503, which is beyond the RPC-error scope in the PR title.**

Registry-scan saturation is not an RPC error, and it previously returned a timing-normalized 401. I think the change is correct, it genuinely is retryable, but it is a behavior change the PR body does not mention, and it makes scan saturation externally observable.

Low severity. Worth a line in the description.

services/server/src/auth.rs
```diff
@@ -856,6 +919,31 @@ mod tests {
...
+        assert!(
+            resp.headers().get("x-auth-error").is_none(),
+            "503 must not look like an identity reject"
+        );
+        let body = axum::body::to_bytes(resp.into_body(), 64).await.unwrap();
+        assert_eq!(&body[..], b"upstream unavailable");
+    }
+
+    #[test]
+    fn rpc_errors_are_unavailable_not_stale_cache() {
```
harrymove-ctrl: **This test leaves `OnchainVerifyError` as an unused non-test import and duplicates the `sui.rs` assertions.**

The old production use in `auth.rs` was removed with the `ScanCapExceeded` match arm. The remaining reference is this test through `use super::*`, while `storage/sui.rs` already tests the same variant classification.

A normal build/test run emits the unused-import warning, but CI does not fail because warnings are not denied. Either remove this duplicate test and the import, or move `use crate::storage::sui::OnchainVerifyError;` into the test module.

More importantly, replace this duplicate classification test with a `resolve_account` regression test covering cache hit + `RpcError`, because that is the behavior WALM-429 actually changes.

harrymove-ctrl (commented):
## Review: correct diagnosis, but the cache fallback needs a security decision

I traced this independently from the production Railway logs and reached the same root cause: a transient Sui gRPC failure was classified as delegate revocation, the cache row was evicted, and signed HTTP auth returned an empty 401. Splitting transient upstream failure from definitive identity rejection is the right direction.

### What is right

1. The new classification correctly separates `RpcError` / `ScanCapExceeded` from `KeyNotFound`, `AccountDeactivated`, and `WrongObjectType` for the five variants that exist today.
2. HTTP auth and the MCP proxy now share the same `OnchainVerifyError` availability classification.
3. TypeScript and Python SDK changes are versioned and tested, including empty-body 503.
4. Release manifests, changelogs, and the manual release verifier are synchronized.
5. The troubleshooting entry accurately explains that this production incident was not fixed by signing in again.

### Findings

**1. Security: cached authorization remains valid during a Sui outage.**

On a cache hit, any error classified as unavailable now authenticates with the cached account and owner. The row can remain valid for up to 24 hours from its last insert or refresh. If the delegate is revoked or the account is deactivated while Sui cannot be consulted, the old delegate can continue to call remember, recall, restore, analyze, and other protected routes.

Keeping the row is correct because a transient 429 must not delete it. Using that row as the authorization authority is a separate security posture change. The fail-closed alternative is: retain the cache row, return 503, and retry verification later. If stale-cache authorization is intentional, it needs explicit security approval and a documented maximum staleness bound before merge.

**2. SDKs classify every HTTP 503 as credential-verification failure.**

The server also uses 503 for Redis/rate-limiter outages, LLM or embedder failures, and other upstream dependencies. The new status-only branches in both SDKs replace those distinct failures with a credential-specific message. Auth should emit a machine-readable code or header, and the SDKs should use the credential copy only for that code. Other 503 responses should remain generic or retain their sanitized server message.

**3. The changed cache behavior has no regression test.**

The new Rust tests cover error classification and the response helper, but not `resolve_account`. There is no test proving that a transient failure retains the cache row, that definitive rejection evicts it, or that the selected response is 503 versus authenticated success. This is the exact regression surface of WALM-429.

**4. Build hygiene: `OnchainVerifyError` is now an unused non-test import in `auth.rs`.**

The warning is visible in a normal build/test run but does not fail CI because warnings are not denied. Remove the duplicate auth test/import or move the import into the test module.

**5. The auth-specific 503 response should provide a backoff signal.**

Add a short `Retry-After` header and ensure SDK/MCP retry behavior uses bounded exponential backoff. The existing 100 ms sleep is a timing-normalization control for 401 identity rejection, not a substitute for rate limiting or backoff.

**6. `is_unavailable()` should use an exhaustive match.**

This function is now shared classification policy. An exhaustive arm for every enum variant forces future variants to receive an explicit security decision instead of silently defaulting to 401.

**7. Registry-scan semaphore exhaustion is also reclassified from 401 to 503.**

That is probably correct because it is retryable, but it is beyond the RPC-only wording in the title and should be documented as an intentional behavior change.

### Scope and follow-ups

- Playground/dashboard: receives the workspace SDK fix.
- TypeScript SDK: version 0.1.6.
- Python SDK: version 0.1.9.
- `memwal-mcp`: has no SDK dependency; its SSE bridge surfaces generic HTTP 503 but does not currently retry it.
- This PR corrects the false 401 classification; it does not eliminate the underlying Sui 429 pressure. The unkeyed endpoint and retry amplification need a separate production follow-up.
- The known old-package account remains a genuine `WrongObjectType` 401 and needs separate migration.

### Disposition

I would not merge the stale-cache authorization behavior without either:

1. changing cache-hit RPC unavailability to retain the row but return 503, or
2. recording explicit security approval for bounded stale-cache authorization and testing that bound.

The SDK-wide 503 misclassification and the missing `resolve_account` regression test should also be fixed in this PR.

packages/sdk/src/utils.ts
```diff
@@ -321,6 +321,18 @@ export function sanitizeServerError(
         };
     }
 
+    // Relayer 503 = Sui RPC/gRPC could not be consulted (WALM-429). This is
+    // retryable and is not a sign-in failure — do not send callers to
+    // memwal_login.
+    if (Number(status) === 503) {
```
harrymove-ctrl: **Do not classify every HTTP 503 as a credential-verification failure.**

This helper is used for all SDK requests, while the relayer also returns 503 for Redis or rate-limiter outages, LLM/embedder failures, and other upstream dependencies. For example, a 503 body saying `Rate limiter temporarily unavailable` is replaced with `cannot verify credentials`.

Please make the auth middleware emit a machine-readable code or header such as `AUTH_UPSTREAM_UNAVAILABLE`, and select this credential-specific copy only when that signal is present. Other 503 responses should keep a generic retryable message or their existing sanitized server message. The Python SDK has the same status-only issue in `_HttpStatusError`.

services/server/src/auth.rs
ducnmm: Addressed in 27781e0e: fail-closed. Cache hit + `RpcError` / `ScanCapExceeded` keeps the row (so a later verify can succeed) but returns 503 — we do not authenticate from the cached mapping while Sui is unreachable. Revoked keys therefore cannot ride the 24h TTL through an outage.

services/server/src/auth.rs
ducnmm: Addressed in 27781e0e: auth 503 now sends `Retry-After: 5` and `x-auth-error: AUTH_UPSTREAM_UNAVAILABLE`. No 100ms 401 pad on this path. TS/Python attach `retryAfterSeconds` / `retry_after` on the thrown error so callers can bound backoff; remember-job polling already uses jittered exponential backoff on 5xx.

services/server/src/storage/sui.rs
ducnmm: Addressed in 27781e0e: `is_unavailable` is an exhaustive `match` over all five variants, so a new `Timeout` / `Transport` arm is a compile error instead of a silent 401.

services/server/src/auth.rs
ducnmm: Agreed — registry-scan semaphore exhaustion is retryable, not a revoke, so it stays 503. Called out in the PR body in 27781e0e's description update.

services/server/src/auth.rs
ducnmm: Addressed in 27781e0e: dropped the duplicate `is_unavailable` assertions (those live in `sui.rs`). Added `cache_reverify_action` tests for cache-hit + `RpcError` (keep row, do not auth), `KeyNotFound` (evict), and `Ok` (authenticate).

packages/sdk/src/utils.ts
ducnmm: Addressed in 27781e0e: auth 503 emits `x-auth-error: AUTH_UPSTREAM_UNAVAILABLE`. TS `sanitizeServerError` / Python `_HttpStatusError` only use the credential-verification copy when that header is present. A Redis or rate-limiter 503 keeps the generic sanitized body.

jessiemongeon1:
## Style Guide Audit

Audited **3** file(s) against the [Sui Documentation Style Guide](https://docs.sui.io/references/contribute/style-guide).

### `docs/sdk/changelog.mdx` (1 violation(s))

_1 violation(s) (0 regex, 1 claude)_

- **Line 45** — passive_voice
  - Current: `HTTP 503 from the relayer is reported as a retryable upstream outage`
  - Fix: `The SDK reports HTTP 503 from the relayer as a retryable upstream outage`

---

ducnmm:
Rebased onto `origin/dev` from a worktree (`MemWal-prs/walm-429`). Python stays **0.1.9**; the AUTH_UPSTREAM_UNAVAILABLE changelog bullet sits next to the existing `remember_bulk_async` job_ids fix. TS stays **0.1.6**. Conflicts were changelog-only.

services/server/src/storage/sui.rs
```diff
@@ -998,6 +998,22 @@ impl std::fmt::Display for OnchainVerifyError {
...
+    /// True when the chain could not be consulted, as opposed to a definitive
+    /// "this key is not registered / this account is dead" answer.
+    ///
+    /// HTTP signed auth and the MCP proxy must not treat these as a revoke:
+    /// a Sui gRPC 429 is `RpcError`, and logging it as "revoked on-chain"
+    /// produced intermittent empty 401s that the SDK mapped to memwal_login
+    /// (WALM-429).
+    pub fn is_unavailable(&self) -> bool {
+        match self {
+            Self::RpcError(_) | Self::ScanCapExceeded(_) => true,
```
harrymove-ctrl, line 1016: `RpcError` is not only "the chain could not be consulted" — it is also the catch-all for several permanent conditions, so `is_unavailable() == true` is wrong for them:

```rust
// sui.rs:460 — caller sent a malformed x-account-id
.map_err(|e| OnchainVerifyError::RpcError(format!("invalid object id: {}", e)))?
// sui.rs:480 — includes tonic NOT_FOUND: the object does not exist
.map_err(|e| OnchainVerifyError::RpcError(format!("gRPC GetObject failed: {}", e)))?
// sui.rs:483
.ok_or_else(|| OnchainVerifyError::RpcError("gRPC response missing object".into()))?;
```

Two concrete consequences after this PR:

1. A client with a wrong or typo'd `x-account-id` gets a 503 and the SDK tells them "Retry; this is not a sign-in failure" — indefinitely — instead of the 401 they used to get. That is precisely the failure mode WALM-435 is trying to make legible.
2. A deleted account object (gRPC `NOT_FOUND`) now keeps its cache row and 503s until the 24h TTL expires. Before this PR it evicted and returned 401, which is the correct handling for a definitive miss.

The variant-level match is right; the granularity underneath it is not. Cleanest fix is to classify on `tonic::Status::code()` at the call site — `NOT_FOUND` / `INVALID_ARGUMENT` are definitive, `UNAVAILABLE` / `RESOURCE_EXHAUSTED` / `DEADLINE_EXCEEDED` are unavailable — or, minimally, add a `NotFound` variant so the exhaustive match forces the decision.

services/server/src/auth.rs
```diff
@@ -73,6 +72,73 @@ fn unsupported_legacy_sdk() -> StatusCode {
...
+/// failure. Empty 401 here is what made the SDK print memwal_login (WALM-429).
+///
+/// Fail-closed: do not authenticate from a cached mapping while the chain
+/// is unreachable. Keep the cache row (do not evict), return 503 with
+/// `x-auth-error: AUTH_UPSTREAM_UNAVAILABLE` and `Retry-After`.
+pub(crate) fn upstream_unavailable() -> Response {
+    (
+        StatusCode::SERVICE_UNAVAILABLE,
+        [
+            ("retry-after", "5"),
```
harrymove-ctrl: Two things on this response:

**Blocking:** `Retry-After` never reaches browser callers. `relayer_cors()` (`main.rs:93`) exposes only `x-auth-error`, and `relayer_cors_exposes_only_x_auth_error` pins that with `assert_eq!(names.len(), 1, ...)`. `Retry-After` is not CORS-safelisted, so Fetch hides it and `err.retryAfterSeconds` stays `undefined` for the dashboard and the browser SDK — the two clients most able to hammer a 429ing fullnode. Add `retry-after` to `expose_headers` and update that assertion to 2. The message copy is fine either way since `x-auth-error` is already exposed.

**Nit:** `AUTH_UPSTREAM_RETRY_AFTER_SECS: u64 = 5` is declared just above but this literal `"5"` is what actually ships; the constant is referenced only from `upstream_unavailable_is_503_not_401`. The test asserts the constant equals 5 and separately asserts the header equals `"5"`, so a future edit to one will not fail on the other. Build the header value from the constant instead.

harrymove-ctrl (changes requested):
## Request changes: two narrow fixes, then this is good to merge

I re-reviewed the fail-closed rework (`9fee45f0`) against the code, not just the diff. All six findings from my first pass are genuinely addressed — `is_unavailable` is an exhaustive `match`, the SDKs gate the credential copy on `x-auth-error`, the duplicate classification test is gone and replaced with `cache_reverify_action` coverage, and the fail-closed posture removes the 24h stale-auth exposure entirely. CI is green across all 23 checks, and the version story checks out (npm latest is 0.1.5 so 0.1.6 is new; PyPI latest is 0.1.8 so the 0.1.9 bullet lands in an unreleased section).

Two things still need fixing before merge. Both are small and both sit inside the scope this PR already claims.

### Blocking

**1. `Retry-After` is not CORS-exposed, so the backoff signal is a no-op for browser clients.**

`relayer_cors()` in `services/server/src/main.rs:93` exposes exactly one header, and `relayer_cors_exposes_only_x_auth_error` asserts `names.len() == 1`:

```rust
.expose_headers(["x-auth-error".parse::<header::HeaderName>().unwrap()])
```

`Retry-After` is not a CORS-safelisted response header, so `res.headers.get("retry-after")` returns `null` in a browser and `err.retryAfterSeconds` is never set for the dashboard or the browser SDK. `x-auth-error` is exposed, so the *message* is correct — only the bounded-backoff signal is lost, and it is lost for the client population most likely to retry-storm. Node and Python are unaffected.

Fix is two lines: add `retry-after` to `expose_headers` and bump the `names.len()` assertion to 2.

**2. `is_unavailable()` classifies definitive failures as retryable.**

The exhaustive `match` makes the *variant* mapping deliberate, which was the point of my earlier comment — but `RpcError` itself conflates "the chain could not be consulted" with several permanent conditions. See the inline comment on `storage/sui.rs`. Net effect: a typo'd `x-account-id` and a gRPC `NOT_FOUND` both now return a 503 telling the caller "Retry; this is not a sign-in failure" forever, and on a cache hit the row is retained instead of evicted — the opposite of what those cases want. Since this PR promotes `is_unavailable` to "the single source of truth for both the HTTP and MCP auth paths", the granularity matters more now than it did before.

### Not blocking, but please file follow-ups

**3. `routes/oauth.rs:686` still has the exact bug this PR fixes.**

```rust
let verified_owner = verify_delegate_key_onchain(...)
    .await
    .map_err(|e| OAuthError::invalid_request(format!("delegate key is not registered on-chain: {e}")))?;
```

A Sui 429 here tells the user their delegate key is not registered. Worse than the recall path: `consume_oauth_session()` runs immediately above and has already atomically burned the session, so the user cannot retry — they have to restart the whole authorize flow. Different error shape (OAuth errors are not HTTP-503-shaped), so I would not widen this PR for it, but it should not be lost.

**4. The reported symptom is reclassified, not resolved.**

WALM-429 reports "recall works some of the time and fails some of the time". After this PR, recall still fails during a Sui 429 burst — as a 503 instead of a 401. That is within the scope of the PR title and it is the right first step, but WALM-429 should not close on merge.

The pressure source is that the relayer issues a `GetObject` on *every* signed request; the 24h cache only saves the registry scan, not the RPC. A short per-key re-verify window (re-verify at most every 30-60s) would cut on-chain calls by roughly two orders of magnitude and bound revocation staleness to a minute — strictly better than either 24h stale-auth or failing closed on every transient blip.

One related note for the PR body: fail-closed also removes an accidental retry. Previously a cache-hit RPC failure evicted and fell through to Strategy 2, which issued a second `GetObject` that sometimes succeeded. That path is gone now. I think the trade is correct (it halves amplification during exactly the burst that caused this), but it is a second behavior change beyond the title, alongside the semaphore reclassification you already documented.

### Nits

- `AUTH_UPSTREAM_RETRY_AFTER_SECS` is declared but `upstream_unavailable()` hardcodes `"5"`; the constant is only referenced from the test, so the two can drift silently. Have the response build its header from the constant.
- `cache_reverify_action` is well tested, but the wiring in `resolve_account` that turns `UnavailableKeepCache` into a 503 is still uncovered. A single middleware-level test would close the loop on the actual WALM-429 regression surface.
- Branch is 8 commits behind `dev` (mergeable, no conflicts) — worth a rebase before merge.

ducnmm:
Addressed the 3 Sep CHANGES_REQUESTED on `f5243f7b` (rebased on `dev`).

**Blocking**
1. CORS now exposes `Retry-After` alongside `x-auth-error` (`relayer_cors`; test asserts both).
2. GetObject is classified by gRPC **code**: `NOT_FOUND` / `INVALID_ARGUMENT` / invalid object id / missing object → `NotFound` (evict, 401). `UNAVAILABLE` / `RESOURCE_EXHAUSTED` / `DEADLINE_EXCEEDED` → `RpcError` (503, keep cache, fail-closed). Field/json parse on an object that *was* returned stays `RpcError` so a live key is not 401+evicted.

**Nit:** `Retry-After` is built from `AUTH_UPSTREAM_RETRY_AFTER_SECS`.

Out of scope (not in this PR): oauth.rs 429-as-unregistered; per-key re-verify window.

Ready for re-review. @harrymove-ctrl

ducnmm:
@harrymove-ctrl rebased onto `origin/dev` after #862 (`ba1ad54c`). Changelog conflicts only — WALM-429 bullets sit under existing 0.1.6. Please re-review; still CHANGES_REQUESTED from 3 Sep.

ducnmm:
@harrymove-ctrl gentle ping — still waiting on review of `ba1ad54c1f8011bc9bbf353bfbd5fcfc3da90162`.

harrymove-ctrl (approved):
## Approve — both blocking items are fixed, verified against `ba1ad54c`

Re-reviewed the code, not just the diff. Both blockers from my 3 Sep pass are genuinely resolved, and two of the three fixes went further than what I asked for.

### Blocking items — resolved

**1. `Retry-After` is now CORS-exposed.** `relayer_cors()` exposes `header::RETRY_AFTER` alongside `x-auth-error`, and the pinning test was renamed to `relayer_cors_exposes_x_auth_error_and_retry_after` with an explicit presence assertion for each plus `names.len() == 2`. The dashboard and browser SDK can now read `err.retryAfterSeconds`, which was the population most able to retry-storm a 429ing fullnode.

**2. `is_unavailable()` granularity is fixed at the right layer.** You added a `NotFound` variant *and* classified on `tonic::Status::code()` rather than picking one. `map_get_object_status` maps `NOT_FOUND` / `INVALID_ARGUMENT` to definitive and everything else to `RpcError`, and the comment about why message-text matching would misfire on `INTERNAL` errors that merely mention a missing object is exactly the right reason to have written it that way. `parse_object_id` and the JSON-RPC missing-`result` / missing-`content` paths are `NotFound` too, so the typo'd `x-account-id` case gets its 401 back instead of an indefinite "retry this" 503. `get_object_status_classifies_by_grpc_code` covers all six codes including the `INTERNAL`-mentions-not-found trap.

**Nit resolved:** `upstream_unavailable()` builds the header from `AUTH_UPSTREAM_RETRY_AFTER_SECS`, so the constant and the wire value can no longer drift.

### Three improvements not mentioned in the PR body

All good, just noting them so they are not invisible in the history:

- `grpc_get_object` dedups the GetObject call and its `observe_external` between the verify and list paths. Those were two copies that could have diverged.
- `mcp_proxy.rs` switched to `err.is_unavailable()` and the shared `crate::auth::upstream_unavailable()`. MCP 503s previously shipped as a bare `(503, "upstream unavailable")` with no headers; they now carry `Retry-After` and `x-auth-error` like the HTTP path. That is a real behavior improvement for MCP clients.
- `find_account_by_delegate_key` now treats `NotFound` like `KeyNotFound` and continues the scan. Previously a single deleted account object in the registry aborted the whole scan with an error for every caller.

### Deliberate call worth recording

`fe49ee4b` keeps field-parse failures (`json_account_active` on an object missing `active`) as `RpcError`, i.e. retryable. That is structurally the same thing I flagged in blocker 2: a permanent condition advertised as retryable. I am fine with it here, and the reasoning holds — the object was returned and already passed `ensure_memwal_account_type`, so the ambiguity is genuine, and not evicting a live user's cache row is the safer side to err on. Flagging it only so the trade is written down: if the Move contract ever renames `active`, every user gets a 503 forever with no eviction and no 401 to make the cause legible.

### Follow-ups — not blocking this merge

1. `routes/oauth.rs:686` still maps a Sui 429 to "delegate key is not registered on-chain". Worse than the recall path was, because `consume_oauth_session()` has already burned the session, so the user cannot retry without restarting the authorize flow. Out of scope here (different error shape), but it should not be lost.
2. **WALM-429 should not close on merge.** Recall still fails during a Sui 429 burst — now as an honest 503 rather than a misleading 401. That is the right first step and matches the PR title, but the reported symptom is reclassified, not resolved. The pressure source is the `GetObject` on every signed request; a short per-key re-verify window (30-60s) would cut on-chain calls by roughly two orders of magnitude and bound revocation staleness to a minute.
3. The `UnavailableKeepCache` -> 503 wiring in `resolve_account` is still only covered through the `upstream_unavailable()` helper test. One middleware-level test would close the loop on the actual WALM-429 regression surface.
4. The style-guide audit comment from 27 Aug is still outstanding — `docs/sdk/changelog.mdx` line 45, passive voice, and the same phrasing appears in the frontmatter `answer`. Not a required check, so it does not gate CI, but the bot asks for it before merge.

CI is green on `ba1ad54c` (20 pass, 1 skipped), no conflicts. Branch is 3 commits behind `dev` — worth updating before merge if branch-up-to-date is enforced.

PR #812, opened 2026-08-27
fix(plugin): bundle the memwal skill so plugin install ships it

harrymove-ctrl:
## Problem

`memwal@memwal-plugins` shipped **zero skills**. `SKILL.md` sat at the repo root where:

1. no plugin manifest referenced it,
2. it was outside `packages/mcp/plugin/`, so `plugin install` never delivered it,
3. it was excluded from `packages/mcp/package.json` `files` (`["dist", "plugin", "README.md"]`), so npm never shipped it either.

Net effect: the single-file integration guide reached nobody through the plugin.

This also blocked the documented onboarding path. `apps/app/public/skills/setup` correctly tells agents to install the plugin, but the installed plugin then carried no skill.

## Change

1. `git mv SKILL.md packages/mcp/plugin/skills/memwal/SKILL.md` so it travels with the plugin. History preserved.
2. Bump plugin manifests and both marketplaces `0.0.12` -> `0.0.13`: `plugin.json`, `.claude-plugin/plugin.json`, `.cursor-plugin/plugin.json`, `.codex-plugin/plugin.json`, `.claude-plugin/marketplace.json`, `.cursor-plugin/marketplace.json`. The codex marketplace carries no version field.
3. Update the two links that pointed at the old root path: `README.md` and `docs/contributing/run-repo-locally.md`.

`packages/mcp/package.json` stays at `0.0.12` on purpose. That version belongs to the manual release flow, and the plugin resolves the MCP server unpinned via `npx -y @mysten-incubation/memwal-mcp`, so plugin `0.0.13` works against npm `0.0.12`.

## Verification

Installed from this branch via the plugin CLI and read the component inventory.

Before, installed from the default branch:

```
memwal 0.0.12
  Skills (0)
  Hooks (3)   MCP servers (1)
```

After, installed from this branch:

```
memwal 0.0.13
  Skills (1)  memwal
  Hooks (3)   MCP servers (1)
```

The full documented flow was also exercised end to end from GitHub, marketplace add then install then list, and both steps succeeded.

The three lifecycle hooks were run directly against stdin and both `SessionStart` and `UserPromptSubmit` returned their `additionalContext` payloads, confirming automatic recall and save guidance still arrives without any user-level system-prompt block.

## Note, out of scope

Walrus Memory is not listed in the official plugin Directory, so searching it there returns no results and users must add the marketplace by repository. Getting listed is a separate submission and is not addressed here.


jessiemongeon1:
## Style Guide Audit

harrymove-ctrl:
Closing as superseded by PR #964 (WALM-669) which generates the plugin tree directly from the monorepo.

PR #822, opened 2026-08-28
fix(chatbot): restore chat, titles and artifacts after upstream model retirements (WALM-354)

HoangDucBach:
Closes WALM-354.

## Summary

Restores chat, titles and artifacts after OpenRouter retired four of the seven curated model ids. Replaces those ids, caps `maxOutputTokens` on every model call, resolves a stale `chat-model` cookie to the default instead of forwarding it to `/api/chat`, and points the Playwright mock at the ids the app actually sends. Adds six conversation e2e tests, a guard that counts model calls against capped calls, and a weekly workflow checking the curated ids still exist upstream.

### Four of the seven curated model ids no longer exist

OpenRouter answers a retired id with a 404 at request time, so picking one produced only "Oops, an error occurred!". Two were also hardcoded at their call sites: `getTitleModel` used `google/gemini-2.0-flash-001`, so every chat stayed named "New chat", and `getArtifactModel` used `anthropic/claude-3.5-haiku`, so document creation broke for everyone rather than only whoever picked it.

### No call site capped the output tokens

Every request was quoted against the model's whole output window. OpenRouter reserves that credit up front for Anthropic and Google, and rejected those turns with "requires more credits, you requested up to 65535 tokens" however short the answer would have been. OpenAI and DeepSeek do not reserve, which is why only part of the picker looked broken.

The by-hand sweep that added the caps then missed `requestSuggestions`, which streams through `getArtifactModel` from `lib/ai/tools` rather than beside the artifact handlers. That call is capped here too, and the guard below counts call sites instead of naming them so the next one cannot slip through the same way.

### A stale cookie rejected every send

Both chat pages forwarded `chat-model` verbatim, and `/api/chat` rejects an id missing from `chatModels`. Anyone holding a since-retired selection had every send fail as a 400, with nothing to explain it: the picker's fallback is display-only, so it rendered a valid model name while the request body still carried the dead id.

### The mock provider could not answer a real model id

Under `PLAYWRIGHT`, `getLanguageModel` passed catalog ids to a provider that registers behaviours rather than ids, so every turn threw `NoSuchModelError`. The suite stayed green because no test asserted that a reply arrives. The mock also matched keywords across the whole serialised prompt, where the system prompt contains "hi" inside "this", so it always answered with the greeting.

### The Playwright warm-up compiled nothing

`global-setup.ts` fetched with `redirect: "manual"`, so the proxy's guest redirect ended each request in under 60ms without rendering the route it named. Tests then paid the compile themselves: from a cleared `.next`, 4 to 14 of them failed on first-hit navigations of 40s to 1.3m against a 30s navigation timeout. CI never reported it because by the second of its two retries the routes were warm.

## CI

- `Chatbot / Playwright E2E` already existed on `dev`, so no new e2e job was needed. It now runs the conversation tests (including createDocument artifacts), confirmed in this PR's job log rather than only locally
- `Model ids live on OpenRouter` is new, in `.github/workflows/check-model-ids.yml`, running weekly and on pull requests that touch the curated list
- it gets its own workflow because the failure it catches arrives with no commit attached, and `test.yml` has no schedule trigger to hang a weekly run off
- the catalog endpoint is public, so it needs no key and no environment, and the paths filter keeps unrelated pull requests from taking a network dependency on OpenRouter
- `Chatbot / Lint + Build` picks up the new unit tests, including the guard that counts model calls against capped calls
- the lint step in that job is `continue-on-error: true` on `dev`, because `ultracite` passes a flag the bundled biome rejects, so it is not a gate for this branch either
- the warm-up fix matters most in CI: with two retries configured, a cold runner was passing on the second attempt rather than the first, so the failures were absorbed rather than reported

## Verification

- [x] `test:unit`: 26 pass, covering the retired ids, the `-thinking` strip, cookie resolution, both page call sites, and every capped model call
- [x] the cap guard fails as intended: removing the `requestSuggestions` cap turns it red, restoring it turns it green
- [x] `test:e2e` from a cleared `.next`: 29 pass, 1 skipped, including a new createDocument artifact test. The same cleared-cache run failed 4 to 14 before the warm-up fix
- [x] `check:models`: all 6 ids live on the public catalog
- [x] `next build` in `apps/chatbot`: clean
- [x] CI on this PR: Chatbot Lint + Build, Chatbot Playwright E2E and Model ids live on OpenRouter all pass

## Out of scope

- e2e always runs against the mock provider, so CI never calls OpenRouter. A live smoke test would need a key and a spend decision
- `saveMemory` and the MemWal path have no CI coverage, which needs a relayer and credentials
- whether `providerOptions.anthropic.thinking` is applied through OpenRouter's OpenAI-compatible endpoint is unverified, so reasoning may not behave as designed even though it streams

### DIFF
--- a/apps/chatbot/components/document.tsx
+++ b/apps/chatbot/components/document.tsx
@@ -38,6 +38,7 @@ function PureDocumentToolResult({
   return (
     <button
       className="flex w-fit cursor-pointer flex-row items-start gap-3 rounded-xl border bg-background px-3 py-2"
+      data-testid="artifact-create-result"
       onClick={(event) => {
         if (isReadonly) {
           toast.error(
--- a/apps/chatbot/lib/ai/models.ts
+++ b/apps/chatbot/lib/ai/models.ts
@@ -86,6 +86,17 @@ export function isReasoningModelId(modelId: string): boolean {
   );
 }
 
+// Every id the app can send upstream. OpenRouter retires ids on its own
+// schedule and answers a retired one with a 404 at request time, so keep the
+// full set in one place for check-model-ids.ts to verify against the catalog.
+export const openRouterModelIds = [
+  ...new Set([
+    ...chatModels.map((m) => baseOpenRouterId(m.id)),
+    TITLE_MODEL,
+    ARTIFACT_MODEL,
+  ]),
+];
+
 // Group models by provider for UI
 export const allowedModelIds = new Set(chatModels.map((m) => m.id));
 
--- a/apps/chatbot/lib/ai/models.ts
+++ b/apps/chatbot/lib/ai/models.ts
@@ -24,21 +24,21 @@ export const chatModels: ChatModel[] = [
   },
   // Anthropic
   {
-    id: "anthropic/claude-3.5-haiku",
-    name: "Claude 3.5 Haiku",
+    id: "anthropic/claude-haiku-4.5",
+    name: "Claude Haiku 4.5",
     provider: "anthropic",
     description: "Fast and affordable, great for everyday tasks",
   },
   {
-    id: "anthropic/claude-3.5-sonnet",
-    name: "Claude 3.5 Sonnet",
+    id: "anthropic/claude-sonnet-4.5",
+    name: "Claude Sonnet 4.5",
     provider: "anthropic",
     description: "Best balance of speed and intelligence",
   },
   // Google
   {
-    id: "google/gemini-2.0-flash-001",
-    name: "Gemini 2.0 Flash",
+    id: "google/gemini-2.5-flash",
+    name: "Gemini 2.5 Flash",
     provider: "google",
     description: "Ultra fast and affordable",
   },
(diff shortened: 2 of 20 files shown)


ducnmm:
Need this landed on `dev` before 13:00 ICT today (2026-09-07). CI is green and no review is requested yet — please request review and merge so this can ride the staging → prod promotion this afternoon.

apps/chatbot/lib/ai/models.ts
```diff
@@ -49,18 +49,83 @@ export const chatModels: ChatModel[] = [
...
+
+// An uncapped request is billed as if it will emit the model's whole output
+// window, and OpenRouter reserves that much credit up front for Anthropic and
+// Google models. Left unset, picking one of those rejects the turn outright
+// ("requires more credits ... you requested up to 65535 tokens") however short
+// the answer would have been.
+export const MAX_OUTPUT_TOKENS = 4096;
+// Has to clear the thinking budget the chat route asks for, or the answer is
+// truncated before any of it reaches the reply.
+export const MAX_REASONING_OUTPUT_TOKENS = 12_000;
```
ducnmm, line 83: **[suggestion]** `MAX_REASONING_OUTPUT_TOKENS` is 12_000 while the chat route still requests `budgetTokens: 10_000`. If OpenRouter honors Anthropic thinking, `max_tokens` is the combined thinking+answer budget, so only ~2_000 tokens remain for the visible reply on the model advertised as "extended thinking for complex problems." The unit test only asserts the cap is greater than the budget (and less than 32k), so this headroom can shrink further without failing.

**Suggestion:** Raise the reasoning cap enough to leave a real answer window after the thinking budget (for example 16k–20k, still well under the 32k reservation that was failing). If thinking options are known not to pass through OpenRouter's OpenAI-compatible endpoint, drop the "must clear the budget" coupling and size the cap for the reply alone.

apps/chatbot/lib/ai/models.ts
```diff
@@ -49,18 +49,83 @@ export const chatModels: ChatModel[] = [
...
     description: "Extended thinking for complex problems",
   },
 ];
 
+/**
+ * Models for the two background calls the picker never covers: chat titles and
+ * artifact content.
+ *
+ * Declared beside the curated list rather than hardcoded at the call site. The
+ * previous inline ids ("google/gemini-2.0-flash-001" for titles,
```
ducnmm, line 67: **[suggestion]** The `TITLE_MODEL` / `ARTIFACT_MODEL` JSDoc retells the retirement incident and the previous hardcoded ids. Similar incident-history blocks appear in `models.unit.test.ts`, `.github/workflows/check-model-ids.yml`, and the Playwright warm-up comments. Comments here should state the current invariant (keep these ids in `chatModels`; `-thinking` is local), not the archaeology of how they used to 404.

**Suggestion:** Collapse to one or two lines that say why the constants exist (background calls are not picker-selected; they must stay in the curated set). Move the 2026-08 narrative out of production comments; a changelog or ticket link is enough for the history.

apps/chatbot/components/document.tsx
```diff
@@ -38,6 +38,7 @@ function PureDocumentToolResult({
   return (
     <button
       className="flex w-fit cursor-pointer flex-row items-start gap-3 rounded-xl border bg-background px-3 py-2"
+      data-testid="artifact-create-result"
```
ducnmm, line 41: **[nit]** `data-testid="artifact-create-result"` is on `PureDocumentToolResult`, which also renders update and request-suggestions results. The new conversation e2e never uses this id; it waits on `document-preview` instead.

**Suggestion:** Name the test id after the shared component (for example `artifact-tool-result`) or only set a create-specific id when `type === "create"`. Point the e2e at this id if the intent was to cover the post-open chip, not the inline preview.

ducnmm (approved):
## Summary

This is a solid, well-scoped restore of chatbot chat, titles, and artifacts after OpenRouter retired four curated ids. Title/artifact models now point at live catalog ids, every `streamText`/`streamObject`/`generateText` call is capped, stale `chat-model` cookies are coerced at the page read sites, and the Playwright mock finally answers the ids the app actually sends. CI is green (unit, e2e, and the new catalog check). Residual risk is small: the reasoning cap leaves little room for the visible answer if Anthropic's thinking budget is honored, and several new comments narrate the incident instead of stating a current invariant.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 1

Issue #762, opened 2026-08-24
memwal_restore always returns truncated=true, even for an empty namespace or when total is far below limit


EAZITECH1:
## Summary

`memwal_restore` returns `truncated=true` unconditionally, including when the namespace is empty
and when `total` is far below the requested `limit`. The MCP layer then renders the advisory
*"⚠️ More blobs remain to restore — increase limit and call again."*

The tool's own description instructs the caller to act on that flag: *"If truncated=true, increase
limit and call again."* Since the flag never clears, an agent following the documented contract
loops on a restore that is already complete, and at `limit=100` it has nowhere left to go.

## Environment

- `@mysten-incubation/memwal-mcp` **0.0.7-rc.0**, MCP tools via Claude Code
- Relayer `https://relayer.memory.walrus.xyz`, `memwal_health` → `status=ok version=0.1.0`
- Walrus **Mainnet**, 2026-08-24

## Reproduction

Real namespace, 2 blobs, asking for up to 100:

```
memwal_restore(namespace="walrus-notes::meta", limit=100)
→ total=2  restored=0  skipped=2  truncated=true
  ⚠️ More blobs remain to restore — increase limit and call again.
```

`total=2` and `limit=100`. Nothing remains, yet the flag is set.

A namespace that has never existed:

```
memwal_restore(namespace="walrus-notes-nonexistent-probe", limit=100)
→ total=0  restored=0  skipped=0  truncated=true
```

```
memwal_restore(namespace="walrus-notes-nonexistent-probe", limit=10)
→ total=0  restored=0  skipped=0  truncated=true
```

`total=0`, nothing processed, no blobs anywhere, and `truncated` is still true at both limits. The
value does not appear to be derived from the counts at all.

Three further namespaces on the same account, same call shape, same result: `13/2/2` blobs
respectively, all `truncated=true` at `limit=100`.

## Expected

`truncated` should be true only when the namespace holds more blobs than the call processed —
roughly `total > limit`, or more precisely when a further page exists. With `total=0` or
`total <= limit` it should be false, and the MCP layer should not print the "increase limit"
advisory.

## Impact

- **Agents loop.** The tool description makes the flag actionable, so a compliant agent retries a
  finished restore. At `limit=100` there is no higher limit to try, so the retry is unbounded or
  the agent stalls.
- **Cold-start recovery is where this fires.** `restore` exists for the case where recall returns
  nothing on a new machine, which is exactly when an agent is already uncertain whether memory is
[...]

PR #823, opened 2026-08-28
fix(server): stop restore truncated=true on empty/under-limit namespaces (WALM-431)

ducnmm:
## Summary
- `/api/restore` ORed the sidecar's owner-wide `sourceCapped` into `truncated` on every response.
- That cap saturates at 100 blobs across **all** namespaces, so a moderately-used account always got `truncated=true` — including empty namespaces and namespaces far below `limit`.
- MCP `memwal_restore` tells agents to increase `limit` and retry when truncated. Limit is already clamped to 100, so the retry loops.
- `truncated` is now true only when this namespace was actually sliced, or the caller's page is full **and** the sidecar may still have more.

Fixes WALM-431 / GH #762.

## Test plan
- [x] `cargo test --bin memwal-server restore_truncated`
- [x] `cargo test --bin memwal-server paginate_missing_blobs`


harrymove-ctrl (changes requested):
Direction is right: `limit_truncated || source_capped` was genuinely broken and this is the correct place to fix it. I checked out `7fd9991d`, verified the sidecar cap formula against the source, and ran `cargo test --bin memwal-server restore_truncated` (4 passed).

Requesting changes on one logic bug that leaves the reported loop alive, plus two small cleanups.

---

### 1. `namespace_total >= limit` reproduces the loop this PR is meant to kill (blocking)

`services/server/src/routes/admin.rs:534`

Walk the branch you reach it from:

* past `limit_truncated == false`, so `missing <= limit`
* past `limit < 20`, so `limit >= 20` and the sidecar cap is pinned at 100 (`min(desiredMatches * 5, 100)`, `walrus-query.ts:650`)

Raising `limit` from there changes nothing: same candidate set, same `total`, same `missing`, same page. So in this entire branch a retry provably cannot surface more data, yet we return `true` whenever `namespace_total >= limit`.

Concrete repro: owner with >100 blobs, `default` namespace holding 100 of the 100 candidates, all already local. Call at `limit=100` gives `missing=0`, `source_capped=true`, `total=100 >= 100` so `truncated=true`. MCP says "increase limit and call again"; `clamp_restore_limit` is already at 100 (`admin.rs:544`), so the next call is byte-identical. That is WALM-431 again, just for a heavier account.

There is also a doc/code mismatch here: the function doc says "this namespace filled the requested page" and the PR description says "the caller's page is full", but the code compares `total` (all on-chain blobs found for the namespace), not the page. Those diverge exactly when a namespace is mostly already restored, which is the steady state MCP hits on every repeat call.

Suggested fix at the call site (`admin.rs:684`):

```rust
let truncated = restore_is_truncated(
    limit_truncated,
    source_capped,
    missing_blob_ids.len(),
    limit,
);
```

and rename the parameter `namespace_total` to `page_len` so the signature matches the doc.

Missing test that would have caught this:

```rust
#[test]
fn restore_truncated_false_when_namespace_fully_restored_and_cap_saturated() {
    // missing == 0, sidecar cap pinned at 100, limit already at the clamp
    // ceiling: a retry cannot surface more, so do not tell MCP to retry.
    assert!(!super::restore_is_truncated(false, true, 0, 100));
}
```

One question before you change it: was passing `total` deliberate, as in "this namespace is large, so be conservative"? If so I still think the answer is the separate field in point 2 rather than overloading `truncated`, because the MCP text makes any `true` actionable advice that is provably wrong here.

### 2. `truncated` is now carrying two meanings (follow-up, not blocking)

After this PR, `truncated=false` can hide real incompleteness. Owner with 500 blobs, namespace `work` has 3 blobs inside the first 100 candidates and 40 beyond the cap: at `limit=100` this returns `truncated=false`, the caller stops, and 40 memories are silently absent. Pre-PR that looped, which was wrong, but at least loud. This trades a noisy bug for a quiet one.

The flag is being asked to mean both "the result is incomplete" and "a bigger limit will help", and those have diverged. The honest shape is a separate `sourceCapped` field on `RestoreResponse` so `truncated` can stay truthful about completeness, plus a reworded MCP description at `packages/mcp/src/auth-required.ts:127`, which still says "If truncated=true, increase limit and call again" unconditionally.

That is a wire-format change touching the SDK, so it belongs in its own PR. Could you file it before this merges? Otherwise the silent under-restore ships and nobody comes back to it.

### 3. Duplicated doc comment on the constant (nit)

`admin.rs:500` opens with "Decide restore's public `truncated` flag (WALM-431 / GH #762)." on `SIDECAR_CANDIDATE_CAP_SATURATES_AT_LIMIT`, and the identical line repeats on the function at `admin.rs:508`. The constant does not decide anything; that first line looks like a copy-paste slip. The rest of the constant's comment is good, keep it.

### 4. Test name does not match its assertions (nit)

`restore_truncated_true_when_the_page_is_full_and_source_capped` also asserts `(true, false, 2, 100)` and `(true, true, 0, 100)`. Both are `limit_truncated` short-circuits with no full page, and one has `source_capped == false`. Worth splitting into a `limit_truncated_short_circuits` test.

### 5. Cross-language constant with no guard (optional)

`SIDECAR_CANDIDATE_CAP_SATURATES_AT_LIMIT = 20` is derived from `Math.min(desiredMatches * 5, 100)` in TypeScript, and nothing links them. If someone edits the `5` or the `100` in the sidecar, the Rust side silently goes wrong and the only symptom is this loop returning. A short note in `walrus-query.ts` pointing back at the Rust constant would make the coupling visible from both sides.

ducnmm:
Addressed in 1a2a1214.

1. **Blocking (`page_len` vs `total`):** passing `total` was not deliberate — it disagreed with the helper doc. Call site now passes `missing_blob_ids.len()`, parameter renamed `page_len`. Fully restored namespace at `limit=100` with sidecar cap saturated is `truncated=false` (Harry's repro). `cargo test --bin memwal-server restore_truncated` — 6 passed.

2. **Follow-up, not this PR:** filed [WALM-451](https://linear.app/mysten-labs/issue/WALM-451) for a separate `sourceCapped` field + MCP copy. Wire-format / SDK change stays out of 823.

3. Dropped the copy-pasted "Decide restore's public truncated flag" line from `SIDECAR_CANDIDATE_CAP_SATURATES_AT_LIMIT`.

4. Split `limit_truncated` short-circuits into their own test; added `restore_truncated_false_when_namespace_fully_restored_and_cap_saturated`.

5. Left the sidecar TS coupling note for a later pass (optional).

ducnmm:
Addressed in 7ab00540 (plus the earlier `page_len` change in 1a2a1214).

1. **Blocking (`page_len` vs `total`):** done — `page_len` is `missing_blob_ids.len()`. Passing `total` was not conservative-on-purpose; it disagreed with the helper doc and left Harry's fully-restored + cap-saturated repro looping. `truncated` now means a bigger limit can still help. `restore_truncated_false_when_namespace_fully_restored_and_cap_saturated` covers `(false, true, 0, 100)`.

2. **Follow-up, not this PR:** [WALM-451](https://linear.app/mysten-labs/issue/WALM-451) — separate `sourceCapped` field + MCP copy. Wire format / SDK stays out of 823.

3–5. Done: dropped the copy-pasted first line from `SIDECAR_CANDIDATE_CAP_SATURATES_AT_LIMIT` (kept on the function); split `limit_truncated` short-circuits into `restore_truncated_limit_truncated_short_circuits`; added a one-line note on the sidecar `Math.min(desiredMatches * 5, 100)` formula pointing at Rust `SIDECAR_CANDIDATE_CAP_SATURATES_AT_LIMIT = 20`.

`cargo test --manifest-path services/server/Cargo.toml --bin memwal-server restore_truncated` — 6 passed.

harrymove-ctrl (changes requested):
Thanks for the follow-up. The `page_len` correction, cap-saturated regression coverage, and the focused Rust tests address the original retry-loop bug.

I still need one change before approval:

### [P1] Align the restore truncation contract across clients and docs

The new `restore_is_truncated` logic can return `truncated=false` when the sidecar candidate cap is saturated but this namespace has fewer missing blobs than `limit`. That is an intentional “not known to be retryable” result, but it can still represent an incomplete view because blobs may exist beyond the sidecar’s owner-wide cap.

The public contract has not been updated consistently:

- `docs/relayer/api-reference.md:467` still says any sidecar cap produces `truncated=true`.
- `docs/mcp/reference.md:95` still describes `truncated=true` as proof that the index is incomplete.
- `packages/python-sdk-memwal/memwal/types.py:209-215` still says a source cap can make `truncated=true` even when `total == 0`.
- `services/server/scripts/mcp/tools/restore.ts:41,54` and `packages/mcp/src/auth-required.ts:127` still tell agents to increase `limit` whenever `truncated=true`.

Please update these docs/client descriptions to match the new “known/retryable truncation” semantics, and explicitly document the saturated-cap unknown-completeness case. If that belongs in WALM-451, please link the follow-up and make the current contract explicit here.

I rechecked head `7ab00540`: all reported GitHub checks pass, `restore_truncated` (6 tests) passes, `paginate_missing_blobs` passes, the diff is clean, and the merge is conflict-free. Once the contract/docs are aligned, I’m happy to re-review.

ducnmm:
Addressed in dc42aee3.

Public contract copy now matches `restore_is_truncated` (no wire-format change):

1. **Relayer / MCP / Python RestoreResult** document `truncated=true` as known-retryable-incomplete (`limit` page, or sidecar cap still expandable at `limit < 20`). Saturated cap follows the missing-blob page, not on-chain `total`.
2. **`truncated=false` is not a completeness guarantee** — blobs beyond the owner-wide sidecar candidate cap can still be missing. Follow-up: [WALM-451](https://linear.app/mysten-labs/issue/WALM-451) `sourceCapped`.
3. **MCP** `memwal_restore` description and `formatRestoreResult` no longer treat `truncated=false` as "Restore complete" or tell agents to raise `limit` unconditionally.

Ready for re-review.

harrymove-ctrl:
Pls resolve the conflict @ducnmm

ducnmm:
Rebased onto `origin/dev` (`cc3c92c6`) after #838.

Python stays **0.1.9** (no 0.1.10). The restore `truncated` changelog bullet is under the existing `## 0.1.9` heading next to the `remember_bulk_async` job_ids fix. Conflicts cleared; mergeable again.

jessiemongeon1:
## Style Guide Audit

All 5 file(s) pass the style guide audit.

ducnmm:
Style-guide audit: `onchain` spelling and present tense on the restore `truncated` copy (`docs/mcp/reference.md`, `docs/relayer/api-reference.md`, `docs/python-sdk/api-reference.md`). Contract text from cc3c92c6 is unchanged: `truncated=true` is known-retryable-incomplete; `truncated=false` is not a completeness guarantee; WALM-451 is the `sourceCapped` follow-up.

@harrymove-ctrl ready for re-review when you have a moment.

ducnmm:
Pushed `318cd28` for the remaining MCP copy gap from the 31 Aug review:

- `memwal_restore` only tells the agent to raise `limit` when `truncated && limit < 20`.
- After the sidecar cap saturates, the tool text says truncation follows this call's missing-blob page and is not completeness.
- WALM-451 stays the `sourceCapped` wire-format follow-up (linked in relayer/MCP/Python docs already on this branch).

Please re-review when you can.

ducnmm:
@harrymove-ctrl gentle ping — still waiting on re-review of `318cd28`.

MCP restore copy now only tells the agent to raise `limit` when `truncated && limit < 20`. Earlier items (`page_len`, docs) are on the branch.

harrymove-ctrl (approved):
Approving. Both earlier rounds are addressed; I pushed one doc-only commit (`87697df8`) to close the last gap rather than send it back for a comment.

### Verified on `318cd28`

- `cargo test --manifest-path services/server/Cargo.toml --bin memwal-server restore_truncated` — 6 passed
- `npm test` in `services/server/scripts` — 242 passed, including all four `formatRestoreResult` cases
- `packages/sdk`: `npm run typecheck` clean, 97 tests passed
- 22/23 reported checks pass (1 skipping), `mergeable=MERGEABLE`

### Round 1 items

1. **`page_len` vs `total` (was blocking)** — fixed. The call site passes `missing_blob_ids.len()` and the parameter is `page_len`, matching the helper doc. My repro `restore_is_truncated(false, true, 0, 100)` returns `false`, covered by `restore_truncated_false_when_namespace_fully_restored_and_cap_saturated`. I also checked reachability: `page_len >= limit` is only reachable past the `limit_truncated` short-circuit, so `all_missing.len() <= limit` and the comparison is effectively `page_len == limit` — the exactly-full page. Sound, and the retry terminates because restoring that page empties `all_missing` on the next call.
2. **Two meanings on one flag** — WALM-451 filed and linked from the relayer, MCP, and Python surfaces. Right call keeping the wire-format change out of this PR.
3. Copy-pasted first line dropped from `SIDECAR_CANDIDATE_CAP_SATURATES_AT_LIMIT`.
4. `restore_truncated_limit_truncated_short_circuits` split out.
5. Cross-language note added at `walrus-query.ts:651`.

### Round 2 item

Contract alignment landed on the relayer API reference, MCP reference, Python `RestoreResult`, `formatRestoreResult`, and the `memwal_restore` description. `formatRestoreResult` now takes `limit` and only says "increase limit and call again" while `limit < 20`, which was the last piece of wrong advice.

`packages/sdk/src/types.ts` was the one surface that picked up the retryable semantics but not the completeness caveat — a TS consumer reading that doc comment would still conclude `truncated=false` means done, which is the silent under-restore from round 1 item 2. `87697df8` mirrors the Python wording and points at WALM-451. Doc comment only, no wire or behaviour change; typecheck and SDK tests pass.

### Note, not blocking

`Benchmark Smoke` has a `services/server/scripts/**` path filter that this PR matches, but it did not run here — so the four changed `restore.test.ts` cases were not actually covered by CI on this PR. I ran them locally and they pass. Pre-existing workflow-trigger issue, worth a separate look.

Issue #786, opened 2026-08-24
[Bug] getSuggestions server action returns another user's suggestion text


alfamain:
### Surface

Example app (chatbot, noter, researcher)

### Network

Local / self-hosted

### Package version

dev@caff567, main@1ae17acbc2d3631b4d0e32bfb28e3a43bb59ea38

### What happened?

Calling the `getSuggestions` server action with another user's document ID returns that user's suggestion rows, while the equivalent HTTP route correctly rejects the same request.

### Steps to reproduce

1. Create guest A and a document with one suggestion owned by A.
2. Create guest B; keep B's `authjs.session-token` cookie.
3. Read the compiled action ID for `getSuggestions` from Next's
   `.next/dev/server/server-reference-manifest.json`.
4. POST the server action as B with A's document UUID
   (`Next-Action: <action id>`, body containing `{"documentId":"<A's UUID>"}`).
5. Control: `GET /api/suggestions?documentId=<A's UUID>` with B's cookie.

### Expected

Both calls reject B's request for A's document (matching the HTTP route's `auth()` + `suggestion.userId === session.user.id` check).

### Actual

The server action returns HTTP 200 with A's row:
```json
{
  "documentId":"bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb",
  "originalText":"secret draft paragraph",
  "suggestedText":"rewritten secret draft",
  "userId":"a80ba4b1-829e-4651-8fe6-da7b428192a3"
}
```
The HTTP control with the same cookie/ID returns HTTP 403.
`apps/chatbot/artifacts/actions.ts` calls
`getSuggestionsByDocumentId({ documentId })` with no session check;
`apps/chatbot/lib/db/queries.ts` filters by `documentId` only.

### Logs or error text

```shell
ƒ getSuggestions({"documentId":"bbbbbbbb-…"}) in 112ms artifacts/actions.ts
```

### Checks

- [x] I searched existing issues and this is not a duplicate.
- [x] This report contains no private keys, mnemonics, or other secrets.

HoangDucBach:
[WALM-461](https://linear.app/mysten-labs/issue/WALM-461/bug-getsuggestions-server-action-returns-another-users-suggestion-text)

nikola0x0:
Resolved in https://github.com/MystenLabs/MemWal/pull/824.

PR #824, opened 2026-08-28
fix(chatbot): scope getSuggestions to the calling user (WALM-438)

ducnmm:
## Summary
- `getSuggestions` server action queried suggestions by `documentId` with no session check. Guest B could read guest A's suggestion text.
- The HTTP route already rejected that; the action did not.
- Same class as WALM-309: owner-scoped DB lookup (`documentId` AND `userId`) plus a call-site filter. Unscoped by-documentId lookup is removed.
- Example chatbot only — not the relayer.

Fixes WALM-438 / GH #786.

## Test plan
- [x] `pnpm --filter @memwal/chatbot test:unit` (ownership + SQL-scope suites)

### DIFF
--- a/apps/chatbot/app/(chat)/api/suggestions/route.ts
+++ b/apps/chatbot/app/(chat)/api/suggestions/route.ts
@@ -1,5 +1,5 @@
 import { auth } from "@/app/(auth)/auth";
-import { getSuggestionsByDocumentId } from "@/lib/db/queries";
+import { getSuggestionsByDocumentIdForUser } from "@/lib/db/queries";
 import { ChatbotError } from "@/lib/errors";
 
 export async function GET(request: Request) {
@@ -14,22 +14,20 @@ export async function GET(request: Request) {
   }
 
   const session = await auth();
+  const userId = session?.user?.id;
 
-  if (!session?.user) {
+  if (!userId) {
     return new ChatbotError("unauthorized:suggestions").toResponse();
   }
 
-  const suggestions = await getSuggestionsByDocumentId({
+  const suggestions = await getSuggestionsByDocumentIdForUser({
     documentId,
+    userId,
   });
 
-  const [suggestion] = suggestions;
-
-  if (!suggestion) {
-    return Response.json([], { status: 200 });
-  }
-
-  if (suggestion.userId !== session.user.id) {
+  // Owner-scoped lookup already dropped other users' rows. Re-assert so a
+  // future query refactor cannot leak suggestion text through this route.
+  if (suggestions.some((suggestion) => suggestion.userId !== userId)) {
     return new ChatbotError("forbidden:api").toResponse();
   }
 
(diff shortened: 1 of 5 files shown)


ducnmm:
@harrymove-ctrl gentle ping — still waiting on review of this PR (WALM-438, `getSuggestions` owner scope). Requested 28 Aug.

harrymove-ctrl (approved):
Approving. The fix closes WALM-438 / GH #786 correctly and follows the WALM-309 precedent closely.

**What I verified**

- `getSuggestions` now gates on `auth()` and passes `userId` into an owner-scoped lookup, so the repro in the issue (guest B POSTing the compiled action id with A's document UUID) returns `[]` instead of A's `originalText`/`suggestedText`.
- The unscoped `getSuggestionsByDocumentId` is deleted rather than left available — same "remove the footgun" move as `getDocumentByIdForUser`. `grep` confirms no remaining callers anywhere in `apps/` or `packages/`.
- `getSuggestionsByDocumentIdForUser` filters on `and(documentId, userId)`, and the naming matches the existing `getDocumentByIdForUser` convention.
- Ownership semantics check out: `requestSuggestions` writes `suggestion.userId = session.user.id` after an owner-scoped `getDocumentByIdForUser`, so `suggestion.userId` is always the document owner. Scoping the read by `userId` cannot hide a legitimate row.
- Tests are real coverage, not shape-only. `get-suggestions-ownership.unit.test.ts` asserts the no-session path never reaches the DB, and the leak case proves the call-site filter drops a victim row even if the query layer regresses. `suggestion-query-scope.unit.test.ts` executes the real query body against a mocked builder and asserts the compiled WHERE contains both predicates — that one actually fails if someone drops the `userId` filter, which is the property worth pinning.
- Both suites run in CI (`chatbot-checks` → `pnpm --filter @memwal/chatbot test:unit`), so this is guarded going forward. All checks green, no conflicts, no drift on the touched files since the merge base.

**One non-blocking note (no change requested)**

The HTTP route's cross-user response changes from `403` to `200 []`. Previously the unscoped query returned A's rows and the `suggestion.userId !== session.user.id` check rejected them; now the scoped query returns nothing, so `suggestions.some(...)` can never be true and the `forbidden:api` branch is unreachable.

That's a security improvement — it removes the existence oracle, and it matches the intent already documented on `getDocumentByIdForUser` ("callers cannot distinguish the two, so document existence is not disclosed"). It's also inert in practice: nothing in the app fetches `/api/suggestions`, only the server action is used (`artifacts/text/client.tsx`).

Worth flagging only because the issue's acceptance criteria says "rejected (403/404), not served", and the `403` control in the repro no longer fires. Two options, both fine by me:

- Keep as-is and add a line to the PR description noting the route now returns `200 []` for a non-owned `documentId`, so the AC wording isn't read as unmet.
- Or keep the `.some()` guard as pure defense-in-depth and say so in the comment — it currently reads as if it can fire.

Not worth another round trip. Approving.

Issue #787, opened 2026-08-24
[Bug] NUL byte in recall namespace reaches PostgreSQL and returns HTTP 500


alfamain:
### Surface

Relayer / hosted API

### Network

Local / self-hosted

### Package version

1ae17acbc2d3631b4d0e32bfb28e3a43bb59ea38

### What happened?

Sending a `namespace` containing a NUL byte to `/api/recall` returns an internal server error instead of a validation error.

### Steps to reproduce

Send a correctly signed request as a registered delegate:
```json
POST /api/recall
{"query":"test","namespace":"default\u0000evil","limit":10}
```
Control:
```json
{"query":"test","namespace":"","limit":10}
```

### Expected

A namespace containing control characters is rejected with HTTP 400, the same as an empty or over-length namespace.

### Actual

- NUL namespace → HTTP 500 `Internal server error`.
- Empty namespace → HTTP 200 with no results (also confirms the   read path skips `validate_namespace`, which does reject empty strings).

`services/server/src/types.rs`'s `validate_namespace()` only checks for empty and over-255-byte names; `recall`, `recall_manual`, and `ask` never call it before the namespace reaches PostgreSQL via `storage/db.rs`. Write paths (`remember`, `analyze`, `forget`) do call it. After adding the control-character check and calling it from the read path, the NUL request returned HTTP 400 (local change, reverted after verification).

### Logs or error text

```shell
HTTP 500
{"error":"Internal server error"}
```

### Checks

- [x] I searched existing issues and this is not a duplicate.
- [x] This report contains no private keys, mnemonics, or other secrets.

HoangDucBach:
[WALM-462](https://linear.app/mysten-labs/issue/WALM-462/bug-nul-byte-in-recall-namespace-reaches-postgresql-and-returns-http)

nikola0x0:
Resolved in https://github.com/MystenLabs/MemWal/pull/825.

PR #825, opened 2026-08-28
fix(server): reject control characters in recall namespace with 400 (WALM-439)

ducnmm:
## Summary
- `validate_namespace` only checked empty / over-length. Write paths called it; `recall`, `recall_manual`, and `ask` did not.
- A NUL in `namespace` reached PostgreSQL and returned HTTP 500.
- NUL is now rejected in the shared validator, and the three read paths call it before touching the DB.
- `remember_bulk` hand-rolled its own empty/length checks, so it did not inherit the new rejection — it now delegates to `validate_namespace` too, keeping its `items[{i}].namespace ...` error wording unchanged.

Fixes WALM-439 / GH #787.

## Behavior change — `"namespace": ""` on read paths now returns 400

Because `validate_namespace` also rejects the empty string, wiring it into the read paths changes the response for an **explicitly empty** namespace:

| Request | Before | After |
| --- | --- | --- |
| `POST /api/recall` with `"namespace": ""` | 200, empty result set | **400** `namespace cannot be empty` |
| `namespace` field omitted | 200 (`default` namespace) | 200 (`default` namespace) — unchanged |

This matches what `remember` / `forget` / `stats` have always done, and no in-repo caller or SDK sends `"namespace": ""`, so nothing in this repo is affected. But it is client-visible: a deployed client or UI that sends an empty string to mean "unset" will start getting a 400 instead of degrading to no results. Such callers should **omit the field** and let the `default_namespace` serde default apply.

## Why NUL-only rather than all control characters

An earlier revision rejected every Unicode `Cc` character. That was reverted to a NUL-only check: `\t`, `\n`, `\r`, DEL and C1 are all valid Postgres `text` and were accepted by the write path before this check existed. Since the same validator now also gates the read and delete paths, rejecting them would strand any namespace already written with one — unreadable via `recall`/`ask`/`stats` and undeletable via `forget` (only the blob-level security-delete flow, which takes no namespace, would still reach it). The issue's acceptance criteria only require NUL.

## Test plan
- [x] `cargo test --bin memwal-server namespace_validation`

### DIFF
--- a/services/server/src/routes/admin.rs
+++ b/services/server/src/routes/admin.rs
@@ -291,6 +291,7 @@ pub async fn ask(
     if body.question.is_empty() {
         return Err(AppError::BadRequest("Question cannot be empty".into()));
     }
+    validate_namespace(&body.namespace)?;
 
     // Validate scoring_weights up front — fail fast on malformed input
     // before we burn an embed + vector search + Walrus + SEAL round-trip.
--- a/services/server/src/routes/recall.rs
+++ b/services/server/src/routes/recall.rs
@@ -157,6 +157,7 @@ pub async fn recall(
     if body.query.trim().is_empty() {
         return Err(AppError::BadRequest("Query cannot be empty".into()));
     }
+    validate_namespace(&body.namespace)?;
 
     // Validate scoring_weights up front — fail fast on malformed input
     // (NaN, out-of-range, sub-floor half-life) BEFORE we spend an embed +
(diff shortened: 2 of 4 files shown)


ducnmm:
@harrymove-ctrl gentle ping — still waiting on review of `61ba20b3fa310cb5c06fa90fd6957c4ded347a03`.

harrymove-ctrl (approved):
Updated with a "Behavior change" section (before/after table for "namespace": "" vs. an omitted field, plus the guidance to omit the field rather than send "").

Issue #798, opened 2026-08-26
[Bug] memory_search tool calculates negative relevance percentage when cosine distance > 1.0


sandman-sh:
### Surface
OpenClaw plugin (@mysten-incubation/oc-memwal)

### Network
All networks (Mainnet / Testnet / Local)

### Package version
@mysten-incubation/oc-memwal (main)

### What happened?

In packages/openclaw-memory-memwal/src/tools/search.ts, the memory_search tool formats search hits returned from MemWal's vector search into relevance percentages for the LLM prompt context:

`	ypescript
// packages/openclaw-memory-memwal/src/tools/search.ts (lines 74-80)

// Walrus Memory returns L2 distance â€” convert to similarity % for readability
const formatted = safe
  .map((r: any, i: number) => {
    const relevance = Math.round((1 - r.distance) * 100);
    return ${i + 1}.  (% relevance);
  })
  .join("\n");
`

However, MemWal's PostgreSQL backend (services/server/src/storage/db.rs::search_similar) uses pgvector's cosine distance operator (<=>). Cosine distance ranges from [0, 2]:
- 0.0: identical direction
- 1.0: orthogonal
- 2.0: diametrically opposite

When a search returns memories with cosine distance > 1.0 (e.g. 1.35), 1 - r.distance produces a negative number (e.g. -35%), which is formatted directly into the prompt as 1. ... (-35% relevance).

### Steps to reproduce

1. Execute a memory_search query against a namespace with dissimilar memories resulting in cosine distance d = 1.30.
2. Inspect the returned prompt context in execute():
   `	ext
   Found 1 memories:

   1. Some past memory text (-30% relevance)
   `
3. Inspect details.memories[0].relevance, which evaluates to -0.3.

### Expected

Relevance percentage should be strictly bounded between 0% and 100%, clamping any distance >= 1.0 to 0% relevance (or mapping [0, 2] cosine distance appropriately to [0, 1] similarity).

### Actual

Negative relevance percentages (e.g. (-30% relevance)) are injected into the agent system prompt and returned in tool execution results.

### Impact

- Confuses LLM evaluation when reasoning about memory relevance.
- Violates prompt formatting assumptions and results in degraded agent memory utilization.

### Proposed Fix

Clamp the relevance calculation to non-negative values in packages/openclaw-memory-memwal/src/tools/search.ts:

`	ypescript
const similarity = Math.max(0, 1 - r.distance);
const relevance = Math.round(similarity * 100);
[...]

PR #826, opened 2026-08-28
fix(openclaw): clamp memory_search relevance when cosine distance > 1 (WALM-441)

ducnmm:
## Summary
- `memory_search` used `(1 - distance) * 100` as relevance.
- Relayer cosine distance is `[0, 2]`, so distance > 1.0 printed negative percent into the agent prompt.
- Clamp similarity to `[0, 1]` and reuse the helper in the CLI search command.

Fixes WALM-441 / GH #798.

## Test plan
- [x] `npx tsc && node --test test/**/*.test.mjs` in `packages/openclaw-memory-memwal`


ducnmm:
@harrymove-ctrl gentle ping — still waiting on review of `9fce7c177e1dd7107042fe227efd1705260d16e5`.

ducnmm:
@harrymove-ctrl gentle ping — still waiting on review of `9fce7c177e1dd7107042fe227efd1705260d16e5`.

harrymove-ctrl (approved):
Reviewed against `dev`. Confirmed the premise: `services/server/src/storage/db.rs` `search_similar` uses pgvector cosine distance (`embedding <=> $1`, range [0, 2]) and `001_init.sql` indexes with `vector_cosine_ops`, with no normalization forcing non-negative similarity — so `1 - distance` really can go negative and print `(-35% relevance)` into the prompt. The stale `// Walrus Memory returns L2 distance` comment is corrected here too.

Clamp is right: `min(1, max(0, 1 - distance))` with a `Number.isFinite` guard covers NaN/Infinity and float-negative distances. Reusing the helper in `cli/search.ts` removes the duplicated rounding.

Deliberately leaving `hooks/recall.ts:75` (`(1 - r.distance) >= config.minRelevance`) unclamped is correct and worth stating so nobody "fixes" it later: `minRelevance` allows `0`, so clamping there would turn `max(0, 1-d) >= 0` into always-true and flip the filter from "drop d > 1" to "keep everything".

Two follow-ups, neither blocking:
1. Port the prompt-level regression test from #859 — it asserts the rendered string never matches `/\(-\d+% relevance\)/` and that `details.memories[].relevance` is `[0.87, 0, 0, 0]`. That pins the reported symptom; the helper unit tests here don't.
2. `memory_search` has no `maxDistance` equivalent to the one MCP `recall` got in #850, so `d >= 1` hits still come back in top-K — now showing `0%` instead of a negative number.

PR #828, opened 2026-08-28
fix(analyze): skip the pre-extraction embed on inputs the API will reject [WALM-411]

nikola0x0:
Closes WALM-411.

`/api/analyze` accepts up to 64 KiB and uses the full input for its pre-extraction dedup query. Since #837 (WALM-423), `OpenAiEmbedder::embed` rejects inputs over 16 KiB locally, before any HTTP request. Analyze currently logs that expected rejection as a warning with `pre_extract_status = "embed_failed"`, then continues with plain extraction.

This change skips that query above 16 KiB and reports `skipped_oversized`. It preserves dedup attempts for inputs at or below 16 KiB, including the 8–16 KiB range that the original version of this PR would have skipped. The empty-namespace check still takes precedence and reports `skipped_empty_namespace`.

The analyze threshold matches `MAX_EMBED_INPUT_BYTES`, now `pub(crate)`. A compile-time assertion alongside the guard-reachability assertion prevents the analyze threshold from exceeding the embedder cap. This threshold is independent of remember's summarize-before-store threshold. Byte length is not an exact token count: dense inputs at or below 16 KiB can still exceed the provider's token limit and use the existing failure fallback.

The 31,782-byte production input from 2026-08-27 remains a regression case. Its observed 564 ms API rejection predates #837; current dev rejects it in-process already. This PR improves classification and warning noise for that case and does not claim a network-latency saving over current dev.

Large inputs continue extracting without dedup context. Summarization, fetch timeouts, timeout tuning, and per-fact embeddings are outside this change.

Validation:
- Rebased onto latest `origin/dev` (`b11f0c07`).
- Five guard tests cover the 16 KiB value, reachability and cap drift, eligibility through the boundary (including 8–16 KiB), sizes above the boundary, and the historical production input.
- Targeted tests: all 15 analyze tests and all 3 embedder tests pass.
- `rustfmt --check` on both changed files and `git diff --check` pass.
- `cargo clippy --bins --tests` passes with repository warnings.
- Full library suite: 437 passed, 23 ignored, 17 failed because database-backed job tests time out acquiring a PostgreSQL connection (`PoolTimedOut`). Full binary suite: 648 passed, 37 ignored, 42 failed with PostgreSQL connection timeouts in database-backed tests; targeted tests above pass.

### DIFF
--- a/services/server/src/routes/analyze.rs
+++ b/services/server/src/routes/analyze.rs
@@ -106,6 +106,34 @@ const EMBED_TIMEOUT_MS: u64 = 800;
 const SEARCH_TIMEOUT_MS: u64 = 300;
 const FETCH_TIMEOUT_MS: u64 = 500;
 
+/// Largest input handed straight to the embedder for the pre-extraction dedup
+/// query. `text-embedding-3-small` caps at 8192 tokens and the embedder does
+/// not truncate, so oversized input is rejected outright with
+/// `Invalid 'input': maximum context length is 8192` — after the request has
+/// already cost a round-trip (~560ms observed in production).
+///
+/// Measured bytes-per-token over representative corpora runs ~1.4 (base64)
+/// to ~4.5 (English prose), so the byte length at which 8192 tokens is
+/// reached varies from ~11 KiB to ~36 KiB depending on content. No single
+/// byte threshold is exact; this one sits below the densest case so the
+/// guard never lets a doomed call through.
+///
+/// Deliberately a separate constant from `remember`'s
+/// `SUMMARIZE_THRESHOLD_BYTES`, which holds the same value today. That one
+/// decides when to pay for an LLM summary before storing a *permanent*
+/// embedding; this one decides when to abandon a *throwaway* dedup query
+/// vector. Different trade-offs, so they are free to diverge.
+const MAX_PRE_EXTRACT_EMBED_BYTES: usize = 8 * 1024;
+
+/// Whether to skip the pre-extraction dedup embed for an input of this size.
+///
+/// Skipping costs the extractor its dedup context, which is why the boundary
+/// is inclusive: input of exactly `MAX_PRE_EXTRACT_EMBED_BYTES` is still
+/// handed to the embedder.
+fn should_skip_pre_extract_embed(text_len: usize) -> bool {
+    text_len > MAX_PRE_EXTRACT_EMBED_BYTES
+}
+
 /// One fact that has finished embed + SEAL encrypt and is ready to enqueue:
 /// `(plaintext, importance, embedding, ciphertext)`.
 type PreparedFact = (String, f32, Vec<f32>, Vec<u8>);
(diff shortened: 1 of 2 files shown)


harrymove-ctrl (changes requested):
Thanks for this — the guard itself is clean and well tested, and it merges onto `dev` with all 682 `--bins` tests green. Three things before it lands.

## 1. Rebase onto latest `dev`

This branch is based on a commit that predates #837 (WALM-423, merged 2026-08-31), which added `MAX_EMBED_INPUT_BYTES = 16384` and `reject_oversized_embed_input` at the top of `OpenAiEmbedder::embed`:

```rust
async fn embed(&self, text: &str) -> Result<Vec<f32>, AppError> {
    reject_oversized_embed_input(text)?;   // in-process, no network
```

Anything over 16 KiB now fails in-process, before any HTTP request. The ~560 ms round-trip this PR is built around no longer exists in that range — the 31,782-byte production input pinned as the regression anchor already costs 0 ms on today's `dev`.

## 2. Please reconsider 8 KiB → 16 KiB

After the rebase, the only band where this PR changes real behaviour is 8–16 KiB, and by the bytes-per-token table in the description that band splits badly:

- **base64** (breaches at 11.1 KiB) and **hex** (14.1 KiB) — doomed calls, correctly skipped.
- **CJK** (24 KiB), **JSON** (25.1), **Rust** (33.4), **prose** (35.7), **TypeScript** (36.2) — these embed *successfully* at 8–16 KiB today, and the guard would start discarding their dedup context.

That is 5 of the 7 content types in the table, and the common ones.

The stated rationale — "8 KiB sits below the densest case, so the guard never lets a doomed call through" — optimises for zero doomed calls. But a doomed call costs ~560 ms on a path that already fails soft and degrades to the same empty context, whereas a wrongly skipped call costs the dedup feature itself. At 16 KiB the guard fires exactly where the embedder already rejects for free, so you keep the observability win (`skipped_oversized` instead of a WARN that reads like an incident) without giving up context that was working.

Keeping 8 KiB is still a defensible call — but please make it explicitly against post-#837 behaviour, rather than inheriting the value from a base that did not have that guard.

## 3. Update the comments and the description

Both the doc block on `MAX_PRE_EXTRACT_EMBED_BYTES` and the inline comment at the guard assert the round-trip is spent:

> ...after the request has already cost a round-trip (~560ms observed in production).

Above 16 KiB that is no longer how it fails. The same applies to the Problem and Evidence sections — as written, a reviewer reading this today would draw a materially wrong conclusion about what the change saves.

---

### One small addition worth making while you are in there

The doc block spends a paragraph justifying divergence from `remember`'s `SUMMARIZE_THRESHOLD_BYTES`, but the constant that actually constrains this one is `MAX_EMBED_INPUT_BYTES` — which is not mentioned, and is private. Nothing catches drift: drop the embedder cap below the analyze threshold and this silently starts making doomed calls again, with all five new tests still passing.

Making it `pub(crate)` and adding an assertion next to the existing reachability one would close that:

```rust
const { assert!(MAX_PRE_EXTRACT_EMBED_BYTES <= MAX_EMBED_INPUT_BYTES) }
```

---

Everything else looks right to me:

- the guard sits **after** the empty-namespace branch, so the more informative `skipped_empty_namespace` still wins;
- the boundary is inclusive and tested on both sides;
- `pre_extract_status` is tracing-only and contained to `analyze.rs` — I checked repo-wide, nothing else consumes it, so the new variant needs no updates elsewhere;
- the per-fact embeds at `analyze.rs:491` / `:571` are correctly left alone, since their vectors get stored.

harrymove-ctrl (approved):
Yes, PR #828 can be approved to merge. All requested changes were implemented cleanly and verified.

PR #830, opened 2026-08-28
feat(sdk): listNamespaces() so an agent can discover namespaces [WALM-395]

nikola0x0:
Closes WALM-395 (SDK surface; remaining surfaces tracked below).

## Acceptance criteria

From the ticket:

- [x] **An agent with only delegate credentials can enumerate the namespaces holding at least one memory, without prior knowledge of naming conventions.** Verified live against `relayer.dev.memwal.ai` — see Live verification.

From the ticket's "What's left":

- [x] `listNamespaces()` on the TS SDK
- [ ] `listNamespaces()` on the Python SDK — separate surface, same design applies; not attempted here
- [ ] `memwal_namespaces` MCP tool — **blocked**: `services/server/scripts` installs the *published* SDK (pinned `0.0.3`) with `services/server/` as the Docker build context, so it cannot resolve the workspace SDK. Needs a release first, same blocker as WALM-428
- [ ] `memwal namespaces` CLI command — same publish blocker

## Problem

Recall is similarity-ranked and needs a namespace to search. An agent connecting to an unfamiliar account had to guess names or fall back to `"default"`, which undercuts the cross-session memory portability that is a core value prop.

`GET /v1/owners/{owner}/namespaces` already shipped in `caff5673`. This surfaces it on the TS SDK — metadata only, no blob fetch, no decryption, no SEAL session built or transmitted.

## The blocker the ticket did not anticipate

The ticket scoped this as *"purely client surfacing over the existing endpoint."* It isn't. The read routes take the owner address in the path and reject a mismatch against the caller's credentials (`same_owner`, canonical Sui address, no `me`/`self` alias) — but `MemWalConfig` carries only the delegate key and account id. **The SDK did not know its own owner address**, so it could not build the path at all.

`resolveOwner()` learns it from `POST /api/stats`: same delegate auth scheme, needs only a namespace, rate-limit weight 1, returns the owner the server resolved. Memoised per client, single-flight guarded like the existing compatibility probe.

Using a stats endpoint as a whoami is indirect, and was chosen deliberately over a server-side `me` alias: the SDK ships to npm and self-hosters run their own relayer, so a path-level change would 403 against every older relayer and need a compatibility branch. The indirection sits in one private method, no public API mentions `owner`, and if a self-reference lands later `resolveOwner()` is the only thing that changes.

**This will recur.** WALM-388's `memwal_list` hits the identical wall and can reuse `resolveOwner()`.

## Live verification

Unit tests stub `fetch`, which cannot prove the relayer *accepts* the request. Run against `relayer.dev.memwal.ai` with real delegate credentials:

```
1. listNamespaces()                     PASS  -> 3 namespaces
2. listNamespaces({ limit: 2 })         PASS  -> 2, has_more=true
3. listNamespaces({ cursor, limit: 2 }) PASS  -> 1, has_more=false

POST /api/stats                                       -> 200
GET  /v1/owners/0x4824…/namespaces                    -> 200
GET  …/namespaces?limit=2                             -> 200
GET  …/namespaces?updated_after=eyJ1cGRhdGVk…&limit=2 -> 200

/api/stats calls: 1  (owner memoised correctly)
```

Why this mattered: **signed-GET-with-a-query-string had no prior call site anywhere in the SDK.** Plain signed GET is already proven (`getRememberStatus`), but nothing had signed a query string before. The server verifies against `path_and_query`, not `path`, so the SDK signs the exact string it puts in the URL — symmetric by construction, but `fetch` sits in between and could in principle re-encode. Test 3 is the strongest form: the cursor is base64 JSON containing characters that percent-encode, and it returned 200.

Pagination is self-consistent: 2 + 1 = 3, matching the unpaginated call, with `has_more` flipping correctly.

The live run also caught a bug no stub would have: the relayer returns `snapshot_version: 2` while `MemWalMock` hardcoded `1`. A double that disagrees with the server on a wire-format version is a trap for anyone testing version-gated logic. Fixed test-first.

## Design notes

- Result types mirror the relayer wire shape (snake_case out), matching the existing `RecallResult.dropped_count` convention rather than adding a mapping layer.
- `has_more` is documented as the pagination signal, not page length — the server clamps `limit`, so a caller asking for more than the cap gets exactly the cap back and would wrongly conclude it was done.
- `MemWalMock` gains the same method so the drop-in double keeps parity, aggregating seeded memories by namespace with deterministic timestamps derived from insertion order.

## Checks

- [x] 8 new unit tests (6 client, 2 mock), all written before the implementation and confirmed failing first
- [x] Full SDK suite: 106 passed, 0 failed
- [x] `tsc --noEmit` clean
- [x] `check-docs-code-sync` and `check-docs-freshness` pass
- [x] `docs/sdk/api-reference.md` section added
- [x] Verified against a live relayer, not only stubs

## Follow-up worth filing

A server-side self-reference (accept `me` as the path owner, or a small whoami route) would retire `resolveOwner()`'s indirection and unblock WALM-388 the same way. Not done here to keep this client-only and free of a relayer-deploy dependency.


jessiemongeon1:
## Style Guide Audit

### `docs/sdk/api-reference.md` (4 violation(s))

_4 violation(s) (4 regex, 0 claude)_

- **Line 272** — No em dashes in prose
  - Current: `—`
- **Line 276** — No em dashes in prose
  - Current: `—`
- **Line 277** — No em dashes in prose
  - Current: `—`
- **Line 296** — No em dashes in prose
  - Current: `—`

---

harrymove-ctrl (changes requested):
Nice work here — tests written first and confirmed failing, and running it live against `relayer.dev.memwal.ai` was the right call given that a stubbed `fetch` cannot prove the relayer accepts a signed query string. The description being straight about the ticket mis-scoping this as "purely client surfacing" is appreciated too.

Two things to fix before it goes in.

## 1. Drop `.changeset/list-namespaces.md`

Henry flagged this on WALM-395. I checked it against the repo rather than taking it on faith, and it holds:

- `.github/workflows/release-sdk.yml:51` still runs `pnpm changeset version` on `main` and auto-commits the bump.
- The team has already cleaned up this exact failure twice: `chore: drop changesets to avoid a double bump on main` (21 Aug) and `fix(sdk): drop leftover changeset and bump 0.1.6 by hand (WALM-471)` (4 Sep).

Leaving the file re-creates the double bump WALM-471 cleaned up three days ago. Versions here are bumped by hand.

## 2. Three JSDoc blocks were orphaned by where the new code landed

Same mechanical slip in three places: a new member was inserted *between* an existing doc comment and the member that comment describes. The comment now documents the new code, and the original member is left undocumented. I checked each one against `dev` to confirm the comment belonged to the member below it.

- `packages/sdk/src/memwal.ts:199-209` — the "Keep a generated idempotency key…" block now documents `ownerAddress`; `pendingRememberKeys` loses its docs.
- `packages/sdk/src/memwal.ts:922-1007` — "Check server health. The endpoint is public…" now documents `listNamespaces()`; `health()` loses its docs.
- `packages/sdk/src/types.ts:407-449` — `/** Result from restore() */` now documents `NamespaceSummary`; `RestoreResult` loses its docs.

Move each new declaration below the comment that belongs to the existing member, or move the comment back down onto it.

## Non-blocking

**The mock's cursor disagrees with the live relayer, in exactly the way `snapshot_version` did.** `MemWalMock.listNamespaces` returns `next_cursor` as a bare ISO timestamp and walks with `all.filter(ns => ns.updated_at > cursor)`, but the Live verification section of this PR shows the relayer returning base64 JSON (`eyJ1cGRhdGVk…`). You fixed `snapshot_version` 1→2 on the reasoning that "a double that disagrees with the server on a wire-format version is a trap" — the cursor is the same trap, quieter, and it is the field callers actually round-trip.

**`ONE_PAGE` in `test/list-namespaces.test.mjs` sets `snapshot_version: 1`** while `mock.test.mjs` asserts the real value is 2. Harmless, since the client echoes the body back, but internally contradictory.

**The JSDoc example on `listNamespaces` uses `var more` inside a `do…while`** to escape block scope. The `docs/sdk/api-reference.md` version is clean (`let cursor` with a `while` loop) — worth using that one in both places, since doc examples get copy-pasted.

## Also

The branch is 40 commits behind `dev`. Please update it before merge so CI runs against the current base.

## What I checked and am satisfied with

- `resolveOwner()` single-flight has no race. On success `ownerAddress` is set before the `.finally` clears `ownerPromise`, so there is no window where both are null; on failure the promise clears while the address stays null, so the next call retries. The memoisation test pins the behaviour.
- Choosing `POST /api/stats` as a whoami over a server-side `me` alias is the right trade-off for something that ships to npm and runs against self-hosted relayers, and the indirection stays inside one private method with a clear note on what would replace it.
- Metadata-only read builds no SEAL session, and that is asserted by a test rather than only claimed in prose.

Happy to merge as soon as 1 and 2 land.

harrymove-ctrl (approved):
Approve

PR #831, opened 2026-08-29
fix(chatbot): return 404 when deleting a nonexistent document (WALM-422)

### LINKED ISSUE #784: [Bug] Deleting a nonexistent document returns HTTP 500 instead of 404
### Surface

### Network

Local / self-hosted

### Package version

@caff567

### What happened?

`DELETE /api/document` with a document ID that does not exist throws an unhandled exception instead of returning a normal "not found" error.

### Steps to reproduce

1. Obtain a guest session cookie from `GET /api/auth/guest`.
2. Run:
```bash
curl -i -X DELETE \
  -H 'Cookie: authjs.session-token=<guest session>' \
  'http://127.0.0.1:3001/api/document?id=<unused UUID>&timestamp=2026-01-01T00:00:00.000Z'
```

### Expected

HTTP 404 with `not_found:document`, matching the GET handler in the same file, which already checks `if (!document)`.

### Actual

HTTP 500 with an empty body:
```text
TypeError: Cannot read properties of undefined (reading 'userId')
    at DELETE (app/(chat)/api/document/route.ts:116:16)
```
Applying the same `if (!document)` guard used by GET changed the result to HTTP 404 `not_found:document`; reverting the guard reproduced the HTTP 500 again.

### Logs or error text

```shell
TypeError: Cannot read properties of undefined (reading 'userId')
    at DELETE (app/(chat)/api/document/route.ts:116:16)
DELETE /api/document?id=…&timestamp=… 500
```

### Checks

HoangDucBach:
## Summary

DELETE /api/document answered an empty HTTP 500 for a document id that matches no row. It now returns 404 not_found:document, the same answer GET already gave for that id. A route-level test pins the behavior.

### What happened

DELETE /api/document with an unused UUID threw an unhandled exception instead of reporting a normal not-found error, so the client received a 500 with an empty body.

### Root cause

The handler destructured the first row out of getDocumentsById and read .userId off it without checking that a row came back. For an unknown id the array is empty, so the read landed on undefined:

TypeError: Cannot read properties of undefined (reading 'userId')

### Why GET was unaffected

GET in the same file already ran the `if (!document)` check before the ownership comparison. The two handlers therefore disagreed on the same missing document.

### Fix

Add that same guard to DELETE, placed before the ownership comparison so an unknown id is never reported as 403 or 500.

### Scope of the audit

DELETE was the only unguarded case in the app. auth.ts checks users.length before destructuring, POST in this file checks documents.length, and the suggestions route and chat actions already guard.

## Verification

- [x] pnpm --filter @memwal/chatbot test:unit: 21 pass, 6 of them new
- [x] Removing the guard turns 2 of the new tests red with the exact TypeError from the report
- [x] pnpm exec next build passes, which covers the type-check
- [x] Live reproduction from the report: DELETE returns 404 not_found:document and matches GET on the same id

## Out of scope

- The chatbot lint step already fails on dev because ultracite 7.3.0 passes a --skip flag that biome 2.3.11 rejects. CI marks it continue-on-error, and this branch does not change it.
- lib/ai/models.test.ts matches neither the vitest include nor the Playwright testDir, so no runner executes it today.

### DIFF
--- a/apps/chatbot/app/(chat)/api/document/route.ts
+++ b/apps/chatbot/app/(chat)/api/document/route.ts
@@ -113,6 +113,12 @@ export async function DELETE(request: Request) {
 
   const [document] = documents;
 
+  // Same guard the GET handler above runs. Without it an id that matches no row
+  // reads `.userId` off undefined and the request 500s with an empty body.
+  if (!document) {
+    return new ChatbotError("not_found:document").toResponse();
+  }
+
   if (document.userId !== session.user.id) {
     return new ChatbotError("forbidden:document").toResponse();
   }
--- a/apps/chatbot/app/(chat)/api/document/route.ts
+++ b/apps/chatbot/app/(chat)/api/document/route.ts
@@ -113,8 +113,6 @@ export async function DELETE(request: Request) {
 
   const [document] = documents;
 
-  // Same guard the GET handler above runs. Without it an id that matches no row
-  // reads `.userId` off undefined and the request 500s with an empty body.
   if (!document) {
     return new ChatbotError("not_found:document").toResponse();
   }
--- a/apps/chatbot/app/(chat)/api/document/document-route.unit.test.ts
+++ b/apps/chatbot/app/(chat)/api/document/document-route.unit.test.ts
@@ -1,0 +1,124 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// Regression test for WALM-422: DELETE /api/document read `.userId` off the
+// first row without checking that a row came back, so an id matching no
+// document threw TypeError and the request 500s with an empty body. The GET
+// handler in the same file already had the guard, so the two disagreed on the
+// same missing document.
+//
+// The DB query layer is mocked so no Postgres connection is needed.
+
+const OWNER_ID = "11111111-1111-1111-1111-111111111111";
+const OTHER_ID = "22222222-2222-2222-2222-222222222222";
+const DOC_ID = "33333333-3333-3333-3333-333333333333";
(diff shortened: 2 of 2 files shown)


apps/chatbot/app/(chat)/api/document/route.ts
```diff
@@ -113,6 +113,12 @@ export async function DELETE(request: Request) {
 
   const [document] = documents;
 
+  // Same guard the GET handler above runs. Without it an id that matches no row
```
ducnmm: **[suggestion]** The added comment restates the new `if (!document)` guard, points at GET, and retells the TypeError / empty-500 history. That history already lives in the commit message; GET's identical check has no comment, so DELETE now documents a past bug next to code that is otherwise obvious.

**Suggestion:** Delete the two-line comment and keep the guard, matching GET.

ducnmm:

ducnmm (approved):
## Summary

DELETE `/api/document` now returns `404 not_found:document` when `getDocumentsById` yields no row, matching the GET handler in the same file and the WALM-422 expected contract. The guard is placed before the ownership check, so a missing id cannot 500 on `document.userId` or be reported as 403, and the new route unit tests pin 404, GET parity, 403, the happy path, and the 400/401 short-circuits. The only issue is a bug-history comment on an otherwise self-explanatory check; GET's identical guard has none.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 0

apps/chatbot/app/(chat)/api/document/route.ts
HoangDucBach: fixed

PR #832, opened 2026-08-29
fix(noter): bind getSession and logout to the caller's own session (WALM-420)

### LINKED ISSUE #779: [Bug] Noter `getSession` / `logout` accept any client-supplied sessionId with no binding to the caller
### Surface

### Network

Local / self-hosted

### Package version

`apps/noter` : `main@1ae17acbc2d3631b4d0e32bfb28e3a43bb59ea38`, `dev@caff567`

### What happened?

`auth.getSession` and `auth.logout` are public tRPC procedures that take a `sessionId` and act on it without checking that the caller actually holds that session (no cookie/header binding).

### Steps to reproduce

1. Insert a `users` row and a `wallet_sessions` row with a known UUID,  without setting any cookie for it.
2. Call, with no cookie:
```text
GET  /api/trpc/auth.getSession?input={"json":{"sessionId":"<that UUID>"}}
POST /api/trpc/auth.logout     {"json":{"sessionId":"<that UUID>"}}
```

### Expected

Both calls require proof that the caller owns the session (e.g. an httpOnly session cookie matching the ID), not just knowledge of the ID.

### Actual

```text
getSession → 200  user + suiAddress + expiresAt (no private key)
logout     → 200  {"success":true}   wallet_sessions row count 1 → 0
getSession (again) → 200  {"json":null}
```
uuidv7 IDs are not guessable, so this requires the ID to leak by some other means (logs, XSS, stored client state) to be exploitable.

### Logs or error text

```shell
both calls return normal 200 responses
```

### Checks

HoangDucBach:
Fixes #779

## Summary

auth.getSession and auth.logout took a sessionId as procedure input and acted on it. Both now take the id from the x-session-id header through the tRPC context, the way protectedProcedure and the memory REST routes already do, so the input can no longer steer either call.

### What happened

Both procedures are public and read input.sessionId, ignoring ctx entirely. A caller with no session of its own could read another session's user and Sui address through getSession, or delete that session through logout, purely by naming the id.

### Why the id is not a secret in practice

getSession is a query, so the id travelled in the request URL on every page load and landed in access logs and browser history. The report notes the exploit depends on the id leaking through exactly such a channel, so the input was both the hole and a way to widen it.

### Fix

Context gains a sessionId field, filled from the header whether or not the session is still live, and both procedures read that. The sessionId input and its schema are removed, so there is nothing left to forge. getSession still answers null for a missing, unknown, or expired session, which is how the client detects a stale stored session and clears it.

### Malformed header

The zod input schema used to reject a non-uuid id before it reached a uuid column, where Postgres raises instead of simply not matching. createContext now applies that check to the header, so a malformed value means no credential rather than a 500.

### Client

The getSession query key no longer varies per session, so useAuth resets the cached lookup whenever the stored session changes. Without it a cached null left by an expired session would be read as the answer for the session just established, and the effect that clears stale sessions would wipe a fresh login.

### Scope of the audit

These two were the only procedures in noter taking a sessionId or userId as input. The note router works from ctx.userId and filters on ownership, and the memory REST routes already authenticate from the header only.

## Verification

- [x] pnpm --filter @memwal/noter test:unit: 40 pass, 7 of them new
- [x] Restoring the input-based version turns all 7 new tests red
- [x] Playwright e2e on a fresh database: 8 auth tests and 14 app, memory, and note tests pass
- [x] pnpm exec tsc --noEmit and pnpm exec next build both pass

## Out of scope

- Session ids remain bearer credentials in sessionStorage rather than httpOnly cookies, so possession still grants access. Changing that touches every request path in the app and contradicts the documented header-based design, so it needs its own decision.
- The delegate fixture pool can only be registered once per database, and Playwright recycles fixture indices when it restarts a worker after a failure, so a local rerun against a used database fails on provisioning rather than on the code.

### DIFF
--- a/apps/noter/package/feature/auth/api/route.ts
+++ b/apps/noter/package/feature/auth/api/route.ts
@@ -9,7 +9,7 @@ import { z } from "zod";
 import { verifyPersonalMessageSignature } from "@mysten/sui/verify";
 import { normalizeSuiAddress } from "@mysten/sui/utils";
 import { uuidv7 } from "uuidv7";
-import { validateSessionInput, connectWalletInput } from "./input";
+import { connectWalletInput } from "./input";
 import { AUTH_ERRORS } from "../constant";
 import { walletSessions } from "@/shared/db/schema";
 import * as authService from "../domain/service";
@@ -38,24 +38,29 @@ const suiAddressSchema = z
 
 export const authRouter = router({
   /**
-   * Get current session (for resuming auth state).
+   * Get the caller's own session (for resuming auth state).
    * Resolves wallet / enoki sessions only; the legacy zkLogin table is not trusted.
+   *
+   * The session id is read from the x-session-id header via ctx, never from the
+   * input, so a caller cannot read a session it does not already hold. Returns
+   * null for a missing, unknown, or expired session, which is how the client
+   * detects a stale stored session and clears it.
    */
-  getSession: procedure
-    .input(validateSessionInput)
-    .query(({ ctx, input }) =>
-      authService.getActiveSession(ctx.db, input.sessionId)
-    ),
+  getSession: procedure.query(({ ctx }) =>
+    ctx.sessionId ? authService.getActiveSession(ctx.db, ctx.sessionId) : null
+  ),
 
   /**
-   * Logout - clear session (works for both zkLogin and wallet)
+   * Logout - clear the caller's own session (works for both zkLogin and wallet).
+   * Like getSession, the id comes from the header, so knowing another user's
+   * session id is not enough to end their session.
    */
-  logout: procedure
-    .input(validateSessionInput)
-    .mutation(async ({ ctx, input }) => {
-      await authService.deleteSession(ctx.db, input.sessionId);
-      return { success: true };
-    }),
+  logout: procedure.mutation(async ({ ctx }) => {
+    if (ctx.sessionId) {
+      await authService.deleteSession(ctx.db, ctx.sessionId);
+    }
+    return { success: true };
+  }),
 
   /**
    * Connect wallet - authenticate with Sui wallet (Slush, Sui Wallet)
--- a/apps/noter/package/feature/auth/api/route.ts
+++ b/apps/noter/package/feature/auth/api/route.ts
@@ -51,9 +51,9 @@ export const authRouter = router({
   ),
 
   /**
-   * Logout - clear the caller's own session (works for both zkLogin and wallet).
-   * Like getSession, the id comes from the header, so knowing another user's
-   * session id is not enough to end their session.
+   * Logout - end only the session presented in x-session-id (zkLogin and wallet).
+   * A body/input id is ignored. The header value is still a bearer credential:
+   * presenting a session id there is enough to delete that session.
    */
   logout: procedure.mutation(async ({ ctx }) => {
     if (ctx.sessionId) {
--- a/apps/noter/package/shared/lib/trpc/init.ts
+++ b/apps/noter/package/shared/lib/trpc/init.ts
@@ -8,22 +8,42 @@ import { eq } from "drizzle-orm";
 export type Context = {
   db: typeof db;
   request: Request;
+  /**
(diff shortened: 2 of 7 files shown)


apps/noter/package/feature/auth/api/route.ts
```diff
@@ -38,24 +38,29 @@ const suiAddressSchema = z
...
-      authService.getActiveSession(ctx.db, input.sessionId)
-    ),
+  getSession: procedure.query(({ ctx }) =>
+    ctx.sessionId ? authService.getActiveSession(ctx.db, ctx.sessionId) : null
+  ),
 
   /**
-   * Logout - clear session (works for both zkLogin and wallet)
+   * Logout - clear the caller's own session (works for both zkLogin and wallet).
+   * Like getSession, the id comes from the header, so knowing another user's
```
ducnmm: **[suggestion]** The new logout JSDoc says "knowing another user's session id is not enough to end their session." That overstates the fix. Session ids are still bearer credentials: anyone who can put the victim id in `x-session-id` can still delete that session. The PR body correctly marks httpOnly cookies as out of scope; this comment claims the stronger property the issue asked for and the PR did not implement. The real guarantee is that a JSON/query `sessionId` can no longer steer logout independently of the header.

**Suggestion:** Reword to the actual invariant, e.g. logout ends only the session presented in `x-session-id`; a body/input id is ignored. Do not say that knowledge of the id is insufficient.

apps/noter/package/shared/lib/trpc/init.ts
```diff
@@ -8,22 +8,42 @@ import { eq } from "drizzle-orm";
...
+  /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
 function getSessionIdFromRequest(req: Request): string | null {
-  return req.headers.get("x-session-id");
+  const sessionId = req.headers.get("x-session-id")?.trim();
+
+  // Session ids are compared against uuid columns, where a malformed value makes
+  // Postgres raise rather than simply miss. Anything that is not a uuid cannot
+  // name a session, so treat it as no credential at all.
+  if (!sessionId || !UUID_REGEX.test(sessionId)) {
```
ducnmm, line 30: **[suggestion]** The malformed-header guard is a stated part of the fix (non-uuid `x-session-id` used to reach `eq(walletSessions.id, …)` and make Postgres raise). The seven new tests never call `createContext` / `getSessionIdFromRequest`; they inject `ctx.sessionId` already parsed. Restoring input-based getSession/logout would fail those tests, but dropping or breaking the UUID check would not. `callerFor` also sets `userId` from a hardcoded comparison rather than from header resolution, so it cannot catch a regression that puts a raw header value on the uuid column.

**Suggestion:** Add a unit test that runs `createContext` (db mocked) with `x-session-id` values that are missing, non-uuid, expired, and valid, and assert `sessionId` is null for the malformed cases so the lookup is never attempted.

ducnmm:

ducnmm (approved):
## Summary

The change correctly closes the issue #779 hole: `auth.getSession` and `auth.logout` no longer take a client-supplied `sessionId`, and both act only on `ctx.sessionId` filled from `x-session-id`. Keeping that id on context even when the session is expired/unknown is the right shape for stale-session detection and for deleting an expired row on logout. The client cache reset is necessary now that the query key no longer varies per session. Two suggestions below: the logout JSDoc overstates the remaining bearer-token model, and the new tests never exercise `createContext`'s UUID guard.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

apps/noter/package/feature/auth/api/route.ts
HoangDucBach: fixed

apps/noter/package/shared/lib/trpc/init.ts
HoangDucBach: fixed

PR #833, opened 2026-08-29
fix(researcher): check the destination before fetching a source file url (WAKM-419)

### LINKED ISSUE #778: [Bug] Researcher PDF file URL allows server-side requests to arbitrary HTTP(S) targets
### Surface

### Network

Local / self-hosted

### Package version

@1ae17acbc2d3631b4d0e32bfb28e3a43bb59ea38 - apps/researcher

### What happened?

An authenticated chat request with a PDF file part whose `url` points at `http://127.0.0.1:9999/...` causes the researcher server to make an outbound HTTP request to that address with no destination validation.

### Steps to reproduce

1. Start the researcher app locally with a valid session and database.
2. Start an HTTP listener on `127.0.0.1:9999`.
3. Send an authenticated `POST /api/chat`:
```json
{
  "id": "<chat UUID>",
  "message": {
    "id": "<message UUID>",
    "role": "user",
    "parts": [
      {"type":"text","text":"process this PDF"},
      {
        "type":"file",
        "mediaType":"application/pdf",
        "name":"proof.pdf",
        "url":"http://127.0.0.1:9999/ssrf-proof-token-abc123"
      }
    ]
  },
  "selectedChatModel":"google/gemini-2.5-flash",
  "selectedVisibilityType":"private"
}
```

### Expected

The file URL is resolved and checked against `loopback/RFC1918/link-local/` metadata ranges before any request is made; a private-address target is rejected.

### Actual

The listener receives the request:
```text
CAPTURED REQUEST: /ssrf-proof-token-abc123 node
```
`apps/researcher/lib/rag/ingest/index.ts` calls `fetch(source.fileUrl)` directly. `apps/researcher/lib/ai/source-processing.ts`'s `extractUrlsFromText()` already blocks loopback/RFC1918 for URLs found in chat *text*, but that filter is not applied to file-part URLs, and `apps/researcher/app/(chat)/api/chat/schema.ts` only requires `z.string().url()`.

### Logs or error text

```shell
[ingest] Starting ingestion — type=pdf, text length=...
CAPTURED REQUEST: /ssrf-proof-token-abc123 node
```

### Checks

HoangDucBach:
## Summary

- route the PDF file-part download through a destination check instead of a bare `fetch`
- reject loopback, private, link-local, and reserved targets, by literal or by hostname resolution
- re-check the destination at every redirect hop rather than letting `fetch` follow blindly

Closes #778

### The server fetched whatever url the request named

`processSource` handled a `file` part with `mediaType: application/pdf` by calling `fetch(source.fileUrl)` on the url from the request body. Nothing looked at where that url pointed, so an authenticated chat request turned the server into a probe of its own network. The report proved it with a listener on `127.0.0.1:9999` that logged the incoming request.

### The existing filter did not apply, and would not have been enough

`extractUrlsFromText` has a denylist for urls found in chat text, but it never ran on file parts, and `postRequestBodySchema` only required `z.string().url()`.

That denylist is also prefix matching on a few literals, which misses:

- userinfo in front of the host, as in `http://user@127.0.0.1/`
- loopback outside `127.0.0.1`, such as `127.1.2.3`
- IPv6 forms: `::1`, `0:0:0:0:0:0:0:1`, `::ffff:127.0.0.1`
- link-local and the metadata address `169.254.169.254`, absent from the list
- hostnames that resolve into a blocked range
- redirects, which `fetch` follows on its own

## Fix

`lib/rag/ingest/safe-fetch.ts` holds the check, and ingest calls it:

- `assertPublicUrl` parses the url, requires http or https, and rejects blocked IPv4 and IPv6 ranges. IP literals are checked directly; hostnames are resolved and every returned address has to be public. IPv4-mapped and NAT64 forms are unwrapped so they cannot smuggle a blocked IPv4 destination. Anything unparseable counts as blocked.
- `fetchPublicUrl` runs that check, then fetches with `redirect: "manual"` and re-checks each hop, up to five.

The text-url path is untouched: it reaches its target through Jina Reader rather than from this server, so it is not a request from our network.

## Validation

- `pnpm --filter researcher test:unit` — 31 pass, 12 of them new
- reverting ingest to the bare `fetch` fails the call-site test, so the wiring is pinned rather than assumed
- one test asserts a real loopback listener records nothing, which shows the request is refused before it leaves, not after
- `pnpm exec tsc --noEmit` clean

## Out of scope

- DNS rebinding, where a host answers with a public address for the check and a private one for the connection. Closing it means pinning the connection to the address that was checked, which `fetch` does not expose.
- the prefix denylist in `extractUrlsFromText`, left as is since that path does not fetch from this server
- `/api/research/process-source` with a url body, which applies no denylist at all and reaches the target through Jina for the same reason

### DIFF
--- a/apps/researcher/lib/rag/ingest/safe-fetch.ts
+++ b/apps/researcher/lib/rag/ingest/safe-fetch.ts
@@ -1,0 +1,277 @@
+import { lookup } from "node:dns/promises";
+import { isIP } from "node:net";
+
+import { ChatbotError } from "@/lib/errors";
+
+// No "server-only" marker here, matching every other unit-tested module in lib:
+// the import throws under `node --test`. The node:dns import above keeps this out
+// of a client bundle regardless.
+
+// Outbound fetches for user-supplied source URLs. A file part in a chat request
+// names a URL that the server then downloads, so without a destination check the
+// request doubles as a probe of whatever the server can reach: loopback services,
+// RFC1918 neighbours, and the cloud metadata endpoint on 169.254.169.254.
+//
+// extractUrlsFromText has a prefix-matching denylist for URLs found in chat text,
+// but it only recognises literal 127.0.0.1 and friends at the very start of the
+// string. That misses userinfo (http://x@127.0.0.1), IPv6, 127.x outside .0.1,
+// hostnames that resolve into a private range, and redirects. This module resolves
+// the host and checks every address instead.
+
+const MAX_REDIRECTS = 5;
+
+// [network, prefix length]. Everything a request has no business reaching from a
+// URL a user typed: loopback, the private ranges, link-local (which carries the
+// metadata service), plus the unspecified, multicast, and reserved blocks.
+const BLOCKED_V4_RANGES: [string, number][] = [
+  ["0.0.0.0", 8],
+  ["10.0.0.0", 8],
+  ["100.64.0.0", 10],
+  ["127.0.0.0", 8],
+  ["169.254.0.0", 16],
+  ["172.16.0.0", 12],
+  ["192.0.0.0", 24],
+  ["192.168.0.0", 16],
+  ["198.18.0.0", 15],
+  ["224.0.0.0", 4],
+  ["240.0.0.0", 4],
+];
+
+const BLOCKED_V6_RANGES: [string, number][] = [
+  ["::", 128],
+  ["::1", 128],
+  ["fc00::", 7],
+  ["fe80::", 10],
+  // NAT64 addresses carry an IPv4 destination in their low 32 bits, so they are a
+  // way back into the ranges above.
+  ["64:ff9b::", 96],
+];
+
+function ipv4ToBytes(address: string): number[] | null {
+  const groups = address.split(".");
+
+  if (groups.length !== 4) {
+    return null;
+  }
+
+  const bytes = groups.map((group) =>
+    /^\d{1,3}$/.test(group) ? Number(group) : Number.NaN
+  );
+
+  return bytes.every((byte) => byte >= 0 && byte <= 255) ? bytes : null;
+}
+
+function ipv6GroupsToBytes(part: string): number[] | null {
+  if (part === "") {
+    return [];
+  }
+
+  const groups = part.split(":");
+  const bytes: number[] = [];
+
+  for (const [index, group] of groups.entries()) {
+    // A trailing dotted-quad (::ffff:127.0.0.1) stands for the last four bytes.
+    if (group.includes(".")) {
+      const embedded = index === groups.length - 1 ? ipv4ToBytes(group) : null;
+
+      if (!embedded) {
+        return null;
+      }
+      bytes.push(...embedded);
+      continue;
+    }
+
+    if (!/^[0-9a-f]{1,4}$/i.test(group)) {
+      return null;
+    }
+    const value = Number.parseInt(group, 16);
+    bytes.push(value >> 8, value & 0xff);
+  }
+
+  return bytes;
+}
+
+function ipv6ToBytes(address: string): number[] | null {
+  // Drop any zone id: fe80::1%eth0 addresses the same interface-local target.
+  const [plain] = address.split("%");
+  const halves = plain.split("::");
+
+  if (halves.length > 2) {
+    return null;
+  }
+
+  const head = ipv6GroupsToBytes(halves[0]);
+  const tail = halves.length === 2 ? ipv6GroupsToBytes(halves[1]) : [];
+
+  if (!head || !tail) {
+    return null;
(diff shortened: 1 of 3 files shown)


apps/researcher/lib/rag/ingest/safe-fetch.ts
```diff
@@ -0,0 +1,277 @@
...
+    if (!bytes) {
+      return true;
+    }
+    if (isMappedIpv4(bytes)) {
+      return BLOCKED_V4_RANGES.some(([network, prefix]) =>
+        withinRange(bytes.slice(12), ipv4ToBytes(network) as number[], prefix)
+      );
+    }
+
+    return BLOCKED_V6_RANGES.some(([network, prefix]) =>
```
ducnmm, line 198: **[suggestion]** `isBlockedAddress` unwraps IPv4-mapped `::ffff:0:0/96` and blocks well-known NAT64 `64:ff9b::/96`, but other IPv4-in-IPv6 embeddings are treated as ordinary public v6 (`::7f00:1`, SIIT `::ffff:0:7f00:1`, local-use NAT64 `64:ff9b:1::/48`, 6to4 `2002:7f00:1::`). IPv6 multicast `ff02::1` is also allowed while IPv4 multicast is not. Not a demonstrated loopback bypass on Node 22/26 + undici today (`ENETUNREACH`), but the same smuggling class the PR already handles for mapped/NAT64.

**Suggestion:** Unwrap `::/96` and SIIT through the existing v4 table. Block or unwrap `64:ff9b:1::/48` and `2002::/16` if those translators are in scope, and add `ff00::/8`. Cover canonical `::7f00:1` / `::ffff:7f00:1` in unit tests.

apps/researcher/lib/rag/ingest/safe-fetch.ts
```diff
@@ -0,0 +1,277 @@
...
+  for (let hop = 0; hop <= MAX_REDIRECTS; hop++) {
+    const url = await assertPublicUrl(target);
+    const response = await fetch(url, { ...init, redirect: "manual" });
+    const location = response.headers.get("location");
+
+    if (response.status < 300 || response.status >= 400 || !location) {
+      return response;
+    }
+
+    target = new URL(location, url).toString();
```
ducnmm: **[nit]** After a 3xx, `new URL(location, url)` is uncaught. A malformed `Location` throws `TypeError: Invalid URL` instead of `ChatbotError`. Not an SSRF hole — chat still fails source-processing — just an inconsistent 400 vs generic error.

**Suggestion:** Wrap that `new URL` in the same try/catch as `assertPublicUrl` and throw `ChatbotError("bad_request:api", "Invalid URL format")`.

ducnmm:

ducnmm (approved):
## Summary

PR 833 correctly closes the #778 PDF `fileUrl` SSRF: `processSource` now goes through `fetchPublicUrl` instead of a bare `fetch`. Loopback / RFC1918 / link-local / CGNAT / mapped IPv6 / well-known NAT64 / `localhost` are refused before a request is sent. Redirects are re-checked per hop. Residual DNS-rebinding TOCTOU is already called out.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

apps/researcher/lib/rag/ingest/safe-fetch.ts
HoangDucBach: Took the denylist-consistency half of this.

`isBlockedAddress` now unwraps deprecated IPv4-compatible (`::/96`) and SIIT (`::ffff:0:0/96`) through the existing v4 table, same as mapped `::ffff:0:0/96`. Added `ff00::/8` so v6 multicast matches v4 `224/4`. Tests cover canonical `::7f00:1` / `::ffff:7f00:1`, plus SIIT `::ffff:0:7f00:1` and `ff02::1`. `::ffff:7f00:1` was already blocked (hex spelling of mapped loopback); it was only missing from the list.

Left out local-use NAT64 `64:ff9b:1::/48` and 6to4 `2002::/16`. Those translators are not in scope for this fetch path — well-known `64:ff9b::/96` stays as a full prefix block. Unwrapping `2002::/16` through the v4 table would also be the wrong meaning (6to4 locator ≠ connect to that IPv4).

Re-checked on Node 22 + undici: only mapped forms actually hit a 127.0.0.1 listener. `::7f00:1` / SIIT / NAT64 / 6to4 still `EHOSTUNREACH` (multicast `EAFNOSUPPORT`). Same embedding class, not a demonstrated loopback bypass today.

apps/researcher/lib/rag/ingest/safe-fetch.ts
HoangDucBach: fixed

Issue #777, opened 2026-08-24
[Bug] Researcher session JWT has no minimum length check on AUTH_SECRET


anna-stolbovskaja:
### Surface

Example app (chatbot, noter, researcher)

### Network

Local / self-hosted

### Package version

main@1ae17acbc2d3631b4d0e32bfb28e3a43bb59ea38 (apps/researcher) + confirmed byte-identical to dev@caff567

### What happened?

**Package version:** `main@1ae17acbc2d3631b4d0e32bfb28e3a43bb59ea38` (`apps/researcher`) — confirmed byte-identical to `dev@caff567`, used for the local repro below. Open PR #644 adds a length check for `apps/chatbot` only (`proxy.ts`, `app/(auth)/auth.ts`, `app/(auth)/api/auth/guest/route.ts`, new `lib/auth-secret.ts`); none of those files are under `apps/researcher`.

The researcher app starts and issues valid session JWTs even when `AUTH_SECRET` is far shorter than the 32-character minimum the same codebase already enforces for Enoki.

### Steps to reproduce

1. Start `apps/researcher` locally with `AUTH_SECRET=short` (5 characters).
2. Insert a test user row.
3. Sign an HS256 JWT `{userId, publicKey, accountId}` with that same
   5-byte value using `jose`.
4. `POST /api/chat` with `Cookie: session=<forged JWT>`.

### Expected

Startup fails, or session verification rejects a secret shorter than 32 characters — the same rule `apps/researcher/lib/auth/enoki-challenge.ts` already enforces (`AUTH_SECRET must contain at least 32 characters`).

### Actual

The app starts with no warning and accepts the forged cookie into the authenticated pipeline:
```text
HTTP 200
data: {"type":"start","messageId":"f6edc216-…"}
data: {"type":"error","errorText":"Missing Authentication header"}
```
(The final error is the intentionally invalid local OpenRouter key, not a session-verification failure.) `apps/researcher/lib/auth/session.ts` and `session-token.ts` sign/verify with the raw `process.env.AUTH_SECRET` value, unlike `enoki-challenge.ts`'s `getSecret()`.

### Logs or error text

```shell
data: {"type":"start","messageId":"f6edc216-bfb6-41bc-8dd8-6202bf968aed"}
data: {"type":"error","errorText":"Oops, an error occurred!"}
data: {"type":"error","errorText":"Missing Authentication header"}
```

### Checks

- [x] I searched existing issues and this is not a duplicate.
[...]

PR #834, opened 2026-08-29
fix(researcher): validate AUTH_SECRET before signing or verifying a session

### LINKED ISSUE #781: [Feature Request] Add Scoped Namespace Hard Constraints / Invalidation Flags in MemWal SDK
## 1. Motivation

In complex state tracking agents (like Tabletop RPG World Engines or State Machines on Walrus), certain memories represent **Hard State Invalidation Constraints** (e.g. `[npc_vitality|id=lord_malakor|status=DECEASED]` or `[deprecated_api|v1=true]`).

Currently, agents have to manually parse and filter retrieved vector candidates in prompts. Adding a native `invalidate_previous_keys` or `hard_override` flag in the `remember` SDK would allow agents to automatically tombstone or supersede older embeddings on-chain.

---

## 2. Proposed SDK Method

```typescript
await client.remember({
  text: "[npc|id=malakor|status=DECEASED] Slain in battle",
  namespace: "npc_vitality",
  supersedes: { key: "id", value: "malakor" } // Tombstones previous 'malakor' embeddings in vector search
});
```

HoangDucBach:
## Summary

- validate `AUTH_SECRET` before it is used to sign or verify a session cookie, matching the 32 character rule the app already applied to Enoki challenges
- cover both raw readers, `lib/auth/session.ts` and `proxy.ts`, and drop the duplicate guard in `lib/auth/enoki-challenge.ts` so one rule covers every caller
- keep the check inside the functions so `next build` with the short Docker placeholder still succeeds

Closes #781

### The session path trusted whatever the variable held

`lib/auth/session.ts` and `proxy.ts` both did `new TextEncoder().encode(process.env.AUTH_SECRET)` at module scope. `lib/auth/enoki-challenge.ts`, in the same folder, already rejected anything under 32 characters for that same variable, so the session path was the odd one out.

### A short secret is brute-forceable

The reported 5 character value signs valid HS256 cookies. One captured cookie is enough to recover the secret offline and mint sessions.

### An unset variable is worse, and was not in the report

`new TextEncoder().encode(undefined)` returns zero bytes, and jose signs and verifies HS256 with an empty key without complaint. A deployment that never set the variable therefore issued and accepted cookies signed with a key anyone can reproduce, so forging a session for any user id needed no secret at all.

That state is reachable by accident rather than in theory:

- `.env.example` ships `AUTH_SECRET=` with no value
- the runtime image starts from a fresh base, so it does not inherit the builder stage placeholder

### The report named only part of the surface

It cited `session.ts` and `session-token.ts`. `proxy.ts` carried the identical raw read and gates every page.

## Fix

`lib/auth/auth-secret.ts` holds the rule:

- `getAuthSecret` returns the value only when it is present and at least 32 characters, and throws otherwise
- `getAuthSecretKey` returns that value encoded, for jose
- `session.ts`, `proxy.ts`, and `enoki-challenge.ts` all call it, so the threshold lives in one place

Two deliberate choices:

- Validation happens when the secret is used, not at module scope. The Dockerfile builder stage sets a 22 character `AUTH_SECRET` so `next build` can prerender, and a module scope check would fail the image build instead of a request.
- `proxy.ts` reads the secret whether or not a cookie arrived. Deferring it until a token appears would let a deployment with no secret serve the login page as though the caller were merely signed out.

## Validation

- `pnpm --filter researcher test:unit` — 28 pass, 12 of them new, covering unset, empty, 31 characters, exactly 32, and the reported `short`
- one test signs and verifies a JWT with an empty key to record why the guard exists, then asserts the guard refuses that case
- source-level tests pin the call sites, so reintroducing a raw read or hoisting the check to module scope fails a test rather than passing quietly
- `AUTH_SECRET=build-time-placeholder pnpm next build` succeeds, so the image build is unaffected
- `pnpm exec tsc --noEmit` clean

## Out of scope

- `apps/chatbot`, where open PR #644 adds the same guard through its own `lib/auth-secret.ts`. Touching those files here would collide with that branch.
- promoting the check to startup rather than first use, which would need a boot-time entry point that also runs before `next build` prerendering
- `.env.example` still ships an empty `AUTH_SECRET`; with this change a local run fails loudly instead of accepting an empty key, but generating a value is left to the operator

### DIFF
--- a/apps/researcher/lib/auth/session.ts
+++ b/apps/researcher/lib/auth/session.ts
@@ -3,13 +3,13 @@ import "server-only";
 import { jwtVerify } from "jose";
 import { cookies } from "next/headers";
 import { getUserById } from "@/lib/db/queries";
+import { getAuthSecretKey } from "@/lib/auth/auth-secret";
 import {
   SESSION_MAX_AGE_SECONDS,
   signSessionIdentity,
 } from "@/lib/auth/session-token";
 
 const COOKIE_NAME = "session";
-const secret = new TextEncoder().encode(process.env.AUTH_SECRET);
 
 type SessionUser = {
   id: string;
@@ -28,7 +28,7 @@ export async function getSession(): Promise<{ user: SessionUser } | null> {
   if (!token) return null;
 
   try {
-    const { payload } = await jwtVerify(token, secret);
+    const { payload } = await jwtVerify(token, getAuthSecretKey());
     if (typeof payload.userId !== "string") return null;
 
     const user = await getUserById(payload.userId);
@@ -55,7 +55,7 @@ export async function createSession(
 ): Promise<void> {
   const token = await signSessionIdentity(
     { userId, publicKey, accountId },
-    secret
+    getAuthSecretKey()
   );
   const cookieStore = await cookies();
   cookieStore.set(COOKIE_NAME, token, {
--- a/apps/researcher/lib/auth/session.ts
+++ b/apps/researcher/lib/auth/session.ts
@@ -27,8 +27,13 @@ export async function getSession(): Promise<{ user: SessionUser } | null> {
   const token = cookieStore.get(COOKIE_NAME)?.value;
   if (!token) return null;
 
+  // Same order as proxy.ts: a missing/short AUTH_SECRET must throw, not look
+  // like an invalid cookie. /api/auth/* skips the proxy guard, so this is the
+  // path GET /api/auth/profile takes.
+  const secret = getAuthSecretKey();
+
   try {
-    const { payload } = await jwtVerify(token, getAuthSecretKey());
+    const { payload } = await jwtVerify(token, secret);
     if (typeof payload.userId !== "string") return null;
 
     const user = await getUserById(payload.userId);
--- a/apps/researcher/lib/auth/auth-secret.ts
+++ b/apps/researcher/lib/auth/auth-secret.ts
@@ -1,0 +1,36 @@
+// One rule for AUTH_SECRET across the app. session.ts and proxy.ts used to read
+// the variable raw, which had two silent failure modes:
+//
+//   - unset: `new TextEncoder().encode(undefined)` returns zero bytes, and jose
+//     signs and verifies HS256 with an empty key without complaining. The app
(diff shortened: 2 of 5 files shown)


apps/researcher/lib/auth/session.ts
```diff
@@ -28,7 +28,7 @@ export async function getSession(): Promise<{ user: SessionUser } | null> {
   if (!token) return null;
 
   try {
-    const { payload } = await jwtVerify(token, secret);
+    const { payload } = await jwtVerify(token, getAuthSecretKey());
```
ducnmm: **[suggestion]** `getAuthSecretKey()` sits inside the `try` that treats every throw as an invalid cookie. A missing or short `AUTH_SECRET` therefore looks like "not authenticated" rather than a misconfiguration. `proxy.ts` already calls the guard *before* reading the cookie; `/api/auth/*` skips that proxy check, so `GET /api/auth/profile` returns 401 instead of failing loud.

**Suggestion:** Read the key after the token check and before the verify `try`, matching `proxy.ts`. Same idea in `enoki-challenge.ts` `verifyAndConsumeEnokiChallenge` (short secret currently surfaces as wallet-ownership failure).

ducnmm:
Need this landed on `dev` before 13:00 ICT today (2026-09-07). Covers WALM-417 and WALM-418 (same bug). CI is green — please address review and merge so this can ride the staging → prod promotion this afternoon.

ducnmm (approved):
## Summary

This PR fixes the WALM-417/418 / GitHub #777 hole: researcher session cookies were signed and verified with whatever `AUTH_SECRET` held, including empty and the reported 5-character `short`. Centralizing the Enoki 32-character rule is the right shape. Unit tests and Researcher CI look good.

The forged-cookie path is blocked. One follow-up: the PR body says `Closes #781`, but #781 is an unrelated SDK feature request; the AUTH_SECRET bug is #777.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 0

apps/researcher/lib/auth/session.ts
HoangDucBach: fixed nit

PR #840, opened 2026-08-31
fix(chatbot): rate-limit unauthenticated guest user creation (WALM-415)

### LINKED ISSUE #775: [Bug] Unauthenticated guest login has no rate limit on user-row creation
### Surface

### Network

Local / self-hosted

### Package version

main@1ae17acbc2d3631b4d0e32bfb28e3a43bb59ea38 (apps/chatbot)

### What happened?

Every unauthenticated `GET /api/auth/guest` request inserts a new guest `User` row, with no per-IP or per-window limit observed in local testing.

### Steps to reproduce

Against an empty `User` table, send five unauthenticated requests:
```bash
for i in 1 2 3 4 5; do curl -s -o /dev/null -w '%{http_code}\n' \
  'http://127.0.0.1:3001/api/auth/guest?redirectUrl=/'; done
```

### Expected

Repeated guest logins from one source are throttled after a small budget.

### Actual

Five HTTP 307 redirects and five new `User` rows; no throttling was observed under local `next dev`. `apps/chatbot/app/(auth)/auth.ts`'s guest `authorize()` calls `createGuestUser()` unconditionally on every call, and `apps/chatbot/lib/db/queries.ts`'s `createGuestUser()` has no rate-limit check itself.

### Logs or error text

```shell

```

### Checks

HoangDucBach:
## Ticket

WALM-415 — https://linear.app/mysten-labs/issue/WALM-415
GH #775 — https://github.com/MystenLabs/MemWal/issues/775

## What changed?

- `GET /api/auth/guest` now rate-limits before `signIn("guest")` / `createGuestUser()`.
- Budget: 5 requests per IP and 60 global per 15-minute window. The 6th request returns 429 with `Retry-After`.
- Production requires Redis and fails closed (503) if the limiter is unavailable. Local `next dev` uses Redis when ready, otherwise an in-memory counter so the reported curl repro is throttled.
- Playwright (`PLAYWRIGHT` / `CI_PLAYWRIGHT`) skips the limiter so e2e from a shared IP does not flake.
- Unit tests cover IP extraction, the per-IP budget, and that the guest route calls the limiter before `signIn`.

## Why is this needed?

Every unauthenticated guest login inserted a new User row with no per-IP or per-window cap. The existing chat IP limiter is production-only and a no-op without Redis, so five local curls created five rows.

## Scope

Chatbot guest login route only. `createGuestUser()` stays a plain insert.

### Out of scope

- Noter / researcher guest or public auth limiters
- Rate-limiting NextAuth `authorize()` itself (the public entry is this GET)

## Test plan

- [x] `pnpm test:unit` in `apps/chatbot`
- [x] Against local `next dev`, six unauthenticated `GET /api/auth/guest` from one IP: five 307s, then 429
- [x] Confirm a sixth User row is not created
- [x] Authenticated guest requests still redirect to `/` without consuming the budget

Fixes #775


ducnmm:
Need this landed on `dev` before 13:00 ICT today (2026-09-07). CI is green and no review is requested yet — please request review and merge so this can ride the staging → prod promotion this afternoon.

apps/chatbot/lib/ratelimit.ts
```diff
@@ -46,3 +96,106 @@ export async function checkIpRateLimit(ip: string | undefined) {
...
+ * otherwise a process-local counter so `next dev` is still throttled.
+ */
+export async function checkGuestAuthRateLimit(request: Request): Promise<void> {
+  if (isTestEnvironment) {
+    return;
+  }
+
+  const ip = getClientIp(request) ?? "unknown";
+
+  const redis = getClient();
```
ducnmm: **[bug]** Production fail-closed is keyed off `getClient()` + `redis?.isReady` without waiting for `connect()`. `getClient()` still fires `client.connect()` in the background (existing singleton). On a Railway restart or Vercel isolate cold start, the first unauthenticated request — which `proxy.ts` sends to `/api/auth/guest` — sees `isReady === false` and throws `GuestAuthRateLimitError(503)` even when Redis is healthy. Researcher already avoids this by awaiting `requireSharedRedisClient()` before `eval`. New visitors cannot enter the app until a later request finds the client ready; `/ping` never warms Redis.

**Suggestion:** Await a ready Redis client (same pattern as `apps/researcher/lib/shared-redis.ts`) before deciding the limiter is unavailable. Treat “connecting” as wait-with-timeout; reserve 503 for missing `REDIS_URL`, connect failure, or `eval` errors.

apps/chatbot/app/(auth)/api/auth/guest/route.ts
```diff
@@ -45,5 +50,25 @@ export async function GET(request: Request) {
     return NextResponse.redirect(new URL("/", request.url));
   }
 
+  try {
+    await checkGuestAuthRateLimit(request);
```
ducnmm: **[bug]** The new budget only wraps this GET handler. Guest `authorize()` in `auth.ts` still calls `createGuestUser()` unconditionally. Auth.js credentials sign-in POSTs to `/api/auth/callback/guest` (`next-auth/lib/actions.js` rewrites `signin` → `callback` for credentials). `proxy.ts` lets all `/api/auth/*` through. An attacker can `GET /api/auth/csrf` then `POST /api/auth/callback/guest` with the csrf cookie/token and insert User rows without consuming this limiter. That contradicts the commit claim that one IP cannot farm guest rows; WALM-415 also called out unconditional `authorize()` / `createGuestUser()`.

**Suggestion:** Run `checkGuestAuthRateLimit` inside the guest Credentials `authorize()` (via `headers()` from `next/headers`) or in `createGuestUser()` itself, so every insert path shares the budget. Keep the GET-route check if you want a fast 429 before `signIn`.

apps/chatbot/lib/ratelimit.ts
```diff
@@ -46,3 +96,106 @@ export async function checkIpRateLimit(ip: string | undefined) {
...
+  return true;
+}
+
+/** Test-only: reset the process-local guest limiter. */
+export function resetMemoryGuestAuthRateLimit(): void {
+  memoryGuestCounters.clear();
+}
+
+/**
+ * Cap unauthenticated guest User inserts. Always on outside Playwright:
```
ducnmm: **[suggestion]** The `checkGuestAuthRateLimit` JSDoc narrates the ticket repro and architecture history (“the chat IP limiter is production-only and a no-op without Redis, which is how five local curls each created a row”) instead of a short why.

**Suggestion:** Replace with one or two lines on behavior: always on outside Playwright; production fail-closed on Redis; dev uses Redis if ready else process-local counters.

apps/chatbot/lib/ratelimit.unit.test.ts
```diff
@@ -0,0 +1,141 @@
...
+        )
+      ).toBe("203.0.113.9");
+    } finally {
+      process.env.VERCEL = previous ?? "";
+    }
+  });
+});
+
+describe("checkGuestAuthRateLimit", () => {
+  beforeEach(() => {
```
ducnmm, line 138: **[suggestion]** Unit tests only exercise the in-memory limiter (mocked Redis `isReady: false`, `NODE_ENV !== production`) plus a source-text check that `route.ts` calls the limiter before `signIn`. They do not cover Redis `eval` 429, production 503 when Redis is missing/not ready, the global cap, or `isTestEnvironment` skip. Issue 1 would not fail CI.

**Suggestion:** Add cases for production + Redis not ready → 503, Redis `eval` returning `0` → 429, and Playwright env skipping the limiter. Drive those through `checkGuestAuthRateLimit` with a mock client, not source greps.

apps/chatbot/app/(auth)/api/auth/guest/route.ts
HoangDucBach: fixed in 3c0d3ae199d98b3d6b13cb89f2f595ed4815c07b

apps/chatbot/lib/ratelimit.ts
HoangDucBach: fixed in 3c0d3ae199d98b3d6b13cb89f2f595ed4815c07b

apps/chatbot/lib/ratelimit.unit.test.ts
HoangDucBach: fixed in 3c0d3ae199d98b3d6b13cb89f2f595ed4815c07b

apps/chatbot/lib/ratelimit.ts
```diff
@@ -1,24 +1,145 @@
...
+  client.on("error", () => undefined);
   return client;
 }
 
+function beginConnect(redis: RedisClient): void {
+  if (redis.isOpen || connectPromise) {
+    return;
+  }
+
+  connectPromise = redis.connect().then(
```
ducnmm: **[bug]** `connectPromise` is only cleared in `getReadyRedisClient()`'s `finally` after an await. `getClient()` (still used by `checkIpRateLimit`) calls `beginConnect()` fire-and-forget and never clears it. After authenticated chat warms Redis, `connectPromise` stays as a settled promise. `getReadyRedisClient()` then early-returns on `isReady` and leaves that leftover in place. When the socket later drops (`isReady`/`isOpen` false), `beginConnect` sees a truthy `connectPromise` and skips `redis.connect()`. The next guest login awaits the already-settled promise, returns not-ready, and throws `GuestAuthRateLimitError(503)` even if Redis is healthy again. Recovery happens only on the request after that. Researcher avoids this by always awaiting and nulling `connectPromise` in `getSharedRedisClient()` with no fire-and-forget path.

**Suggestion:** Clear `connectPromise` when `connect()` settles (in the `.then` handlers), not only in `getReadyRedisClient()`'s `finally`. Treat a leftover settled promise as absent in `beginConnect`. Mirror `apps/researcher/lib/shared-redis.ts`: one await-and-clear helper, no early `isReady` return that skips the drain.

apps/chatbot/lib/ratelimit.ts
```diff
@@ -1,24 +1,145 @@
...
 
-function getClient() {
-  if (!client && process.env.REDIS_URL) {
-    client = createClient({ url: process.env.REDIS_URL });
-    client.on("error", () => undefined);
-    client.connect().catch(() => {
-      client = null;
-    });
+export const GUEST_AUTH_RATE_LIMIT_PER_IP = 5;
+export const GUEST_AUTH_RATE_LIMIT_GLOBAL = 60;
```
ducnmm: **[suggestion]** `GUEST_AUTH_RATE_LIMIT_GLOBAL = 60` per 15 minutes applies to every unauthenticated visitor. `proxy.ts` redirects any missing session to `/api/auth/guest`, so this is the only way into the chatbot. 60 site-wide successes in a window is ~4/min; noter/researcher public auth uses 150–300 per 60s. Off-Vercel, `x-real-ip` / `x-forwarded-for` are also spoofable, so 60 crafted IPs exhaust the global key and 429 everyone else for the rest of the window (JSON body, not a page). That is tighter than WALM-415's "small per-source budget" and can lock the public demo under modest bursts.

**Suggestion:** Keep the per-IP 5 if that is the abuse target, and raise the global cap (and/or shorten the window) to something in line with noter/researcher. If 60 is intentional for User-row cost, document it as a site-wide admission cap, not only an anti-farming backstop.

apps/chatbot/app/(auth)/auth.ts
```diff
@@ -68,6 +70,11 @@ export const {
       id: "guest",
       credentials: {},
       async authorize() {
+        await checkGuestAuthRateLimit(
```
ducnmm: **[suggestion]** `GuestAuthRateLimitError` thrown from `authorize()` is not an Auth.js `AuthError`. `@auth/core` wraps it as `CallbackRouteError`. Direct `POST /api/auth/callback/guest` (allowed by `proxy.ts`) then redirects to the Auth error page instead of 429. On the GET path, `signIn()` uses `raw`, so the same wrap is rethrown: a peek-pass / consume-deny (concurrent sixth request, or Redis error between peek and consume) becomes an unhandled 500 from the guest route, whose try/catch only wraps the peek. Inserts are still blocked; the public status/`Retry-After` are not.

**Suggestion:** Catch `GuestAuthRateLimitError` in `authorize()` and rethrow a small Auth.js-safe error, or wrap `signIn("guest", …)` and map `CallbackRouteError` whose `cause` is `GuestAuthRateLimitError` to the same 429/503 JSON as the peek. Sequential GET still relies on the peek for the documented 429.

ducnmm (changes requested):
## Summary

The follow-up commit correctly closes the two holes from the previous review: guest inserts now consume the budget inside Credentials `authorize()` before `createGuestUser()`, and `getReadyRedisClient()` waits for `connect()` instead of 503ing on a cold isolate. Sequential `GET /api/auth/guest` is capped (peek for a JSON 429, consume on the in-process `signIn` → `authorize` path), Playwright is skipped, and production fail-closed / dev memory fallback match the ticket. Dominant remaining risks are a stale `connectPromise` that can 503 a healthy Redis after chat traffic has initialized the singleton, and a global 60/15min cap on a funnel that every unauthenticated visitor hits.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 0

apps/chatbot/lib/ratelimit.ts
HoangDucBach: fixed with sugg — connect() now always await-and-clear; leftover settled promises no longer skip reconnect.

apps/chatbot/lib/ratelimit.ts
HoangDucBach: fixed with sugg — kept per-IP 5, raised global to 150/60s to match noter public auth.

apps/chatbot/app/(auth)/auth.ts
HoangDucBach: fixed with sugg — authorize() rethrows CredentialsSignin; GET /api/auth/guest maps that (and CallbackRouteError cause) to the same 429/503 JSON.

ducnmm (commented):
## Summary

Follow-up on REQUEST_CHANGES for `3c0d3ae1`. Both remaining items are fixed on `695b840f`.

`connectPromise` leftover: `getClient()` is gone; chat and guest auth share `getReadyRedisClient()`, and `startConnect` clears `connectPromise` in `finally`. After reconnect gives up, the next guest call reconnects instead of 503ing.

`GuestAuthRateLimitError` vs Auth.js: `authorize()` maps limiter errors to `CredentialsSignin` (`too_many_requests` / `service_unavailable`). Peek stays `consume: false` on GET; consume stays in `authorize()` before `createGuestUser()`.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

ducnmm (approved):
LGTM, all checks green and previous review feedback fully addressed.

PR #841, opened 2026-08-31
ci(contract): gate SEAL cross-account deny invariant (COMG-714)

ducnmm:
## Summary
CI now fails if the SEAL cross-account deny invariant is deleted.

`sui move test` still runs the full suite, but deleting `test_seal_approve_delegate_requires_matching_owner` would still go green. This gate requires that test to exist and pass.

## Changes
- `scripts/check-compatibility-contract.mjs` asserts `services/contract/tests/account_tests.move` contains `fun test_seal_approve_delegate_requires_matching_owner` and `ENoAccess`.
- `move-contract` job runs `sui move test --filter seal_approve_delegate_requires_matching_owner` after the full suite.
- One-line note in `docs/relayer/versioning-and-compatibility.md`.

COMG-714

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -551,3 +551,7 @@ jobs:
       - name: Run Move unit tests
         working-directory: services/contract
         run: sui move test
+
+      - name: Require SEAL cross-account deny invariant
+        working-directory: services/contract
+        run: sui move test --filter seal_approve_delegate_requires_matching_owner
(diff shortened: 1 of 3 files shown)


jessiemongeon1:
## Style Guide Audit

HoangDucBach (approved):
Approved!
The extra `sui move test` + log grep is the real gate: Sui exits 0 when a filter matches nothing, so the grep is necessary. The Move test still denies a delegate asking for another owner’s id with ENoAccess

PR #843, opened 2026-08-31
feat(server): expose writes ok|paused on /health (COMG-717)

### LINKED ISSUE #474: Ops: no public status page or writes-paused health signal (a multi-day write pause was indistinguishable from an integration bug)
## What happened

Around 2026-07-15, writes (uploads) returned `503` for roughly two days on both `relayer.memwal.ai` and `relayer.memory.walrus.xyz`, while reads and `/health` kept returning `200`. As of 2026-07-17 both `/health` endpoints are back to `200`, so this looks resolved — filing for the durable gap it exposed rather than the incident itself.

## Problem

There was no public status page, changelog, or advance notice, and `/health` returning `200` did not reflect that writes were paused. From an integrator's side, a healthy-looking relayer that rejects every write is indistinguishable from a bug in your own code, which cost real debugging time before we concluded it was upstream. The `503` body said only "security upgrade".

## Ask

- A public status / incident page for the hosted relayers.
- A health signal that distinguishes "relayer up, writes paused" from "relayer up, writes accepted" (e.g. a `writes: ok|paused` field), so clients can fail gracefully instead of erroring once per call.
- Advance notice for planned write pauses.

Related: #402 (writes fail silently when the funding wallet runs out of WAL) is a different cause of the same class of problem — a write that cannot succeed, with no upstream signal a client can read.

ducnmm:
## Summary
`GET /health` now reports `writes: "ok" | "paused"` from `WRITES_PAUSED`, so an intentional write pause is distinguishable from integrator bugs. HTTP 200 and `write_ready` (sidecar liveness) stay unchanged.

## Changes
- `HealthResponse.writes` is `"paused"` when `WRITES_PAUSED` is `1`/`true`/`yes`; empty or unset is `"ok"`.
- Status app treats `writes === "paused"` as degraded / writes-paused even on HTTP 200.
- Relayer API docs distinguish `writes` vs `write_ready`.

COMG-717
Fixes #474


jessiemongeon1:
## Style Guide Audit

ducnmm:
@nikola0x0 gentle ping — still waiting on review of `8da5c37c80b0ddadc5c1d41b71c13ea99ff00a7c`.

services/server/src/routes/admin.rs
```diff
@@ -149,6 +149,7 @@ pub async fn health(State(state): State<Arc<AppState>>) -> Json<HealthResponse>
             ask: ASK_SYSTEM_PROMPT_VERSION.to_string(),
         },
         write_ready: sidecar_write_ready(&state).await,
+        writes: writes_health_status(state.config.writes_paused),
```
nikola0x0, line 153: [P2] Connect the pause signal to actual write admission

`WRITES_PAUSED` is newly introduced here, but its only runtime consumer is this health response: `remember`, `remember_bulk`, `remember_manual`, `analyze`, and the router/middleware never check `config.writes_paused`. Starting the relayer with `WRITES_PAUSED=true` therefore advertises `writes: "paused"` and degrades the public status page while authenticated writes still proceed. Conversely, the field cannot reflect a pause enforced elsewhere unless an operator independently synchronizes it. Please use one source of truth for admission and health (or connect this to the actual pause mechanism), and add a route-level test showing that the reported paused state matches write rejection while reads remain available.

ducnmm:
@nikola0x0 gentle ping — still waiting on review of `8da5c37c80b0ddadc5c1d41b71c13ea99ff00a7c`.

nikola0x0 (commented):
Blocking on connecting the reported pause state to write admission (details inline).

The new flag is consumed by /health but no write handler or middleware reads it, so the advertised state can disagree with whether writes are accepted.

Validation: status TypeScript/Vite build passed; isolated probe checks passed for ok, paused, and older health responses without the field. Inspected the Rust configuration and write routes; Rust tests were not rerun locally (the PR CI reports them passing).

ducnmm:
Addressed the write-admission note on `7f8e7661`: `WRITES_PAUSED` now 503s `remember` / `remember_bulk` / `remember_manual` / `analyze`. `GET /health` stays HTTP 200 with `writes: "paused"`; reads stay open. Please re-review.

nikola0x0 (approved):
LGTM!

PR #845, opened 2026-08-31
fix(server): report restore decrypt failures instead of skipped (COMG-719)

### LINKED ISSUE #399: restore(): report decrypt/embed failures instead of dropping them
## Summary

`restore()` cannot report per-blob download/decrypt/embed failures. In SDK `0.0.7`, the documented response has only `restored`, `skipped`, `total`, `namespace`, and `owner`; decrypt or embed failures are dropped silently and count as neither restored nor skipped.

That makes a recovery operation impossible to audit: a caller cannot distinguish a complete restore from a best-effort restore with missing blobs, identify which Blob IDs failed, or decide whether a retry is safe.

## Source-level reproduction

Installed package: `@mysten-incubation/memwal@0.0.7`.

`dist/memwal.js` documents the current response semantics directly:

```text
restored — completed download → decrypt → embed → DB insert
skipped — already indexed
Decrypt / embed failures are dropped silently and count as neither.
```

`RestoreResult` in `dist/types.d.ts` exposes only:

```ts
interface RestoreResult {
  restored: number;
  skipped: number;
  total: number;
  namespace: string;
  owner: string;
}
```

The existing pagination request in #291 is related to how many blobs are inspected, but does not expose failures within an inspected page.

## Expected behavior

Please expose enough information to audit and retry incomplete recovery, for example:

```ts
interface RestoreResult {
  restored: number;
  skipped: number;
  failed: number;
  failures?: Array<{
    blobId: string;
    stage: "download" | "decrypt" | "embed" | "index";
    code: string;
  }>;
  // existing fields...
}
```

An alternative would be fail-closed behavior by default, with an explicit `bestEffort` option that returns structured failures.

## Why this matters

`restore()` is the disaster-recovery path for rebuilding the local index from Walrus. Silent partial success can make an agent believe its durable memory was fully recovered when some records are unavailable.

ducnmm:
## Summary

`POST /api/restore` no longer folds permanent decrypt/UTF-8 failures into `skipped`. Those blobs stay in the `restore_failed_blobs` negative cache, but the response now reports them as `failed`.

- `skipped` is on-chain blobs already in the local **success** index only
- `failed` is the owner+namespace negative-cache size plus new permanent failures this call (no double-count)
- Additive JSON field on `RestoreResponse`; TypeScript `failed?: number` (SDK defaults omitted to `0`); Python `failed: int = 0`

Fixes #399
COMG-719

## Test plan

- [x] `cargo test --bin memwal-server restore_`
- [x] `node --test packages/sdk/test/restore-truncated.test.mjs`
- [x] `pytest packages/python-sdk-memwal/tests/test_client.py::TestRestore`
- [x] `node scripts/verify-manual-sdk-release.mjs`

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -365,7 +365,7 @@ Cross-namespace and cross-owner reads are not just filtered out of results — t
 |---|---|---|
 | `restored` | Blobs the relayer just rebuilt this call | Pulled from Walrus → SEAL decrypted → re-embedded → inserted as a new row |
 | `skipped` | On-chain blobs already in the local **success** index | No work needed; relayer left them as-is. Does not include decrypt/UTF-8 failures. |
-| `failed` | Permanent decrypt/UTF-8 failures | Negative cache for this owner+namespace plus new permanent failures this call. Older relayers omit the field; SDKs default it to `0`. |
+| `failed` | Permanent decrypt/UTF-8 failures | On-chain blobs in this page that are negative-cached, plus new permanent failures this call. Older relayers omit the field; SDKs default it to `0`. |
 | `total` | All on-chain blobs the relayer saw for `(owner, namespace)` | Before the limit was applied |
 | `namespace` | Echo of the request | |
 | `owner` | Resolved owner address | |
@@ -373,7 +373,7 @@ Cross-namespace and cross-owner reads are not just filtered out of results — t
 
 `truncated=true` means this restore is **known-retryable-incomplete**: more missing blobs than `limit` allowed this call to restore, **or** the sidecar's owner-wide candidate fetch hit its cap **and** raising `limit` can still expand that fetch (`limit < 20`). Once the sidecar cap is saturated (`limit >= 20`, cap pinned at 100), truncation follows this call's missing-blob page length, not onchain `total`. A fully restored namespace does not loop. `truncated=false` is **not** proof the sidecar saw every onchain blob; blobs beyond the owner-wide sidecar candidate cap can still be missing. WALM-451 tracks a `sourceCapped` field for that case. Relayers older than WALM-319 omit `truncated`; SDKs default it to `false`.
 
-Permanent decrypt or invalid-UTF-8 failures count in `failed`, not `skipped`. Transient download/decrypt/embed errors are still not counted in `restored`, `skipped`, or `failed` and may be retried. `restored + skipped + failed` is therefore a lower bound on inspected blobs, not a strict equality with `total`.
+Permanent decrypt or invalid-UTF-8 failures count in `failed`, not `skipped`. Transient download/decrypt/embed errors are still not counted in `restored`, `skipped`, or `failed` and may be retried (`truncated=true` when a page yields only those). `restored + skipped + failed` therefore never exceeds `total`, and falls short of it whenever transient errors leave blobs uncounted.
 
 #### Default and limit
 
--- a/SKILL.md
+++ b/SKILL.md
@@ -155,7 +155,7 @@ const stored = await memwal.waitForRememberJob(accepted.job_id, {
 | `recall({ query, limit?, topK?, namespace?, maxDistance? })` *(preferred)* or `recall(query, limit?, namespace?)` | Semantic search for memories | `{ results: [{ blob_id, text, distance }], total }` |
 | `analyze(text, namespace?)` | Extract facts and accept one memory job per fact | `{ job_ids, facts, fact_count, status, owner }` |
 | `analyzeAndWait(text, namespace?, opts?)` | Extract facts and wait for all fact jobs to complete | `{ results, facts, total, succeeded, failed, owner }` |
-| `restore(namespace, limit?)` | Rebuild missing index entries from Walrus | `{ restored, skipped, total, namespace, owner, truncated }` |
+| `restore(namespace, limit?)` | Rebuild missing index entries from Walrus | `{ restored, skipped, failed, total, namespace, owner, truncated }` |
 | `health()` | Check relayer health | `{ status, version }` |
 | `getPublicKeyHex()` | Get hex-encoded public key | `string` |
 
@@ -271,6 +271,7 @@ interface EmbedResult {
 interface RestoreResult {
   restored: number;
   skipped: number;
+  failed?: number;
   total: number;
   namespace: string;
   owner: string;
@@ -363,7 +364,8 @@ Cross-namespace and cross-owner reads are not just filtered out of results — t
 | Field | Counts | Notes |
 |---|---|---|
 | `restored` | Blobs the relayer just rebuilt this call | Pulled from Walrus → SEAL decrypted → re-embedded → inserted as a new row |
-| `skipped` | On-chain blobs already in the local index | No work needed; relayer left them as-is |
+| `skipped` | On-chain blobs already in the local **success** index | No work needed; relayer left them as-is. Does not include decrypt/UTF-8 failures. |
+| `failed` | Permanent decrypt/UTF-8 failures | Negative cache for this owner+namespace plus new permanent failures this call. Older relayers omit the field; SDKs default it to `0`. |
 | `total` | All on-chain blobs the relayer saw for `(owner, namespace)` | Before the limit was applied |
 | `namespace` | Echo of the request | |
 | `owner` | Resolved owner address | |
@@ -371,7 +373,7 @@ Cross-namespace and cross-owner reads are not just filtered out of results — t
 
 `truncated=true` means this restore is **known-retryable-incomplete**: more missing blobs than `limit` allowed this call to restore, **or** the sidecar's owner-wide candidate fetch hit its cap **and** raising `limit` can still expand that fetch (`limit < 20`). Once the sidecar cap is saturated (`limit >= 20`, cap pinned at 100), truncation follows this call's missing-blob page length, not onchain `total`. A fully restored namespace does not loop. `truncated=false` is **not** proof the sidecar saw every onchain blob; blobs beyond the owner-wide sidecar candidate cap can still be missing. WALM-451 tracks a `sourceCapped` field for that case. Relayers older than WALM-319 omit `truncated`; SDKs default it to `false`.
 
-**Silent drops.** A blob that *cannot* be decrypted or embedded (e.g. wrong delegate key, malformed ciphertext, embedding API down) is dropped without counting in `restored` *or* `skipped`. `restored + skipped` is therefore a lower bound on healthy entries, not a strict equality with `total`.
+Permanent decrypt or invalid-UTF-8 failures count in `failed`, not `skipped`. Transient download/decrypt/embed errors are still not counted in `restored`, `skipped`, or `failed` and may be retried. `restored + skipped + failed` is therefore a lower bound on inspected blobs, not a strict equality with `total`.
 
 #### Default and limit
 
--- a/SKILL.md
+++ b/SKILL.md
@@ -271,7 +271,7 @@ interface EmbedResult {
 interface RestoreResult {
   restored: number;
   skipped: number;
-  failed?: number;
+  failed: number;
   total: number;
   namespace: string;
   owner: string;
--- a/packages/sdk/src/types.ts
+++ b/packages/sdk/src/types.ts
@@ -449,6 +449,12 @@ export interface ListNamespacesOptions {
 export interface RestoreResult {
     restored: number;
     skipped: number;
+    /**
+     * Permanent decrypt/UTF-8 failures: the owner+namespace negative cache
+     * plus any new permanent failures this call. Relayers older than
+     * COMG-719 omit this field; the SDK defaults it to `0`.
+     */
+    failed?: number;
     total: number;
     namespace: string;
     owner: string;
--- a/packages/sdk/src/types.ts
+++ b/packages/sdk/src/types.ts
@@ -450,11 +450,11 @@ export interface RestoreResult {
     restored: number;
     skipped: number;
     /**
-     * Permanent decrypt/UTF-8 failures: the owner+namespace negative cache
-     * plus any new permanent failures this call. Relayers older than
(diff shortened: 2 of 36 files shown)


jessiemongeon1:
## Style Guide Audit

Audited **6** file(s) against the [Sui Documentation Style Guide](https://docs.sui.io/references/contribute/style-guide).

**2** violation(s) found. All must be fixed before merge.

### `docs/mcp/changelog.mdx` (2 violation(s))

_2 violation(s) (0 regex, 2 claude)_

- **Line 36** — passive-voice
  - Current: `This release reports restore `failed` counts and retries the same page when truncation is a transient download or embed blip.`
  - Fix: `This release reports restore `failed` counts and retries the same page when a transient download or embed blip causes truncation.`
- **Line 40** — no-slashes
  - Current: `download/embed blip`
  - Fix: `download or embed blip`

---

ducnmm:
Rebased onto origin/dev. Python stays 0.1.9; the restore `failed` changelog bullet sits under the existing 0.1.9 heading.

Style-guide audit: `onchain` spelling and `might` on the restore skipped/failed copy.

ducnmm:
@harrymove-ctrl gentle ping — still waiting on review of `a9f1364714dc7eb345fb9cca9b8100dd336701c0`.

ducnmm:
@harrymove-ctrl gentle ping — still waiting on review of `a9f1364714dc7eb345fb9cca9b8100dd336701c0`.

ducnmm:
Rebased onto origin/dev (`48a4b22e`). Restore decrypt failures still report as `failed` (not `skipped`); kept WALM-431 `truncated`. Python 0.1.9 / TS 0.1.6 unchanged.

@harrymove-ctrl please review this head.

ducnmm:
@harrymove-ctrl gentle ping — still waiting on review of `48a4b22ecb346d7cb7d07181f7a25823b94707de`.

services/server/src/routes/admin.rs
```diff
@@ -546,6 +546,37 @@ fn clamp_restore_limit(limit: usize) -> usize {
...
+    newly_failed: usize,
+) -> (usize, usize) {
+    let existing_set: std::collections::HashSet<&str> =
+        existing_blob_ids.iter().map(|s| s.as_str()).collect();
+    let skipped = on_chain_blob_ids
+        .iter()
+        .filter(|id| existing_set.contains(id.as_str()))
+        .count();
+    (skipped, failed_blob_ids.len() + newly_failed)
+}
```
harrymove-ctrl: **`failed` can exceed `total` — it's a lifetime counter in a per-page response.**

`total = all_blob_ids.len()`, and `all_blob_ids` comes from `query_blobs_by_owner(..., Some(limit))` — the limit is forwarded to the sidecar (`storage/walrus.rs:838`), so `total <= limit <= 100`, default 10.

`get_failed_blob_ids` (`storage/db.rs:2981`) is `SELECT blob_id FROM restore_failed_blobs WHERE owner=$1 AND namespace=$2` — no `LIMIT`, and never intersected with the on-chain page. A namespace with 60 historical permanent failures returns `total: 10, failed: 60`.

The asymmetry lives inside this one helper: `skipped` **is** intersected with the on-chain list (and `restore_skip_fail_counts_does_not_count_off_chain_existing` proves it), `failed` is not. An operator triaging a namespace can't tell how much of `failed` relates to the page they just asked about.

Suggestion below also drops `newly_failed`, see the separate comment on line 921.

```suggestion
/// Count restore `skipped` / `failed` over the on-chain page.
///
/// `skipped` is on-chain blobs already in the local **success** index
/// (`existing_blob_ids`). Negative-cached blob IDs are not skipped.
///
/// `failed` is on-chain blobs in `failed_blob_ids` (the owner+namespace
/// negative cache). Both counts share `on_chain_blob_ids` as their domain,
/// so neither can exceed `total`.
fn restore_skip_fail_counts(
    on_chain_blob_ids: &[String],
    existing_blob_ids: &[String],
    failed_blob_ids: &[String],
) -> (usize, usize) {
    let existing_set: std::collections::HashSet<&str> =
        existing_blob_ids.iter().map(|s| s.as_str()).collect();
    let failed_set: std::collections::HashSet<&str> =
        failed_blob_ids.iter().map(|s| s.as_str()).collect();
    let skipped = on_chain_blob_ids
        .iter()
        .filter(|id| existing_set.contains(id.as_str()))
        .count();
    let failed = on_chain_blob_ids
        .iter()
        .filter(|id| failed_set.contains(id.as_str()))
        .count();
    (skipped, failed)
}
```

This intentionally breaks `restore_skip_fail_counts_uses_full_negative_cache_len`, which codifies the un-intersected behavior — that test is the thing I'm questioning, so it would need to flip.

If the cumulative number is genuinely wanted for triage, it deserves its own field (`failed_cached_total`) rather than sharing a name with a per-page count.

services/server/src/routes/admin.rs
```diff
@@ -693,12 +725,15 @@ async fn restore_unbounded(
         missing_blob_ids.len(),
         limit,
     );
-    let skipped = total - missing_blob_ids.len();
+    let (skipped, failed) =
+        restore_skip_fail_counts(&all_blob_ids, &existing_blob_ids, &failed_blob_ids, 0);
```
harrymove-ctrl: **`newly_failed` is dead at the only call site.**

This passes `0`, and the real addition happens ~190 lines later as `let failed = failed + newly_failed;` (line 921). So the parameter exists only to be exercised by unit tests — the no-double-count invariant its doc comment describes ("Callers must not include blobs already in `failed_blob_ids` in `newly_failed`") is enforced nowhere near the code that actually performs the addition. The tests give false confidence about an invariant they don't guard.

Pairs with the suggestion on the helper above:

```suggestion
        restore_skip_fail_counts(&all_blob_ids, &existing_blob_ids, &failed_blob_ids);
```

Line 921's `let failed = failed + newly_failed;` then stays as-is and becomes the single honest place the two sources combine.

SKILL.md
```diff
@@ -363,15 +364,16 @@ Cross-namespace and cross-owner reads are not just filtered out of results — t
...
 
-**Silent drops.** A blob that *cannot* be decrypted or embedded (e.g. wrong delegate key, malformed ciphertext, embedding API down) is dropped without counting in `restored` *or* `skipped`. `restored + skipped` is therefore a lower bound on healthy entries, not a strict equality with `total`.
+Permanent decrypt or invalid-UTF-8 failures count in `failed`, not `skipped`. Transient download/decrypt/embed errors are still not counted in `restored`, `skipped`, or `failed` and may be retried. `restored + skipped + failed` is therefore a lower bound on inspected blobs, not a strict equality with `total`.
```
harrymove-ctrl: This sentence is wrong in both directions today.

`failed` includes blobs excluded **before** download (negative-cached, never inspected this call), so the sum over-counts inspected blobs — and with the unbounded negative cache it can exceed `total` outright.

If the helper suggestion lands:

```suggestion
Permanent decrypt or invalid-UTF-8 failures count in `failed`, not `skipped`. Transient download/decrypt/embed errors are still not counted in `restored`, `skipped`, or `failed` and may be retried. `restored + skipped + failed` therefore never exceeds `total`, and falls short of it whenever transient errors leave blobs uncounted.
```

packages/sdk/src/types.ts
```diff
@@ -449,6 +449,12 @@ export interface ListNamespacesOptions {
 export interface RestoreResult {
     restored: number;
     skipped: number;
+    /**
+     * Permanent decrypt/UTF-8 failures: the owner+namespace negative cache
+     * plus any new permanent failures this call. Relayers older than
+     * COMG-719 omit this field; the SDK defaults it to `0`.
+     */
+    failed?: number;
```
harrymove-ctrl: Nit: `failed` is optional here while `truncated` two fields down is required — yet `restore()` fills both with a `??` default, so the SDK guarantees both. Consumers now have to write `result.failed ?? 0` for a field that is always present.

`truncated: boolean` already proves the codebase tolerates the required-field-plus-`??`-default pattern at the wire boundary, so:

```suggestion
    failed: number;
```

Non-blocking — just the inconsistency between the two fields, both introduced the same way.

harrymove-ctrl (changes requested):
Reviewed against [WALM-480](https://linear.app/mysten-labs/issue/WALM-480/postmortem-ai-6-triage-structured-per-blob-restore-failures) (postmortem AI-6). CI is green and the decrypt half is well-executed — `RestoreDecrypt` is the right shape, the negative-cache read happens before any write this call so there's genuinely no double-count, docs coverage across relayer/TS/Python/`SKILL.md`/`llms-full.txt` is consistent, and both SDKs handle the omitted-field path with tests.

Requesting changes on two things that block "Fixes #399", plus inline suggestions.

---

### 1. Embed failures are still silently dropped — AI-6's other half

AI-6 reads *"so decrypt/**embed** failures cannot be silently omitted"*, and this PR claims `Fixes #399`. But `admin.rs:946` is unchanged: an embed error logs a warning and returns `None`, landing in neither `restored`, `skipped`, nor `failed`.

With the embedder down and `source_capped=false`, restore returns:

```json
{ "restored": 0, "skipped": 0, "failed": 0, "total": 10, "truncated": false }
```

`restore_is_truncated` returns early when `!source_capped` (`admin.rs:530`), so there is no retry signal either. That is a fully silent no-op — exactly the failure mode AI-6 exists to eliminate.

Not negative-caching embed failures is correct (they're transient). Counting them is a separate concern. Either would close this:

- a `transient_failed` count alongside `failed`, or
- force `truncated=true` when an inspected page yields neither a success nor a permanent failure, so the caller at least knows to retry.

If embed is intentionally deferred, could we drop `Fixes #399` and file the follow-up, so AI-6 doesn't auto-close half-done?

### 2. No test covers the actual behavior change

The added Rust tests exercise the pure helper and `RestoreResponse` serialization. Nothing covers the `PermanentFail` vs `TransientFail` classification, which *is* the change. Swapping the two branches at `admin.rs:886` (`permanent_from_error`) leaves every test green while inverting the semantics — and inverting it wrongly negative-caches blobs during a SEAL infra blip, the exact scenario the comment there warns about.

The SDK tests are good but only cover wire compat.

---

Inline: `failed` can currently exceed `total` (comment on the helper), the `newly_failed` parameter is dead at its only call site, one `SKILL.md` sentence, and a TS type nit.

ducnmm:
Addressed review on `3bf3e606`. `failed` is page ∩ cache; embed-down forces `truncated` for retry; classification test pins permanent vs transient; MCP prints `failed` and splits retry vs raise-limit. @harrymove-ctrl please re-review.

harrymove-ctrl (approved):
Re-reviewed `3bf3e606`. Both blockers are resolved and CI is green (24/24). Approving.

### 1. `failed` bounded to the page — resolved

`restore_skip_fail_counts` (`admin.rs:558`) now intersects both counts with `on_chain_blob_ids`, and the dead `newly_failed` parameter is gone, so `let failed = failed + newly_failed;` (line 996) is the single place the two sources combine.

I re-derived the invariant against the exclusion set: `restored ⊆ missing_blob_ids = page \ (existing ∪ cache)`, `skipped = page ∩ existing`, `failed = (page ∩ cache) + newly_failed` with `newly_failed ⊆ missing_blob_ids`. Three disjoint subsets of the page, so `restored + skipped + failed <= total` holds and the negative-cache read still can't double-count. `restore_skip_fail_counts_intersects_failed_with_page` pins exactly that.

### 2. Classification is now covered — resolved

Extracting `restore_fail_class` as a pure helper is the right call. `restore_fail_class_pins_permanent_vs_transient` pins all four arms — InvalidUtf8, `InvalidCiphertext`, and the `NoAccessError` string as permanent, `TimeoutError` as transient. Swapping the branches in the `Decrypt` arm now fails CI, which was the thing I couldn't test before.

### 3. Embed failures — resolved as specified

`transient_unresolved` covers download transients and embed failures, and `restore_truncated_after_page` forces `truncated`. The original failure mode (embedder down, `source_capped=false`, response `{restored:0, skipped:0, failed:0, truncated:false}`) now returns `truncated=true`, and MCP prints "retry the same limit" rather than "increase limit". The `transientPage` guard in `formatRestoreResult` is correctly gated on `truncated`, so the WALM-431 cap path keeps its own hint.

All four inline comments are closed, and the docs/SDK surface is consistent — mocks, `manual.ts`, `memwal.ts`, `auth-required.ts`, Python `types.py`/`client.py` all carry `failed` with the omitted-field default.

---

### Follow-up (non-blocking, no action needed here)

The retry signal only fires when the page is *entirely* transient:

```rust
truncated || (restored == 0 && newly_failed == 0 && transient_unresolved > 0)
```

So a partial embed outage stays silent — 10 missing, 5 embedded, 5 embed-failed returns `{restored:5, skipped:0, failed:0, total:10, truncated:false}`. `restore_truncated_after_page_signals_retry_on_transients_only` pins the sharper case too: `assert!(!restore_truncated_after_page(false, 0, 1, 9))` — one permanent failure masks nine retryable blobs.

A caller can't recover that from the shortfall, because `restored + skipped + failed < total` is also permanently true once a blob expires (`cleanup_expired_blob` drops the local row while the blob stays in on-chain `total`), so retrying on shortfall would loop forever.

Comparing against the page length would close it:

```rust
truncated || (transient_unresolved > 0 && restored + newly_failed < inspected_page_len)
```

This is a stricter reading than the option I offered last round ("force `truncated=true` when an inspected page yields neither a success nor a permanent failure"), which you implemented as written — so it's a separate ticket, not a change request. Worth filing off WALM-480 so the partial case doesn't get lost now that the total-blackout case is fixed.

ducnmm:
Version dump: this PR also changes SDK / Python / MCP while `dev == main`, so it now bumps **TS 0.1.7**, **Python 0.1.10**, **MCP 0.0.13** (all manifests + both changelogs each + verify-script pins). Restore `failed` notes moved out of the already-released headings. OpenClaw unchanged. `node scripts/verify-manual-sdk-release.mjs` passes on `b27cfbb0`.

PR #848, opened 2026-08-31
docs: align the MCP docs and the setup skill with actual plugin behavior

harrymove-ctrl:
## Why

Support asked for install instructions we can hand to users, because the required user behaviour differs per client. Kenton then asked three clarifying questions about what the plugin guarantees, what plain MCP guarantees, and whether the setup skill still reflects either.

Answering those surfaced one underlying problem in two places: **what we tell users no longer matches how the product behaves.** This PR fixes both places, so the docs, the homepage setup command, and existing-user comms can all quote the same thing.

## What changed, and why it is one PR

The two surfaces carry the same message and would break in the same way if only one were updated:

- `docs/mcp/*` is what people read.
- `apps/app/public/skills/setup` is what an agent **executes**. It is served by `memory.walrus.xyz/skills/setup`, which is also the homepage setup command.

### 1. Plugin vs MCP-only positioning (`docs/mcp/overview.md`)

The overview claimed the proactive behavior "works on both installation paths". That overstates MCP-only and understates the plugin:

- **Plugin** is the tested path for reliable automatic memory, and it needs no extra instructions. Its `SessionStart` hook already injects the routing text, "prefer the `memwal_*` tools over any built-in or local memory feature", plus the save rubric ([`on_session_start.mjs`](https://github.com/MystenLabs/MemWal/blob/dev/packages/mcp/plugin/scripts/on_session_start.mjs)).
- **MCP-only** gives the tools on every client, but proactive behavior is best-effort: it varies by client and model, and a client's own built-in memory commonly wins.

### 2. Per-client install matrix (`docs/mcp/overview.md`)

Adds the one-command setup path and a matrix covering hooks support and the exact user action for Claude Code, Codex, Antigravity, Cursor, Claude Desktop, OpenCode, and both ChatGPT surfaces. States next to it that on plugin rows, the install command is the whole setup, so nobody stacks a redundant `CLAUDE.md` or `AGENTS.md` block on top.

### 3. ChatGPT means two different things (`docs/mcp/overview.md`)

The macOS **desktop app** ships the Codex CLI and reads the same `~/.codex`, so the Codex plugin install works there unchanged. **ChatGPT web Connectors** remain unsupported: a single bearer field cannot supply the required `x-memwal-account-id` header. Saying flatly "ChatGPT is not supported" sent desktop users away from a path that already works.

### 4. Claude Desktop instructions (`docs/mcp/claude-desktop.md`)

Adds an **Add memory instructions** block for users to paste, and states plainly that proactive behavior is best-effort there, because Claude Desktop cannot run hooks and its built-in memory can win.

### 5. The setup skill (`apps/app/public/skills/setup`)

The skill still described the world from before the hooks carried the routing. It said only Claude Code needed a plugin, steered Codex and Cursor users to hand-edited MCP config, never mentioned Antigravity, and required a `~/.claude/CLAUDE.md` block on every Claude Code install.

- **Plugin-first on every client that has one** (Agent Goal, rules 4 and 10), with MCP-only kept as an explicit fallback.
- **Codex** leads with `codex plugin marketplace add` / `codex plugin add` plus the `/hooks` trust step.
- **Antigravity** gets a section. It had zero mentions in the previous 412 lines.
- **Cursor** gets a verified install path (below).
- **`~/.claude/CLAUDE.md` demoted to a fallback**, for MCP-only clients and for when a user reports the agent still writing `MEMORY.md`.
- **ChatGPT scoped** to web connectors, with the desktop app routed to the Codex section.

## Verified on a real machine

The Cursor path was not written from documentation. Cursor loads plugins from `~/.cursor/plugins/local/<name>/`, so:

```bash
npx -y degit MystenLabs/MemWal/packages/mcp/plugin ~/.cursor/plugins/local/memwal --force
```

- The copy produces a valid `.cursor-plugin/plugin.json` declaring `hooks` and `mcpServers`.
- `CURSOR_PLUGIN_ROOT=<path> node scripts/on_session_start.mjs` returns the expected `hookSpecificOutput` carrying the routing text, exit 0.
- After restarting Cursor, it wrote `mcp-server-plugin-memwal-memwal.log` and logged `connection:connect_success`. The `plugin-` prefix is Cursor's own naming for a plugin-scoped server, so the plugin genuinely loaded.

Cursor ships no plugin CLI, and importing a third-party marketplace needs a team admin, so that copy is the install. The skill says both, so nobody hunts for a `cursor plugin` command.

That run also surfaced a failure mode worth documenting: the relayer refused the session with `MCP rate limit: ip_active_cap` (HTTP 429), then 503, while `/health` returned 200 throughout. The cause was a duplicate server, the plugin's own plus a leftover manual `memwal` entry in `~/.cursor/mcp.json`, stacked on stale `memwal-mcp` processes from earlier days. It reads like a broken install when it is not.

## Troubleshooting rows added to the skill

- Stale `~/.codex/.tmp/marketplaces/memwal-plugins/` clone, which makes `codex plugin marketplace add` fail while still exiting 0, so scripted installs do not notice.
- Cursor not showing the plugin, verified from the log files rather than a CLI, including how a sibling `mcp-server-user-memwal.log` reveals a duplicate entry.
- `ip_active_cap` 429, telling the agent to remove the duplicate, then list leftover processes and ask the user before ending any, since they can belong to other open clients.
- The `MEMORY.md` row rewritten to check the plugin before reaching for `CLAUDE.md`.

## Test plan

- [x] `node scripts/check-docs-code-sync.mjs` passes
- [x] `node scripts/check-docs-freshness.mjs` passes
- [x] Style guide audit findings cleared: no em dashes in prose, "might" not "may", "through" not "via", second person not first
- [x] Cursor plugin install, hook execution, and plugin load after restart verified
- [ ] Mintlify preview renders both docs pages
- [ ] `curl -sL https://memory.walrus.xyz/skills/setup` serves the updated skill after deploy
- [ ] One end-to-end run of the skill against a fresh Codex install


jessiemongeon1:
## Style Guide Audit

ducnmm (approved):
Looks good. Re-read including the later setup-skill commits.

The overview now matches reality: plugin is the tested automatic path; MCP-only is best-effort. Distinguishing ChatGPT desktop (Codex plugin) from ChatGPT web Connectors (`x-memwal-account-id`) is the important user-facing fix. Setup skill no longer forces a `CLAUDE.md` routing block on plugin installs; Cursor degit + log verification is documented.

Nits, not blocking:
- Follow-up still worth a ticket if none exists: `codex plugin marketplace add` duplicate-source still exits 0 (you mentioned it).
- `ip_active_cap` troubleshooting is useful; keep asking before killing processes.

Approve.

Issue #373, opened 2026-07-08
Expose Similarity threshold / MaxDistance parameters to the MCP Recall tool


Kafiirr:
The memwal_recall MCP tool uses vector database queries but lacks a parameter for LLMs to define the relevance threshold ( maxDistance ). As a result, the agent is often fed weakly related memories (due to pgvector returning top-k regardless of distance), which can trigger LLM hallucinations.

• Suggested: Expose  maxDistance  in the  memwal_recall  tool schema, allowing agents to filter out low-confidence matches before context ingestion.

nikola0x0:
Resolved in https://github.com/MystenLabs/MemWal/pull/850.

PR #849, opened 2026-09-03
docs: clarify MCP score vs SDK distance polarity (WALM-458)

ducnmm:
## Ticket

WALM-458 — https://linear.app/mysten-labs/issue/WALM-458/dx-document-mcp-score-vs-sdk-distance-polarity

## What changed?

- SDK API reference: MCP `memwal_recall` prints `score = 1 − distance`; do not apply SDK `maxDistance` to those scores.
- MCP overview: same polarity in one sentence.
- Troubleshooting FAQ: score vs distance.

## Why is this needed?

Community builders treated MCP `score` (high = similar) as SDK `distance` (low = similar) and inverted relevance filters.

## Scope

Document the two polarities. No wire-format or ranking change.

### Out of scope

- MCP `maxDistance` param (WALM-457 / GH #373)
- OpenClaw negative % (WALM-441)

## How was this tested?

- [ ] Unit tests
- [ ] Integration tests
- [ ] End-to-end tests
- [ ] Manual testing
- [x] Not applicable

Commands: docs-only review of the three files vs `origin/dev`.

## How can the reviewer verify it?

1. `docs/sdk/api-reference.md` — `recall` says MCP score polarity is inverted vs `maxDistance`.
2. `docs/mcp/overview.md` — one sentence that MCP score is `1 − cosine distance`.
3. `docs/troubleshooting/overview.md` — "Score vs distance" FAQ.

## Risks and dependencies

None

## Author checklist

- [x] This pull request maps to one ticket and one logical outcome.
- [x] I reviewed the complete diff myself.
- [x] I removed unrelated, debug, and temporary changes.
- [x] I ran the relevant tests.
- [ ] CI is green.
- [x] The branch is up to date with its target branch.
- [x] I added or updated tests where appropriate.
- [x] I documented any important risk, dependency, rollout, or follow-up.
- [x] I provided clear verification steps.
- [x] The pull request is ready for review and is no longer a Draft.

PR #850, opened 2026-09-03
feat(mcp): expose maxDistance on memwal_recall (WALM-457)

ducnmm:
## Ticket

WALM-457 — https://linear.app/mysten-labs/issue/WALM-457/mcp-expose-maxdistance-on-memwal-recall
GH #373 — https://github.com/MystenLabs/MemWal/issues/373

## What changed?

- Optional `maxDistance` on MCP `memwal_recall` (sidecar + cold-start schema).
- Client-side cosine-distance filter (`distance < maxDistance`) after positional `recall(query, limit, namespace)`, before duplicate collapse. Does **not** bump the sidecar SDK pin (`0.0.3`).
- Result lines include both `score` (`1 − distance`) and `distance`.
- Empty copy distinguishes “all hits outside maxDistance” from decrypt failures.
- Docs + changelog under MCP `0.0.12` (no version bump; `dev` already ahead of `main`).

## Why is this needed?

Recall returned top-K with no relevance cutoff, so agents ingested weak matches. SDK already had `maxDistance`; MCP did not (GH #373).

## Scope

MCP `memwal_recall` cutoff + polarity in the tool output/docs.

### Out of scope

- Bumping `@mysten-incubation/memwal` in `services/server/scripts` (WALM-428)
- Server-side `/api/recall` filter
- OpenClaw relevance % (WALM-441)

## How was this tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] End-to-end tests
- [ ] Manual testing
- [ ] Not applicable

Commands:

```bash
cd services/server/scripts && node --test --import tsx './mcp/__tests__/recall-*.test.ts'
pnpm --filter @mysten-incubation/memwal-mcp test
node scripts/verify-manual-sdk-release.mjs
```

1. Sidecar `RECALL_INPUT` and `packages/mcp/src/auth-required.ts` both expose optional `maxDistance`.
2. Filter keeps `distance < maxDistance`; omit = no filter.
3. Result line looks like `[score=0.750 distance=0.250]`.
4. All hits outside cutoff → `All matching memories were outside maxDistance.` (not decrypt-failure copy).
5. `packages/mcp` version still `0.0.12`.

## Risks and dependencies

None. Sidecar stays on SDK `0.0.3`; filter is local.

## Author checklist

- [ ] CI is green.

### DIFF
--- a/docs/mcp/reference.md
+++ b/docs/mcp/reference.md
@@ -75,12 +75,14 @@ Search the user's Walrus Memory for facts relevant to a query. The agent calls t
 
 Each result line includes `score` and `distance`. `distance` is cosine distance (low = similar). Displayed `score` is `1 − cosine distance` (high = similar). Optional `maxDistance` is a cosine-distance cutoff: hits with `distance >= maxDistance` are dropped. Omit `maxDistance` to apply no cutoff.
 
+The cutoff is applied to the `limit` nearest matches, not before them, so a tight `maxDistance` returns fewer than `limit` results — that is the cutoff biting, not an empty namespace. Raise `limit` to widen the candidate pool the cutoff sees.
+
 | **Parameter** | **Type** | **Required** | **Description** |
 | --- | --- | --- | --- |
 | `query` | string | yes | Natural-language query to match against stored memories. |
 | `limit` | integer (1–100) | no | Max memories to return. Default `10`. |
 | `namespace` | string | no | Namespace bucket to search. |
-| `maxDistance` | number | no | Cosine-distance cutoff (low = similar). Hits with `distance >= maxDistance` are dropped. Omit for no cutoff. |
+| `maxDistance` | number (>= 0) | no | Cosine-distance cutoff (low = similar). Hits with `distance >= maxDistance` are dropped. Omit for no cutoff. |
 
 ### memwal_analyze
 
--- a/docs/mcp/reference.md
+++ b/docs/mcp/reference.md
@@ -73,11 +73,14 @@ Save several durable facts in one batched call. Prefer this over repeated `memwa
 
 Search the user's Walrus Memory for facts relevant to a query. The agent calls this **proactively** at the start of a task or when the user references past work, decisions, or preferences. Returns matches ranked by relevance.
 
+Each result line includes `score` and `distance`. `distance` is cosine distance (low = similar). Displayed `score` is `1 − cosine distance` (high = similar). Optional `maxDistance` is a cosine-distance cutoff: hits with `distance >= maxDistance` are dropped. Omit `maxDistance` to apply no cutoff.
+
 | **Parameter** | **Type** | **Required** | **Description** |
 | --- | --- | --- | --- |
 | `query` | string | yes | Natural-language query to match against stored memories. |
 | `limit` | integer (1–100) | no | Max memories to return. Default `10`. |
 | `namespace` | string | no | Namespace bucket to search. |
+| `maxDistance` | number | no | Cosine-distance cutoff (low = similar). Hits with `distance >= maxDistance` are dropped. Omit for no cutoff. |
 
 ### memwal_analyze
 
--- a/packages/mcp/src/auth-required.ts
+++ b/packages/mcp/src/auth-required.ts
@@ -98,6 +98,11 @@ function buildToolDefinitions(proactive: boolean) {
                 query: { type: "string", minLength: 1 },
                 limit: { type: "integer", minimum: 1, maximum: 100, default: 10 },
                 namespace: { type: "string" },
+                maxDistance: {
+                    type: "number",
+                    description:
+                        "Optional cosine-distance cutoff (low = similar; 0 = identical). Hits with distance >= maxDistance are dropped. Omit to apply no cutoff. Displayed score is 1 - distance (high = similar); do not treat score as the cutoff.",
+                },
             },
             required: ["query"],
             additionalProperties: false,
--- a/packages/mcp/src/auth-required.ts
+++ b/packages/mcp/src/auth-required.ts
@@ -100,6 +100,7 @@ function buildToolDefinitions(proactive: boolean) {
                 namespace: { type: "string" },
                 maxDistance: {
                     type: "number",
+                    minimum: 0,
                     description:
                         "Optional cosine-distance cutoff (low = similar; 0 = identical). Hits with distance >= maxDistance are dropped. Omit to apply no cutoff. Displayed score is 1 - distance (high = similar); do not treat score as the cutoff.",
                 },
--- a/services/server/scripts/mcp/__tests__/recall-max-distance.test.ts
+++ b/services/server/scripts/mcp/__tests__/recall-max-distance.test.ts
@@ -8,7 +8,7 @@
 import test from "node:test";
 import assert from "node:assert/strict";
 
-import { filterByMaxDistance, formatRecallLine } from "../tools/recall.js";
+import { emptyRecallText, filterByMaxDistance, formatRecallLine } from "../tools/recall.js";
 
 const row = (text: string, distance: number) => ({ text, distance });
 
@@ -45,6 +45,16 @@ test("keeps hits whose distance is below maxDistance", () => {
             { text: "ok", distance: 0.499 },
         ],
     );
+});
+
+test("all hits outside maxDistance does not use decrypt-failure wording", () => {
+    const results = [row("far", 0.9), row("farther", 1.1)];
+    const filtered = filterByMaxDistance(results, 0.5);
+    assert.equal(filtered.length, 0);
+    const text = emptyRecallText(results.length, 3);
+    assert.match(text, /outside maxDistance/);
+    assert.doesNotMatch(text, /decrypt/);
+    assert.doesNotMatch(text, /download/);
 });
 
 test("displayed score is 1 minus cosine distance", () => {
--- a/services/server/scripts/mcp/__tests__/recall-max-distance.test.ts
+++ b/services/server/scripts/mcp/__tests__/recall-max-distance.test.ts
@@ -21,40 +21,38 @@ test("omitting maxDistance leaves every hit in place", () => {
     assert.deepEqual(filterByMaxDistance(results, undefined), results);
 });
 
-test("drops hits whose distance is at or above maxDistance", () => {
+test("the cutoff is exclusive: keeps below maxDistance, drops at or above", () => {
     const kept = filterByMaxDistance(
-        [row("kept", 0.49), row("equal", 0.5), row("worse", 0.9)],
+        [row("best", 0.1), row("ok", 0.499), row("boundary", 0.5), row("worse", 0.9)],
         0.5,
     );
     assert.deepEqual(
         kept.map((m) => m.text),
-        ["kept"],
+        ["best", "ok"],
     );
 });
 
-test("keeps hits whose distance is below maxDistance", () => {
-    const kept = filterByMaxDistance([
-        row("best", 0.1),
-        row("ok", 0.499),
-        row("boundary", 0.5),
-    ], 0.5);
-    assert.deepEqual(
-        kept.map((m) => ({ text: m.text, distance: m.distance })),
-        [
-            { text: "best", distance: 0.1 },
-            { text: "ok", distance: 0.499 },
-        ],
-    );
-});
-
-test("all hits outside maxDistance does not use decrypt-failure wording", () => {
+test("all hits outside maxDistance is not reported as a decrypt failure", () => {
     const results = [row("far", 0.9), row("farther", 1.1)];
     const filtered = filterByMaxDistance(results, 0.5);
     assert.equal(filtered.length, 0);
-    const text = emptyRecallText(results.length, 3);
+    const text = emptyRecallText(results.length, 0);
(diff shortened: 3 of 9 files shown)


jessiemongeon1:
## Style Guide Audit

Audited **3** file(s) against the [Sui Documentation Style Guide](https://docs.sui.io/references/contribute/style-guide).

**3** violation(s) found. All must be fixed before merge.

### `docs/mcp/reference.md` (3 violation(s))

_3 violation(s) (1 regex, 2 claude)_

- **Line 78** — No em dashes in prose
  - Current: `—`
- **Line 78** — no em dashes in prose
  - Current: `returns fewer than `limit` results — that is the cutoff biting`
  - Fix: `returns fewer than `limit` results. That is the cutoff biting`
- **Line 85** — passive voice
  - Current: `Hits with `distance >= maxDistance` are dropped.`
  - Fix: `The tool drops hits with `distance >= maxDistance`.`

---

services/server/scripts/mcp/tools/recall.ts
```diff
@@ -60,6 +66,37 @@ export function collapseDuplicates<T extends { text: string }>(
...
+
+/**
+ * Empty-result copy after the maxDistance filter.
+ *
+ * Decrypted hits that all missed the cutoff are not a download/decrypt
+ * failure, even when `dropped_count` is also > 0.
+ */
+export function emptyRecallText(resultCount: number, dropped: number): string {
+    if (resultCount > 0) {
+        return "All matching memories were outside maxDistance.";
```
harrymove-ctrl: **Requested change — this message asserts more than the code knows.**

With `results.length = 2` (both outside the cutoff) and `dropped_count = 3`, the agent is told everything was outside `maxDistance`. But the 3 dropped hits failed to decrypt, so their distances were never computed — some may well have been *inside* the cutoff. The agent concludes "nothing relevant here" when in fact 3 matches were silently unavailable.

The non-empty path does report `dropped` (`recall.ts:181`), so the information is only lost in this one branch. The docstring says the conflation is deliberate, and I agree the decrypt-failure copy alone would be wrong here — but appending is strictly better than replacing:

```ts
export function emptyRecallText(resultCount: number, dropped: number): string {
    if (resultCount > 0) {
        const suffix =
            dropped > 0
                ? ` (${dropped} additional ${dropped === 1 ? "match" : "matches"} could not be decrypted and were not checked against the cutoff.)`
                : "";
        return `All matching memories were outside maxDistance.${suffix}`;
    }
    ...
}
```

The existing test at `recall-max-distance.test.ts:50` passes `dropped = 3` and asserts `doesNotMatch(/decrypt/)`, so it would need to flip to asserting both facts are present — which is the behaviour you actually want to pin.

services/server/scripts/mcp/tools/recall.ts
```diff
@@ -23,6 +23,12 @@ const RECALL_INPUT = {
         .describe(
             "Optional namespace bucket to search within. Defaults to the session's namespace."
         ),
+    maxDistance: z
+        .number()
```
harrymove-ctrl, line 27: **Suggestion — no lower bound.** `z.number()` accepts `0`, negatives, and `Infinity`. `maxDistance: 0` or `-1` drops every hit and returns "All matching memories were outside maxDistance." with no hint the input was nonsense.

The likeliest model error is reading this as a *similarity* threshold — the description guards against that verbally, but `.nonnegative()` makes the schema reject the worst version of that mistake instead of returning a plausible-looking empty result:

```ts
maxDistance: z
    .number()
    .nonnegative()
    .optional()
```

packages/mcp/src/auth-required.ts
```diff
@@ -98,6 +98,11 @@ function buildToolDefinitions(proactive: boolean) {
                 query: { type: "string", minLength: 1 },
                 limit: { type: "integer", minimum: 1, maximum: 100, default: 10 },
                 namespace: { type: "string" },
+                maxDistance: {
+                    type: "number",
```
harrymove-ctrl, line 102: Same as the zod side — worth adding `minimum: 0` here so the cold-start schema and the sidecar schema stay in lockstep:

```js
maxDistance: {
    type: "number",
    minimum: 0,
    description: "...",
},
```

`tool-definitions.test.mjs:47` already walks both lists, so asserting the bound there is a one-line addition.

docs/mcp/reference.md
```diff
@@ -73,11 +73,14 @@ Save several durable facts in one batched call. Prefer this over repeated `memwa
 
 Search the user's Walrus Memory for facts relevant to a query. The agent calls this **proactively** at the start of a task or when the user references past work, decisions, or preferences. Returns matches ranked by relevance.
 
+Each result line includes `score` and `distance`. `distance` is cosine distance (low = similar). Displayed `score` is `1 − cosine distance` (high = similar). Optional `maxDistance` is a cosine-distance cutoff: hits with `distance >= maxDistance` are dropped. Omit `maxDistance` to apply no cutoff.
```
harrymove-ctrl, line 76: **Suggestion — say that you can get back fewer than `limit`.**

The cutoff is applied client-side *after* the top-K retrieval, so `limit: 10, maxDistance: 0.3` can legitimately return 2 rows. Without a note, an agent that gets 2 results is likely to conclude the namespace is nearly empty rather than that its cutoff was tight.

One sentence covers it:

> The cutoff is applied to the `limit` nearest matches, so a tight `maxDistance` returns fewer than `limit` results. Raise `limit` to widen the candidate pool before the cutoff.

services/server/scripts/mcp/__tests__/recall-max-distance.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    assert.doesNotMatch(text, /download/);
+});
+
+test("displayed score is 1 minus cosine distance", () => {
+    const line = formatRecallLine(
+        { text: "shipped the composite ranker", distance: 0.25 },
+        0,
+    );
+    assert.match(line, /\[score=0\.750 distance=0\.250\]/);
+    assert.equal((1 - 0.25).toFixed(3), "0.750");
```
harrymove-ctrl: Nit: this asserts JS float arithmetic rather than anything in `recall.ts` — the `assert.match` on the line above already pins the behaviour. Safe to drop.

Also minor: "drops hits whose distance is at or above maxDistance" (`:24`) and "keeps hits whose distance is below maxDistance" (`:35`) both exercise the same `0.5` boundary from opposite sides. One of them would do.

services/server/scripts/mcp/tools/recall.ts
```diff
@@ -60,6 +66,37 @@ export function collapseDuplicates<T extends { text: string }>(
     return { unique, collapsed: results.length - unique.length };
 }
 
+/**
+ * Drop hits whose cosine distance is at or above `maxDistance`.
+ *
+ * Same polarity as the SDK: cosine distance is low-is-similar, keep
+ * `distance < maxDistance`. Omit `maxDistance` (or pass a non-number) to
+ * leave the list unchanged.
+ */
```
harrymove-ctrl, line 79: Nit: consider naming the ticket that makes this redundant, e.g. `TODO(WALM-428): delete once the sidecar SDK pin moves to 0.1.x, which filters on maxDistance itself.`

The test-file header explains *why* the filter lives here, but someone doing the SDK bump will be reading this file, not the test.

harrymove-ctrl (changes requested):
Solid, well-scoped change — one ticket, 136/10 lines, good docstrings. I checked out the branch, ran the tests (10/10 pass) and CI is green on all 18 checks. Requesting changes on two small things (**1** and **3**); the rest are non-blocking suggestions.

## Verified

- **Polarity matches the SDK exactly.** `packages/sdk/src/memwal.ts:607` does `memory.distance < options.maxDistance` — identical to the new `filterByMaxDistance`. Confirmed.
- **The client-side filter is genuinely necessary.** The pinned `@mysten-incubation/memwal@0.0.3` only exposes `recall(query, limit?, namespace?)`, and `maxDistance` appears nowhere in its `dist/`. Worth noting the workspace SDK's own `maxDistance` is *also* a client-side filter, so no server-side cutoff is being bypassed.
- **`RecallMemory.distance` is a required `number` in 0.0.3**, so `T extends { distance: number }` typechecks and there is no undefined-drops-everything hazard.
- **No downstream parser of `[score=...]`.** Only `services/server/benchmarks/core/client.py` touches score, and it builds its own from the API response — the output-format change is safe.
- **`emptyRecallText` is passed the pre-filter count**, so the "outside maxDistance" branch is correctly unreachable when `maxDistance` is omitted.

## Requested changes

**1. `emptyRecallText` overclaims when decrypt failures coexist** — see inline on `recall.ts:92`.

**3. `docs/mcp/overview.md` is now stale, and the branch is behind `dev`.** #849 landed on `dev` after this branch forked, adding:

> `memwal_recall` prints `score = 1 - cosine distance` (higher = more similar); the wire and SDK field is `distance`.

That last clause implies `distance` is *not* printed — which this PR changes. Please rebase and update that line (the author checklist has "branch is up to date with its target" ticked, but it no longer is). `git merge-tree` shows it merges cleanly, so it is just the doc sentence.

## Non-blocking

**2.** No lower bound on `maxDistance` — inline on `recall.ts:27` and `auth-required.ts:102`.
**4.** Docs don't say results can be fewer than `limit` — inline on `reference.md:76`.
**5.** Test nit — inline on `recall-max-distance.test.ts:66`.
**6.** `TODO(WALM-428)` marker — inline on `recall.ts:75`.

## Operational note (no action needed)

The npm MCP package now advertises `maxDistance` in its cold-start list. A client on the new package talking to an older deployed relayer will have the key stripped by zod's default object parsing — the agent believes a cutoff applied and gets an unfiltered list. Low risk since the sidecar ships server-side, but it is a silent failure mode rather than a loud one. Worth knowing when this rolls out.

harrymove-ctrl:
@ducnmm heads up — I pushed the review fixes directly to this branch rather than leaving you to apply them, and **rebased onto `dev` in the process**, so your local copy needs a `git fetch && git reset --hard origin/henrynguyen/walm-457-mcp-expose-maxdistance-on-memwal_recall` before you touch it again. Sorry for the rewrite; shout if you'd rather I had opened a follow-up PR instead.

Your two commits are unchanged — the rebase was needed to pick up #849, and everything else is one commit on top (`778771e2`).

**The one behavioural change:**

`emptyRecallText` previously returned only `"All matching memories were outside maxDistance."` whenever `results.length > 0`. But undecrypted matches never had a distance computed, so with `results = 2` (both outside the cutoff) and `dropped_count = 3`, that sentence claimed the 3 decrypt failures were also outside the cutoff — they may well have been inside it. It now reports both:

```
All matching memories were outside maxDistance. (3 further matches were never checked against the cutoff: they failed to download or decrypt.)
```

**The rest:**

- `.nonnegative()` on the zod schema and `minimum: 0` in the cold-start JSON Schema. `maxDistance: -1` used to drop every hit and read as an empty namespace.
- `reference.md`: noted that the cutoff runs on the `limit` nearest matches, so a tight cutoff returns fewer than `limit` rows — otherwise an agent reads a short list as an empty namespace.
- `overview.md`: refreshed the line from #849, which said "the wire and SDK field is `distance`" and predated this PR printing `distance` on the result line.
- `TODO(WALM-428)` on `filterByMaxDistance` so whoever bumps the SDK pin finds it.
- Tests: folded the two 0.5-boundary cases into one that pins the whole partition, dropped `assert.equal((1 - 0.25).toFixed(3), "0.750")` (it asserted JS float arithmetic, not the formatter), and added coverage for the cutoff/undecrypted overlap plus the singular wording.

Verified locally: sidecar recall tests 11/11, `packages/mcp` 61/61, `tsc --noEmit` clean on every touched file.

---

**Two design calls I left alone — your judgement, not review nits:**

1. **Printing both `score` and `distance` on every line.** The tool description now has to end with *"do not treat score as the cutoff"* — that warning only exists because the line carries two numbers meaning the same thing inversely. Printing `distance` alone would match the wire field, the SDK, and the cutoff parameter, and the warning could go. Against it: #849 just documented `score`, and it costs a re-doc. Fine either way, but worth deciding rather than defaulting.

2. **`wrapTool<{ query; limit; namespace?; maxDistance? }>` duplicates `RECALL_INPUT` by hand.** This PR had to edit both places; miss one and the handler silently never sees the new argument, with no typecheck failure to catch it. Inferring the type from the zod shape would close that — probably its own small PR.

Also worth knowing at rollout (no action here): the npm MCP package now advertises `maxDistance` in its cold-start list, so a client on the new package talking to an older deployed relayer has the key stripped by zod's default object parsing — the agent believes a cutoff applied and gets an unfiltered list. Low risk since the sidecar ships server-side, but it fails silently rather than loudly.

ducnmm:
Addressed the 3 Sep review on `f825ed93` (rebased on `dev`):

1. `emptyRecallText` now keeps the outside-`maxDistance` sentence and appends decrypt-drop count when `dropped > 0`.
3. Overview no longer implies `distance` is hidden; result lines print both `score` and `distance`.

Also took the non-blocking items: `maxDistance` `minimum: 0` / `.nonnegative()`, fewer-than-`limit` note, dropped the float-arithmetic assert, `TODO(WALM-428)` on the local filter.

Please re-review.

harrymove-ctrl (approved):
Approved

PR #852, opened 2026-09-03
docs(skill): install the plugin on every client that has one

harrymove-ctrl:
## Why

Follow-up to [#848](https://github.com/MystenLabs/MemWal/pull/848), which fixed the docs. This fixes the other surface: the setup skill at `apps/app/public/skills/setup`, served by `memory.walrus.xyz/skills/setup`, which is also what the homepage setup command runs. It matters more than the docs, because an agent *executes* it.

The skill still described the world from before the plugin hooks carried the memory routing themselves. It said only Claude Code needed a plugin, steered Codex and Cursor users to hand-edited MCP config, never mentioned Antigravity, and required a `~/.claude/CLAUDE.md` routing block on every Claude Code install.

## What

**Plugin-first on every client that has one** (Agent Goal, rules 4 and 10). Claude Code, Codex, Cursor, Antigravity, and the ChatGPT desktop app each get a one-command plugin install, with MCP-only kept as an explicit fallback.

**Codex** leads with `codex plugin marketplace add` / `codex plugin add`, plus the `/hooks` trust step, instead of only hand-editing `~/.codex/config.toml`.

**Antigravity** gets a section. It had zero mentions in the previous 412 lines, although the plugin and a docs page both exist.

**Cursor** gets a verified install path. Cursor loads plugins from `~/.cursor/plugins/local/<name>/`, so:

```bash
npx -y degit MystenLabs/MemWal/packages/mcp/plugin ~/.cursor/plugins/local/memwal --force
```

Cursor ships no plugin CLI, and importing a third-party marketplace needs a team admin, so the copy above is the install. The section says both, so nobody hunts for a `cursor plugin` command.

**`~/.claude/CLAUDE.md` demoted to a fallback** (rule 11 and the routing block itself). The plugin's `SessionStart` hook already injects "prefer the `memwal_*` tools over any built-in or local memory feature" plus the save rules on every session, so a plugin install needs no extra instructions. The block still applies for MCP-only clients, and when a user reports the agent still writing `MEMORY.md`.

**ChatGPT scoped** (rule 1 and the Remote MCP section). Unsupported now refers to chatgpt.com web connectors, which expose a single bearer field and cannot send `x-memwal-account-id`. The desktop app ships the Codex CLI and reads `~/.codex`, so it routes to the Codex section.

## Verified on a real machine

The Cursor path was not written from the docs. It was installed and observed:

- The degit copy produces a valid `.cursor-plugin/plugin.json` declaring `hooks: ./hooks/cursor-hooks.json` and `mcpServers: .cursor-mcp.json`.
- `CURSOR_PLUGIN_ROOT=<path> node scripts/on_session_start.mjs` returns the expected `hookSpecificOutput` carrying the routing text, exit 0.
- After restarting Cursor, it wrote `mcp-server-plugin-memwal-memwal.log` and logged `connection:connect_success`. The `plugin-` prefix is Cursor's own naming for a plugin-scoped server, so the plugin was genuinely loaded.

That run also surfaced a failure mode worth documenting. The relayer refused the session with `MCP rate limit: ip_active_cap` (HTTP 429), then 503, while `/health` returned 200 the whole time. The cause was a duplicate server, the plugin's own plus a leftover manual `memwal` entry in `~/.cursor/mcp.json`, stacked on top of stale `memwal-mcp` processes from earlier days. It reads like a broken install when it is not.

## Troubleshooting rows added

- Stale `~/.codex/.tmp/marketplaces/memwal-plugins/` clone, which makes `codex plugin marketplace add` fail while still exiting 0.
- Cursor not showing the plugin, verified from the log files rather than a CLI, including how a sibling `mcp-server-user-memwal.log` reveals a duplicate entry.
- `ip_active_cap` 429, with instructions to remove the duplicate and to list leftover processes and ask the user before ending any, since they can belong to other open clients.
- The `MEMORY.md` row rewritten to check the plugin before reaching for `CLAUDE.md`.

## Test plan

- [x] `node scripts/check-docs-code-sync.mjs` passes
- [x] `node scripts/check-docs-freshness.mjs` passes
- [x] Cursor plugin install, hook execution, and plugin load after restart verified
- [ ] `curl -sL https://memory.walrus.xyz/skills/setup` serves the updated text after deploy
- [ ] One end-to-end run of the skill against a fresh Codex install


harrymove-ctrl:
Folded into #848 instead. There is no CODEOWNERS in this repo, so `docs/` and `apps/app/` route to the same reviewers, and both changes carry the same message: making what we tell users match how the plugin actually behaves. Splitting them would have meant two review cycles for one story, and would have left #848 answering "is this the only change needed?" with a no.

The two commits here were cherry-picked onto `docs/mcp-per-platform-install` unchanged (`c62e0d09`, `eed1cd6f`); the skill file is byte-identical to this branch. Closing.

PR #851, opened 2026-09-03
chore: promote dev to staging (WALM-460)

ducnmm:
## Ticket

WALM-460 — https://linear.app/mysten-labs/issue/WALM-460/promote-dev-to-staging-sep-2026

## What changed?

Promote `dev` → `staging` (56 commits as of 2026-09-03). Highlights already on `dev`:

- WALM-458 — MCP score vs SDK distance polarity docs
- WALM-453 — Python `remember_bulk_async` job_ids alignment
- WALM-423 — oversized recall embedding → HTTP 400
- WALM-413 — `OWNER_TOKEN_SECRET` minimum length
- WALM-383 / #810 — `sort=recent` + recall write-time
- MCP 0.0.12 — client identity on sessions
- COMG-726 / COMG-714 — security-sweep docs + SEAL CI invariant
- Relayer job-error sanitization; login-callback copy; system prompt templates

## Why is this needed?

Staging is behind a week of merged `dev` work. Henry asked to promote so Harry can review the train before it sits only on `dev`.

## Scope

One promotion PR. No new feature work in this PR.

### Out of scope

- V2 Oyster (WALM-459) / WALM-352 D11 design gap
- `staging` → `main`
- Open `dev` PRs not yet merged (WALM-457 #850, WALM-431 #823, …)

## How was this tested?

- [ ] Unit tests
- [ ] Integration tests
- [ ] End-to-end tests
- [ ] Manual testing
- [x] Not applicable

Commands: `git log origin/staging..origin/dev` (56 commits). Individual PRs already CI'd on `dev`.

## How can the reviewer verify it?

1. Diff base `staging` … head `dev`.
2. Confirm no unexpected commits (no V2 Oyster, no unmerged feature branches).
3. After merge: staging deploy health / MCP smoke (Harry).

## Risks and dependencies

Staging relayer/docs will pick up the above. MCP package on staging follows the 0.0.12 changelog on `dev`.

## Author checklist

- [x] This pull request maps to one ticket and one logical outcome.
- [x] I reviewed the complete diff myself.
- [x] I removed unrelated, debug, and temporary changes.
- [x] I ran the relevant tests.
- [ ] CI is green.
- [x] The branch is up to date with its target branch.
- [x] I added or updated tests where appropriate.
- [x] I documented any important risk, dependency, rollout, or follow-up.
- [x] I provided clear verification steps.
- [x] The pull request is ready for review and is no longer a Draft.

PR #855, opened 2026-09-03
fix(server): scope the embedding input guard to the API path

harrymove-ctrl:
Follow-up to [WALM-423](https://linear.app/mysten-labs/issue/WALM-423) (#837), found while reviewing the `dev` → `staging` promotion in [#851](https://github.com/MystenLabs/MemWal/pull/851). Two problems, both in `services/server/src/services/embedder.rs`.

## 1. The 16 KiB ceiling applied to the mock path

`reject_oversized_embed_input` ran before the `openai_api_key` match, so it gated the mock branch too. That branch hashes the input locally and has no context window at all.

Nothing shortens the text before it gets there: `MAX_REMEMBER_TEXT_BYTES` is 1 MiB, and `needs_summary` in `remember.rs:114` requires `openai_api_key.is_some()`. So on a deployment without an API key — local dev, CI, self-hosted — every `remember` between 16 KiB and 1 MiB started returning `BadRequest` for text that had always worked.

Fixed by moving the check into the `Some(api_key)` arm. The limit describes `EMBEDDING_MODEL`'s context window, so it belongs where a model is actually called.

## 2. Every upstream 400 was blamed on the caller

```rust
if status == reqwest::StatusCode::BAD_REQUEST {
    return Err(AppError::BadRequest(
        "embedding input exceeds the model context limit".into(),
    ));
}
```

An unknown model id, a malformed request body, or a provider-side auth shape we got wrong all come back as 400. All three reached the caller as "embedding input exceeds the model context limit", which tells a client to shorten a query that was never too long — and files a server misconfiguration under a 4xx that no dashboard alerts on. The real body only reached `tracing::warn!`.

Now gated on `is_context_length_error`, which reads the body for the shapes OpenAI-compatible providers actually emit (`context_length_exceeded`, "maximum context length", "reduce the length", "too many tokens", "string too long"). Anything else stays `AppError::Internal`.

## Not changed

`analyze` (64 KiB `MAX_ANALYZE_TEXT_BYTES`) and the admin restore path both embed unsummarized text and can exceed the model window. Both already degraded silently before WALM-423 — the provider returned 400, the old code mapped it to `Internal`, and each caller caught it and continued with no vector. That behaviour is unchanged here; it predates #837 and wants its own ticket.

## Testing

`cargo test --bins` (the invocation CI uses) — 672 passed, 0 failed. Two new tests pin both directions of the body check:

- `context_length_bodies_are_attributed_to_the_caller` — the five provider shapes that must still surface as 400.
- `other_400_bodies_are_not_blamed_on_the_caller` — unknown model, bad API key, unrecognized argument, empty body; all must stay `Internal`.

---

## Review findings status

From the two review passes over [#851](https://github.com/MystenLabs/MemWal/pull/851). This PR is scoped to `embedder.rs` only — it resolves 2 of them.

**Resolved here**

| Finding | Where |
| --- | --- |
| 16 KiB guard ran before the `openai_api_key` match, gating the mock path too | `embedder.rs:70` |
| Every upstream 400 mapped to "input exceeds the model context limit" | `embedder.rs:114` |

**Proposed in [#856](https://github.com/MystenLabs/MemWal/pull/856)** — draft. WALM-383 is In Review and owned by @nikola.le, so that PR is offered to them rather than queued to merge; these are not resolved yet.

| Finding |
| --- |
| `sort=recent` silently overridden when `scoring_weights` set `recency > 0` |
| `RecallResult.created_at` doc claimed the over-fetch mode was still unshipped |
| Changeset declared `patch` for a release adding three public API surfaces |

**Withdrawn after verification**

| Finding | Why |
| --- | --- |
| Short `OWNER_TOKEN_SECRET` panics at boot | Checked Railway: dev is 64 bytes, staging and production are UNSET (empty short-circuits before the length check). No environment can hit it. |
| `restore` silently drops oversized vectors | Pre-existing, not a WALM-423 regression — the provider already returned 400 and the old code mapped it to `Internal`, which the same handler caught and discarded. Behaviour identical before and after. |
| `analyze` degrades above 16 KiB | Same: pre-existing, unchanged by WALM-423. |

**Still open, filed separately** — none blocking the promotion

`candidate_limit` stops over-fetching at `limit >= 50` · `recall_manual` has no `sort` · MCP recall line renders a UTC date-only stamp · Python `remember_bulk` raise discards accepted `job_ids` · `sanitize_job_error_for_client` keys on `status == "failed"` so a `done` job is told to keep polling · `ConnectMcp` treats any `TypeError` as "link expired" · sidecar re-identifies the session from a client-controlled header on every POST · SEAL invariant CI gate only greps for a test name · live JS SDK e2e went 19 → 11 tests (`rememberBulkAndWait` / `analyzeAndWait` lost live coverage) · `wait-for-relayer` keeps the SHA wait when it cannot diff, guaranteeing a 480 s timeout · `test-sdk.yml` paths exclude `services/server/**`

### DIFF
--- a/services/server/src/services/embedder.rs
+++ b/services/server/src/services/embedder.rs
@@ -27,6 +27,12 @@ pub const EMBEDDING_MODEL: &str = "openai/text-embedding-3-small";
 pub const EMBEDDING_DIMS: usize = 1536;
 
 /// 16384 = 8192 tokens × 2 chars/token under cl100k; do not use admin 64KiB.
+///
+/// This is a property of [`EMBEDDING_MODEL`]'s context window, not of
+/// embedding in general — so it is enforced only on the API path. The mock
+/// fallback hashes the input locally and has no such ceiling; rejecting there
+/// would break key-less deployments that legitimately embed up to
+/// `MAX_REMEMBER_TEXT_BYTES`.
 const MAX_EMBED_INPUT_BYTES: usize = 16384;
 
 fn reject_oversized_embed_input(text: &str) -> Result<(), AppError> {
@@ -36,6 +42,23 @@ fn reject_oversized_embed_input(text: &str) -> Result<(), AppError> {
         )));
     }
     Ok(())
+}
+
+/// True when an upstream 400 blames the *input's length*.
+///
+/// OpenAI-compatible providers report that as `context_length_exceeded`, or
+/// in prose. Every other 400 — unknown model id, malformed body, a
+/// provider-side auth shape we got wrong — is our bug, not the caller's, and
+/// must stay [`AppError::Internal`]: returning 400 there tells a client to
+/// shorten a query that was never the problem, and hides a misconfiguration
+/// behind a 4xx that no dashboard alerts on.
+fn is_context_length_error(body: &str) -> bool {
+    let lower = body.to_ascii_lowercase();
+    lower.contains("context_length_exceeded")
+        || lower.contains("maximum context length")
+        || lower.contains("reduce the length")
+        || lower.contains("too many tokens")
+        || lower.contains("string too long")
 }
 
 #[async_trait]
@@ -67,9 +90,11 @@ impl OpenAiEmbedder {
 impl Embedder for OpenAiEmbedder {
     #[tracing::instrument(name = "embedder.embed", skip_all, fields(text_len = text.len()))]
     async fn embed(&self, text: &str) -> Result<Vec<f32>, AppError> {
-        reject_oversized_embed_input(text)?;
         match &self.config.openai_api_key {
             Some(api_key) => {
+                // Guard the model's context window, not `embed` as a whole:
+                // only this arm reaches a model. See `MAX_EMBED_INPUT_BYTES`.
+                reject_oversized_embed_input(text)?;
                 // Real embedding via OpenRouter/OpenAI-compatible API
                 let url = format!("{}/embeddings", self.config.openai_api_base);
 
@@ -111,8 +136,9 @@ impl Embedder for OpenAiEmbedder {
                             status, body
                         )));
                     }
-                    if status == reqwest::StatusCode::BAD_REQUEST {
-                        tracing::warn!(%status, body, "embedding API returned 400");
+                    if status == reqwest::StatusCode::BAD_REQUEST && is_context_length_error(&body)
+                    {
+                        tracing::warn!(%status, body, "embedding API rejected the input length");
                         return Err(AppError::BadRequest(
                             "embedding input exceeds the model context limit".into(),
                         ));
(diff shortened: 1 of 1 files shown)


services/server/src/services/embedder.rs
```diff
@@ -67,9 +90,11 @@ impl OpenAiEmbedder {
 impl Embedder for OpenAiEmbedder {
     #[tracing::instrument(name = "embedder.embed", skip_all, fields(text_len = text.len()))]
     async fn embed(&self, text: &str) -> Result<Vec<f32>, AppError> {
-        reject_oversized_embed_input(text)?;
         match &self.config.openai_api_key {
             Some(api_key) => {
+                // Guard the model's context window, not `embed` as a whole:
+                // only this arm reaches a model. See `MAX_EMBED_INPUT_BYTES`.
```
ducnmm: **[suggestion]** The headline regression — mock `embed()` must accept inputs up to `MAX_REMEMBER_TEXT_BYTES` (1 MiB), and only length-shaped upstream 400s may become `BadRequest` — is not locked at the call site. Existing tests still call `reject_oversized_embed_input` directly, and the new tests only exercise `is_context_length_error`. Moving the guard back above the `openai_api_key` match, or mapping every HTTP 400 to `BadRequest` again, would leave CI green. The mock path needs no HTTP mock: `OpenAiEmbedder` with `openai_api_key: None` plus a 16385-byte string is enough. The 400 mapping is harder without a fake upstream, but even a thin wrapper around the status/body branch would pin the intended `BadRequest` vs `Internal` split.

**Suggestion:** Add a tokio test that mock-embeds >16 KiB (and ideally 1 MiB) and asserts `Ok` with `EMBEDDING_DIMS`. If HTTP mocking is too heavy, extract the 400-classification arm so a unit test can assert length-shaped bodies → `BadRequest` and the existing non-length fixtures → `Internal`.

services/server/src/services/embedder.rs
```diff
@@ -38,6 +44,23 @@ fn reject_oversized_embed_input(text: &str) -> Result<(), AppError> {
...
+/// must stay [`AppError::Internal`]: returning 400 there tells a client to
+/// shorten a query that was never the problem, and hides a misconfiguration
+/// behind a 4xx that no dashboard alerts on.
+fn is_context_length_error(body: &str) -> bool {
+    let lower = body.to_ascii_lowercase();
+    lower.contains("context_length_exceeded")
+        || lower.contains("maximum context length")
+        || lower.contains("reduce the length")
+        || lower.contains("too many tokens")
+        || lower.contains("string too long")
```
ducnmm: **[suggestion]** `is_context_length_error` is an OR of five substrings and will both miss and over-match. It matches the prose `"string too long"` but not the OpenRouter/OpenAI code form `string_too_long` (the same list already special-cases `context_length_exceeded` with underscores). It also misses documented OpenAI-compatible phrasing such as `"the input length exceeds the context length"` and `"Input length X exceeds the maximum allowed input length"`, which contain neither `maximum context length` nor `context_length_exceeded`. Those inputs are exactly the ≤16 KiB but token-dense cases this backstop exists for; they would now surface as Internal 500 and invite retries against an unrecoverable limit. `"string too long"` is also the broadest needle — harmless on this request shape today (`model` is a constant, `input` is the only caller string) but it will classify any gateway 400 that happens to include that phrase as a caller-length error.

**Suggestion:** Prefer stable codes (`context_length_exceeded`, `string_too_long`, `max_tokens_exceeded`, `token_limit_exceeded`) and a tighter length phrase such as `"context length"` / `"input length exceeds"`. Drop or keep `"string too long"` only if you also match `string_too_long`. Add the FN bodies above to `context_length_bodies_are_attributed_to_the_caller`.

services/server/src/services/embedder.rs
```diff
@@ -67,9 +90,11 @@ impl OpenAiEmbedder {
 impl Embedder for OpenAiEmbedder {
     #[tracing::instrument(name = "embedder.embed", skip_all, fields(text_len = text.len()))]
     async fn embed(&self, text: &str) -> Result<Vec<f32>, AppError> {
-        reject_oversized_embed_input(text)?;
         match &self.config.openai_api_key {
             Some(api_key) => {
```
ducnmm: **[suggestion]** The inline comment restates what the next line and the `MAX_EMBED_INPUT_BYTES` docs already say (the guard is API-path-only because only this arm calls a model). It narrates the placement rather than adding a constraint that is not visible from the code.

**Suggestion:** Delete the two-line comment. The expanded `MAX_EMBED_INPUT_BYTES` docs already explain why the ceiling is not a property of `embed` as a whole.

harrymove-ctrl:
Closing — not abandoning the findings, just routing them properly.

Both fixes here sit on [WALM-423](https://linear.app/mysten-labs/issue/WALM-423), which is In Review and owned by @henry.nguyen. Opening a PR against someone else's in-review ticket was the wrong move on my part; the findings belong on the ticket, not in a competing branch.

The two issues, for the record:

1. `reject_oversized_embed_input` runs before the `openai_api_key` match, so the 16 KiB ceiling also gates the mock branch — which hashes locally and has no context window. Key-less deployments (local dev, CI, self-hosted) can't store 16 KiB–1 MiB text, since `needs_summary` in remember.rs:114 requires an API key.
2. Every upstream 400 maps to "embedding input exceeds the model context limit", so an unknown model id or bad API key reaches the caller as their input being too long.

Branch `fix/embedder-oversize-guard-placement` is left in place (1 file, 16/16 checks were green) if anyone wants the diff.

PR #856, opened 2026-09-03
fix(server,sdk): an explicit recall sort suppresses scoring_weights (WALM-383)

harrymove-ctrl:
> **Draft — needs @nikola.le's call before it merges.**
> [WALM-383](https://linear.app/mysten-labs/issue/WALM-383/memwal-recall-has-no-date-aware-ordering-or-write-time-field-newest) is In Review and owned by Nikola. This PR is offered as one way to close a gap found in it, not a claim on the ticket. Fold it into `nikolale/walm-383-recall-date-aware`, take the diff, or tell me to close it — all fine.

Found while reviewing the `dev` → `staging` promotion in [#851](https://github.com/MystenLabs/MemWal/pull/851). Not a promotion blocker.

## The gap

[#810](https://github.com/MystenLabs/MemWal/pull/810) is merged, and the design it implements is exactly what Henry approved on 2026-08-27. The gap is one level below that design.

In the WALM-383 thread Henry ruled out option **C** — *"both, with `scoring_weights` overriding `sort`"* — with *"Not B, not C. C adds a precedence rule the model will get wrong."* But at the server layer the merged code behaves like C anyway, because nothing enforces the choice:

1. `select_hits_for_sort` (`recall.rs:201`) orders candidates by write-time descending and truncates to `limit`.
2. `CompositeRanker::rank` (`recall.rs:262`) then reorders the survivors by composite score.

So `{sort: "recent", scoring_weights: {recency: 0.3}}` returns neither ordering — the newest row is no longer first. Henry restated the same precedence on [WALM-460](https://linear.app/mysten-labs/issue/WALM-460/promote-dev-to-staging-sep-2026#comment-4db9eaf8) yesterday:

> **`sort` wins.** If the caller sets `sort=recent` (or any explicit `sort`), that mode is the order. `scoring_weights` must not override it. Weights only apply when `sort` is omitted.

The SDK exposes both on `RecallOptions`, so the combination is one object literal away. No test covered it.

## What this does

`RecallRequest.sort` becomes `Option<RecallSort>`.

The rule distinguishes *omitted* from *explicit*, which a `#[serde(default)]` bare enum cannot express — `{}` and `{"sort":"relevance"}` both deserialize to `Relevance`. With `Option`, `effective_scoring_weights(sort, requested)` collapses the caller's weights to the default whenever `sort` is present.

Suppressing the weights rather than skipping the ranker keeps one code path: at `ScoringWeights::default()` the ranker short-circuits and returns its input order with `score: None`, which is exactly "no ranker ran".

Validation still runs on what the caller **sent**, so a malformed weight vector is a 400 whether or not `sort` would have discarded it.

## Two extras, same feature

- **`RecallResult.created_at` doc was stale.** It said the server-side over-fetching recency mode "is still open" — it shipped in #810 forty lines above. The field's own docs were pointing integrators at the client-side sort WALM-383 exists to replace.
- **Changeset was `patch`.** #810 adds three public API surfaces (`created_at`, `sort`, `scoringWeights`); bumped to `minor` so a consumer can depend on `sort`.

## Not changed

`recall_manual` — `RecallManualRequest` has no `sort`, so there is no precedence to resolve. Whether the manual path should get `sort` is a separate question, as is [WALM-428](https://linear.app/mysten-labs/issue/WALM-428/memwal-recall-expose-sortrecent-on-the-mcp-tool) (MCP knob).

## Proof the defect is real

`recall_sort_precedence_regression_tests` replays the handler's actual ordering pipeline — `select_hits_for_sort` → `CompositeRanker::rank` — over the two rows from the WALM-383 report, with a switch for whether precedence is applied. `apply_precedence: false` **is `dev` today**.

Inputs: `old-but-close` (distance 0.10, written 2026-06-06) and `newest-but-far` (distance 0.90, written 2026-07-06), `now = 2026-07-06`, weights `semantic 1.0 / recency 0.3 / half-life 30d`.

```
old-but-close    1.0*(1-0.10) + 0.3*2^(-30/30) = 0.90 + 0.15 = 1.05
newest-but-far   1.0*(1-0.90) + 0.3*2^(0/30)   = 0.10 + 0.30 = 0.40
```

| Test | Asserts |
| --- | --- |
| `selection_alone_puts_the_newest_row_first` | `select_hits_for_sort` does its job — newest first, before the ranker |
| `dev_today_lets_scoring_weights_undo_sort_recent` | **the defect**: the ranker reorders and `old-but-close` is returned first |
| `precedence_keeps_the_newest_row_first` | with the fix, `newest-but-far` is first |
| `omitted_sort_still_lets_weights_reorder` | with no `sort`, both paths produce identical output — weighted recall is not regressed |

Worth noting the weights above *favour* recency. The defect is not an artefact of hostile inputs: it survives weights a caller would pick specifically to reinforce `sort=recent`, because at a 30-day half-life the semantic gap (0.80) dwarfs the recency gap (0.15).

## Testing

`cargo test --bins` (CI's invocation): **678 passed, 0 failed**. Clippy: 0 errors. The SDK change is JSDoc only.

Four further unit tests pin the rule on `effective_scoring_weights` itself:

- `omitted_sort_lets_the_caller_weights_through`
- `explicit_recent_suppresses_the_weights`
- `explicit_relevance_also_suppresses_the_weights` — "any explicit sort", per the decision
- `omitted_sort_with_default_weights_is_still_the_plain_cosine_path`

### DIFF
--- a/.changeset/recall-write-time-and-recency-weights.md
+++ b/.changeset/recall-write-time-and-recency-weights.md
@@ -1,5 +1,5 @@
 ---
-"@mysten-incubation/memwal": patch
+"@mysten-incubation/memwal": minor
 ---
 
 Expose each memory's write-time on `recall()`, and let callers ask for a newest-first ordering (#621).
@@ -9,5 +9,7 @@ Expose each memory's write-time on `recall()`, and let callers ask for a newest-
 `created_at` is the piece a "newest wins" protocol was missing: previously the only date available was whatever the memory text happened to mention, so callers could neither order nor verify by write-time.
 
 `sort: "recent"` is the newest-wins mode: the relayer over-fetches semantic candidates (5x `limit`, capped at 50), orders them by write-time descending, then truncates to `limit`. `scoringWeights` does something narrower — it re-ranks the candidates the vector search already returned and does not widen the search, so a record that fell outside the cosine top-`limit` cannot be recovered by any weighting.
+
+An explicit `sort` suppresses `scoringWeights` for that request: the two are alternatives, not layers. Weights apply only when `sort` is omitted.
 
 Fully backward-compatible: omitting `sort` and `scoringWeights` leaves the request byte-identical to a plain cosine sort, and `created_at` is optional so relayers that don't send it behave exactly as before.
--- a/packages/sdk/src/types.ts
+++ b/packages/sdk/src/types.ts
@@ -165,6 +165,11 @@ export interface RecallOptions {
      *
      * For newest-wins, use `sort: "recent"` instead. It over-fetches
      * candidates server-side before ordering them by write-time.
+     *
+     * IGNORED when `sort` is set. An explicit `sort` — either value — decides
+     * the order outright; weights apply only when `sort` is omitted. Otherwise
+     * `sort: "recent"` would stop meaning newest-first as soon as the same
+     * request carried `recency > 0`.
      */
     scoringWeights?: ScoringWeights;
     /**
@@ -179,6 +184,8 @@ export interface RecallOptions {
      * widens the candidate set; weights only re-rank the set that was already
      * returned, so weights alone cannot surface a record that fell outside
      * the window.
+     *
+     * Setting `sort` at all suppresses `scoringWeights` for that request.
      */
     sort?: "relevance" | "recent";
 }
--- a/services/server/src/routes/mod.rs
+++ b/services/server/src/routes/mod.rs
@@ -236,6 +236,29 @@ pub(super) fn select_hits_for_sort(
     hits
 }
 
+/// Which scoring weights actually reach the ranker (WALM-383 precedence).
+///
+/// An explicit `sort` **is** the order: it suppresses the caller's weights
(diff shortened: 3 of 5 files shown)


harrymove-ctrl:
Closing — routing this to the ticket owner instead.

[WALM-383](https://linear.app/mysten-labs/issue/WALM-383) is In Review and owned by @nikola.le. A PR from outside on an in-review ticket is the wrong shape; this should have been a comment on WALM-383 from the start.

The defect, for the record: at the server layer `select_hits_for_sort` orders by write-time and truncates, then `CompositeRanker::rank` reorders the survivors by composite score — so `{sort: "recent", scoring_weights: {recency: 0.3}}` returns the stale row first. That is behaviour @henry.nguyen ruled out as option C on WALM-383 on 2026-08-27 ("Not B, not C").

Notably the weights that trigger it *favour* recency: at a 30-day half-life the semantic gap (0.80) dwarfs the recency gap (0.15), which is the same point Henry made in that thread.

Branch `fix/walm-383-sort-precedence` is left in place if it's useful — it carries a regression test that reproduces the ordering on current `dev`.

PR #860, opened 2026-09-04
docs(mcp): position the plugin as the automatic-memory path

harrymove-ctrl:
Docs-only. Cherry-picks the four documentation files from #848 onto `main` directly.

**This is the narrow alternative to #861** (`staging → main`, the full promotion). Merge one or the other, not both — #861 already contains these same four files. If the team wants the plugin-positioning docs live without shipping this cycle's server/SDK/MCP code, merge this and close #861. Otherwise close this one.

## What changed

The docs presented the plugin and the bare MCP server as interchangeable install options. They are not.

- **Plugin** ships lifecycle hooks that carry the memory instructions, so proactive save and recall works with no further setup from the user. Available on Claude Code, Codex, Antigravity and Cursor.
- **MCP-only** gives the agent the same tools, but proactive behaviour depends on the client and the system prompt. It varies by client and model, and on a client that ships its own memory feature the built-in one commonly wins. Positioned as compatible / tool access.

Three corrections fall out of that:

- **Cursor** was documented as MCP-only. It installs a plugin from a local directory, so its page is rewritten plugin-first.
- **Claude Desktop** has no hook mechanism, so it gets an explicit instruction block to paste.
- **ChatGPT** was being read as one product. The macOS desktop app ships Codex and takes the plugin; ChatGPT web Connectors cannot use it. These are now distinguished.

The setup skill served from the product page is updated to match — it installs the plugin on every client that has one, and verifies the Cursor plugin from logs.

## Scope

| File | |
|---|---|
| `docs/mcp/overview.md` | per-client install table, plugin-vs-MCP-only framing |
| `docs/mcp/cursor.md` | rewritten plugin-first |
| `docs/mcp/claude-desktop.md` | instruction block to paste |
| `apps/app/public/skills/setup` | setup skill matched to the above |

Deliberately excludes the other 10 docs files that differ between `main` and `staging`, because they document code that is not on `main`: `docs/sdk/api-reference.md` covers `sort=recent` and `created_at`, `docs/api/owner-token-auth.md` covers the 32-byte `OWNER_TOKEN_SECRET` rule, and the two changelog pages cover unreleased versions. Those ship with #861.

Verified before opening:

- No reference to anything unreleased on `main` — no `sort=recent`, no `created_at`, no `OWNER_TOKEN_SECRET` length rule, no version numbers.
- All seven internal doc links (`/mcp/antigravity`, `/mcp/claude-code`, `/mcp/claude-desktop`, `/mcp/codex`, `/mcp/cursor`, `/mcp/opencode`, `/mcp/reference`) resolve against `main`.
- No new pages, so `docs/docs.json` needs no nav entry.

## Note for the merger

`apps/app/public/skills/setup` sits under `apps/app/**`, the path filter on `deploy-app-walrus.yml`. Merging triggers a mainnet Walrus Site deploy. Intended — the setup skill is served from the product page — but worth knowing.

Unlike #861, this PR does **not** carry `.changeset/recall-write-time-and-recency-weights.md`, so it cannot trip the `release-sdk.yml` version-drift bug described there.

### DIFF
--- a/apps/app/public/skills/setup
+++ b/apps/app/public/skills/setup
@@ -15,8 +15,11 @@ the exact config and terminal commands to run.
 ## Agent Goal
 
 Connect Walrus Memory MCP to the user's AI client, sign the user in, and verify
-that memory tools work. On Claude Code, the plugin is required: MCP-only is an
-incomplete setup and auto-remember/recall will not be reliable.
+that memory tools work. Install the plugin wherever one exists: Claude Code,
+Codex, Cursor, Antigravity, and the ChatGPT desktop app. The plugin ships
+lifecycle hooks that make save and recall automatic with no extra instructions,
+so MCP-only is an incomplete setup on those clients. On clients with no plugin,
+MCP-only is the whole setup, and proactive behavior is best-effort.
 
 Do not install the Walrus Memory SDK or edit application source code unless the
 user explicitly asks for developer integration.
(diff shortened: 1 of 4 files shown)


jessiemongeon1:
## Style Guide Audit

PR #53, opened 2026-03-26
feat(noter): server-issued challenge nonce for wallet auth replay protection

Ashwin-3cS:
### Description
Closes https://github.com/MystenLabs/MemWal/issues/771
Resolves Linear WALM-414

This PR resolves the merge conflicts with `dev` on PR #53 and implements replay protection for `noter` wallet authentication using the repository's standard Redis-backed single-use challenge pattern (`enoki-challenge.ts`).

---

### Key Changes
1. **Server-Issued Single-Use Challenge (Redis)**:
   - Implements `issueWalletChallenge` (and backwards-compatible `getChallenge` alias) generating an unguessable 32-byte hex nonce stored in Redis with atomic `SET NX EX` (5 min TTL).
   - In `connectWallet`, consumes the challenge via atomic `GETDEL` (`verifyAndConsumeEnokiChallenge`), preventing replay attacks.
   - If Redis is unavailable or down, fails closed (500) to prevent bypass.

2. **Clean DB Schema & Monorepo Standard**:
   - Reverted custom PostgreSQL `wallet_challenges` table and migrations to eliminate DB bloat and schema drift, adhering to the codebase's existing Redis challenge architecture.

3. **Input Validation & Sanitization**:
   - Enforces `suiAddressSchema` (normalizing to 0x-prefixed 64-character hex) and strictly requires `challengeId`.
   - Replaces client-supplied arbitrary message payloads with server-issued nonce verification.
   - Safe user serialization via `toSafeUser()`.

4. **Testing**:
   - Added unit test suite `connect-wallet-replay.unit.test.ts` covering:
     - Replay attacks (second use of same challengeId fails with 400).
     - Redis downtime fail-closed behavior (500).
     - Address mismatch & invalid input formats (400).
   - All 15 monorepo CI checks passing green.

PR #868, opened 2026-09-07
fix(noter): require a server-issued single-use challenge for connectWallet (WALM-414)

harrymove-ctrl:
## Problem

`auth.connectWallet` verified a wallet signature over a message the **caller** supplied, and consumed nothing:

```ts
const { walletType, address, signature, message } = input;
const signerAddress = await verifyPersonalMessageSignature(
  new TextEncoder().encode(message),
  signature,
);
if (signerAddress.toSuiAddress() !== address) { /* 401 */ }
```

That proves only that the address holder signed *some* string at *some* point — no freshness, no server intent, no single-use property. The same `{message, signature, address}` triple verified an unbounded number of times, and every replay inserted a new `wallet_sessions` row with a fresh 24-hour expiry. `connectWallet` is a bare (unauthenticated) `procedure` on the publicly mounted auth router, so this was remotely reachable. Both the signature and the exact signed message were persisted in plaintext, so one DB read, log line, or network capture yielded a permanently reusable credential.

## Fix

`connectWallet` now uses the same ownership gate `connectEnoki` and `exportDelegateKey` already use:

- **`issueWalletChallenge`** issues a server-built, single-use `"signin"` challenge through the existing `enoki-challenge` module (Redis `SET NX EX` at issue, atomic `GETDEL` at verify).
- **`connectWallet`** calls `verifyAndConsumeEnokiChallenge` *before* any user upsert or session insert; a failed verify maps to `UNAUTHORIZED`, an unreachable challenge store to `SERVICE_UNAVAILABLE` (fail-closed, matching `connectEnoki`).
- **`connectWalletInput`** drops `message` and requires `challengeId`. Leaving the caller-chosen message accepted would have kept the vulnerable path alive. The address is tightened from a bare string to the canonical `0x` + 64-hex schema, promoted to a shared `suiAddressSchema` export, and normalized once so the challenge check and every DB op key on the same value.
- **`wallet_sessions.signedMessage`** records the consumed challenge id instead of a caller-supplied string.
- Removes the uncalled `generateAuthMessage()` (`Date.now() + Math.random()`) from `wallet-client.ts`.

Because the consume is an atomic `GETDEL` that runs *before* signature verification, there is no TOCTOU window: of any set of parallel replays, at most one caller can obtain the challenge. The address is bound three independent ways (`issued.address` equality, the `{ address }` option inside `verifyPersonalMessageSignature`, and `publicKey.toSuiAddress()` equality). Reusing the `"signin"` purpose preserves the existing guarantee that a sign-in signature can never authorize a delegate-key export.

## Testing

`apps/noter/package/feature/auth/api/connect-wallet-replay.unit.test.ts` drives the **real** router via `createCaller`, with the challenge layer mocked by a stub that reproduces the `GETDEL` single-use contract. It pins the reported symptom (a replayed call is `UNAUTHORIZED` and exactly one session row is written) plus the unknown-challenge, fail-closed, legacy-input, and malformed-address paths. The real store's single-use semantics remain covered by `lib/enoki-challenge.unit.test.ts`.

All three `noter-checks` gates pass:

```
Test Files  10 passed (10)
     Tests  54 passed (54)
tsc --noEmit  → exit 0, no output
next build    → Compiled successfully
```

Non-vacuity was verified by mutation: deleting the `if (!ownershipVerified) throw` gate fails exactly two tests (the replay assertion and the unknown-challenge assertion), then the suite returns green once restored.

## Reviewer notes

- **Breaking wire change.** Dropping `message` and requiring `challengeId` changes the contract of a public unauthenticated tRPC endpoint. `auth.connectWallet` has no caller anywhere in the repo and `generateAuthMessage` was never imported, so nothing in this monorepo breaks — an external consumer of the old shape would now get a 400.
- **Redis is now a fail-closed dependency for wallet sign-in.** A deployment without `REDIS_URL` returns `SERVICE_UNAVAILABLE`. This already applies to `connectEnoki`, so it is consistent rather than new.
- **Shared verifier and plain wallets.** `verifyAndConsumeEnokiChallenge` passes `{ address, client }` to `verifyPersonalMessageSignature`, which was written for zkLogin. That is safe for an ed25519 wallet signature: in `@mysten/sui` the `client` option is read only by `ZkLoginPublicIdentifier.fromBytes`, so the `ED25519` branch (`new Ed25519PublicKey(bytes)`) ignores it, while `address` is applied generically and supplies exactly the binding this path needs.
- **Follow-up worth filing:** `lib/wallet-verify.ts` still exports a dead `verifyWalletSignature(message, signature, address)` that accepts `address` and never uses it, returning `true` for any valid signature by any key. It has zero callers today, but it is the same vulnerability class and should be deleted.

---
Linear: WALM-414

### DIFF
--- a/apps/noter/doc/zklogin-integration.md
+++ b/apps/noter/doc/zklogin-integration.md
@@ -6,8 +6,10 @@ auth tRPC router (`package/feature/auth/api/route.ts`) and consumed by the `useA
 1. **Enoki zkLogin** (`connectEnoki`) — the primary sign-in. The Enoki flow runs client-side
    (`app/components/enoki-login-card.tsx` / `sui-providers.tsx`); on completion the app registers or
    looks up the user by Sui address and creates a session. This is the login the UI drives.
-2. **Sui wallet** (`connectWallet`) — sign-in with a Sui wallet (e.g. Slush). The server verifies the
-   personal-message signature before creating a session.
+2. **Sui wallet** (`connectWallet`) — sign-in with a Sui wallet (e.g. Slush). Two-step, like the Enoki
+   flow: the client calls `issueWalletChallenge` for a server-issued single-use message, signs it with
+   the wallet, and returns `{ challengeId, signature }`. The server verifies the signature against the
+   challenge and consumes it before creating a session, so a captured signature cannot be replayed.
 3. **Delegate key** (`connectDelegateKey`) — manual login with a delegate private key + account id.
 
 Sessions are stored in the `wallet_sessions` table (Enoki reuses it with `walletType = "enoki"`).
--- a/apps/noter/package/feature/auth/api/connect-wallet-replay.unit.test.ts
+++ b/apps/noter/package/feature/auth/api/connect-wallet-replay.unit.test.ts
@@ -1,0 +1,200 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// Replay-protection tests for the Sui-wallet sign-in path. These exercise the
+// REAL authRouter procedures via createCaller, asserting the route:
+//   - accepts only a server-issued challenge, so the caller can no longer pick
+//     the message it signs;
+//   - consumes that challenge, so replaying the same {challengeId, signature}
+//     is UNAUTHORIZED and mints exactly one 24-hour wallet_sessions row;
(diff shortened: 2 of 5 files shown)


harrymove-ctrl:
Superseded by #53 (conflicts resolved on original PR, CI all green).

PR #869, opened 2026-09-07
fix(mcp): honour Retry-After on a throttled SSE handshake (WALM-386)

harrymove-ctrl:
## What

`openSseStream` read the relayer's `retry-after` header and threw it away — it survived only as text interpolated into an `Error` message string, so nothing machine-readable reached the retry loops. Both `connectInBackground` and `reconnect` fell back to the generic geometric backoff, so the first retry after a 429 landed ~500ms later, well inside the window the relayer had just asked for. Against `ip_active_cap` — a *concurrent* cap that clears only when some other session closes, and which deliberately sends no `retry-after` because it has no honest ETA — a sub-second retry is pure noise against a condition it cannot affect.

This is the **client half only**. It makes the bridge stop making things worse and stop looking broken. It does not fix the outage; see *Not addressed* below.

## Changes

- The 429 branch throws a typed `RelayerThrottledError` carrying a resolved numeric `retryAfterMs`. `Retry-After` is parsed defensively (delta-seconds or HTTP-date; unparseable values yield `null`, never a `NaN` that would poison the backoff arithmetic), clamped to `MAX_THROTTLE_WAIT_MS` (60s) so a misconfigured or hostile header cannot park the bridge, and falls back to `DEFAULT_THROTTLE_FLOOR_MS` (5s, overridable via `MEMWAL_MCP_THROTTLE_FLOOR_MS`) when absent — the `ip_active_cap` shape.
- Both retry loops floor their backoff at the recorded throttle deadline, held in `throttledUntilMs` on the `runBridge` scope so it survives the separate `reconnect()` calls that would otherwise each restart from 500ms. Cleared once a session opens. The `immediate` reconnect path (login credential swap) still bypasses everything, so re-login gains no multi-second stall.
- One `note()` per throttle episode says this is a rate limit rather than a bad config or bad credentials, with the next attempt's ETA. Previously the 429 went only to a JSON stderr log line, so a throttled bridge was indistinguishable from a broken one — the reporter's headline symptom.
- The non-OK handshake exits keep **draining** the body and still do **not** abort the socket. 45b0ad87 removed those aborts on purpose because that is the path that fired the Windows libuv assertion in this same ticket; the comment above them now records why. The only behavioural change there is a `.catch()` around the drain so a truncated error body cannot mask the real status with a read failure.

## Tests

`packages/mcp/test/sse-handshake-429.test.mjs` spawns the real bridge binary against a mock relayer that 429s the SSE GET and timestamps every attempt. It pins the header-honoured case, the header-less floor, no fatal exit, exactly one local `initialize` reply, recovery of a tool call buffered during the throttle, and the throttle-vs-misconfiguration wording.

Verified as a genuine guard, not a tautology: reverting `bridge.ts` to `origin/dev` fails both with `attempt 2 came 504ms after attempt 1` and `retry came after 503ms`. Full package suite: 63/63, `tsc` clean.

## Not addressed (server side, needs its own change)

Investigated and verified against source rather than guessed:

1. **The cap is keyed on IP only.** `acquire(ip: string)` (`services/server/scripts/mcp/rateLimit.ts:108`) is its sole signature, and it runs before `resolveAuth`, by design. The reporter's inference that it tracks the account or delegate key is **wrong** — but their observation is real.
2. **Why a second public IP hit the same cap — likely root cause.** `TRUSTED_PROXY_HOPS` defaults to `0` (`services/server/src/types.rs:599`), and at 0 `canonical_client_ip` returns the TCP peer and ignores XFF (`services/server/src/client_ip.rs:21-23`). Behind an ingress the peer *is* the ingress, collapsing every user onto one IP and degenerating `maxSessionsPerIp: 16` into a **global** 16-session cap. **Action: check the deployed `TRUSTED_PROXY_HOPS` for the prod relayer** — possibly a one-line env fix, not determinable from the repo.
3. **Phantom connections have a concrete mechanism.** The comment at `services/server/src/mcp_proxy.rs:341` promises a `read_timeout` per-chunk watchdog. That `read_timeout` does not exist; the only timeout in the file is `.timeout(86_400)` at :370. A wedged proxy→sidecar stream holds the sidecar response, and therefore the rate-limit slot, for up to a day.
4. **Nothing reaps a leaked slot.** `active` decrements only in `release()` (`rateLimit.ts:147-155`) — no TTL, no sweep. Any missed release is permanent until restart, which is the structural reason the condition "never clears".
5. **The Windows libuv assertion is not fixed here** and was not reproducible from the code. Needs a Windows repro.

## Caveats

- Two of the ticket's three headline symptoms no longer reproduce on `dev`: the fatal exit was fixed by 71ad573c and the misleading "30s" wording by 45b0ad87. Confirm the reporter's installed `@mysten-incubation/memwal-mcp` version.
- Accepted trade-off: a 429 now waits at least the 5s floor instead of retrying in 500ms, slowing recovery from a transient 429 that is not a concurrent cap. The floor applies to 429s only; every other failure keeps today's fast geometric backoff.

---
Linear: WALM-386 · Refs #624 (client half only; server-side findings in the *Not addressed* section)


harrymove-ctrl:
Superseded by #903, which carries this branch's commits unchanged plus the other half.

This PR and the other one both rewrote `packages/mcp/src/bridge.ts` and conflicted with each other, so merging either would have forced the other to rebase. They are also two halves of the same user-visible failure — a bridge that waits in silence and then names the wrong layer — so #903 reviews them together. All tests from both branches pass side by side there (90 passed, 0 failed).

Not deleting the branch, in case the history is wanted.

PR #876, opened 2026-09-07
fix(mcp): warn on unrecognised flags, document env presets, report relayer in health (WALM-390)

### LINKED ISSUE #630: Unrecognised flags silently ignored (e.g. --prod); server never reports which network it is on
### Summary

`--prod` appears in MemWal MCP configurations in circulation, including the one I was handed to work
from:

    "args": ["-y", "@mysten-incubation/memwal-mcp", "--prod", "--namespace", "ck2-unrouted"]

It is not a recognised flag. `--help` on v0.0.6 documents only `--relayer`, `--web-url`, `--label`,
`--namespace`/`--ns`, plus `login`, `--logout` and `--help`. The flag is accepted without warning and
has no effect.

### Why this matters

**Unknown flags fail silently.** Anyone believing `--prod` selects an environment is wrong, and
nothing tells them. The same applies to any typo — `--namespac work` would be swallowed whole and the
user would quietly write to the relayer default. A single "unrecognised option" warning on stderr
prevents the entire class.

**The active network is never reported.** Network selection happens through `--relayer`, defaulting to
`https://relayer.memory.walrus.xyz`, and the running server never states which relayer or network it
is using outside of `MEMWAL_MCP_DEBUG=1`. So the destination is production by default, chosen
implicitly, and unconfirmed at runtime.

That combination is uncomfortable given writes cost real WAL and gas, are permanent and public, and
cannot be retracted (see companion issue re: tombstones). A user who never formed an intention about
which network they were on makes an irreversible paid public write and finds out afterwards.

### Expected

- Warn on unrecognised CLI flags instead of ignoring them
- Report the active relayer and network on startup, not only under a debug env var
- Preferably an explicit `--network mainnet|testnet`, so the destination is stated rather than
  inherited from a URL default
- If `--prod` is circulating in official material, either implement it or correct the source

### Actual

Unknown flags are silently swallowed, and the active network is invisible without debug logging.

nikola0x0:
Closes [WALM-390](https://linear.app/mysten-labs/issue/WALM-390). Fixes #630.

`--prod` landing as a real flag left three gaps. This closes all three.

### 1. Unknown flags no longer vanish

`parseArgs`' default branch dropped anything unmatched (`// Unknown flag: ignore silently.`), so a typo'd `--namesapce work` wrote to the `default` namespace with nothing on stderr. It now collects what it didn't match, and `main()` names each one:

```
{"level":"warn","event":"cli.unrecognised_arg","arg":"--namesapce"}
[memwal-mcp] Unrecognised option `--namesapce` — ignored. Run `memwal-mcp --help` for the supported options.
```

Warns rather than exits — a flag from a newer config must not brick the server.

One decision worth a look: **an unknown flag consumes a following non-flag token as its presumed value.** Without it, `--namesapce work` warned twice, the second naming the user's data. It also keeps a mistyped secret (`--tokenn hunter2`) out of stderr and the structured logs; verified `hunter2` reaches neither.

### 2. Presets documented

`--help` gained a `Network presets` section listing all four with their resolved URLs — **rendered from `ENV_PRESETS`, not retyped**, so a new preset can't ship undocumented the way `--prod` did. `printHelp` was split into a testable `helpText()`; the test walks each preset through `parseArgs` *and* the help text, so they can't drift apart again.

Drive-by while in that block: help gave the `--label` default as `"Walrus Memory MCP"`, code uses `"MCP Client"` (`index.ts:158`). Corrected.

### 3. Health names the relayer

```
Walrus Memory is reachable. status=ok version=1.2.3 relayer=https://relayer-staging.memory.walrus.xyz
```

`relayerUrl` is threaded onto `MemWalSession` from the URL `resolveAuth` already receives. Deliberately **not** an SDK getter: `MemWal.serverUrl` is private, and `services/server` imports the *published* `@mysten-incubation/memwal`, so an SDK change wouldn't land until a release. Tool description updated so the agent knows it can use this to confirm which network it's on.

### Verification

- `packages/mcp`: 70/70
- `services/server/scripts` MCP suite: 80/80
- `tsc --noEmit` clean both sides
- Built binary smoke-tested for the warning and the `--help` output

New coverage: `packages/mcp/test/unknown-flags.test.mjs` (9 cases), `services/server/scripts/mcp/__tests__/health-relayer.test.ts` (2 cases).

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -28,8 +28,8 @@ interface ParsedArgs {
     webUrl?: string;
     label?: string;
     namespace?: string;
-    /** Args parseArgs did not recognise, in the order seen. WALM-390: these
-     *  used to be dropped on the floor; main() now names them on stderr. */
+    /** Args parseArgs did not recognise, in the order seen. For a flag
+     *  written `--key=value`, only `--key` is recorded — see parseArgs. */
     unknown: string[];
 }
 
@@ -41,6 +41,10 @@ const ENV_PRESETS: Record<string, { relayer: string; web: string }> = {
     staging: { relayer: "https://relayer-staging.memory.walrus.xyz", web: "https://staging.memory.walrus.xyz" },
     local: { relayer: "http://127.0.0.1:8000", web: "http://localhost:5173" },
 };
+
+/** Bare words that are commands rather than values. An unknown flag must not
+ *  swallow one as its argument. */
+const POSITIONALS = new Set(["login"]);
 
 export function parseArgs(argv: string[]): ParsedArgs {
     const out: ParsedArgs = { help: false, logout: false, forceLogin: false, unknown: [] };
@@ -92,19 +96,30 @@ export function parseArgs(argv: string[]): ParsedArgs {
                 else if (a?.startsWith("--label=")) out.label = a.split("=", 2)[1];
                 else if (a?.startsWith("--namespace=")) out.namespace = a.split("=", 2)[1];
                 else if (a?.startsWith("--ns=")) out.namespace = a.split("=", 2)[1];
-                // WALM-390: anything still unmatched is a typo, or a flag
-                // from a newer build. Record it. Swallowing it here is what
-                // let `--prod` look supported for months while doing nothing.
-                // Values of KNOWN value-taking flags never reach this branch
-                // — `next()` already consumed them.
+                // Anything still unmatched is a typo, or a flag from a newer
+                // build. Values of KNOWN value-taking flags never reach this
+                // branch — `next()` already consumed them.
                 else if (a !== undefined) {
-                    out.unknown.push(a);
-                    // An unknown flag may take a value too. Consume a
-                    // following non-flag token as that value so `--namesapce
-                    // work` warns once about `--namesapce` rather than twice,
-                    // the second naming the user's data. Keeps a mistyped
-                    // secret (`--tokenn hunter2`) out of stderr and the logs.
-                    if (a.startsWith("-") && argv[i + 1] !== undefined && !argv[i + 1].startsWith("-")) {
+                    // Record the key only. A mistyped value-taking flag written
+                    // `--tokenn=hunter2` would otherwise put the user's secret
+                    // on stderr, which is the one place this warning must not
+                    // put it.
+                    const eq = a.indexOf("=");
+                    out.unknown.push(eq === -1 ? a : a.slice(0, eq));
+                    // An unknown flag may take its value as the next token, so
+                    // consume one — `--namesapce work` should warn once about
+                    // `--namesapce`, not a second time naming the user's data.
+                    // POSITIONALS are exempt: they are commands, not values, and
+                    // swallowing one would turn `memwal-mcp --typo login` into a
+                    // run that never logs in.
+                    const value = argv[i + 1];
+                    if (
+                        a.startsWith("-") &&
+                        eq === -1 &&
+                        value !== undefined &&
+                        !value.startsWith("-") &&
+                        !POSITIONALS.has(value)
+                    ) {
                         i++;
                     }
                 }
@@ -117,9 +132,9 @@ export function parseArgs(argv: string[]): ParsedArgs {
 export async function main(argv: string[] = process.argv.slice(2)): Promise<void> {
     const args = parseArgs(argv);
 
-    // WALM-390: name what we could not parse. Runs before the --help branch
-    // so `memwal-mcp --typo --help` still calls the typo out. Warn, never
-    // exit: an unknown flag from a newer config must not brick the server.
+    // Runs before the --help branch so `memwal-mcp --typo --help` still calls
+    // the typo out. Warn, never exit: an unknown flag from a newer config must
+    // not brick the server.
     for (const flag of args.unknown) {
         log.warn("cli.unrecognised_arg", { arg: flag });
         note(
@@ -288,11 +303,10 @@ function printHelp(): void {
 }
 
 /** The `--help` body. Exported so tests can assert it stays in step with the
- *  flags parseArgs actually accepts — WALM-390 was reported precisely because
- *  it had drifted. */
+ *  flags parseArgs actually accepts. */
 export function helpText(): string {
     // Rendered from ENV_PRESETS rather than retyped, so a new preset cannot
-    // ship undocumented the way --prod did.
+    // ship undocumented.
     const presetLines = Object.entries(ENV_PRESETS).flatMap(([name, urls]) => [
         `  ${`--${name}`.padEnd(33)}relayer: ${urls.relayer}`,
         `  ${"".padEnd(33)}web:     ${urls.web}`,
@@ -334,7 +348,8 @@ export function helpText(): string {
         ...presetLines,
         "",
         "                                   An explicit --relayer or --web-url",
-        "                                   overrides the preset it follows.",
+        "                                   wins over a preset, whichever",
+        "                                   order they are written in.",
         "",
         "Environment (equivalent to options):",
         "  MEMWAL_SERVER_URL                same as --relayer",
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -28,6 +28,9 @@ interface ParsedArgs {
     webUrl?: string;
     label?: string;
     namespace?: string;
+    /** Args parseArgs did not recognise, in the order seen. WALM-390: these
+     *  used to be dropped on the floor; main() now names them on stderr. */
+    unknown: string[];
 }
 
 /** Per-environment URL shortcuts. `--dev`/`--staging`/`--local` set both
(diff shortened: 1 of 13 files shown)


services/server/scripts/mcp/tools/health.ts
```diff
@@ -30,11 +30,16 @@ export function registerHealthTool(
                     : extra.write_ready === true
                       ? " write_ready=true"
                       : "";
+            // WALM-390: name the relayer that actually answered. A config on
+            // the wrong network is otherwise invisible here.
+            const relayerNote = session.relayerUrl
```
ducnmm: **[bug]** `memwal_health` now appends `relayer=${session.relayerUrl}` and the tool description tells agents to use that field to confirm prod / staging / dev / local. `session.relayerUrl` is not the URL the MCP client connected to. It is copied from `resolveAuth(..., serverUrl)` (`auth.ts:162`), and `serverUrl` is `process.env.MEMWAL_RELAYER_URL ?? "http://localhost:3001"` in `sidecar/app.ts`. The Rust parent fills that env with `http://127.0.0.1:$PORT` when it is unset (`services/server/src/main.rs:708-713`); `runtime.env.example` does not override it. For any deployment on that default, health will print a loopback address on every network, which is exactly the “looks healthy on the wrong network” failure this ticket is fixing. The new test injects `https://relayer-staging.memory.walrus.xyz` into a mock session and never calls `resolveAuth`, so it cannot catch this. Hosted OAuth happens to require a public `MEMWAL_RELAYER_URL` (issuer), but that is an accidental overlap, not something this change guarantees, and self-host / local still report loopback.

**Suggestion:** Print a value that is actually a network identity. Either (a) have the relayer put its public origin on `GET /health` and echo that, (b) pass a dedicated public-base URL into the sidecar separately from the loopback SDK URL, or (c) have the stdio package annotate health with the `--relayer` / `MEMWAL_SERVER_URL` it used to connect. Add a test that builds a session through `resolveAuth(headers, "http://127.0.0.1:8000")` (or `mountMcpRoutes`’s default) and asserts the health text — today that would print the loopback URL.

packages/mcp/src/index.ts
```diff
@@ -89,7 +92,22 @@ function parseArgs(argv: string[]): ParsedArgs {
...
+                // Values of KNOWN value-taking flags never reach this branch
+                // — `next()` already consumed them.
+                else if (a !== undefined) {
+                    out.unknown.push(a);
+                    // An unknown flag may take a value too. Consume a
+                    // following non-flag token as that value so `--namesapce
+                    // work` warns once about `--namesapce` rather than twice,
+                    // the second naming the user's data. Keeps a mistyped
+                    // secret (`--tokenn hunter2`) out of stderr and the logs.
+                    if (a.startsWith("-") && argv[i + 1] !== undefined && !argv[i + 1].startsWith("-")) {
```
ducnmm: **[suggestion]** An unknown flag now consumes the next argv token when that token does not start with `-`. That is the right call for `--namesapce work` and `--tokenn hunter2`, but it is a behavior change for the only positional command this parser has. `memwal-mcp --typo login` previously still set `forceLogin`; it now records `--typo`, swallows `login`, and never enters the login path (while `memwal-mcp login --typo` still works). The same happens for `--flag=value` forms: `--tokenn=hunter2` is pushed to `unknown` in full and then written to stderr via `log.warn` / `note`, which is the leak the space-separated consume was meant to avoid.

**Suggestion:** Do not consume the next token when it is a known positional (`login`). When recording an unknown token that contains `=`, log only the key side (`split("=", 2)[0]`) so a mistyped `--tokenn=hunter2` does not land in stderr.

packages/mcp/src/index.ts
```diff
@@ -89,7 +92,22 @@ function parseArgs(argv: string[]): ParsedArgs {
                 else if (a?.startsWith("--label=")) out.label = a.split("=", 2)[1];
                 else if (a?.startsWith("--namespace=")) out.namespace = a.split("=", 2)[1];
                 else if (a?.startsWith("--ns=")) out.namespace = a.split("=", 2)[1];
-                // Unknown flag: ignore silently.
+                // WALM-390: anything still unmatched is a typo, or a flag
```
ducnmm: **[suggestion]** Several new comments restate the patch and embed WALM-390 history rather than a non-obvious constraint: the `ParsedArgs.unknown` JSDoc, this default-branch block, the `main()` warning loop, `helpText()`’s export JSDoc, the `MemWalSession.relayerUrl` JSDoc in `auth.ts`, and the comment above `relayerNote` in `health.ts`. The consume-next-token comment is the one that actually explains a why (do not log the user’s value); the rest narrates the ticket.

**Suggestion:** Keep a short why where the code is surprising (unknown flags must not exit; skip the following non-flag token so values stay off stderr). Drop the ticket-history / “used to be dropped on the floor” / “how the bug got reported” prose.

packages/mcp/src/index.ts
```diff
@@ -288,6 +330,12 @@ function printHelp(): void {
         '                                   relayer uses its "default".',
         "                                   Alias: --ns",
         "",
+        "Network presets (set --relayer and --web-url together):",
+        ...presetLines,
+        "",
+        "                                   An explicit --relayer or --web-url",
```
ducnmm, line 350: **[nit]** Help now says an explicit `--relayer` / `--web-url` “overrides the preset it follows.” Preset application uses `??=`, so an explicit flag also wins when it precedes the preset (`--relayer https://custom --prod` keeps the custom URL). The new wording implies order dependence that the parser does not have.

**Suggestion:** Say that an explicit `--relayer` or `--web-url` wins over a preset, regardless of order — matching the existing `ENV_PRESETS` comment.

ducnmm (dismissed):
## Summary

The CLI half of WALM-390 looks correct: `parseArgs` now records unknown tokens, swallows a following non-flag value so user data is not logged, `--help` is generated from `ENV_PRESETS` (so `--prod` cannot drift out of the help text again), and the default-label line is brought in line with the actual `"MCP Client"` fallback. Tests cover the interesting parser cases. The health change is the weak spot — it prints `session.relayerUrl`, which `resolveAuth` fills from the sidecar’s SDK `serverUrl` (`MEMWAL_RELAYER_URL`, default loopback), not from a public network identity, and the new test never goes through that wiring.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 1

jessiemongeon1:
## Style Guide Audit

services/server/scripts/mcp/tools/health.ts
nikola0x0: Fixed in 81e4391. Split the two meanings apart: `relayerUrl` stays the address the sidecar dials, and a new `publicRelayerUrl` carries an origin only when an operator supplied one. `main.rs` forwards `MEMWAL_RELAYER_URL` on as `MEMWAL_PUBLIC_RELAYER_URL` only when it was actually set, since it otherwise injects the loopback value itself — health prints no `relayer=` field at all when it is absent.

The stdio package covers everyone else: the bridge always knows the URL it dialled, which is exactly what `--prod` / `--relayer` / `MEMWAL_SERVER_URL` selected, so it annotates the health reply on the way through.

`health-relayer.test.ts` now builds its session through `resolveAuth(headers, "http://127.0.0.1:8000")` instead of injecting one. It asserts the absence of `relayer=` rather than the loopback URL — that case fails against the old code.

packages/mcp/src/index.ts
nikola0x0: Both fixed. `login` is exempt from the consume-next-token rule now (`POSITIONALS`), so `memwal-mcp --typo login` reaches the login path again. For the `=` form only the key side is recorded — `--tokenn=hunter2` reports `--tokenn` — and it no longer consumes a following token either, since its value is already attached.

Four cases added to `unknown-flags.test.mjs` covering both.

packages/mcp/src/index.ts
nikola0x0: Dropped at all six sites. The consume-next-token comment kept its why, and the `--prod`-shipped-undocumented reasoning stayed on `helpText` where it explains why the presets are rendered rather than retyped.

packages/mcp/src/index.ts
nikola0x0: Reworded to "wins over a preset, whichever order they are written in". Added a test asserting both orderings resolve to the explicit URL, and that the old phrasing is gone.

ducnmm (commented):
## Summary

Follow-up on REQUEST_CHANGES for `0bd5d8975b`. Head `624baf14` addresses the prior items: health `relayer=` is a network identity (or absent), unknown flags no longer swallow `login` or log `--key=value` secrets, help no longer implies preset order matters.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

harrymove-ctrl (approved):
Approving. Second round is resolved: all four threads are closed and @ducnmm's 08:52 follow-up reports 0 bugs / 0 suggestions / 0 nits, leaving only the earlier REQUEST_CHANGES standing. CI is green.

Checked the two changes that carried real risk:

- The `relayerUrl` / `publicRelayerUrl` split is the correct fix for the original defect. `main.rs` injecting loopback when the operator set nothing was exactly why the old field could not be a network identity, and printing no `relayer=` at all beats printing `127.0.0.1` on staging. Building the test session through `resolveAuth` rather than injecting one is what makes it a real regression test.
- `parseArgs` holds up on the cases that matter: `--tokenn=hunter2` records only `--tokenn`, so the secret reaches neither stderr nor the structured log; `--typo login` still reaches the login path via `POSITIONALS`; warn-not-exit keeps a newer config from bricking the server.

One observation, not blocking: `pendingHealthIds` is deleted on a result reply and on purge, but not on an error reply, so a failed `memwal_health` leaves an entry behind. That is the same lifecycle as the existing `pendingListIds` directly above it, so this PR follows the local convention rather than introducing anything — worth a separate cleanup for both if it ever matters.

Also note this PR changes `packages/mcp/src` with no changelog entry, while #793 and #806 are being asked to bump the package version. Those want reconciling as one release decision for the batch — see my comment on #793.

PR #877, opened 2026-09-08
fix(sdk): throw when positional recall gets a namespace string (WALM-161)

### LINKED ISSUE #293: bug(sdk): recall() positional overload accepts string as limitOrOptions without error
## Description

The `recall()` method has two overloads:

```ts
recall(params: RecallParams): Promise<RecallResult>;
recall(query: string, limitOrOptions?: number | RecallOptions, namespace?: string): Promise<RecallResult>;
```

The deprecated positional form accepts `number | RecallOptions` as the second argument, but TypeScript's overload resolution allows passing a plain `string` without a compile-time error in some configurations:

```ts
// Developer intends: recall("food allergies", { namespace: "profile" })
// Actually writes:
const result = await memwal.recall("food allergies", "profile");
// compiles without error — "profile" is silently accepted as limitOrOptions
//   and ignored at runtime. Namespace is not set.
```

This happens because the first overload (`RecallParams`) matches `string` as a valid `RecallParams` key in loose type contexts, and the fallback to the positional overload treats the second arg as `number | RecallOptions` — a string is neither but is not caught by the union type check in all tsconfig strictness levels.

## Expected behavior

Either:
1. A runtime warning when positional form is used (since it is already `@deprecated`)
2. Remove the positional overload in the next minor — the object form `recall({ query, limit, namespace })` is strictly superior
3. At minimum, add a runtime type check: `if (typeof limitOrOptions === "string") throw new TypeError(...)`

## Steps to reproduce

```ts

const client = MemWal.create({ key: "...", accountId: "..." });
// No compile error, no runtime error, silently wrong:
const result = await client.recall("test query", "my-namespace");
// result uses default namespace, not "my-namespace"
```

## Impact

Silent bugs in namespace isolation — queries go to the wrong namespace without any indication.

**Environment:** SDK v0.0.7, TypeScript 5.4, `strict: true` in tsconfig.

HoangDucBach:
## Ticket

WALM-161 — https://linear.app/mysten-labs/issue/WALM-161
GH #293 — https://github.com/MystenLabs/MemWal/issues/293

## What changed?

- Positional `recall(query, "namespace")` now throws `TypeError` instead of treating the string as `RecallOptions` and silently querying the default namespace.
- Shared `resolveRecallCall` / `resolveLimitOrOptions` used by `MemWal`, `MemWalMock`, and `MemWalManual.recallManual`.
- Deprecated overloads split into `recall(query, options)` and `recall(query, limit?, namespace?)` so a string second argument does not type-check.
- Valid calls unchanged: `recall({ query, namespace })`, `recall(query, limit, namespace)`, `recall(query, { namespace })`.

## Why is this needed?

`recall("food allergies", "profile")` compiled and ran against the default namespace with no error, so namespace isolation failed silently.

## Scope

TypeScript SDK `recall()` and `MemWalManual.recallManual()` argument parsing.

### Out of scope

- Removing positional `recall` (already deprecated; future major)
- Python SDK positional `recall(query, limit, namespace)`
- MCP / `withMemWal` (already pass a numeric limit)

## Test plan

- [x] `pnpm test` in `packages/sdk` (115 pass)
- [x] `recall("q", "profile")` throws `TypeError`
- [x] `recall("q", 5, "profile")`, `recall("q", { namespace })`, and object form still isolate namespaces

Fixes #293

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/sdk/api-reference.md
+++ b/docs/sdk/api-reference.md
@@ -112,7 +112,7 @@ Search for memories matching a natural language query, scoped to `owner + namesp
 
 - Preferred form: `recall({ query, limit?, topK?, namespace?, maxDistance?, sort?, scoringWeights? })`
 - `limit` defaults to `10`; `topK` is an alias and wins when both are set
-- Legacy positional forms still work: `recall(query)`, `recall(query, limit)`, `recall(query, limit, namespace)`, and `recall(query, options)`
+- Legacy positional forms still work: `recall(query)`, `recall(query, limit)`, `recall(query, limit, namespace)`, and `recall(query, options)`. `recall(query, namespace)` throws `TypeError` — namespace is not a valid second argument.
 - `maxDistance` filters weak matches client-side by dropping results where `distance >= maxDistance`
 - `sort` picks the ordering: `"relevance"` (default) or `"recent"` for newest-among-matches
 - `scoringWeights` blends recency and importance into the ranking (see [Ordering](#ordering) below)
--- a/docs/sdk/api-reference.md
+++ b/docs/sdk/api-reference.md
@@ -112,7 +112,7 @@ Search for memories matching a natural language query, scoped to `owner + namesp
 
 - Preferred form: `recall({ query, limit?, topK?, namespace?, maxDistance?, sort?, scoringWeights? })`
 - `limit` defaults to `10`; `topK` is an alias and wins when both are set
-- Legacy positional forms still work: `recall(query)`, `recall(query, limit)`, `recall(query, limit, namespace)`, and `recall(query, options)`. `recall(query, namespace)` throws `TypeError` — namespace is not a valid second argument.
+- Legacy positional forms still work: `recall(query)`, `recall(query, limit)`, `recall(query, limit, namespace)`, and `recall(query, options)`. `recall(query, namespace)` throws `TypeError`; namespace is not a valid second argument.
 - `maxDistance` filters weak matches client-side by dropping results where `distance >= maxDistance`
 - `sort` picks the ordering: `"relevance"` (default) or `"recent"` for newest-among-matches
 - `scoringWeights` blends recency and importance into the ranking (see [Ordering](#ordering) below)
@@ -347,6 +347,9 @@ Embed locally, SEAL encrypt locally, send encrypted payload + vector to relayer 
 ### `recallManual(query, limit?, namespace?): Promise<RecallManualResult>`
 
 Embed locally, search via relayer, download from Walrus, SEAL decrypt locally. Returns decrypted text results.
+
+- `recallManual(query, { limit?, namespace?, scoringWeights? })` is valid
+- `recallManual(query, namespace)` throws `TypeError`; namespace is not a valid second argument
 
 ### `restore(namespace, limit?): Promise<RestoreResult>`
 
--- a/packages/sdk/src/memwal.ts
+++ b/packages/sdk/src/memwal.ts
@@ -74,6 +74,7 @@ import {
     compatibilityErrorFromStatus,
 } from "./compatibility.js";
 import { applyTokenBudget, estimateTokens } from "./tokens.js";
+import { resolveRecallCall } from "./recall-args.js";
 
 // ============================================================
 // Ed25519 Signing (lazy-loaded)
@@ -591,6 +592,8 @@ export class MemWal {
      * - `recall(query, limit, namespace)`
      * - `recall(query, { limit, namespace, maxDistance, topK })`
      *
+     * `recall(query, namespace)` is invalid and throws `TypeError`.
+     *
      * `topK` and `limit` are aliases; if both are set, `topK` wins.
      *
      * @returns RecallResult with decrypted text results
@@ -637,9 +640,16 @@ export class MemWal {
      * `recall({ query, limit, namespace })`. Positional will be removed in a
      * future major version of the SDK.
      */
+    async recall(query: string, options: RecallOptions): Promise<RecallResult>;
+    /**
+     * @deprecated Positional `recall(query, limit, namespace)` is easy to
+     * misread as `recall(query, namespace)`. Prefer the object form
+     * `recall({ query, limit, namespace })`. Positional will be removed in a
+     * future major version of the SDK.
+     */
     async recall(
         query: string,
-        limitOrOptions?: number | RecallOptions,
+        limit?: number,
         namespace?: string,
     ): Promise<RecallResult>;
     async recall(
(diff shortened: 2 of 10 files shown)


jessiemongeon1:
## Style Guide Audit

packages/sdk/src/recall-args.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+    if (typeof limitOrOptions === "number") {
+        return { limit: limitOrOptions, namespace } as T;
+    }
+    if (typeof limitOrOptions === "object" && !Array.isArray(limitOrOptions)) {
+        return limitOrOptions;
+    }
+    throw new TypeError(
+        `${method}() second argument must be a number (limit) or an options object, not a string. ` +
+            `Namespace goes in ${method}({ query, namespace }) or as the third argument: ` +
```
ducnmm: **[bug]** The `TypeError` always interpolates `${method}({ query, namespace })` as the recovery form. That is valid for `recall()`, but `MemWalManual.recallManual` has no object-first overload — only `recallManual(query, options)` and `recallManual(query, limit?, namespace?)`. A caller who follows the hint (`recallManual({ query, namespace })`) then fails at `query.trim()` with `TypeError: query.trim is not a function`. The new `recallManual` test only asserts `/recallManual/` in the message, so this copy bug is uncaught.

**Suggestion:** Phrase the hint so it is true for both APIs, e.g. namespace belongs in an options object (`${method}(query, { namespace })`) or as the third argument (`${method}(query, limit, namespace)`). For `recall` you can still mention the preferred `recall({ query, namespace })` form separately. Assert the full recovery text in the `recallManual` test.

packages/sdk/src/recall-args.ts
```diff
@@ -0,0 +1,52 @@
+import type { RecallOptions, RecallParams } from "./types.js";
+
+/**
+ * Parse `limit` / options for deprecated positional recall-style calls.
+ * A string second argument is the `recall(query, namespace)` footgun (#293).
```
ducnmm: **[suggestion]** The helper JSDoc restates what the function does, then embeds ticket/design history (`footgun (#293)`). That does not explain a non-obvious constraint.

**Suggestion:** Drop the ticket and the narration. If a comment stays, keep one short WHY line (string second arg is namespace, not options).

packages/sdk/src/memwal.ts
```diff
@@ -591,6 +592,8 @@ export class MemWal {
      * - `recall(query, limit, namespace)`
      * - `recall(query, { limit, namespace, maxDistance, topK })`
      *
+     * `recall(query, namespace)` is invalid and throws `TypeError`.
```
ducnmm, line 614: **[suggestion]** The new sentence documenting that `recall(query, namespace)` throws sits in the large recall JSDoc immediately above `countTokens()`, so it is not attached to any `recall` overload. Hover/generated docs on `recall()` still only show the `@deprecated` overload notes, which do not mention the throw.

**Suggestion:** Move this block (or at least the throw note) onto the `recall` overloads, or move `countTokens` out from between the docs and the method.

docs/sdk/api-reference.md
```diff
@@ -112,7 +112,7 @@ Search for memories matching a natural language query, scoped to `owner + namesp
 
 - Preferred form: `recall({ query, limit?, topK?, namespace?, maxDistance?, sort?, scoringWeights? })`
 - `limit` defaults to `10`; `topK` is an alias and wins when both are set
-- Legacy positional forms still work: `recall(query)`, `recall(query, limit)`, `recall(query, limit, namespace)`, and `recall(query, options)`
+- Legacy positional forms still work: `recall(query)`, `recall(query, limit)`, `recall(query, limit, namespace)`, and `recall(query, options)`. `recall(query, namespace)` throws `TypeError` — namespace is not a valid second argument.
```
ducnmm: **[nit]** The new sentence uses an em dash (`—`) (`throws \`TypeError\` — namespace is not a valid second argument`). The Sui docs style guide used on this repo forbids em dashes in prose.

**Suggestion:** Use a period, comma, or parentheses, e.g. `throws TypeError; namespace is not a valid second argument`.

ducnmm (changes requested):
## Summary

The shared `resolveLimitOrOptions` / `resolveRecallCall` helper is the right fix for WALM-161 / #293: `recall(query, "namespace")` no longer silently searches the default namespace, valid object / `limit` / 3-arg forms stay intact, and `MemWal`, `MemWalMock`, and `MemWalManual.recallManual` now share one parser. Changelog placement is correct (under existing `## 0.1.6`, no version bump, no changeset). The throw path itself is sound; the main defect is that the new `TypeError` text tells `recallManual` callers to use a call shape that does not exist.

## Issue counts by severity

- bugs: 1
- suggestions: 3
- nits: 1

## Issues outside the diff

- **[suggestion]** docs/sdk/api-reference.md:347 — `MemWal.recall` now documents the `TypeError`, but `MemWalManual.recallManual` was given the same runtime check and still reads only as `recallManual(query, limit?, namespace?)` with no note that a string second argument throws, and no mention of the options-object form.
  - **Suggestion:** Mirror the recall note: options object is valid, `recallManual(query, namespace)` throws.

packages/sdk/src/recall-args.ts
HoangDucBach: fixed with sugg — recovery text is now method(query, { namespace }) / third arg; recall still mentions recall({ query, namespace }).

packages/sdk/src/recall-args.ts
HoangDucBach: fixed with sugg — dropped the ticket; comment is now that a string second arg is namespace, not options.

packages/sdk/src/memwal.ts
HoangDucBach: fixed with sugg — moved the recall docs (including the throw) onto the overloads.

docs/sdk/api-reference.md
HoangDucBach: fixed with sugg — period instead of em dash.

ducnmm (commented):
## Summary

`7f6323b7` fixes the blocking TypeError copy from `f21d6920`. Shared recovery text is now `${method}(query, { namespace })` or `${method}(query, limit, namespace)` — both exist on `MemWal.recall` / `MemWalMock.recall` and on `MemWalManual.recallManual`. `recallManual` callers are no longer sent to a non-existent object-first shape.

The rest of WALM-161 still holds: string second args throw instead of being treated as options; object / `limit` / 3-arg forms still parse.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

HoangDucBach:
> ## Summary
> `7f6323b7` fixes the blocking TypeError copy from `f21d6920`. Shared recovery text is now `${method}(query, { namespace })` or `${method}(query, limit, namespace)` — both exist on `MemWal.recall` / `MemWalMock.recall` and on `MemWalManual.recallManual`. `recallManual` callers are no longer sent to a non-existent object-first shape.
> 
> The rest of WALM-161 still holds: string second args throw instead of being treated as options; object / `limit` / 3-arg forms still parse.
> 
> ## Issue counts by severity
> * bugs: 0
> * suggestions: 0
> * nits: 0

Let approve to we can merge

ducnmm (approved):
LGTM, all checks green and previous review feedback fully addressed.

PR #878, opened 2026-09-08
fix(app): retry missing account object instead of fatal 404 (WALM-596)

### LINKED ISSUE #390: 404 error fetching account object during key registration — reproduced on Testnet, Mainnet, and Devnet
When completing wallet setup and clicking "Register key onchain & continue", the app fails with Unexpected status code: 404. This blocks users from completing setup entirely.

Steps to Reproduce:
1. Go to localhost:5173/
2. Generate a new key pair (private + public key shown)
3. Check "I've saved my private key... I understand Walrus Memory can't show it again"
4. Click "Register key onchain & continue"
5. Observe error banner: Unexpected status code: 404

Console Error:
Failed to fetch account object ID: SuiHTTPStatusError: Unexpected status code: 404 ()
  at async fetchObjectJson (suiClientCompat.ts:46:17)
  at async fetchAccountIdForOwner (suiClientCompat.ts:64:30)
  at Dashboard.tsx:324:31
Expected behavior:
Key registers onchain successfully and user proceeds past setup.

Actual behavior:
Registration fails with 404; user is completely blocked from completing setup.

Additional context — reproduced across all networks:
Tested on Testnet, Mainnet, and Devnet — same 404 error occurs on all three. This suggests the issue is not a network-specific RPC problem, but rather in the account-object-fetch logic itself (fetchAccountIdForOwner), likely querying an object ID that doesn't exist yet for a newly generated key.

Environment:
- Browser: Chrome
- Dev server: localhost:5173 (Vite)
- OS: Windows 10

Screenshots:

Image

Image
Image
Image

HoangDucBach:
## Ticket

WALM-596 — https://linear.app/mysten-labs/issue/WALM-596
GH #390 — https://github.com/MystenLabs/MemWal/issues/390

## What changed?

- `fetchObjectJson` / `fetchAccountIdForOwner` treat GetObject and dynamic-field 404 / NotExists as not readable yet (`null`) instead of throwing.
- After `create_account`, the setup wizard reads `AccountCreated` / object changes from `waitForTransaction`, then polls the registry until the account id appears.
- Pre-create lookup and dashboard account resolve poll a few times so a lagging RPC does not look like "no account".
- Unit tests cover nested JSON-RPC table id unwrap, 404 → null, retry-then-success, and the AccountCreated fallback.

## Why is this needed?

The onboarding block in #390 was a bad registry lookup: `getDynamicFieldObject` was called with `accounts.id = { id: tableId }` and returned `Invalid params`, so the UI treated an existing account as missing. That unwrap is already on `origin/dev`.

What was still missing: after a successful `create_account`, SetupWizard did a single registry read and did not take `account_id` from `AccountCreated`. A late GetObject could still surface as a hard 404. This PR is that remaining retry / event fallback, not a re-fix of the table-id bug.

## Scope

Dashboard web app account lookup (`suiClientCompat`, SetupWizard, Dashboard).

### Out of scope

- Relayer / SDK on-chain account verification
- Changing create_account Move logic

## Test plan

- [x] `pnpm test src/utils/suiClientCompat.test.ts` in `apps/app`
- [x] Wallet setup on testnet: generate key → Register key onchain & continue completes without a 404 banner
- [x] After create_account, a lagging GetObject is retried rather than shown as a fatal 404

### Manual test (testnet, `https://relayer.dev.memwal.ai`)

Existing-account path (wallet already had a `MemWalAccount`): **Register key onchain & continue** showed `account found! adding delegate key...`, then dashboard with 1 **Web App** / **This browser** key. No 404 banner.

New-account path (fresh MemWal owner `0x6f2b7d20ac8bba49e67a0a5ca4ffcdb64270a6af462eb440e401d427806c6a39`): two sponsored txs, no console errors, dashboard 1 key.

| | |
|---|---|
| `create_account` | [`4Wq2MR2kfXNFmLdarap4YZECmJH1MESB4ete2UyrUzUz`](https://suiscan.xyz/testnet/tx/4Wq2MR2kfXNFmLdarap4YZECmJH1MESB4ete2UyrUzUz) — `AccountCreated` |
| `add_delegate_key` | [`3AKBaoGFcAdgc74ikZChT5WV5X3xF4Eag5NaA74HkQ4q`](https://suiscan.xyz/testnet/tx/3AKBaoGFcAdgc74ikZChT5WV5X3xF4Eag5NaA74HkQ4q) — `DelegateKeyAdded` |
| account object | `0xd0c1ee22568bc7916a55dcf4f48ec858ed00b0938319f09f7805d517c44c2b38` |

RPC returned the object in time, so a delayed GetObject was not forced. Create-path still completed without a 404 / "object ID not found" banner.

Fixes #390

### DIFF
--- a/apps/app/src/pages/Dashboard.tsx
+++ b/apps/app/src/pages/Dashboard.tsx
@@ -28,7 +28,7 @@ import { SecretValueInput } from '../components/SecretValueInput'
 import { config } from '../config'
 import { getAnalyticsErrorType, trackEvent } from '../utils/analytics'
 import { apiGet } from '../utils/api'
-import { fetchAccountIdForOwner, fetchObjectJson, publicKeyToHex } from '../utils/suiClientCompat'
+import { fetchAccountIdForOwner, fetchObjectJson, isMissingObjectError, pollAccountIdForOwner, publicKeyToHex } from '../utils/suiClientCompat'
 
 function DelegateKeyCtaIcon(props: SVGProps<SVGSVGElement>) {
     return (
@@ -346,12 +346,16 @@ export default function Dashboard({
         setAccountLookupComplete(false)
         setLoadingAccount(true)
         try {
-            const accountId = await fetchAccountIdForOwner(suiClient, config.memwalRegistryId, address)
+            const accountId = await pollAccountIdForOwner(suiClient, config.memwalRegistryId, address, { attempts: 4 })
             if (accountId) {
                 setResolvedAccountObjectId(accountId)
             }
         } catch (err) {
-            console.error('Failed to fetch account object ID:', err)
+            if (isMissingObjectError(err)) {
+                console.warn('Account object not readable yet:', err)
+            } else {
+                console.error('Failed to fetch account object ID:', err)
+            }
             setResolvedAccountObjectId(null)
         } finally {
             setAccountLookupAddress(address)
--- a/apps/app/src/pages/Dashboard.tsx
+++ b/apps/app/src/pages/Dashboard.tsx
@@ -28,7 +28,7 @@ import { SecretValueInput } from '../components/SecretValueInput'
 import { config } from '../config'
 import { getAnalyticsErrorType, trackEvent } from '../utils/analytics'
 import { apiGet } from '../utils/api'
-import { fetchAccountIdForOwner, fetchObjectJson, isMissingObjectError, pollAccountIdForOwner, publicKeyToHex } from '../utils/suiClientCompat'
+import { fetchAccountIdForOwner, fetchObjectJson, pollAccountIdForOwner, publicKeyToHex } from '../utils/suiClientCompat'
 
 function DelegateKeyCtaIcon(props: SVGProps<SVGSVGElement>) {
     return (
@@ -351,11 +351,7 @@ export default function Dashboard({
                 setResolvedAccountObjectId(accountId)
             }
         } catch (err) {
-            if (isMissingObjectError(err)) {
-                console.warn('Account object not readable yet:', err)
-            } else {
-                console.error('Failed to fetch account object ID:', err)
-            }
+            console.error('Failed to fetch account object ID:', err)
             setResolvedAccountObjectId(null)
         } finally {
             setAccountLookupAddress(address)
--- a/apps/app/src/pages/SetupWizard.tsx
+++ b/apps/app/src/pages/SetupWizard.tsx
@@ -92,12 +92,10 @@ async function accountIdAfterCreate(
     try {
         const waited = await suiClient.waitForTransaction({
             digest,
+            include: { events: true, effects: true, objectTypes: true },
             options: { showEvents: true, showObjectChanges: true },
-        })
-        const fromTx = findCreatedAccountId(waited as {
-            objectChanges?: Array<Record<string, unknown>> | null
-            events?: Array<Record<string, unknown>> | null
-        })
+        } as Parameters<typeof suiClient.waitForTransaction>[0])
+        const fromTx = findCreatedAccountId(waited)
         if (fromTx) return fromTx
     } catch {
         await suiClient.waitForTransaction({ digest })
--- a/apps/app/src/pages/SetupWizard.tsx
+++ b/apps/app/src/pages/SetupWizard.tsx
@@ -22,7 +22,7 @@ import { Link, useNavigate } from 'react-router-dom'
 import { LogOut, Copy, TriangleAlert } from 'lucide-react'
 import { config } from '../config'
 import { getAnalyticsErrorType, trackEvent } from '../utils/analytics'
-import { fetchAccountIdForOwner, fetchObjectJson, publicKeyToHex } from '../utils/suiClientCompat'
+import { fetchObjectJson, findCreatedAccountId, pollAccountIdForOwner, publicKeyToHex } from '../utils/suiClientCompat'
 
 type Step = 'intro' | 'import-key' | 'generating' | 'show-key' | 'onchain' | 'done' | 'error'
 
@@ -77,11 +77,32 @@ function hexToBytes(hex: string): Uint8Array {
 
 async function getAccountObjectId(suiClient: ReturnType<typeof useSuiClient>, ownerAddress: string): Promise<string | null> {
     try {
-        return await fetchAccountIdForOwner(suiClient, config.memwalRegistryId, ownerAddress)
+        return await pollAccountIdForOwner(suiClient, config.memwalRegistryId, ownerAddress, { attempts: 3 })
     } catch (err) {
         console.warn('[getAccountObjectId] lookup failed, treating as no-account', err)
         return null
     }
+}
+
+async function accountIdAfterCreate(
+    suiClient: ReturnType<typeof useSuiClient>,
+    ownerAddress: string,
+    digest: string,
+): Promise<string | null> {
+    try {
+        const waited = await suiClient.waitForTransaction({
+            digest,
+            options: { showEvents: true, showObjectChanges: true },
+        })
+        const fromTx = findCreatedAccountId(waited as {
+            objectChanges?: Array<Record<string, unknown>> | null
+            events?: Array<Record<string, unknown>> | null
+        })
+        if (fromTx) return fromTx
+    } catch {
+        await suiClient.waitForTransaction({ digest })
+    }
+    return pollAccountIdForOwner(suiClient, config.memwalRegistryId, ownerAddress)
 }
 
 type AccountJson = { delegate_keys?: { public_key?: unknown }[] }
@@ -191,15 +212,10 @@ export default function SetupWizard() {
                 ],
             })
             const createResult = await signAndExecute({ transaction: tx })
-            await suiClient.waitForTransaction({ digest: createResult.digest })
-
-            // Resolve through AccountRegistry after finality. This works for
-            // both the app-wide gRPC client and the local JSON-RPC E2E client;
-            // getTransactionBlock() is JSON-RPC-only.
-            knownAccountId = await getAccountObjectId(suiClient, ownerAddress)
+            knownAccountId = await accountIdAfterCreate(suiClient, ownerAddress, createResult.digest)
 
             if (!knownAccountId) {
-                throw new Error('Account created but object ID was not found in the registry. Please try again.')
+                throw new Error('Account created but the object is not readable yet. Please try again.')
             }
 
             setTxStatus('adding delegate key...')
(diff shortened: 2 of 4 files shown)


apps/app/src/utils/suiClientCompat.ts
```diff
@@ -46,58 +46,138 @@ function unwrapJsonRpcFields(value: unknown): unknown {
...
 
+/** RPC 404 / NotExists — the object is not readable yet, not a fatal setup failure. */
+export function isMissingObjectError(error: unknown): boolean {
+    const status =
+        error && typeof error === 'object' && 'status' in error
+            ? Number((error as { status: unknown }).status)
+            : undefined
+    if (status === 404) return true
+    const message = error instanceof Error ? error.message : String(error)
+    return /unexpected status code:\s*404|status code:\s*404|notExists|not found/i.test(message)
```
ducnmm: **[suggestion]** `isMissingObjectError` treats any message matching `/not found/i` as a lagging object. That is broader than 404 / `notExists`. Public fullnodes now reject JSON-RPC with `Method not found. JSON-RPC on public fullnodes has been deprecated...`, which matches and is converted to `null` by `fetchObjectJson` / `fetchAccountIdForOwner`. Callers then poll as if the wallet had no account instead of surfacing a transport/config failure. The reported error (`Unexpected status code: 404 ()`) is already covered by the status check.

**Suggestion:** Drop the bare `not found` alternative. Match `status === 404`, `notExists`, `dynamicFieldNotFound`, `object not found`, and `unexpected status code: 404`. Add a unit test that `Method not found` is not a miss.

apps/app/src/pages/SetupWizard.tsx
```diff
@@ -77,13 +77,34 @@ function hexToBytes(hex: string): Uint8Array {
...
     }
 }
 
+async function accountIdAfterCreate(
+    suiClient: ReturnType<typeof useSuiClient>,
+    ownerAddress: string,
+    digest: string,
+): Promise<string | null> {
+    try {
+        const waited = await suiClient.waitForTransaction({
```
ducnmm, line 93: **[suggestion]** `accountIdAfterCreate` requests JSON-RPC `options: { showEvents, showObjectChanges }` and `findCreatedAccountId` reads `objectChanges` plus `events[].type` / `parsedJson.account_id`. The app’s default client is `SuiGrpcClient` when `VITE_SUI_GRPC_URL` is set. gRPC `waitForTransaction` takes `include`, not `options`; events use `eventType` + `bcs`, not `type` + `parsedJson`. On the shipped client the shortcut always returns null and the create path falls through to `pollAccountIdForOwner`. Poll after `waitForTransaction` is enough for the ticket — the new helper just does not run where the app actually runs.

**Suggestion:** Pass both dialects (`include` and `options`), and in `findCreatedAccountId` also accept gRPC `eventType` / BCS the way `packages/sdk/src/account.ts` and noter already do.

ducnmm (commented):
## Summary

This is a correct fix for WALM-596 / GH #390. GetObject and dynamic-field HTTP 404s are treated as unreadability (`null`) instead of a thrown fatal, the wizard and dashboard poll the registry, and after `create_account` the create digest is mined for `MemWalAccount` / `AccountCreated` before a longer poll. Two non-blocking gaps remain: `isMissingObjectError` matches any `"not found"` message, and the new AccountCreated shortcut is JSON-RPC-shaped while the app’s default client is gRPC.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

apps/app/src/utils/suiClientCompat.ts
HoangDucBach: fixed with sugg

apps/app/src/pages/SetupWizard.tsx
HoangDucBach: fixed with sugg

apps/app/src/utils/suiClientCompat.ts
```diff
@@ -46,58 +52,202 @@ function unwrapJsonRpcFields(value: unknown): unknown {
...
+export function isMissingObjectError(error: unknown): boolean {
+    const status =
+        error && typeof error === 'object' && 'status' in error
+            ? Number((error as { status: unknown }).status)
+            : undefined
+    if (status === 404) return true
+    const message = error instanceof Error ? error.message : String(error)
+    // Do not match a bare "not found" — that also hits JSON-RPC's
+    // "Method not found" deprecation error on public fullnodes.
+    return /unexpected status code:\s*404|status code:\s*404|notExists|dynamicFieldNotFound|object not found/i.test(message)
```
ducnmm: **[bug]** `isMissingObjectError` now requires the contiguous phrase `object not found` (plus HTTP 404 / `notExists` / `dynamicFieldNotFound`). That correctly rejects `Method not found. JSON-RPC on public fullnodes has been deprecated.`, but it no longer matches the error `@mysten/sui` gRPC actually throws for a missing object or dynamic field.

`SuiGrpcClient.getObject` / `getDynamicField` go through `BatchGetObjects`; a missing item becomes `new Error(object.result.error.message)` with no `status` field. Sui's `ObjectNotFoundError` formats that message as `Object {id} not found` (optional ` with version {n}`), so the ID sits between `Object` and `not found`. The regex therefore returns false, `fetchObjectJson` / `fetchAccountIdForOwner` rethrow instead of returning null, and `pollAccountIdForOwner` aborts on the first attempt instead of sleeping/retrying.

That is a regression from bfb74466, whose bare `not found` *would* have matched this gRPC string. It matters here: `App.tsx` uses `SuiGrpcClient` whenever `VITE_SUI_GRPC_URL` is set (testnet `.env.example` and the Docker build arg), and Dashboard/Setup now depend on poll-on-miss. `accountIdAfterCreate`'s tx-event shortcut can still save the create path when events/objectTypes are present, but the registry fallback and Dashboard lookup will surface the raw `Object … not found` instead of retrying. The new unit test only asserts the literal `'object not found'`, not the gRPC wording.

**Suggestion:** Match the interpolated gRPC/ObjectError wording without bringing back bare `not found`, e.g. `/object(?:\s+\S+)+\s+not found/i` (and keep the HTTP 404 / `notExists` / `dynamicFieldNotFound` arms). Add a test for `new Error('Object 0xabc… not found')` (and optionally `Object {id} with version 1 not found`) staying true while `Method not found…` stays false. Defense in depth: `pollAccountIdForOwner` should also treat a still-unrecognized miss as null/retry rather than letting the first throw eat the remaining attempts.

ducnmm (changes requested):
## Summary

The follow-up correctly addresses the gRPC AccountCreated gap: `findCreatedAccountId` now unwraps `{ Transaction | FailedTransaction }`, reads `eventType` + BCS, and falls back to `effects.changedObjects` + `objectTypes`, with tests for those shapes. Tightening `isMissingObjectError` does stop JSON-RPC `"Method not found"` from being treated as a miss, but the new `object not found` needle does not match the gRPC miss the app actually throws — `Object {id} not found` — so registry polling still cannot retry on the default gRPC client.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

apps/app/src/utils/suiClientCompat.ts
HoangDucBach: fixed with sugg — matcher now hits Object {id} not found / with version, and poll retries an unrecognized miss instead of aborting the first throw.

apps/app/src/utils/suiClientCompat.ts
```diff
@@ -46,58 +52,218 @@ function unwrapJsonRpcFields(value: unknown): unknown {
...
+    },
+): Promise<string | null> {
+    const attempts = options?.attempts ?? 6
+    const sleep = options?.sleep ?? ((ms) => new Promise((resolve) => setTimeout(resolve, ms)))
+    for (let i = 0; i < attempts; i++) {
+        try {
+            const accountId = await fetchAccountIdForOwner(suiClient, registryId, ownerAddress)
+            if (accountId) return accountId
+        } catch (error) {
+            // Known miss: treat as unreadability and keep polling.
```
ducnmm: **[suggestion]** The catch comment narrates the previous miss-variant abort and describes a path that cannot happen. `fetchAccountIdForOwner` already returns `null` for `isMissingObjectError`, so this `catch` only sees unrecognized throws.

**Suggestion:** Drop the comment, or keep a one-liner that unrecognized errors still spend the remaining attempts and rethrow on the last try.

apps/app/src/pages/Dashboard.tsx
```diff
@@ -346,12 +346,16 @@ export default function Dashboard({
...
         setLoadingAccount(true)
         try {
-            const accountId = await fetchAccountIdForOwner(suiClient, config.memwalRegistryId, address)
+            const accountId = await pollAccountIdForOwner(suiClient, config.memwalRegistryId, address, { attempts: 4 })
             if (accountId) {
                 setResolvedAccountObjectId(accountId)
             }
         } catch (err) {
-            console.error('Failed to fetch account object ID:', err)
+            if (isMissingObjectError(err)) {
```
ducnmm: **[nit]** `pollAccountIdForOwner` never throws a missing-object error (misses become `null` inside `fetchAccountIdForOwner`). The `isMissingObjectError` branch in this `catch` is dead.

**Suggestion:** Catch any throw as a lookup failure, or drop the `isMissingObjectError` split.

ducnmm (commented):
## Summary

The gRPC miss-matching bug is fixed. `isMissingObjectError` now treats `Object {id} not found` / `Object {id} with version n not found` as a registry miss and still rejects JSON-RPC `Method not found`. Unit tests cover those cases plus `fetchObjectJson` / `pollAccountIdForOwner` retries. The rest of the PR still looks correct.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

apps/app/src/utils/suiClientCompat.ts
HoangDucBach: fixed with sugg, one-liner now: unrecognized throws still spend remaining attempts, rethrow on the last try.

ducnmm (commented):
## Summary

Nits on `90068f98` are addressed: poll catch comment is one line, Dashboard miss-branch is gone (`pollAccountIdForOwner` never throws a miss). No remaining issues on this head.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

HoangDucBach:
> ## Summary
> Nits on `90068f98` are addressed: poll catch comment is one line, Dashboard miss-branch is gone (`pollAccountIdForOwner` never throws a miss). No remaining issues on this head.
> 
> ## Issue counts by severity
> * bugs: 0
> * suggestions: 0
> * nits: 0

Let approve PR to we can merge

PR #879, opened 2026-09-08
fix(oauth): don't burn the session when Sui is unavailable (WALM-605)

harrymove-ctrl:
Fixes WALM-605 — https://linear.app/mysten-labs/issue/WALM-605

Follow-up to WALM-429 / #811, which taught signed HTTP auth that a Sui `GetObject` 429 is a retryable 503 rather than a revoked key. The OAuth callback never learned the same lesson.

## Problem

`session_complete` in `services/server/src/routes/oauth.rs` did two things in the wrong order:

1. `consume_oauth_session(&session_id)` — an irreversible `UPDATE ... WHERE status = 'pending' RETURNING *` that spends the one-time authorization session.
2. `verify_delegate_key_onchain(...)`, whose every error — including a throttled fullnode — was mapped to `OAuthError::invalid_request("delegate key is not registered on-chain: {e}")`.

So a Sui 429 was indistinguishable from a genuinely unregistered delegate key, **and** the session was already gone by the time we found out. The consent UI showed "Something went wrong" with no retry, and the user had to restart `/oauth/authorize` from the beginning. The error mapping alone was not the bug; the ordering was.

## The fix

**Classification.** Added a small pure decision function over `Result<String, OnchainVerifyError>`:

```
Ok(owner)                    -> Verified    (definitive)
Err(e) if e.is_unavailable() -> Retryable   (RpcError | ScanCapExceeded)
Err(e)                       -> Unregistered(definitive)
```

It delegates to `OnchainVerifyError::is_unavailable()` — the exact primitive `auth::cache_reverify_action` uses — rather than re-deriving gRPC-code logic, so the OAuth callback and signed recall cannot drift apart again.

**Ordering.** The on-chain read now happens *before* the claim, and `DelegateVerifyOutcome::consumes_session()` decides whether to claim at all:

- `Verified` / `Unregistered` — definitive, so the session is spent exactly as before this change.
- `Retryable` — the row is left `pending`, so the user can press "connect" again on the same session.

Single-use is unaffected. `claim_oauth_session` is still the only gate on minting an authorization code, and it is still the atomic `WHERE status = 'pending'` update, so two concurrent completers still leave exactly one winner (the loser gets "session already used"). Everything above the claim was already gated on a fresh, session-bound wallet ownership proof, so moving two reads above it opens no new surface. The pre-read also filters on `status == "pending"` so an already-consumed session fails fast without an on-chain call.

**HTTP shape.** A retryable failure now returns RFC 6749 `temporarily_unavailable` with `503` and `Retry-After: 5` instead of a dead-end `400 invalid_request`. `OAuthError` gained an optional `retry_after_secs` that `IntoResponse` emits; `OAUTH_UPSTREAM_RETRY_AFTER_SECS = 5` mirrors signed auth's `AUTH_UPSTREAM_RETRY_AFTER_SECS`. `Retry-After` is already in the CORS `expose_headers` list, so the browser consent UI can read it.

**`session_cancel`** has the second `consume_oauth_session` call. It is *not* affected: cancelling consults nothing off-box, so there is no retryable failure that can strand the user — the denial is itself the definitive outcome and burning the session is the point. It was switched to the shared `claim_oauth_session` helper and the reasoning recorded in a comment.

## Tests

`services/server` has no dev-dependencies and no HTTP/gRPC mock harness, so these follow the established pattern (`auth.rs`'s `resolve_account_cache_hit_rpc_error_keeps_row_without_authenticating`, `storage/sui.rs`'s `get_object_status_classifies_by_grpc_code`) — pure unit tests over the extracted decision, in `#[cfg(test)] mod tests` in `routes/oauth.rs`:

- `rpc_error_during_verify_is_retryable_and_spares_the_session` — the reported case: `RpcError("gRPC GetObject failed: 429 Too Many Requests")` classifies as `Retryable` and `consumes_session()` is false.
- `key_not_found_during_verify_is_unregistered_and_consumes_the_session` — the contrast case: a genuine unregistered key is `Unregistered` and still spends the session.
- `scan_cap_exceeded_during_verify_is_retryable`, `definitive_verify_failures_are_unregistered_and_consume_the_session` (`NotFound` / `AccountDeactivated` / `WrongObjectType`), `verified_key_consumes_the_session` — the rest of the variant matrix.
- `classification_tracks_is_unavailable_exactly` — asserts `Retryable` is exactly `OnchainVerifyError::is_unavailable()` across all six variants, so a new variant added to the enum cannot silently pick the wrong side here.
- `retryable_verify_maps_to_a_503_the_client_can_retry` / `unregistered_verify_stays_a_definitive_400` — response-shape checks over `IntoResponse`: 503 + `Retry-After: 5` + `"error": "temporarily_unavailable"` vs 400 with no `Retry-After`.

### What these tests cannot cover

They do not exercise `session_complete` end to end. That handler needs a Postgres pool, a Sui gRPC/JSON-RPC client and a wallet-signature verifier, and the crate has no harness to fake any of them. Specifically **not** covered by automated tests:

- That the row really is still `pending` in Postgres after a 429 (asserted only via `consumes_session()` gating the single `claim_oauth_session` call site).
- The concurrent-completer race on the atomic claim.
- That a real Sui 429 surfaces as `RpcError` — that lives in `map_get_object_status`, already covered by `get_object_status_classifies_by_grpc_code` in `storage/sui.rs`.

The handler has exactly one `consumes_session()` call site and one `claim_oauth_session` call in the success path, which is what keeps the tested decision equivalent to the handler's behaviour.

## Verification

Run in `services/server`:

- `cargo fmt` — clean (an unrelated pre-existing rustfmt drift in `routes/memory_read.rs` was reverted rather than folded into this PR).
- `cargo clippy --all-targets` (the project's CI invocation) — compiles; **zero** warnings in `src/oauth.rs` or `src/routes/oauth.rs`. The crate's ~630 pre-existing warnings elsewhere are unchanged; CI marks this step `continue-on-error` for exactly that reason. With `-D warnings` the crate does not build on `dev` either, unrelated to this change.
- `cargo test` — `454 passed; 0 failed; 23 ignored` (lib) and `698 passed; 0 failed; 37 ignored` (bin, the target CI runs). The two live integration suites (`security_delete_limits_live`, `sui_client_live`) are `#[ignore]`d as usual. All 8 new tests pass.

## Not in scope

The consent UI (`apps/app/src/pages/ConnectClaude.tsx`) renders `error_description` under a generic "Something went wrong" with no retry button. The session now survives, so a page reload works — but a proper "retry" affordance keyed on the 503 would be a nicer follow-up.

### DIFF
--- a/apps/app/src/pages/ConnectClaude.tsx
+++ b/apps/app/src/pages/ConnectClaude.tsx
@@ -335,12 +335,22 @@ export default function ConnectClaude() {
                         tx.object('0x6'),
                     ],
                 })
+                // `account.move` abort codes: EDelegateKeyAlreadyExists = 0,
+                // ETooManyDelegateKeys = 2, ENotOwner = 4.
                 let result
+                let alreadyOnChain = false
                 try {
                     result = await signAndExecute({ transaction: tx })
                 } catch (txErr: unknown) {
                     const m = txErr instanceof Error ? txErr.message : String(txErr)
                     if (m.includes('abort code: 0') && m.includes('add_delegate_key')) {
+                        // This session's key is already registered, so a previous
+                        // attempt landed the tx and then failed further along —
+                        // WALM-605's Sui 429 during verify is exactly that. There
+                        // is nothing left to submit: go straight to /complete
+                        // rather than reporting the abort as an error.
+                        alreadyOnChain = true
+                    } else if (m.includes('abort code: 4') && m.includes('add_delegate_key')) {
                         setErrorMsg(
                             `This wallet (${currentAccount.address.slice(0, 10)}…${currentAccount.address.slice(-6)}) is not the owner of Walrus Memory account ${accountId.slice(0, 10)}…${accountId.slice(-6)}. ` +
                             `Switch to the wallet that created this account, or run /setup for a new one.`
@@ -348,19 +358,21 @@ export default function ConnectClaude() {
                         trackEvent('claude_connect_failed', { error_type: 'owner_mismatch' })
                         setStep('error')
                         return
-                    }
-                    if (m.includes('abort code: 2') && m.includes('add_delegate_key')) {
+                    } else if (m.includes('abort code: 2') && m.includes('add_delegate_key')) {
                         setErrorMsg(
                             `This account already has the maximum number of delegate keys (20). Go to /dashboard and revoke an unused key, then try again.`
                         )
                         trackEvent('claude_connect_failed', { error_type: 'max_delegate_keys' })
                         setStep('error')
                         return
+                    } else {
+                        throw txErr
                     }
-                    throw txErr
                 }
-                await suiClient.waitForTransaction({ digest: result.digest })
-                txDigest = result.digest
+                if (!alreadyOnChain && result) {
+                    await suiClient.waitForTransaction({ digest: result.digest })
+                    txDigest = result.digest
+                }
             }
 
             setStep('finishing')
--- a/apps/app/src/pages/ConnectClaude.tsx
+++ b/apps/app/src/pages/ConnectClaude.tsx
@@ -206,11 +206,21 @@ export default function ConnectClaude() {
     // request. Re-running `handleConnect` instead would re-submit
     // `add_delegate_key` for a key already on chain (abort code 0).
     //
-    // Mirrored into the existing sessionStorage breadcrumb because a ref does
-    // not survive a reload, and a reload is the obvious thing to try when the
-    // relayer says "retry". Without it the first-time path re-runs
-    // `handleConnect` and dead-ends on that abort.
+    // A ref does not survive a reload, and a reload is the obvious thing to try
+    // when the relayer says "retry"; without a durable copy the first-time path
+    // re-runs `handleConnect` and dead-ends on that abort. It gets its own key
+    // rather than joining `memwal_claude_connect`, which the mount effect
+    // rewrites to `{ session }` on every load and would drop the payload.
     const pendingComplete = useRef<CompletePayload | null>(null)
+
+    const forgetPendingComplete = useCallback(() => {
+        pendingComplete.current = null
+        try {
+            sessionStorage.removeItem(PENDING_COMPLETE_KEY)
+        } catch {
+            // Nothing to clean up if storage is unavailable.
+        }
+    }, [])
 
     const rememberPendingComplete = useCallback(
         (payload: CompletePayload) => {
@@ -235,7 +245,9 @@ export default function ConnectClaude() {
     }, [sessionId])
 
     const completeSession = useCallback(async (payload: CompletePayload) => {
-        rememberPendingComplete(payload)
+        // The ref covers an in-flight /complete; only a retryable failure earns
+        // a durable crumb, so a definitive 400 cannot leave one behind.
+        pendingComplete.current = payload
         setStep('finishing')
         try {
             const { redirect_url } = await fetchJson<{ redirect_url: string }>(
@@ -256,14 +268,20 @@ export default function ConnectClaude() {
             trackEvent('claude_connect_complete', { reused_delegate: payload.reused_delegate })
             window.location.replace(redirect_url)
         } catch (err) {
-            if (!isRetryableComplete(err)) throw err
+            if (!isRetryableComplete(err)) {
+                // Definitive: the session is spent. Clear the crumb so a reload
+                // lands on the error rather than looping on "Sui is busy".
+                forgetPendingComplete()
+                throw err
+            }
             // The session is still pending, so offer this one request again
             // rather than dropping into the terminal error state.
+            rememberPendingComplete(payload)
             setErrorMsg(err instanceof Error ? err.message : String(err))
             setStep('retry-complete')
             trackEvent('claude_connect_failed', { error_type: 'sui_unavailable' })
         }
-    }, [apiBase, sessionId, rememberPendingComplete])
+    }, [apiBase, sessionId, rememberPendingComplete, forgetPendingComplete])
 
     const handleRetryComplete = useCallback(async () => {
         const payload = readPendingComplete()
--- a/apps/app/src/pages/ConnectClaude.tsx
+++ b/apps/app/src/pages/ConnectClaude.tsx
@@ -53,6 +53,31 @@ type CompletePayload = {
     reused_delegate: boolean
 }
 
+/** Where the in-flight `POST /complete` payload survives a reload. */
+const PENDING_COMPLETE_KEY = 'memwal_claude_pending_complete'
+
+/**
+ * The saved `POST /complete` payload for this session, if any.
+ *
+ * Session-scoped on purpose: one authorization's wallet signature must never be
+ * replayed against another.
+ */
+function loadStoredComplete(sessionId: string): CompletePayload | null {
+    try {
+        const raw = sessionStorage.getItem(PENDING_COMPLETE_KEY)
+        if (!raw) return null
+        const saved = JSON.parse(raw) as { session?: string; payload?: CompletePayload }
+        if (saved.session !== sessionId || !saved.payload?.owner_signature) return null
+        return saved.payload
+    } catch {
+        return null
+    }
+}
+
+function hasStoredComplete(sessionId: string): boolean {
+    return loadStoredComplete(sessionId) !== null
+}
+
 type Step =
     | 'loading'
     | 'consent'
@@ -153,7 +178,10 @@ export default function ConnectClaude() {
             .then((view) => {
                 if (cancelled) return
                 setSession(view)
-                setStep('consent')
+                // A reload mid-recovery resumes at the retry step. Landing on
+                // consent would re-run handleConnect, and on the first-time path
+                // that re-submits add_delegate_key for a key already on chain.
+                setStep(hasStoredComplete(sessionId) ? 'retry-complete' : 'consent')
             })
             .catch((err) => {
                 if (cancelled) return
@@ -173,13 +201,41 @@ export default function ConnectClaude() {
         sessionStorage.setItem('memwal_claude_connect', JSON.stringify({ session: sessionId }))
     }, [sessionValid, sessionId])
 
+
     // Everything `POST /complete` needs, kept so a retry re-sends exactly this
     // request. Re-running `handleConnect` instead would re-submit
     // `add_delegate_key` for a key already on chain (abort code 0).
+    //
+    // Mirrored into the existing sessionStorage breadcrumb because a ref does
+    // not survive a reload, and a reload is the obvious thing to try when the
+    // relayer says "retry". Without it the first-time path re-runs
+    // `handleConnect` and dead-ends on that abort.
     const pendingComplete = useRef<CompletePayload | null>(null)
 
+    const rememberPendingComplete = useCallback(
+        (payload: CompletePayload) => {
+            pendingComplete.current = payload
+            try {
+                sessionStorage.setItem(
+                    PENDING_COMPLETE_KEY,
+                    JSON.stringify({ session: sessionId, payload }),
+                )
+            } catch {
+                // Private mode or a full quota: the in-page button still works.
+            }
+        },
+        [sessionId],
+    )
+
+    const readPendingComplete = useCallback((): CompletePayload | null => {
+        if (pendingComplete.current) return pendingComplete.current
+        const saved = loadStoredComplete(sessionId)
+        if (saved) pendingComplete.current = saved
+        return saved
+    }, [sessionId])
+
     const completeSession = useCallback(async (payload: CompletePayload) => {
-        pendingComplete.current = payload
+        rememberPendingComplete(payload)
         setStep('finishing')
         try {
             const { redirect_url } = await fetchJson<{ redirect_url: string }>(
@@ -195,6 +251,7 @@ export default function ConnectClaude() {
                 },
             )
             sessionStorage.removeItem('memwal_claude_connect')
+            sessionStorage.removeItem(PENDING_COMPLETE_KEY)
             setStep('redirecting')
             trackEvent('claude_connect_complete', { reused_delegate: payload.reused_delegate })
             window.location.replace(redirect_url)
@@ -206,10 +263,10 @@ export default function ConnectClaude() {
             setStep('retry-complete')
             trackEvent('claude_connect_failed', { error_type: 'sui_unavailable' })
         }
-    }, [apiBase, sessionId])
+    }, [apiBase, sessionId, rememberPendingComplete])
 
     const handleRetryComplete = useCallback(async () => {
-        const payload = pendingComplete.current
+        const payload = readPendingComplete()
         if (!payload) return
         try {
             await completeSession(payload)
@@ -218,8 +275,7 @@ export default function ConnectClaude() {
             setStep('error')
             trackEvent('claude_connect_failed', { error_type: getAnalyticsErrorType(err) })
         }
-    }, [completeSession])
-
+    }, [completeSession, readPendingComplete])
     const handleConnect = useCallback(async () => {
         if (!session) return
         if (!currentAccount) {
@@ -325,6 +381,7 @@ export default function ConnectClaude() {
     // Auto-proceed once the wallet popup resolves.
     useEffect(() => {
         if (!walletPickerOpen && currentAccount && step === 'consent') {
+            if (hasStoredComplete(sessionId)) return
             void handleConnect()
         }
         // eslint-disable-next-line react-hooks/exhaustive-deps
@@ -391,8 +448,8 @@ export default function ConnectClaude() {
                         <div className="setup-classic-intro">
                             <h2 className="setup-classic-title">Sui is busy — try again</h2>
                             <p className="setup-classic-description">
-                                Your delegate key is registered and this authorization is still valid. We just
-                                couldn't reach Sui to verify it. Nothing needs redoing — press the button to
+                                We couldn't reach Sui to verify your delegate key, so this authorization is
+                                still pending rather than spent. Nothing needs redoing — press the button to
                                 finish.
                             </p>
                             <p className="setup-classic-description" style={errorTextStyle}>{errorMsg}</p>
--- a/apps/app/src/pages/ConnectClaude.tsx
+++ b/apps/app/src/pages/ConnectClaude.tsx
@@ -142,8 +142,9 @@ async function fetchJson<T>(url: string, init?: RequestInit): Promise<T> {
 /**
  * The relayer could not reach Sui, so it left the authorization session
  * `pending` instead of burning it (WALM-605). Re-POSTing `/complete` alone is
- * the recovery — restarting `handleConnect` would re-submit `add_delegate_key`
- * for a key that is already on chain, which aborts with code 0.
+ * the recovery: the wallet signature and the on-chain registration from the
+ * first attempt are both still good, so restarting `handleConnect` would only
+ * re-ask the wallet and re-submit a transaction the chain has already taken.
  */
 function isRetryableComplete(err: unknown): boolean {
     const e = err as Partial<OAuthRequestError> | null
@@ -179,8 +180,8 @@ export default function ConnectClaude() {
                 if (cancelled) return
                 setSession(view)
                 // A reload mid-recovery resumes at the retry step. Landing on
-                // consent would re-run handleConnect, and on the first-time path
-                // that re-submits add_delegate_key for a key already on chain.
+                // consent would re-run handleConnect, which re-asks the wallet
+                // to sign and re-sends a transaction the chain already took.
                 setStep(hasStoredComplete(sessionId) ? 'retry-complete' : 'consent')
             })
             .catch((err) => {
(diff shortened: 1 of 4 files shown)


services/server/src/routes/oauth.rs
```diff
@@ -675,26 +735,54 @@ pub async fn session_complete(
...
+    };
+    let verified_owner = match outcome {
+        DelegateVerifyOutcome::Retryable { reason } => {
+            // Deliberately no session_id: possession of one plus a wallet
+            // signature mints a code, so it does not belong in logs.
+            tracing::warn!(
+                %reason,
+                "oauth session_complete: Sui unavailable, leaving session pending for retry"
+            );
+            return Err(OAuthError::temporarily_unavailable(format!(
```
ducnmm, line 838: **[suggestion]** The 503 asks the user to retry the same request, and `consumes_session` documents that they can hit "connect" again without restarting `/oauth/authorize`. That is true only if the client re-POSTs `/complete`. `ConnectClaude.tsx` does not: any `/complete` failure sets `step` to `error` (no Approve button), and a page reload re-runs the full `handleConnect` path, including `POST …/account`. `find_reusable_oauth_delegate` only returns `status = 'active'` rows, and `activate_oauth_delegate` runs *after* a successful verify, so a first-time connect that already landed `add_delegate_key` still gets `needs_onchain_registration: true`. Reloading then re-submits the same pending pubkey, which aborts `EDelegateKeyAlreadyExists` (Move code `0`). The SPA maps abort code `0` to "this wallet is not the owner" (`ENotOwner` is actually `4`), so the likely WALM-605 journey (write key → immediate GetObject 429) still dead-ends, just with a more misleading error. Reconnects that already have an active reusable delegate *can* recover on reload; first-time cannot. Relayer CORS already exposes `Retry-After`, but the SPA never reads it.

**Suggestion:** Either make `session_account` skip a second `add_delegate_key` when this session's pending pubkey is already the one being verified (or treat abort-0 as "already on-chain, proceed to complete"), or have the consent UI retry only `POST /complete` on `temporarily_unavailable` / 503. Do not claim reload works for the first-time path until one of those exists.

services/server/src/routes/oauth.rs
```diff
@@ -659,12 +714,17 @@ pub async fn session_complete(
         .await
         .map_err(|_| OAuthError::invalid_request("wallet ownership proof is invalid"))?;
 
-    // Atomically claim the session so it can only ever be completed once —
-    // Postgres analog of the WALM-30 prior art's Redis GETDEL guarantee.
-    let session = state
+    // Read the session WITHOUT claiming it. The claim moved below the on-chain
```
ducnmm: **[suggestion]** Several new comments narrate the refactor and ticket history instead of a short invariant. The block at 717–722 ("The claim moved below… consuming first meant a Sui 429… before WALM-605"), 757–760 ("exactly as this handler did before WALM-605"), 852–854, and the test preamble at 1211–1218 restates what the code does and embeds architecture history. The useful WHY is already in `DelegateVerifyOutcome` / `is_unavailable()`; the rest will rot.

**Suggestion:** Keep a one-liner on the irreversible claim (`claim only after a definitive on-chain answer; Retryable must leave status='pending'`). Drop the before/after WALM-605 narration and the "whole of the fix's logic" test essay.

services/server/src/routes/oauth.rs
```diff
@@ -675,26 +735,54 @@ pub async fn session_complete(
...
+            )));
+        }
+        DelegateVerifyOutcome::Unregistered { reason } => {
+            return Err(OAuthError::invalid_request(format!(
+                "delegate key is not registered on-chain: {reason}"
+            )));
+        }
+        DelegateVerifyOutcome::Verified { owner } => owner,
+    };
+    let session = claimed.expect("a verified outcome always claims the session");
```
ducnmm: **[nit]** `claimed.expect("a verified outcome always claims the session")` is a request-path panic. Today it holds because `consumes_session()` is true for `Verified` and the `Unregistered`/`Retryable` arms return first, but the compiler does not enforce that coupling — a new consuming variant, or swapping match order, becomes a 500 panic instead of an `OAuthError`.

**Suggestion:** Match `outcome` first; call `claim_oauth_session` inside the `Verified` (and `Unregistered`) arms and drop the `Option`/`expect`. That also makes the "Retryable does not claim" path obvious without a boolean helper.

ducnmm (commented):
## Summary

The WALM-605 fix is correctly scoped: `session_complete` now classifies `verify_delegate_key_onchain` through `OnchainVerifyError::is_unavailable()` (the same primitive as `auth::cache_reverify_action`), claims the one-time session only on a definitive Verified/Unregistered answer, and returns RFC-shaped `temporarily_unavailable` with `503` + `Retry-After: 5` when Sui cannot be consulted. Single-use minting is still gated on the atomic `UPDATE … WHERE status = 'pending'`, so concurrent completers cannot double-issue a code. The remaining risk is recovery, not classification: the only client (`ConnectClaude`) cannot retry `POST /complete` in isolation, so the first-time path that just submitted `add_delegate_key` still may not actually recover despite the session surviving.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 1

ducnmm (commented):
## Summary

Head `1e3d8e17` (after last review on `6aff6841`) addresses all three prior comments. `session_complete` still classifies `verify_delegate_key_onchain` through `OnchainVerifyError::is_unavailable()` — the same primitive as `auth::cache_reverify_action` — so a Sui `GetObject` 429 / `RpcError` / `ScanCapExceeded` is `DelegateVerifyOutcome::Retryable`, does not claim the one-time row, and returns RFC `temporarily_unavailable` with `503` + `Retry-After: 5`. `KeyNotFound` / `NotFound` / `AccountDeactivated` / `WrongObjectType` stay `Unregistered`, still consume the session, and stay a definitive `400 invalid_request`. 429/unavailable is not treated as revoked or unregistered.

The follow-up also: claims inside the match arms (no request-path `expect`), trims the WALM-605 before/after comments, and teaches `ConnectClaude` to re-POST `/complete` alone. `fetchJson` now keeps HTTP status and the OAuth `error` code; `completeSession()` stashes the payload and, on `temporarily_unavailable` / 503, shows a retry step instead of the terminal error. Tests pin that the retry hits `/complete` once more with no second `/account` and no `add_delegate_key`. Single-use minting is still the atomic `UPDATE … WHERE status = 'pending'`. No SDK/MCP dump (app + relayer only).

The leftover hole is first-time connect after `add_delegate_key` already landed: the retry button recovers, but a reload still re-runs `handleConnect` and can abort code 0 as "not the owner".

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

apps/app/src/pages/ConnectClaude.tsx
```diff
@@ -146,6 +173,53 @@ export default function ConnectClaude() {
         sessionStorage.setItem('memwal_claude_connect', JSON.stringify({ session: sessionId }))
     }, [sessionValid, sessionId])
 
+    // Everything `POST /complete` needs, kept so a retry re-sends exactly this
+    // request. Re-running `handleConnect` instead would re-submit
+    // `add_delegate_key` for a key already on chain (abort code 0).
+    const pendingComplete = useRef<CompletePayload | null>(null)
```
ducnmm, line 216: **[suggestion]** `pendingComplete` is a `useRef`. The in-page "Finish connecting" button re-sends `/complete` and does not re-submit `add_delegate_key`, which is the recovery WALM-605 needed. A reload (or anything that remounts the page) drops that ref, returns the user to consent, and `handleConnect` POSTs `/account` again. `find_reusable_oauth_delegate` only returns `status = 'active'` rows, and `activate_oauth_delegate` still runs only after a successful verify, so a first-time connect that already landed `add_delegate_key` still gets `needs_onchain_registration: true`, re-submits the same pubkey, aborts `EDelegateKeyAlreadyExists` (code 0), and the SPA still maps that to "this wallet is not the owner". Reused-delegate reloads are fine; the reported first-time + 429 path is not, if the user refreshes instead of pressing the new button.

**Suggestion:** Persist the complete payload (the existing `memwal_claude_connect` sessionStorage breadcrumb is the natural place) so a reload can re-POST `/complete` only, and/or treat `add_delegate_key` abort code 0 as "already on-chain, proceed to complete" instead of owner-mismatch. Do not claim reload works for first-time until one of those exists.

apps/app/src/pages/ConnectClaude.tsx
```diff
@@ -322,6 +387,23 @@ export default function ConnectClaude() {
                         </div>
                     )}
 
+                    {step === 'retry-complete' && (
+                        <div className="setup-classic-intro">
+                            <h2 className="setup-classic-title">Sui is busy — try again</h2>
+                            <p className="setup-classic-description">
+                                Your delegate key is registered and this authorization is still valid. We just
```
ducnmm: **[nit]** The retry copy says "Your delegate key is registered and this authorization is still valid." A 503 is "Sui could not be consulted", not a positive on-chain answer. For the happy first-time path (`waitForTransaction` already succeeded) it is likely true; it is not what `classify_delegate_verify` actually knows. If the later retry gets `KeyNotFound`, the user is told the opposite in the terminal error.

**Suggestion:** Phrase it as "we could not reach Sui to verify the key; this authorization is still pending — press the button to finish" and leave registration as unknown until `/complete` returns 200.

ducnmm (commented):
## Summary

Head `c3f3269c` (after last review on `1e3d8e17`) addresses both prior comments. The 503 recovery payload is now mirrored into `sessionStorage` under `memwal_claude_pending_complete`, scoped to the session id, so a remount resumes at `retry-complete` and re-POSTs `/complete` only — no second `/account`, no second `add_delegate_key`. Auto-connect is suppressed while that payload exists. Retry copy no longer claims the key is registered. Tests pin reload-resume and cross-session ignore.

Server-side WALM-605 is unchanged and still correct: `classify_delegate_verify` uses `OnchainVerifyError::is_unavailable()`, Retryable does not claim, Unregistered/Verified still consume via the atomic `WHERE status = 'pending'` update, 503 + `Retry-After: 5` vs definitive 400. 429/unavailable is not treated as revoked. Residual: `rememberPendingComplete` writes storage at the start of every `/complete` and only clears on 200, so a terminal 400 plus reload shows the Sui-busy retry step for a session that is already spent.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

apps/app/src/pages/ConnectClaude.tsx
```diff
@@ -146,6 +201,81 @@ export default function ConnectClaude() {
...
+
+    const readPendingComplete = useCallback((): CompletePayload | null => {
+        if (pendingComplete.current) return pendingComplete.current
+        const saved = loadStoredComplete(sessionId)
+        if (saved) pendingComplete.current = saved
+        return saved
+    }, [sessionId])
+
+    const completeSession = useCallback(async (payload: CompletePayload) => {
+        rememberPendingComplete(payload)
```
ducnmm, line 261: **[suggestion]** `rememberPendingComplete` runs at the start of every `completeSession`, and `PENDING_COMPLETE_KEY` is removed only on 200. A definitive `/complete` failure (400 `invalid_request`, including Unregistered which has already claimed the row) therefore leaves the payload in sessionStorage. `session_view` does not filter on `status = 'pending'`, so a reload of the same URL treats `hasStoredComplete` as "Sui is busy" and shows `retry-complete` for a session that is already spent. Further reloads loop on that screen because the key is never cleared. The in-page path is fine (terminal catch sets `error`); only remount is wrong.

**Suggestion:** Keep the ref write at the start (so a hung in-flight `/complete` can still resume), but `sessionStorage.removeItem(PENDING_COMPLETE_KEY)` in the terminal `catch` of `handleConnect` / `handleRetryComplete`. Or persist to sessionStorage only in the `isRetryableComplete` branch. Pin a 400-then-remount test that stays on the error step, next to the existing 503-reload test.

apps/app/src/pages/ConnectClaude.tsx
```diff
@@ -146,6 +201,81 @@ export default function ConnectClaude() {
         sessionStorage.setItem('memwal_claude_connect', JSON.stringify({ session: sessionId }))
     }, [sessionValid, sessionId])
 
+
+    // Everything `POST /complete` needs, kept so a retry re-sends exactly this
+    // request. Re-running `handleConnect` instead would re-submit
+    // `add_delegate_key` for a key already on chain (abort code 0).
+    //
+    // Mirrored into the existing sessionStorage breadcrumb because a ref does
```
ducnmm: **[nit]** The comment says the payload is "Mirrored into the existing sessionStorage breadcrumb", but this is a new key (`memwal_claude_pending_complete`). That split is load-bearing: the effect at 199–202 overwrites `memwal_claude_connect` with `{ session }` on every mount, so stuffing the payload into that key would drop it before `hasStoredComplete` runs.

**Suggestion:** One line: persist under a separate key because `memwal_claude_connect` is rewritten on mount. Otherwise a later cleanup that merges the two breadcrumbs re-breaks reload recovery.

harrymove-ctrl:
@ducnmm addressed both comments from the `1e3d8e17` review, head is now `c3f3269c`.

**suggestion (reload drops the ref):** you were right that the button alone did not close the reported path. The `/complete` payload is now mirrored into sessionStorage keyed by session id, so a remount resumes at the retry step and re-sends only `/complete`. Auto-connect is suppressed while a recovery payload exists, so the resume cannot fall through into a second `/account`. Two tests: a reload completes with no second `/account` and no second `add_delegate_key`; a payload stored under a different session id is ignored.

**nit (retry copy):** dropped the "your delegate key is registered" claim. A 503 only means Sui could not be consulted, so it now says the authorization is still pending.

I went with the client-side fix rather than treating abort code 0 as "already on-chain" — happy to do the Move-abort mapping as a follow-up if you would rather have both.

apps/app/src/pages/ConnectClaude.tsx
```diff
@@ -187,61 +335,64 @@ export default function ConnectClaude() {
...
                 })
+                // `account.move` abort codes: EDelegateKeyAlreadyExists = 0,
+                // ETooManyDelegateKeys = 2, ENotOwner = 4.
                 let result
+                let alreadyOnChain = false
                 try {
                     result = await signAndExecute({ transaction: tx })
                 } catch (txErr: unknown) {
                     const m = txErr instanceof Error ? txErr.message : String(txErr)
                     if (m.includes('abort code: 0') && m.includes('add_delegate_key')) {
```
ducnmm: **[suggestion]** `alreadyOnChain` is set only when `signAndExecute` throws a message containing both `abort code: 0` and `add_delegate_key`. ConnectClaude always goes through `useSponsoredTransaction`, which `throw new Error(sponsorFailureMessage(lastError))`. Enoki/sidecar dry-run of a duplicate `add_delegate_key` is an upstream 5xx whose body would contain the Move abort; `sponsor.rs` `mask_upstream` maps that to HTTP 502 + `code: sponsor_upstream_error` + `"Sponsor service error"` and never echoes the upstream body. `sponsorFailureMessage` then turns any 502 into `"The transaction was rejected by the sponsor (it may be invalid or already applied). Please refresh and try again."` — no abort code, no function name. Abort 4 / 2 are the same dead match (pre-existing). The new test (`ConnectClaude.test.tsx:232`) mocks `new Error('MoveAbort ... add_delegate_key ... abort code: 0')`, which is not that wrapped 502, so it cannot catch this. 503 retry and 503-reload still work (sessionStorage crumb). Re-entering `handleConnect` after the first-time tx already landed — in-flight `/complete` reload before the crumb is written, or Approve again from consent — still 502s and never reaches `/complete`.

**Suggestion:** Do not treat string-matching abort 0 as the re-submit fix until that text actually survives the sponsor wrapper. Either (1) persist `PENDING_COMPLETE_KEY` at the start of `completeSession` (terminal 400 now clears it, so the old spent-session loop is gone) so an in-flight reload re-POSTs `/complete` only, and (2) have `session_account` skip a second `add_delegate_key` when this session's pending pubkey is already on the account — the original alternative, and the one that does not need Move text in a masked 502. Do not fold every sponsor 502 into `alreadyOnChain`: that would send `ENotOwner` / `ETooManyDelegateKeys` / `EAccountDeactivated` into `/complete` and burn the session as Unregistered. If the SPA must key off the tx error, the relayer has to emit a stable `delegate_key_already_exists` code; `mask_upstream` is explicitly not allowed to forward abort strings. Pin a test whose `signAndExecute` rejection is the real wrapped 502 / `sponsor_upstream_error` message; the current raw `MoveAbort` fixture will keep passing while production still dead-ends.

ducnmm (commented):
## Summary

Head `9145aa6c` (after last review on `c3f3269c`) addresses both prior comments. A definitive `/complete` 400 no longer leaves `memwal_claude_pending_complete` behind: the ref is set up front, sessionStorage is written only on `temporarily_unavailable` / 503, and the terminal branch calls `forgetPendingComplete()`. The 400-then-remount test lands on Approve, not the Sui-busy retry step. The breadcrumb comment now says the payload has its own key because `memwal_claude_connect` is rewritten to `{ session }` on mount.

Server-side WALM-605 is unchanged and still correct. `classify_delegate_verify` uses `OnchainVerifyError::is_unavailable()` — the same primitive as `auth::cache_reverify_action` — so a Sui `GetObject` 429 / `RpcError` / `ScanCapExceeded` is `DelegateVerifyOutcome::Retryable`, does not claim the one-time row, and returns RFC `temporarily_unavailable` with `503` + `Retry-After: 5`. `KeyNotFound` / `NotFound` / `AccountDeactivated` / `WrongObjectType` stay `Unregistered`, still consume via the atomic `WHERE status = 'pending'` update, and stay a definitive `400 invalid_request`. 429/unavailable is not treated as revoked. Claims live in the match arms. Single-use minting is still that update.

The leftover is the new abort-0 mapping: the contract is right (`EDelegateKeyAlreadyExists = 0`, `ENotOwner = 4`) but it never sees those strings. `useSponsoredTransaction` wraps Enoki/sidecar failures through `mask_upstream` into a 502 / `sponsor_upstream_error` message with no Move abort text. The new test injects a raw `MoveAbort ... abort code: 0` the hook does not throw. 503 retry and 503-reload still work; re-entering `handleConnect` after the first-time tx already landed still 502s.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 0

harrymove-ctrl:
@ducnmm you were right, and the new tests say so too — head is now `42d61c82`.

**suggestion (`alreadyOnChain` never fires):** confirmed the whole chain. `signAndExecute` here is `useSponsoredTransaction`, which ends in `throw new Error(sponsorFailureMessage(lastError))`; `mask_upstream` maps a 5xx to `(502, "sponsor_upstream_error", "Sponsor service error")` and drops the body; `sponsorFailureMessage` turns any 502 into *"The transaction was rejected by the sponsor (it may be invalid or already applied). Please refresh and try again."*. No abort code, no function name — so abort 0, 4 and 2 were all dead, and the `MoveAbort ... abort code: 0` fixture was testing a string production never produces.

I took both of your options rather than either alone, and did **not** fold 502 into `alreadyOnChain`:

1. **`/account` answers it on chain.** It now verifies this session's own pending key against the account and returns `resume_pending_registration`. Verified means an earlier attempt landed the tx, so the UI goes straight to `/complete` and never builds a second `add_delegate_key`. Classification goes through `classify_delegate_verify`, so this and `session_complete` cannot drift on what a 429 means.

   A Sui outage during that check is a retryable 503, not a guess. Registering walks into the masked 502; skipping sends `/complete` at a key that is not on the account, which is definitive `Unregistered` and burns the session. The preflight claims nothing, so refusing costs the user one retry. Trade-off worth naming: a Sui outage can now block the preflight, where before it could not. The DB reuse fast path is checked first, so reconnects are untouched, and it is one `GetObject` per connect flow.

2. **The crumb is written before the request.** `rememberPendingComplete` moved to the top of `completeSession`. The in-flight-reload case has no 503 to react to, which is exactly the gap the ref cannot cover. The spent-session loop you caught earlier stays closed because both terminal exits still clear it — success, and the definitive 400.

**Dropped, not kept as decoration:** all three abort branches. The abort-4 owner-mismatch copy was unreachable; a real mismatch still surfaces, as `session_complete`'s 400 `verified owner does not match connected account`.

**Tests** — the fixture is now built from the real `SponsorHttpError` / `sponsorFailureMessage` exports via `importActual`, so it cannot drift from the wrapper, and it first asserts the message contains no `abort code` / `add_delegate_key` before asserting the page **does not POST `/complete`**. That is the guard against the wrong fix you flagged: folding every 502 in would send `ENotOwner` / `ETooManyDelegateKeys` / `EAccountDeactivated` into `/complete` and burn the session. Plus `/account` 503 submits no transaction; a reload with `/complete` still in flight resumes at retry with no second signature and no second preflight; and four Rust tests pin the preflight decision, one of which asserts it and `session_complete` agree on which errors are definitive.

app: 100 tests pass, `tsc -b` clean, no new eslint output. server: 13 `routes::oauth` tests pass.

Two follow-ups I did **not** fold in, happy to file either:
- A stable `delegate_key_already_exists` code from the relayer, as you suggested. Not needed for this path any more now that `/account` decides, but it is the only way the SPA could ever distinguish `ENotOwner` from `ETooManyDelegateKeys` for a message better than the generic sponsor one.
- `AccountDeactivated` still classifies as "register", so a deactivated account is told to submit a tx that will abort. Pre-existing, and orthogonal to WALM-605.

apps/app/src/pages/ConnectClaude.tsx
```diff
@@ -220,28 +370,23 @@ export default function ConnectClaude() {
...
+                owner_address: currentAccount.address,
+                owner_signature: ownerSignature,
+                // `/complete` requires a non-empty digest. When no tx was sent
+                // this call names why, rather than claiming a reused delegate
+                // for a key this session registered on an earlier attempt.
+                tx_digest: txDigest || (resuming ? 'already-on-chain' : 'reused-delegate'),
+                reused_delegate: !preflight.needs_onchain_registration && !resuming,
+            })
         } catch (err) {
             setErrorMsg(err instanceof Error ? err.message : String(err))
```
ducnmm, line 385: **[suggestion]** `/account` 503 is thrown into `handleConnect`'s catch, which always sets `step` to `error` and hides Approve. The relayer left the row `pending` and the copy says “please retry”, but the only way back is a full reload (no crumb yet). `/complete` 503 correctly stays on `retry-complete`; `/account` 503 does not. Not a burn.

**Suggestion:** On `isRetryableComplete(err)` before any payload exists, return to `consent` (keep Approve visible) instead of `error`. Do not POST `/complete` from this branch — there is no signature yet. The existing “does not submit a transaction when /account cannot reach Sui” test should still assert no `signAndExecute` and no `/complete`.

ducnmm (commented):
## Summary

Head `eb89add7` fixes the last review: the duplicate-key skip is decided by `/account` against the chain, not by matching a Move abort that `mask_upstream` never forwards. `classify_pending_key_preflight` reuses `classify_delegate_verify` / `OnchainVerifyError::is_unavailable()`, so a Sui 429 is `Unknown` → RFC `temporarily_unavailable` 503 and `session_account` never claims. `Verified` returns `resume_pending_registration`; the SPA skips `add_delegate_key` and POSTs `/complete`, which still claims only on a definitive arm.

A 429 on the preflight cannot be mistaken for `Unregistered` and cannot burn the session. Writing `PENDING_COMPLETE_KEY` before every `/complete` does not re-introduce the spent-session retry-screen loop: terminal 400 still `forgetPendingComplete()`, and the 400-then-remount test lands on Approve.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 0

PR #880, opened 2026-09-08
fix(server): make namespace restore persist its batch atomically (WALM-591)

### LINKED ISSUE #566: [Bug] Non-Atomic PostgreSQL Transaction Rollback During Namespace Restore Timeouts
### Problem Statement
During namespace `restore` operations across large memory datasets (>10,000 Walrus blobs), if a network timeout occurs midway through batch retrieval from Walrus aggregators, the MemWal relayer persists partial state records into PostgreSQL without atomic rollback.

### Impact
- Subsequent `recall` queries return duplicated vector chunks or truncated metadata.
- Database index state diverges from actual underlying Walrus blob availability certificates.

### Proposed Fix
1. Wrap namespace restoration loops inside explicit database transaction blocks (`BEGIN...COMMIT`).
2. Implement cursor pagination with idempotency keys bound to `blobId`.
3. Add a verification checksum step before marking a restored namespace as `active`.

harrymove-ctrl:
Fixes [WALM-591](https://linear.app/mysten-labs/issue/WALM-591) · Closes #566

## The bug

`POST /api/restore` wraps the entire restore in a deadline:

```rust
// services/server/src/routes/admin.rs
tokio::time::timeout(Duration::from_secs(55), restore_unbounded(state, auth, body)).await
```

Step 6 (persist) then wrote its results with a plain `for` loop of
**autocommitted** `insert_vector` calls — one transaction per row:

```rust
// Step 6: Insert only new entries (no delete!)
let restored = results.len();
for (blob_id, vector) in &results {
    ...
    state.db.insert_vector(&id, owner, namespace, blob_id, vector, blob_size, ...).await?;
}
```

Two ways that leaves a silently half-written index:

1. **Timeout (the reported case).** When the 55s deadline elapses, `timeout`
   **drops the in-flight future**. The loop stops wherever it was, but every
   row already inserted is *already committed* — each iteration had its own
   transaction. The caller gets `UpstreamUnavailable("Restore timed out; …")`
   and never learns how many rows landed, because `restored` is only reported
   on the success path.
2. **Mid-loop DB error.** `?` returns after earlier iterations have committed,
   with the same result.

Either way pgvector ends up holding a partial batch that a later `recall`
reads back as duplicated or truncated results — the "silent half-written
index" from the issue. Nothing reconciles it, and a retry of `restore` skips
the rows that did land (Step 2 filters out already-present blobs), so the
namespace stays inconsistent.

## The fix

Persist the whole batch in **one transaction**.

- `storage/db.rs`: new `VectorInsert` row descriptor and
  `VectorDb::insert_vectors_atomic`, which begins a single transaction, writes
  every row, and commits once. The statement pair (upsert + tombstone clear)
  is factored out of the existing `insert_vector` into `insert_vector_in_tx`,
  so single-row and batch inserts run identical SQL; `insert_vector` keeps its
  signature and behaviour and just wraps the helper in its own transaction.
- `routes/admin.rs`: Step 6 builds the batch up front and hands it over in one
  call.

Both failure modes now roll back:

- **Mid-loop error** — `?` drops the `Transaction`.
- **Dropped future (the timeout)** — SQLx queues a `ROLLBACK` when a dropped
  `Transaction` returns its connection to the pool. This is the same
  cancellation-safety property `jobs::JobUploadLock` already documents and
  relies on for its advisory-lock transaction, which is exactly why a
  transaction is the right fix for a cancelled future rather than, say,
  per-row compensation.

Nothing else changed: discovery, download, SEAL decrypt, re-embed, the
skip/negative-cache behaviour, and the `RestoreResponse { restored, skipped,
total, namespace, owner, truncated }` shape are all untouched.

### Why all-or-nothing rather than "explicit partial"

The AC permits a consistent, queryable partial state with an explicit error,
but all-or-nothing is strictly better here. Restore is idempotent and
re-derivable — Walrus is the source of truth and Step 2 re-computes what is
missing on the next call — so rolling back costs nothing but the re-download,
while a partial commit is *worse* than useless: the rows that landed cause the
retry to skip exactly the blobs it should re-check. Zero rows plus the
existing error is both consistent and trivially queryable, and it needs no new
response field or reconciliation path.

### Why one transaction is safe at this batch size

`clamp_restore_limit` clamps `body.limit` to `1..=100` (the ceiling mirrors
`/api/ask`), and `results` is a subset of that page — rows that were missing,
downloaded, decrypted and re-embedded. So the worst case is 100 single-row
upserts into `vector_entries` in one transaction: a few hundred KB of vectors,
sub-second, touching only rows this owner+namespace is restoring. It takes no
table-level locks and holds no locks another writer contends on. Restore is
also owner-rate-limited before any work starts.

## Tests

Three regression tests in `storage::db::tests`, following the existing
`DATABASE_URL` gate used by the neighbouring DB integration tests
(`forget_clears_idempotency_keys_so_the_same_text_can_be_rewritten`):

```rust
let Some(db) = test_db().await else {
    eprintln!("skipping DB integration test: DATABASE_URL is not configured");
    return;
};
```

- `restore_batch_insert_rolls_back_every_row_when_one_fails` — a 3-row batch
  whose middle row has a wrong-dimension embedding (rejected by the
  `vector(1536)` column). Asserts an explicit error **and zero rows**.
- `restore_batch_insert_commits_every_row_on_success` — the fix must not turn
  restore into a no-op; all rows must be visible.
- `restore_batch_insert_is_all_or_nothing_when_the_future_is_cancelled` —
  reproduces the actual trigger. It first times an uncancelled 40-row batch,
  then cancels a second one at half that duration so the drop genuinely lands
  mid-batch (a fixed 1ms deadline aborts before the first row and passes
  trivially). Asserts none-or-all, never a partial count.

Each test uses UUID-suffixed owners and cleans up its rows, like the tests
around it.

**These tests were verified to actually catch the bug.** Reverting
`insert_vectors_atomic` to the old per-row autocommit loop and re-running:

```
thread '...restore_batch_insert_rolls_back_every_row_when_one_fails' panicked:
assertion `left == right` failed: a failed restore batch must leave zero rows, not a partial index
  left: 1
 right: 0

thread '...restore_batch_insert_is_all_or_nothing_when_the_future_is_cancelled' panicked:
a cancelled restore batch must be all-or-nothing, saw 26 of 40 rows

test result: FAILED. 1 passed; 2 failed
```

`26 of 40 rows` is precisely the silent half-written index this PR fixes.

### Results

Run against `pgvector/pgvector:pg17` (the same image CI's `server-checks` job
uses), with `DATABASE_URL` set — so these tests really executed, they did not
skip:

```
test storage::db::tests::restore_batch_insert_commits_every_row_on_success ... ok
test storage::db::tests::restore_batch_insert_is_all_or_nothing_when_the_future_is_cancelled ... ok
test storage::db::tests::restore_batch_insert_rolls_back_every_row_when_one_fails ... ok

test result: ok. 693 passed; 0 failed; 37 ignored; 0 measured; 0 filtered out
```

Full `cargo test --bins` (CI's exact command). The cancellation test was run 5
consecutive times to check for timing flakiness — 5/5 green. Note the skip
guard still matters: without `DATABASE_URL` all three print the skip line and
return, which is what happens on a laptop with no local Postgres.

`rustfmt --check` is clean on both changed files.

`cargo clippy --all-targets -- -D warnings` reports **613 errors on this
branch vs 612 on `dev` (3da15d36)** — i.e. all pre-existing. CI runs
`cargo clippy --all-targets` with `continue-on-error: true` and an explicit
comment that the crate "emits pre-existing clippy warnings with `-D
warnings`". The single new line is `struct VectorInsert is never constructed`,
which is a false positive from that same pre-existing cascade: `lib.rs`
deliberately declares a reduced module tree that excludes `routes`, so in the
lib target *everything* in `db.rs` is reported unused — including
`VectorDb::new`, `pool()`, `db_status` and `TOMBSTONE_RETENTION`.
`VectorInsert` is constructed in `routes/admin.rs` and in the tests.

`cargo fmt --check` fails on `dev` as well, entirely in
`services/server/src/routes/memory_read.rs` (9 diffs) — pre-existing and
untouched here.

### DIFF
--- a/services/server/src/routes/admin.rs
+++ b/services/server/src/routes/admin.rs
@@ -1162,22 +1162,10 @@ async fn restore_unbounded(
         .flatten()
         .collect();
 
-    // Step 6: Insert only new entries (no delete!), atomically.
-    //
-    // GH #566 / WALM-591: this used to be a plain `for` loop of autocommitted
-    // `insert_vector` calls. `restore()` wraps this future in a 55s
-    // `tokio::time::timeout`, so a slow batch had its future DROPPED mid-loop
-    // and every row inserted up to that point stayed committed — a silently
-    // half-written index that a later `recall` read back as duplicated or
-    // truncated results, with `restored` never reported to the caller. A
-    // mid-loop DB error behaved identically.
-    //
-    // Building the whole batch first and handing it to `insert_vectors_atomic`
-    // makes the step all-or-nothing: SQLx rolls the transaction back both when
-    // a row errors and when the future is dropped, so a failed restore leaves
-    // zero rows and the client can simply retry. `limit` is clamped to at most
-    // 100 (`clamp_restore_limit`) and `results` is a subset of that page, so
-    // this is a bounded, short-lived transaction.
+    // Step 6: Insert only new entries (no delete!), all-or-nothing via
+    // `insert_vectors_atomic` — restore runs under a 55s timeout, and a
+    // partially committed batch is a half-written index (GH #566 / WALM-591).
+    // `limit` is clamped to 100 (`clamp_restore_limit`), so the batch is small.
     transient_unresolved += decrypted_texts.len().saturating_sub(results.len());
     let restored = results.len();
     let truncated =
--- a/services/server/src/routes/admin.rs
+++ b/services/server/src/routes/admin.rs
@@ -15,6 +15,7 @@ use futures::stream::{self, StreamExt};
 use std::sync::Arc;
 
 use crate::services::llm_chat::{ChatCompletionRequest, ChatCompletionResponse, ChatMessage};
+use crate::storage::db::VectorInsert;
 use crate::storage::{seal, walrus};
 use crate::types::*;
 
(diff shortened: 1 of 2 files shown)


services/server/src/routes/admin.rs
```diff
@@ -909,52 +910,79 @@ async fn restore_unbounded(
         .flatten()
         .collect();
 
-    // Step 6: Insert only new entries (no delete!)
+    // Step 6: Insert only new entries (no delete!), atomically.
```
ducnmm: **[suggestion]** The GH #566 / WALM-591 story is copied three times (this 16-line block, `insert_vectors_atomic`'s doc at `services/server/src/storage/db.rs:1777`, and the test module header around `db.rs:968`). It narrates the old loop, the timeout, Drop, and the clamp, which is design history rather than a local invariant. The `insert_vectors_atomic` copy also overstates cancellation (`cancelling this future at any point leaves zero rows behind`): a drop after Postgres has processed `COMMIT` leaves the full batch, which the cancellation test already accepts.

**Suggestion:** Keep one short WHY on `insert_vectors_atomic` (one transaction so restore timeout/`?` cannot commit a prefix; Drop queues ROLLBACK). In `restore_unbounded`, a single line that Step 6 is all-or-nothing via that helper is enough. Drop the repeated ticket narrative from the tests.

services/server/src/storage/db.rs
```diff
@@ -901,6 +920,209 @@ mod tests {
...
+
+        // Cancel the real run halfway through that measured duration, so the
+        // drop lands while rows are still being written.
+        let (owner, ids, blob_ids) = build(&format!("0xcancel-{suffix}"), "cancel");
+        let batch: Vec<VectorInsert<'_>> = ids
+            .iter()
+            .zip(blob_ids.iter())
+            .map(|(id, blob_id)| restore_row(id, &owner, &namespace, blob_id, &vector))
+            .collect();
+        let deadline = (full_batch_time / 2).max(Duration::from_micros(1));
```
ducnmm: **[nit]** `deadline = (full_batch_time / 2).max(Duration::from_micros(1))` can still miss the mid-batch window. On a fast pool the warm-up is sub-millisecond, so the second run is cancelled before `BEGIN` and `landed == 0` passes vacuously. Against a regression to per-row autocommit, that would not reproduce the "26 of 40" failure the PR description used to validate the test.

**Suggestion:** Floor the deadline at a few milliseconds, or insert a short `tokio::time::sleep` between rows only under `#[cfg(test)]` so cancellation is deterministic rather than load-dependent.

ducnmm (commented):
## Summary

This correctly fixes the WALM-591 / GH #566 restore hole: Step 6 used to autocommit one `insert_vector` per row, so the 55s `tokio::time::timeout` (or a mid-loop `?`) could leave a partial `vector_entries` page that a later restore would skip. Folding the upsert + tombstone delete into `insert_vector_in_tx` and committing once via `insert_vectors_atomic` is the right model — SQLx 0.8 queues `ROLLBACK` on `Transaction` drop, the same property `JobUploadLock` already relies on — and `insert_vector`'s signature/behavior is preserved. Tests cover mid-batch SQL failure, the success path, and dropped-future all-or-nothing; the main residual risk is comment duplication plus a timing-based cancellation test that can pass without actually interrupting a write.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

harrymove-ctrl:
@ducnmm ready for another look when you have a moment — head `7c777d30` addresses both comments from the `8444822d` review and has not been re-reviewed yet.

- **suggestion (duplicated rationale):** the GH #566 story is told once, on `insert_vectors_atomic`. One line at the `restore_unbounded` call site, none in the tests.
- Also corrected the overstated cancellation claim you flagged: a drop after Postgres has processed `COMMIT` lands the whole batch, which the test already accepts. The property is "never partial", not "cancelling at any point leaves zero rows".
- **nit (vacuous test):** floored the cancellation deadline at 5ms. Half a sub-millisecond warm-up cancelled before `BEGIN` on a fast pool, so `landed == 0` passed without exercising the rollback.

harrymove-ctrl:
@ducnmm triaged both open review threads against the current head `31e9240c`. Both are already addressed by `31e9240c`, so there is nothing new to push here. Line numbers below are at that head.

**Already addressed**

1. *The GH #566 / WALM-591 story is copied three times, and `insert_vectors_atomic` overstates cancellation* — `31e9240c`. One WHY now lives on `insert_vectors_atomic` (`services/server/src/storage/db.rs:1805-1817`): one transaction so neither a mid-batch `?` nor a dropped future can commit a prefix, SQLx queues the `ROLLBACK`, batch bounded by `clamp_restore_limit`. `restore_unbounded` is down to four lines saying Step 6 is all-or-nothing via that helper (`services/server/src/routes/admin.rs:1165-1168`). The test module header and the rollback test's ticket narrative are gone (`db.rs:994` is a single line). The overstated claim went with it: a drop after Postgres has processed `COMMIT` lands the whole batch, so the documented property is "never partial", not "cancelling at any point leaves zero rows".
2. *`max(Duration::from_micros(1))` can pass vacuously on a fast pool* — `31e9240c`. The deadline is floored at 5ms (`db.rs:1132`), and the match now has an explicit `Ok(Ok(()))` arm asserting a full commit when the batch beats the deadline, so neither outcome is a free pass.

Nothing declined.

**What I could and could not run** — `cargo check --all-targets` is clean on this branch. The only output is dead-code warnings, all of them in `src/sui/*` and `src/routes/security_delete.rs`, which this branch does not touch. `cargo test --bin memwal-server restore_batch_insert` compiles and runs all three tests, but every one of them takes the `DATABASE_URL is not configured` early return and prints `skipping DB integration test`, so they pass without touching Postgres. The sandbox I am in cannot open a TCP connection, so the rollback, commit, and cancellation behaviour these tests exist to pin was **not** actually exercised. That still needs a run against a real database.

ducnmm (approved):
LGTM, test cancellation and rationale comments cleaned up; all checks pass.

Issue #564, opened 2026-08-09
Race Conditions and State Loss Risks in waitForRememberJob / Bulk Polling


gclgcl-87:
Both waitForRememberJob and waitForRememberJobs use exponential backoff + jitter polling. Key logic:Only treats done / failed / not_found as terminal states.
Continues polling on pending / running / uploaded.
On timeout, throws an error and may leave blob_id as an empty string.
Potential problems:If the server briefly returns inconsistent states between uploaded → done, or if a job is cleaned up (not_found), the client loses intermediate state with no callback/event mechanism to notify the caller.
The bulk version uses a Map + shift() to handle duplicate job_ids. This is fragile — if the server returns duplicates or out-of-order results, some jobs may be incorrectly marked as timed out.
Default timeouts (60s for single, 120s for bulk) may be insufficient for large embeddings or network jitter, and there is no adaptive extension.

PR #881, opened 2026-09-08
fix(sdk): keep remember job state across polling timeouts (WALM-597)

harrymove-ctrl:
## The state-loss mechanism

`waitForRememberJob` and `waitForRememberJobs` both poll with backoff and both **discarded every non-terminal observation**.

The job state machine is `pending → running → uploaded → done` (`services/server/src/types.rs:470`), and `uploaded` is the point where the relayer has already written the blob — `jobs.rs:944` sets `status='uploaded', blob_id=$1`. Both wait helpers correctly treat `uploaded` as non-terminal and keep polling, but neither kept the observation. So when the deadline hit:

- **single** — threw a 504 carrying only `jobId`. The blob id it had already seen was gone.
- **bulk** — returned the pre-seeded entry with `blob_id: ""`, indistinguishable from a job that never started.

Either way the caller could not tell "still uploading, blob exists" from "nothing happened", and recovering the blob needed a manual `getRememberStatus` round trip that most callers do not make.

Three further bugs in the same code compounded it, each of which loses or misreports state on its own.

### 1. `done` with no `blob_id` resolved successfully with `""`

```ts
blob_id: status.blob_id ?? "",
```

The relayer sets `status='done'` and `blob_id` in a single `UPDATE` (`jobs.rs:891`), so a `done` without a blob id is a contract violation, not a valid empty result. A caller persisting that empty string has silently lost the memory. It now rejects with `502` (single) / a `failed` result entry (bulk).

### 2. Duplicate job ids corrupted bulk results

```ts
const idx = jobIds.indexOf(jobId);
```

`indexOf` always resolves to the **first** slot, so `["dup", "dup"]` wrote both observations into `results[0]` and left `results[1]` stranded on its pre-seeded timeout entry — a completed write reported as a timeout. Duplicates are a real case: the `statusById` bucket-array-with-`.shift()` right above it already anticipates duplicate responses.

Pending work is now tracked **per occurrence** rather than per id. As a side effect this also handles a relayer that dedupes its response: the unmatched slot stays pending and is re-polled instead of being abandoned.

### 3. `failed` counted timeouts as failures

```ts
failed: results.length - succeeded,
```

A timeout is an *unknown* outcome — the job normally completes server-side after the client stops waiting. Reporting it as `failed` told callers the write was lost. `failed` now counts only server-reported failures, and timeouts get their own `timedOut` field.

## New shape

Single — the thrown 504 keeps `jobId` and gains the last observation. Exported as `RememberJobTimeoutError`:

```ts
try {
    await memwal.rememberAndWait(text)
} catch (err) {
    const timeout = err as RememberJobTimeoutError
    if (timeout.status === 504) {
        timeout.jobId       // "job-abc"   — unchanged
        timeout.lastStatus  // "uploaded"  — new
        timeout.lastBlobId  // "0xblob..." — new: the blob exists
    }
}
```

Bulk — the last observation is folded into the timeout entry:

```ts
{
    id: "job-abc",
    status: "timeout",        // discriminator unchanged
    blob_id: "0xblob...",     // was "" — now the blob id seen at "uploaded"
    last_status: "uploaded",  // new: how far it got
    namespace: "notes",
    error: "polling timed out after 120000ms",
}
```

`RememberBulkResult` gains `timedOut` alongside `succeeded` / `failed`.

Both loops now also poll **once before the first backoff**, so an already-finished job returns immediately instead of paying 75–125 ms of dead latency.

## Backwards compatibility

- The single-job timeout still throws, still `status: 504`, still carries `.jobId` — existing catch-and-read-`.jobId` callers are unaffected.
- Bulk still returns per-item results and still uses `status: "timeout"` as the discriminator, so existing branching keeps working.
- `lastStatus` / `lastBlobId` / `last_status` are additive and optional.
- `timedOut` is a new required field on `RememberBulkResult`; it is additive for readers, and the only in-repo constructor besides the SDK itself (`mock.ts`) is updated.
- The one behavioural break is deliberate and is the fix: `done` with an empty `blob_id` no longer resolves successfully.

Nothing in-repo reads `RememberBulkResult.failed` off the workspace SDK. The sidecar MCP tools that print `failed=` (`services/server/scripts/mcp/tools/{analyze,remember-bulk}.ts`) pin the **published** `@mysten-incubation/memwal@0.0.3` from npm, not the workspace source, so they are unaffected until that pin is bumped.

## Tests

New `packages/sdk/test/remember-job-state-loss.test.mjs` (10 tests, `node:test`, `globalThis.fetch` path router). **7 of the 10 fail against the pre-fix implementation**; the other 3 lock in behaviour that was already correct (job id present on timeout, `uploaded` not treated as terminal).

- timeout after `uploaded` reports `jobId` + `lastStatus` + `lastBlobId` (single) and `blob_id` + `last_status` (bulk)
- timeout with no successful poll leaves `lastStatus` / `lastBlobId` undefined rather than lying with `""`
- `uploaded → done` transition resolves with the real blob id
- `done` with no blob id errors (502) instead of returning `""` — single and bulk
- duplicate job ids map to the correct result slots, with namespaces still aligned to input positions
- a duplicated id is re-polled when the relayer dedupes its response
- `timeout` is counted in `timedOut`, not `failed`
- an already-done job returns without a first-poll sleep

## Verification

```
pnpm --filter @mysten-incubation/memwal build      # clean
pnpm --filter @mysten-incubation/memwal typecheck  # clean
pnpm --filter @mysten-incubation/memwal test       # 121/121 pass (111 before, +10 new)
```

Also typechecked clean: `@mysten-incubation/oc-memwal`, `apps/app` (`tsc -b`), and `services/server/scripts`.

## Note for the reviewer

This touches `packages/sdk/src/memwal.ts`, which two sibling PRs (WALM-598, WALM-162) also touch in different regions of the same file. Expect a possible rebase depending on merge order.

Fixes WALM-597
https://linear.app/mysten-labs/issue/WALM-597
Source issue: https://github.com/MystenLabs/MemWal/issues/564

### DIFF
--- a/docs/sdk/production-readiness.md
+++ b/docs/sdk/production-readiness.md
@@ -96,11 +96,19 @@ An autonomous agent should not make a decision that assumes a write persisted un
 const accepted = await memwal.rememberBulkAsync(items);
 const settled = await memwal.waitForRememberJobs(accepted.job_ids);
 
-if (settled.failed > 0) {
-  const bad = settled.results.filter((r) => r.status !== "done");
-  throw new Error(`${settled.failed} writes did not persist: ${bad.map((b) => b.status).join(", ")}`);
+const bad = settled.results.filter((r) => r.status !== "done");
+if (bad.length > 0) {
+  throw new Error(`${bad.length} writes did not persist: ${bad.map((b) => b.status).join(", ")}`);
 }
 ```
+
+`settled.failed` counts only jobs the relayer itself reported as failed. A job
+that was still running when the poll deadline expired is counted in
+`settled.timedOut` instead, because its outcome is unknown rather than known-bad
+— it may still complete server-side. Gate on `failed + timedOut` (or, as above,
+on every result that is not `done`); `failed > 0` alone would treat a timed-out
+batch as persisted. A timed-out item carries `last_status` and, once the job
+reached `uploaded`, `blob_id`, so you can poll it again instead of rewriting it.
 
 <Warning>
 A job reaching `done` confirms that the relayer stored the memory, but the vector index can briefly lag behind that signal, so a `recall` fired in the same instant might not return the memory yet. For read-after-write critical paths, tolerate a short delay or re-query rather than treating an empty first result as a missing memory.
--- a/docs/sdk/production-readiness.md
+++ b/docs/sdk/production-readiness.md
@@ -103,12 +103,12 @@ if (bad.length > 0) {
 ```
 
 `settled.failed` counts only jobs the relayer itself reported as failed. A job
-that was still running when the poll deadline expired is counted in
-`settled.timedOut` instead, because its outcome is unknown rather than known-bad
-— it may still complete server-side. Gate on `failed + timedOut` (or, as above,
-on every result that is not `done`); `failed > 0` alone would treat a timed-out
-batch as persisted. A timed-out item carries `last_status` and, once the job
-reached `uploaded`, `blob_id`, so you can poll it again instead of rewriting it.
+that still ran when the poll deadline expired is counted in `settled.timedOut`
+instead, because its outcome is unknown rather than known-bad. It might still
+complete server-side. Gate on `failed + timedOut`, or, as above, on every result
+that is not `done`. `failed > 0` alone would treat a timed-out batch as
+persisted. A timed-out item carries `last_status` and, once the job reached
+`uploaded`, `blob_id`, so you can poll it again instead of rewriting it.
 
 <Warning>
 A job reaching `done` confirms that the relayer stored the memory, but the vector index can briefly lag behind that signal, so a `recall` fired in the same instant might not return the memory yet. For read-after-write critical paths, tolerate a short delay or re-query rather than treating an empty first result as a missing memory.
--- a/packages/sdk/CHANGELOG.md
+++ b/packages/sdk/CHANGELOG.md
@@ -5,6 +5,13 @@
 ### Added
 
 - `restore()` results include `failed` (required like `truncated`; SDK defaults omitted to `0`) for permanent decrypt/UTF-8 failures instead of folding them into `skipped` or dropping them silently.
+- `RememberBulkResult.timedOut` counts jobs that were still running when the poll deadline expired, separately from `failed`.
+- `RememberJobTimeoutError` types the `waitForRememberJob` poll timeout (`status` 504) and carries `jobId` plus the last observed `lastStatus` / `lastBlobId`.
+
+### Changed
+
+- **`RememberBulkResult.failed` now counts only relayer-reported failures.** Jobs still running when the poll deadline expired are counted in the new `timedOut` instead of being folded into `failed`. A timeout is an unknown outcome, not a lost write, so `if (settled.failed > 0)` is no longer a complete "did every write persist" check. Gate on `failed + timedOut`, or on `results.some((r) => r.status !== "done")`. `docs/sdk/production-readiness.md` is updated to match.
+- **A job reporting `done` without a `blob_id` is now an error, not a result.** `waitForRememberJob` throws `502` instead of resolving with `blob_id: ""`, which a caller could store as a valid handle to a memory that was never written. The bulk path reports the same case as `failed`.
 
 ### Fixed
 
@@ -12,6 +19,9 @@
 - Declare `engines.node >= 20.0.0`, matching `memwal-mcp` and `openclaw-memory-memwal`. The SDK was the only published package without a floor. (WALM-599)
 - Empty-body 401s now use the same AUTH_REJECTED troubleshooting message as credential 401s instead of telling callers to run `memwal_login`. Headless SDK clients do not have that MCP tool.
 - `account.ts` and `manual.ts` PTBs use typed `tx.pure` helpers instead of the legacy untyped moveCall argument syntax that fails under modern `@mysten/sui`.
+- A `waitForRememberJob` timeout keeps the job id and the last observed `status` / `blob_id` instead of throwing them away, so a caller can poll the job again rather than reissuing a paid write. The 404 (job cleaned up) and 500 (job failed) throws carry the same fields.
+- `waitForRememberJobs` no longer resets a `blob_id` an earlier `uploaded` poll observed when a later poll reports `failed`, `not_found`, or `done` without a blob id.
+- A job id repeated in one `waitForRememberJobs` call resolves each occurrence instead of writing every status into the first slot.
 
 ## 0.1.6
 
--- a/packages/sdk/node_modules
+++ b/packages/sdk/node_modules
@@ -1,1 +1,0 @@
-/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules
--- a/packages/sdk/node_modules
+++ b/packages/sdk/node_modules
@@ -1,0 +1,1 @@
+/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules
--- a/packages/sdk/src/memwal.ts
+++ b/packages/sdk/src/memwal.ts
@@ -363,9 +363,13 @@ export class MemWal {
             }
 
             if (!("status" in status) || status.status === "not_found") {
+                // Job cleanup can retire a row an earlier poll already saw
+                // `uploaded`. Carry that observation out, same as 502/504.
                 throw Object.assign(new Error(`remember job not found: ${jobId}`), {
                     status: 404,
                     jobId,
+                    lastStatus,
+                    lastBlobId,
                 });
             }
 
@@ -376,12 +380,9 @@ export class MemWal {
 
             if (status.status === "done") {
                 if (!status.blob_id) {
-                    // The server sets blob_id and status='done' in the same
-                    // UPDATE, so this cannot happen against a healthy relayer.
-                    // Surface it rather than resolving with blob_id: "" — a
-                    // caller that stores the empty string has silently lost the
-                    // memory. lastBlobId lets a caller recover deliberately if
-                    // an earlier "uploaded" poll did carry the blob id.
+                    // Reject `done` without a blob id rather than resolving
+                    // with blob_id: "". `lastBlobId` carries anything an
+                    // earlier poll saw, so a caller can still recover.
                     throw Object.assign(
                         new Error(
                             `remember job reported done without a blob_id (job_id=${jobId})`,
@@ -402,7 +403,7 @@ export class MemWal {
                     new Error(
                         `remember job failed: ${redactInternalUrls(status.error ?? "unknown error")}`,
                     ),
-                    { status: 500, jobId },
+                    { status: 500, jobId, lastStatus, lastBlobId },
                 );
             }
         }
@@ -537,10 +538,9 @@ export class MemWal {
             namespace: namespaces[idx] ?? this.namespace,
             error: `polling timed out after ${timeoutMs}ms`,
         }));
-        // Track pending work per *occurrence*, not per id. The same job id may
-        // legitimately appear twice in `jobIds`, and resolving it by
-        // `jobIds.indexOf(jobId)` always wrote to the first slot — leaving the
-        // duplicate stuck on its pre-seeded timeout entry.
+        // Track pending work per *occurrence*, not per id: the same job id may
+        // legitimately appear twice in `jobIds`, and each slot needs its own
+        // result.
         let pendingSlots = jobIds.map((jobId, idx) => ({ jobId, idx }));
         let attempt = 0;
 
@@ -593,8 +593,8 @@ export class MemWal {
                     // returns per-item results instead of throwing.
                     if (!status.blob_id) {
                         results[slot.idx] = {
+                            ...results[slot.idx],
                             id: slot.jobId,
-                            blob_id: "",
                             status: "failed",
                             namespace,
                             error: "job reported done without a blob_id",
@@ -610,9 +610,12 @@ export class MemWal {
                     continue;
                 }
                 if (status.status === "failed" || status.status === "not_found") {
+                    // Spread first: a blob id an earlier `uploaded` poll folded
+                    // in is the caller's only handle on a write that may have
+                    // landed. Do not reset it to "".
                     results[slot.idx] = {
+                        ...results[slot.idx],
                         id: slot.jobId,
-                        blob_id: "",
                         status: "failed",
                         namespace,
                         error:
@@ -637,9 +640,9 @@ export class MemWal {
         }
 
         const succeeded = results.filter((r) => r.status === "done").length;
-        // A timeout is an unknown outcome, not a known failure — the job may
-        // well still complete server-side. Counting it as `failed` (which
-        // `results.length - succeeded` did) told callers the write was lost.
+        // A timeout is an unknown outcome, not a known failure: the job may
+        // still complete server-side. Callers who need "did every write land"
+        // must check `failed + timedOut`, not `failed`.
         const failed = results.filter((r) => r.status === "failed").length;
         const timedOut = results.filter((r) => r.status === "timeout").length;
 
--- a/packages/sdk/src/memwal.ts
+++ b/packages/sdk/src/memwal.ts
@@ -373,9 +373,6 @@ export class MemWal {
                 });
             }
 
-            // Remember what we saw before branching, so a later timeout can
-            // report the last known state even if this poll is non-terminal.
-            lastStatus = status.status;
             if (status.blob_id) lastBlobId = status.blob_id;
 
             if (status.status === "done") {
@@ -406,6 +403,13 @@ export class MemWal {
                     { status: 500, jobId, lastStatus, lastBlobId },
                 );
             }
+
+            // Non-terminal (pending | running | uploaded). Record it only here,
+            // so `lastStatus` means "how far the job got", the same as the bulk
+            // path's `last_status`. Assigning before the branches made a 500
+            // report "failed" and a 502 report "done" — the terminal poll, not
+            // the progress a caller needs to decide whether to recover.
+            lastStatus = status.status;
         }
 
         const observed = [
--- a/packages/sdk/src/memwal.ts
+++ b/packages/sdk/src/memwal.ts
@@ -616,10 +616,13 @@ export class MemWal {
                 if (status.status === "failed" || status.status === "not_found") {
                     // Spread first: a blob id an earlier `uploaded` poll folded
                     // in is the caller's only handle on a write that may have
-                    // landed. Do not reset it to "".
+                    // landed. Do not reset it to "". `mark_remember_job_failed`
+                    // leaves `blob_id` in place, so the failing poll itself can
+                    // also be the first one to carry it.
                     results[slot.idx] = {
                         ...results[slot.idx],
                         id: slot.jobId,
+                        blob_id: status.blob_id || results[slot.idx].blob_id,
                         status: "failed",
                         namespace,
                         error:
--- a/packages/sdk/src/memwal.ts
+++ b/packages/sdk/src/memwal.ts
@@ -311,6 +311,22 @@ export class MemWal {
 
     /**
      * Poll an accepted remember job until it reaches a terminal state.
+     *
+     * Non-terminal states (`pending`, `running`, `uploaded`) keep polling — the
+     * job is still progressing server-side. Every observation is retained so a
+     * timeout can report where the job had got to instead of discarding it:
+     * `uploaded` already carries the Walrus `blob_id`, and losing that on
+     * timeout is what made a timed-out job look indistinguishable from one that
+     * never started.
+     *
+     * @throws A {@link RememberJobTimeoutError} (`status: 504`) carrying
+     * `jobId`, plus `lastStatus` / `lastBlobId` when the job was observed at
+     * least once. The job is *not* cancelled — it keeps running server-side, so
+     * callers can resume with `getRememberStatus(err.jobId)`.
+     * @throws `status: 404` when the job id is unknown to the server.
+     * @throws `status: 500` when the server reports the job as failed.
+     * @throws `status: 502` when the server reports `done` without a `blob_id`
+     * — a contract violation that used to resolve as a silent empty string.
      */
     async waitForRememberJob(
         jobId: string,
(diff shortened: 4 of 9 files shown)


packages/sdk/src/memwal.ts
```diff
@@ -515,44 +576,79 @@ export class MemWal {
...
-                    pending.delete(jobId);
-                } else if (status.status === "failed" || status.status === "not_found") {
-                    results[idx] = {
-                        id: jobId,
+                    continue;
+                }
+                if (status.status === "failed" || status.status === "not_found") {
+                    results[slot.idx] = {
+                        id: slot.jobId,
                         blob_id: "",
```
ducnmm: **[suggestion]** Timeout is no longer the only path that throws away an `uploaded` blob id. After a non-terminal poll folds `blob_id` / `last_status` into the pre-seeded timeout entry (lines 629–633), a later terminal overwrite replaces the whole object with `blob_id: ""` for `failed`, `not_found`, and `done` without `blob_id` (line 597). The single-job 404 at line 366 similarly throws `{ status: 404, jobId }` without `lastStatus` / `lastBlobId`, even if an earlier poll already saw `uploaded`. That is the job-cleanup case called out in GH #564, and it is inconsistent with the new 502 path at line 389, which does keep `lastBlobId` for recovery.

**Suggestion:** Attach `lastStatus` / `lastBlobId` to the 404 (and 500) throws the same way as 504/502. On the bulk side, when overwriting a folded timeout entry with `failed` / `not_found` / done-without-blob, keep the last observed `blob_id` (and `last_status`) instead of forcing `""`.

packages/sdk/src/memwal.ts
```diff
@@ -515,44 +576,79 @@ export class MemWal {
...
+                stillPending.push(slot);
             }
+            pendingSlots = stillPending;
         }
 
         const succeeded = results.filter((r) => r.status === "done").length;
+        // A timeout is an unknown outcome, not a known failure — the job may
+        // well still complete server-side. Counting it as `failed` (which
+        // `results.length - succeeded` did) told callers the write was lost.
+        const failed = results.filter((r) => r.status === "failed").length;
```
ducnmm, line 653: **[suggestion]** `failed` is now only server-reported failures, with timeouts counted in `timedOut`. That is a clearer model, but it is a silent break for the documented durability check `if (settled.failed > 0)` in `docs/sdk/production-readiness.md` (line 99). After this change a batch that only timed out has `failed === 0`, so that snippet treats the writes as persisted. The PR text says nothing in-repo reads `RememberBulkResult.failed`; the production-readiness and agent-storage-loop docs do (the latter already filters `status !== "done"`, which stays correct).

**Suggestion:** Update those docs to gate on `failed + timedOut` or `results.some((r) => r.status !== "done")`, and note the `failed` semantic change in `packages/sdk/CHANGELOG.md` when this ships. Do not leave `if (settled.failed > 0)` as the official “writes did not persist” check.

packages/sdk/src/memwal.ts
```diff
@@ -347,11 +369,30 @@ export class MemWal {
...
             }
 
+            // Remember what we saw before branching, so a later timeout can
+            // report the last known state even if this poll is non-terminal.
+            lastStatus = status.status;
+            if (status.blob_id) lastBlobId = status.blob_id;
+
             if (status.status === "done") {
+                if (!status.blob_id) {
+                    // The server sets blob_id and status='done' in the same
```
ducnmm: **[suggestion]** The six-line comment explains a historical SDK bug and the relayer’s `UPDATE` shape (`status='done'` and `blob_id` in one statement). That is server internals leaking into the client, and the same story is already in the PR / ticket.

**Suggestion:** Keep one sentence: reject `done` without `blob_id` rather than returning `""`. Drop the `UPDATE` and “silently lost the memory” narration. Same for the historical `indexOf` comment at line 540.

packages/sdk/test/remember-job-state-loss.test.mjs
```diff
@@ -0,0 +1,352 @@
...
+function makeClient() {
+    const client = MemWal.create({
+        key: new Uint8Array(32).fill(1),
+        accountId: "0x1",
+        serverUrl: "https://relayer.example",
+    });
+    client.buildSealSession = async () => "test-session";
+    return client;
+}
+
```
ducnmm, line 51: **[nit]** The `ONE_POLL` comment says a 20ms budget “guarantees exactly one poll” because the first backoff is 75–125ms. In the new loops the deadline is checked *before* the sleep, so a stub that returns in a couple of milliseconds still enters a second iteration, sleeps, and polls again. The tests only assert `polls > 0`, so they pass, but the comment is wrong and would mislead a later `polls === 1` assertion.

**Suggestion:** Describe `ONE_POLL` as “at least one immediate poll, then the deadline expires during the first backoff,” or assert `polls >= 1` explicitly and drop “exactly one.”

ducnmm (commented):
## Summary

The timeout state-loss fix is correct and well-targeted: both wait loops keep the last non-terminal observation, `uploaded` stays non-terminal, `jobId` is always on the 504, and bulk timeouts retain `blob_id` / `last_status` instead of resetting to `""`. Duplicate slot tracking, the `done`-without-`blob_id` rejection, the `failed` / `timedOut` split, and first-poll-before-backoff are real related fixes, and the new tests lock in the ticket behavior. Remaining gaps are that non-timeout terminal paths still drop a previously seen blob id, and the `failed` semantic change makes the in-repo `settled.failed > 0` durability check a false success on timeout.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 1

jessiemongeon1:
## Style Guide Audit

### `docs/sdk/production-readiness.md` (4 violation(s))

_4 violation(s) (2 regex, 2 claude)_

- **Line 108** — Use "might" not "may"
  - Current: `may`
  - Fix: `might`
- **Line 108** — No em dashes in prose
  - Current: `—`
- **Line 106** — passive voice
  - Current: `that was still running when the poll deadline expired`
  - Fix: `that still ran when the poll deadline expired`
- **Line 108** — word preference: use 'might' not 'may'
  - Current: `it may still complete server-side`
  - Fix: `it might still complete server-side`

---

packages/sdk/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # @mysten-incubation/memwal
 
+## Unreleased
```
ducnmm: **[bug]** Follow-up added notes under `## Unreleased` on a branch still sitting on `## 0.1.6`. `origin/dev` already has unreleased TS `## 0.1.7` (`restore()` `failed`, empty-body 401 AUTH_REJECTED, typed `tx.pure`). Merge/rebase that takes this side drops those landed bullets; keeping both headings leaves an `Unreleased` section this repo does not use. `docs/sdk/changelog.mdx` (FAQ answer + `## 0.1.7` body) is still the pre-WALM-597 0.1.7 text. The new `timedOut` field and the deliberate `done`-without-`blob_id` 502 break are also missing from the notes.

**Suggestion:** Rebase onto current `origin/dev`. Leave `packages/sdk/package.json` at `0.1.7` (do not bump). Fold these bullets under the existing `## 0.1.7` — `timedOut` / `RememberJobTimeoutError` under `### Added`, `failed` semantics under `### Changed`, state-loss fixes under `### Fixed` — without deleting the restore / 401 / `tx.pure` lines. Mirror the same additions in `docs/sdk/changelog.mdx` (FAQ `answer` and the 0.1.7 section intro).

packages/sdk/src/memwal.ts
```diff
@@ -341,17 +363,37 @@ export class MemWal {
...
                     status: 404,
                     jobId,
+                    lastStatus,
+                    lastBlobId,
                 });
             }
 
+            // Remember what we saw before branching, so a later timeout can
+            // report the last known state even if this poll is non-terminal.
+            lastStatus = status.status;
```
ducnmm: **[suggestion]** `lastStatus = status.status` runs before the `done` / `failed` branches, so a 500 throw reports `lastStatus: "failed"` and a 502 reports `"done"` — the terminal poll, not how far the job got. 404 is checked *before* that assignment, so cleanup after `uploaded` correctly keeps `lastStatus: "uploaded"`. Bulk does the useful thing too: the follow-up spread leaves `last_status: "uploaded"` on a later `failed` (test at `remember-job-state-loss.test.mjs:356`). The new 500 test documents the single-job split as intentional (`lastStatus === "failed"`, `lastBlobId` survives), but a caller who gates recovery on `lastStatus === "uploaded"` will skip the 500 path even when `lastBlobId` is set. `RememberJobTimeoutError` still says “last non-terminal status,” which is true only for 504.

**Suggestion:** Assign `lastStatus` only on non-terminal polls (same as bulk `last_status`, same as 404). Then 500/502 keep `"uploaded"` and `lastBlobId` remains the recovery handle. Update the 500 test accordingly, or if “most recent poll” is the single-job contract, say that on the 500/502 throws and stop implying `lastStatus` matches bulk `last_status`.

packages/sdk/src/types.ts
```diff
@@ -241,14 +274,26 @@ export interface RememberBulkOptions {
 export interface RememberBulkItemResult {
     /** job_id returned by the server */
     id: string;
-    /** Walrus blob_id once the job completes ("" if failed) */
+    /**
+     * Walrus blob_id once the job completes ("" if failed).
```
ducnmm: **[suggestion]** `RememberBulkItemResult.blob_id` still says `"" if failed` and that a non-empty last blob id is only a `timeout` feature. After the follow-up spread, a `failed` / `not_found` / done-without-blob item can carry the `uploaded` blob id, and `last_status` is no longer “only set when `status` is `timeout`” (comment at line 292; asserted at `remember-job-state-loss.test.mjs:355-356`). Callers who still treat empty `blob_id` as “nothing landed” or who only read `last_status` on timeouts will miss the recovery handle this commit just preserved.

**Suggestion:** Document that `blob_id` / `last_status` are the last observation on any non-`done` item, including `failed` and `timeout`. Empty `blob_id` means never seen `uploaded`, not “status is failed.”

docs/sdk/production-readiness.md
```diff
@@ -96,12 +96,20 @@ An autonomous agent should not make a decision that assumes a write persisted un
...
+const bad = settled.results.filter((r) => r.status !== "done");
+if (bad.length > 0) {
+  throw new Error(`${bad.length} writes did not persist: ${bad.map((b) => b.status).join(", ")}`);
 }
 ```
 
+`settled.failed` counts only jobs the relayer itself reported as failed. A job
+that was still running when the poll deadline expired is counted in
+`settled.timedOut` instead, because its outcome is unknown rather than known-bad
+— it may still complete server-side. Gate on `failed + timedOut` (or, as above,
```
ducnmm: **[nit]** The new `timedOut` paragraph still trips the Sui style-guide audit on this follow-up: em dash, “may” instead of “might”, and passive “that was still running when the poll deadline expired.” The snippet change itself is the right durability check.

**Suggestion:** Split the sentence (comma or period, no em dash), use “might”, and say the job “still ran when the poll deadline expired.”

ducnmm (dismissed):
## Summary

Follow-up `c492b23d` addresses the four prior comments on `c0953d73`: 404/500 throws now carry `lastStatus` / `lastBlobId`; bulk `failed` / `not_found` / done-without-blob overwrites spread the folded entry so an earlier `uploaded` blob id survives; `docs/sdk/production-readiness.md` gates on `status !== "done"` instead of `settled.failed > 0`; relayer-UPDATE / `indexOf` narration is trimmed; `ONE_POLL` no longer claims exactly one poll. New tests pin 404-after-uploaded, 500-after-uploaded, and bulk fail-after-uploaded.

The recovery shape is right: 404 (the GH #564 cleanup case) keeps the pre-cleanup observation because `not_found` is checked before `lastStatus` is assigned, and bulk terminal overwrites no longer reset `blob_id` to `""`. Do not bump TS (`origin/dev` is already 0.1.7 vs `main` 0.1.6). Remaining issues are the changelog heading, a single-vs-bulk `lastStatus` split introduced on the 500 path, and type comments the spread made false.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 1

packages/sdk/node_modules
```diff
@@ -0,0 +1 @@
+/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules
```
ducnmm: **[bug]** Commit `2ad15250` adds `packages/sdk/node_modules` as a git symlink (mode 120000) whose target is `/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules`. Root `.gitignore` is `node_modules/`, which ignores directories only, so a symlink of that name is not ignored. A checkout on CI or another machine gets a broken link that can block `pnpm install` in `packages/sdk`, and it leaks the author's local path. npm publish would not ship it (`files: ["dist", "README.md"]`), but the tree still must not contain it.

**Suggestion:** `git rm --cached packages/sdk/node_modules` and drop it from the branch. Do not force-add anything named `node_modules`.

packages/sdk/src/memwal.ts
```diff
@@ -515,44 +580,82 @@ export class MemWal {
...
-                    results[idx] = {
-                        id: jobId,
-                        blob_id: "",
+                    continue;
+                }
+                if (status.status === "failed" || status.status === "not_found") {
+                    // Spread first: a blob id an earlier `uploaded` poll folded
+                    // in is the caller's only handle on a write that may have
+                    // landed. Do not reset it to "".
+                    results[slot.idx] = {
```
ducnmm, line 622: **[suggestion]** Single-job polling copies `lastBlobId` from every poll, including `failed` (`memwal.ts:376`), so a first observation of `failed` that still carries `blob_id` is recoverable. The bulk `failed` / `not_found` overwrite spreads the pre-seeded timeout entry and never reads `status.blob_id`. Relayer `mark_remember_job_failed` (`jobs.rs:266`) updates status/error only, so an `uploaded` then `failed` row still has `blob_id`, and both status endpoints return it. `waitForRememberJobs` after other work (the agent-storage-loop pattern) can therefore first-see `{ status: "failed", blob_id: "0x..." }` and report `blob_id: ""`. That contradicts the `types.ts` comment this commit added: empty means never seen at `uploaded`. The uploaded-then-failed-across-polls test still passes because the earlier fold supplies the id.

**Suggestion:** On the failed overwrite, set `blob_id: status.blob_id || results[slot.idx].blob_id` (same as the non-terminal fold). Pin a first-poll `failed` + `blob_id` case next to the existing late-fail test.

ducnmm (dismissed):
## Summary

Follow-up `2ad15250` addresses the four comments on `c492b23d`: notes fold into existing `## 0.1.7` on a rebase of `origin/dev` (restore / 401 / `tx.pure` kept; `timedOut` / 502 break added; mirrored in `docs/sdk/changelog.mdx`); `lastStatus` is assigned only on non-terminal polls so a 500 keeps `"uploaded"`; `blob_id` / `last_status` comments match the spread behavior; the production-readiness paragraph drops the em dash and uses "might". `package.json` stays 0.1.7.

The uploaded-then-terminal sequences the tests pin now keep the recovery handle on both single and bulk. Two leftovers: this follow-up committed a machine-local `packages/sdk/node_modules` symlink (gitignore `node_modules/` does not match a symlink), and a bulk `failed` overwrite still ignores `status.blob_id` on the failed poll itself, so a first observation of `failed` + blob id reports `""`.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

harrymove-ctrl:
@ducnmm addressed all four points from the `c492b23d` review, head is now `2ad15250`. Rebased onto `origin/dev`, so the file conflict is gone.

**bug (changelog):** rebased and folded the notes into the existing `## 0.1.7` next to the landed restore / 401 / `tx.pure` bullets — no `## Unreleased`, no version bump, `package.json` stays 0.1.7. Added the two entries that were missing: `timedOut`, and the deliberate `done`-without-`blob_id` 502 break. Mirrored into `docs/sdk/changelog.mdx` (FAQ answer, section intro, Added/Changed/Fixed).

**suggestion (`lastStatus` split):** took your first option. `lastStatus` is assigned only on a non-terminal poll, so it now means how far the job got rather than the poll that ended the wait, matching the bulk `last_status`. 500 and 502 keep `"uploaded"` with `lastBlobId` intact; the 500 test asserts that instead of documenting the old split.

**suggestion (type comments):** `blob_id` and `last_status` now say they carry the last observation on any non-`done` item, and that an empty `blob_id` means never seen `uploaded`, not "status is failed".

**nit:** style-guide pass on the new production-readiness paragraph — no em dash, "might", active voice.

harrymove-ctrl:

```
```

harrymove-ctrl:
@ducnmm triaged all 10 open review threads against the current head `74829bd8`. Every one is already addressed by a commit pushed after the review that raised it, so there is nothing new to push here. Line numbers below are at `74829bd8`.

**Already addressed**

1. *404/500 throws drop `lastStatus` / `lastBlobId`; bulk terminal overwrite resets `blob_id` to `""`* — `e75101fd`. 404 throws with both fields (`memwal.ts:365-372`), 500 does too (`memwal.ts:399-404`). The bulk done-without-blob (`:598-606`) and `failed` / `not_found` (`:616-634`) branches spread the existing entry instead of replacing it.
2. *`failed` semantics silently break the documented durability check* — `e75101fd`. `docs/sdk/production-readiness.md:99` now gates on `settled.results.filter((r) => r.status !== "done")`, and the paragraph below it spells out `failed` vs `timedOut`. `packages/sdk/CHANGELOG.md` records the semantic change under `## 0.1.7` → `### Changed`.
3. *Relayer `UPDATE` / `indexOf` narration* — `e75101fd`. The 502 comment is three lines and says only "reject `done` without a blob id rather than resolving with `blob_id: \"\"`"; the duplicate-slot comment (`memwal.ts:545-547`) states the invariant with no history.
4. *`ONE_POLL` "exactly one poll"* — `e75101fd`. The comment now says at least one immediate poll, notes the deadline is checked before the sleep, and states the tests never assert an exact count.
5. *`## Unreleased` changelog heading* — `e75101fd` + `681b7cac`. Rebased onto `dev`; notes folded into the existing `## 0.1.7` alongside the restore / 401 / `tx.pure` bullets, `timedOut` and the 502 break included, `package.json` still `0.1.7`, and `docs/sdk/changelog.mdx` mirrors it (FAQ answer, section intro, Added/Changed/Fixed).
6. *`lastStatus` assigned before the terminal branches* — `681b7cac`. It is now assigned only on a non-terminal poll (`memwal.ts:412`), so 500 and 502 keep `"uploaded"` and `lastBlobId`.
7. *`types.ts` comments made false by the spread* — `681b7cac`. `blob_id` and `last_status` both document "last observation on any non-`done` item", and that an empty `blob_id` means never seen at `uploaded`, not "status is failed".
8. *Style-guide nit on the new paragraph* — `681b7cac`. No em dash, "might", "that still ran when the poll deadline expired".
9. *`packages/sdk/node_modules` symlink* — `74829bd8`. `git ls-tree -r 74829bd8 --name-only | grep node_modules` returns nothing.
10. *Bulk `failed` overwrite never reads `status.blob_id`* — `74829bd8`. `memwal.ts:625` is now `blob_id: status.blob_id || results[slot.idx].blob_id`, with a first-poll `failed` + `blob_id` test at `remember-job-state-loss.test.mjs:359`.

Nothing declined.

**Tests** — `packages/sdk`: `tsc` clean, `node --test "test/**/*.test.mjs"` gives 133 pass, 0 fail. The `dev` baseline is 128 pass, 0 fail; this branch adds 5 tests.

**Two things outside the review threads**

- The branch no longer merges cleanly with `dev`. `git merge-tree w597 origin/dev` conflicts on `packages/sdk/CHANGELOG.md` and `docs/sdk/changelog.mdx` only; `memwal.ts` and `types.ts` auto-merge. It needs another rebase before it can land.
- `74829bd8` put a relayer internal back into a code comment: `memwal.ts:619` cites `mark_remember_job_failed` to justify reading `status.blob_id` on the failing poll. That is the shape thread 3 asked to keep out of the client. Say the word and I will cut it back to the invariant alone.

**On #953** — none of this feedback touches the "a job id missing from a poll response stays pending" behaviour (`memwal.ts:585-588`), which is the pattern the Python SDK mirrors in #953. No divergence introduced by anything above.

packages/sdk/src/memwal.ts
```diff
@@ -515,44 +580,85 @@ export class MemWal {
                 }
             }
 
-            for (const jobId of pendingIds) {
-                const status = statusById.get(jobId)?.shift();
+            const stillPending: typeof pendingSlots = [];
+            for (const slot of pendingSlots) {
+                const status = statusById.get(slot.jobId)?.shift();
                 if (!status) {
```
ducnmm, line 586: **[suggestion]** An unmatched slot is pushed back pending without the observation an earlier slot with the same job id just consumed. Re-polling does not help while that earlier slot stays non-terminal: every later response is `shift()`ed by the first slot, so the duplicate remains the pre-seeded timeout entry (`blob_id: ""`, no `last_status`) until the deadline.

The new dedupe test only covers an immediately `done` job, where slot 0 leaves pending on the first poll and slot 1 is fetched on the next. It misses the timeout case this PR is fixing. Production is fine — `build_bulk_status_results` emits one clone per requested id — so this is only the deduping server the comment and test claim to handle.

**Suggestion:** When a response has fewer entries than pending slots for an id, copy that poll's `blob_id` and `last_status` onto the unmatched slots (leave them pending), or apply the single status to every still-pending slot of that id. Add a timeout test where a deduping server stays on `uploaded` and assert every duplicate slot keeps `blob_id`.

ducnmm (commented):
## Summary

Head `74829bd84ac8d93311568a64e0676cdbe6380816` fixes the state loss from the dismissed `c492b23d` and `2ad15250` reviews: non-terminal observations survive timeout, 404/500/502 carry `lastStatus` / `lastBlobId`, a first-seen `failed` poll keeps `blob_id`, `timedOut` is split out of `failed`, duplicate slots are tracked per occurrence, and the stray `packages/sdk/node_modules` symlink is gone. No blocking bugs.

Rebase onto current `dev` before merge. `dev` is 0.1.8 (npm `0.1.7` does not include this fix) and `memwal.ts` conflicts in both wait-loop headers — keep dev's `retryAfterMs` / `pollDeadlineMs` together with this state retention, and move the changelog bullets off the already-published `## 0.1.7`.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 0

PR #882, opened 2026-09-08
fix(auth): stop re-verifying delegate keys on every signed request (WALM-606)

harrymove-ctrl:
Fixes WALM-606 — https://linear.app/mysten-labs/issue/WALM-606

Follow-up to WALM-429 / #811.

## The amplification

`auth::resolve_account` (Strategy 1) called `verify_delegate_key_onchain` — one Sui `GetObject` — on **every** signed request, including every hit on the 24h Postgres `delegate_key_cache`. That table caches `public_key -> (account_id, owner)` so a hit skips the AccountRegistry *scan*; it never skipped the per-request RPC.

So the relayer turned one fullnode `429` into one user-visible auth failure, per request, for as long as the burst lasted. #811 reclassified those from a misleading `401` to an honest `503` — the symptom got truthful, but the pressure source was untouched and recall still failed during a 429 burst.

Fail-closed on that RPC is the right call for auth. The problem is that we were paying for it on requests where we already knew the answer seconds ago.

## What changed

A short in-memory re-verify window, keyed on the exact **`(delegate public key, account object id)`** pair that verified, storing the `owner` that verify returned.

- Inside the window: authenticate from the last good verify. No `GetObject` leaves the process.
- Outside the window: unchanged from today — verify; on `is_unavailable()` return `503` and keep the Postgres row (the #811 contract); on a definitive miss evict the row.

Strategies 2 (signed `x-account-id` hint) and 3 (registry scan) also open a window on success, so the request that populates the DB cache doesn't force the *next* one to burn a second `GetObject` before the window ever exists.

## Interval: 45s, hard-capped at 60s

`MEMWAL_AUTH_REVERIFY_INTERVAL_SECS`, default `45`, clamped to `0..=60`; out-of-range or unparseable values warn and fall back to the default. Follows the existing `AUTH_MAX_CLOCK_DRIFT_SECS` pattern in `types.rs` (documented default + hard ceiling + fallback).

45s is the midpoint of the reviewed 30–60s band. The ceiling is not a tuning knob, it is the **revocation-visibility bound** — an operator must not be able to widen how long a revoked key keeps working. `0` restores verify-every-request as an incident kill switch.

Documented in `docs/reference/environment-variables.md` and `services/server/.env.example`, and in doc comments on the constants themselves.

## Why skipping the GetObject inside the window is safe

1. **Only a successful verify may open a window.** The new `CacheHitOutcome` splits *authenticated from the window* from *verified on-chain just now*, and only the latter carries the "record this" instruction. An `RpcError` maps to `UnavailableKeepCache`, which has no path to the recorder — so it is structurally impossible for an unreachable fullnode to open a window and let an unverifiable key authenticate. That was the main fail-open to avoid.
2. **Keyed on the pair, not the public key.** A cache row rewritten to point the same key at a different account cannot ride a window opened against the original account — the lookup misses and we re-verify.
3. **The window carries the verified owner.** A live window authenticates as the identity the chain actually returned, not as the `owner` column of a mutable 24h Postgres row.
4. **A definitive miss forgets the key's windows.** `Evict` calls `forget_verified_delegate` (all accounts for that key) alongside `delete_cached_key`, so a revocation observed by one request cannot keep authenticating on another that is still inside its window.
5. **The Postgres read still happens every request.** The window only removes the *RPC*, not the mapping lookup — so a row evicted by a concurrent revocation still drops the request out of Strategy 1 immediately.
6. **Restarts and replicas fail closed.** The map is in-memory, so a restart re-verifies every key once, and each replica maintains its own window — the bound holds per process, never longer.

**Revocation-visibility bound:** a delegate key removed on-chain keeps authenticating for at most one interval (≤60s, 45s by default) before the next request re-verifies, gets `KeyNotFound`, evicts the Postgres row and forgets the window. That is four orders of magnitude tighter than the 24h mapping cache this sits in front of.

## Memory

Mirrors the existing `DelegateKeysCache` in `storage/sui.rs`, including the lesson recorded there: a TTL alone gates *trust on read* but never frees the map slot. `VERIFIED_DELEGATE_CACHE_MAX_AGE` is 10 minutes and is swept by the same 5-minute `main.rs` task (in-memory `retain`, no I/O). 10 minutes is ≥10x the 60s interval ceiling, so a sweep can never close a live window — worst case is one extra `GetObject` for a key idle ten minutes.

`storage/sui.rs` now documents the three-way boundary between `delegate_key_cache` (Postgres, 24h, skips the scan), `DelegateKeysCache` (30s, `/agents` list, not on the auth path) and `VerifiedDelegateCache` (this one, the only one that may skip the per-request `GetObject`).

## Tests — 18 new, and what they don't cover

`resolve_account` needs a live `AppState` (Postgres, Redis, a Sui client) and this crate has no HTTP/gRPC mock harness, so tests target the extracted decision point and the cache mechanics, using the no-network technique already established at `storage/sui.rs:1495-1646` (construct the cache directly, backdate the timestamp).

`auth.rs` — `cache_hit_outcome`, the decision the Strategy 1 cache-hit branch delegates to. The verify argument is a closure with a call counter, so "no RPC inside the window" is asserted, not assumed:

- `window_hit_authenticates_without_any_onchain_call` — counter stays 0; owner comes from the window.
- `window_hit_survives_a_sui_429_without_evicting_or_failing` — `RpcError` inside the window: counter 0, neither `UnavailableKeepCache` (503) nor `Evict`.
- `closed_window_still_503s_on_rpc_error_and_keeps_the_row` — the unchanged #811 path.
- `closed_window_observes_a_revocation_and_evicts` — `KeyNotFound` after expiry → `Evict`.
- `closed_window_success_asks_the_caller_to_open_a_window` — only this variant records.
- `deactivated_account_inside_window_is_bounded_by_the_interval` — documents the tradeoff explicitly.

`storage/sui.rs` — window mechanics: `None` never opens a window; boundary is exclusive; interval `0` never opens; owner served inside the window; expiry forces re-verify; **account-scoping** (a window for account A does not authorize account B, nor a different key); refresh-on-re-verify; `forget_verified_delegate` drops every account for a key while leaving other keys alone; the sweep `retain` predicate; and an assertion that the sweep `MAX_AGE` can never close a window at the configured ceiling.

`types.rs` — the config clamp: default inside the 30–60s band, valid values honoured, `0` accepted, `61`/`86400` rejected back to the default, negative/garbage/empty fall back.

**Not covered** (honest limits):
- No end-to-end test through the axum middleware — no mock harness exists, so the wiring of `recent_verified_owner` / `record_verified_delegate` into `resolve_account` is reviewed by inspection, not exercised.
- No concurrency test for two simultaneous requests racing to open the same window. Both would verify and both would insert the same value; the `RwLock` makes the write safe and the outcome idempotent.
- No test that the `main.rs` sweep task *runs* — only its `retain` predicate is tested, matching the precedent for the existing `/agents` sweep.
- `mcp_proxy.rs::legacy_delegate_registered` also issues a `GetObject` per call on the legacy raw-delegate-bearer path. It is not the signed-request path this ticket scopes, so it is left alone.

## Build & verify (local, `services/server`)

- `cargo test` — **exit 0**. `469 passed; 0 failed; 23 ignored` (lib) and `711 passed; 0 failed; 37 ignored` (bin). All 18 new tests pass.
- `cargo clippy --all-targets` (what CI runs) — **exit 0**, no new warnings from the changed files.
- `rustfmt --check` on the changed files — clean.

`cargo clippy --all-targets -- -D warnings` fails on this branch **and on `dev`** — this is pre-existing. CI documents it at `.github/workflows/test.yml:497` (*"Soft-fail: services/server emits pre-existing clippy warnings with `-D warnings`"*) and runs `cargo clippy --all-targets` with `continue-on-error: true`.

| | `dev` (3da15d36) | this branch |
|---|---|---|
| `--all-targets -- -D warnings` | exit 101 | exit 101 |
| `lib` target errors | 585 | 595 |
| `lib test` target errors | 256 | 256 |
| `bin` target | clean | clean |

The +10 are all `dead_code` in the **`lib`** target only, and all of the same class already there on `dev` (e.g. `constant DEFAULT_AUTH_CLOCK_DRIFT_SECS is never used`, `function configured_auth_clock_drift_secs is never used`). `lib.rs` deliberately compiles a trimmed module tree that excludes `auth` and `main`, so anything reached only from those is dead there:

```
constant DEFAULT_AUTH_REVERIFY_INTERVAL_SECS is never used
constant MAX_AUTH_REVERIFY_INTERVAL_SECS is never used
constant VERIFIED_DELEGATE_CACHE_MAX_AGE is never used
function configured_auth_reverify_interval_secs is never used
function new_verified_delegate_cache is never used
function recent_verified_owner is never used
function record_verified_delegate is never used
function forget_verified_delegate is never used
function verify_window_open is never used
fields `owner` and `verified_at` are never read
associated items `auth_reverify_interval` and `from_env` are never used
```

All of these are live and exercised in the `bin` target, which is what ships. No new warnings appear in the `bin` target from any changed file.

`services/server/src/routes/memory_read.rs` is unformatted on `dev`; `cargo fmt` reformats it. That churn is reverted here to keep the diff scoped.


jessiemongeon1:
## Style Guide Audit

**5** violation(s) found. All must be fixed before merge.

### `docs/reference/environment-variables.md` (5 violation(s))

_5 violation(s) (3 regex, 2 claude)_

- **Line 118** — Use "might" not "may"
  - Current: `may`
  - Fix: `might`
- **Line 118** — US English: behaviour → behavior
  - Current: `behaviour`
  - Fix: `behavior`
- **Line 118** — "on-chain" is one word → "onchain"
  - Current: `on-chain`
  - Fix: `onchain`
- **Line 118** — passive-voice
  - Current: `before the relayer re-issues the Sui `GetObject`. Scoped to the exact`
  - Fix: `Rewrite "Scoped to the exact..." to active voice, for example: "The relayer scopes this to the exact..."`
- **Line 118** — passive-voice
  - Current: `a delegate key revoked on-chain keeps working`
  - Fix: `Rewrite to active voice, for example: "a revoked delegate key keeps working"`

---

services/server/src/storage/sui.rs
```diff
@@ -251,6 +251,144 @@ pub async fn list_delegate_keys_cached(
     Ok(keys)
 }
 
+// ============================================================
+// Verified delegate keys — per-request auth re-verify window (WALM-606)
+// ============================================================
+//
+// `auth::resolve_account` used to call `verify_delegate_key_onchain` — one
```
ducnmm: **[suggestion]** New comments narrate WALM-606 / #811 history, restate the three-cache layout, and embed design rationale that already lives in the PR description. The same pattern repeats on `CacheHitOutcome` (`auth.rs:125`), the Strategy 1 branch (`auth.rs:492`), and `DEFAULT_AUTH_REVERIFY_INTERVAL_SECS` (`types.rs:163`). Comments here should say why a non-obvious constraint exists, not retell the ticket.

**Suggestion:** Keep a short invariant comment (pair-keyed, insert-on-Ok-only, staleness ≤ `MAX_AUTH_REVERIFY_INTERVAL_SECS`) and drop the architecture-history / “this path used to…” narrative.

services/server/src/storage/sui.rs
```diff
@@ -251,6 +251,144 @@ pub async fn list_delegate_keys_cached(
...
+/// `None` means the caller must re-verify on-chain.
+pub async fn recent_verified_owner(
+    cache: &VerifiedDelegateCache,
+    public_key_hex: &str,
+    account_object_id: &str,
+    interval: std::time::Duration,
+) -> Option<String> {
+    let guard = cache.read().await;
+    let entry = guard.get(&(public_key_hex.to_string(), account_object_id.to_string()))?;
+    if verify_window_open(Some(entry.verified_at), interval) {
```
ducnmm, line 377: **[nit]** `recent_verified_owner` is on every signed-request Strategy 1 hit. The lookup builds `(public_key_hex.to_string(), account_object_id.to_string())` while holding the `RwLock` read guard, then clones `owner` on a hit. The two key allocations are unnecessary except as a `HashMap<(String, String)>` API workaround.

**Suggestion:** Build the lookup key before acquiring the lock, or use a key type that can be borrowed as `(&str, &str)` so the hot path does not allocate. `owner.clone()` is required to return an owned value and is fine.

services/server/src/main.rs
```diff
@@ -1401,15 +1402,15 @@ async fn main() {
...
-    // Postgres-backed eviction above, nothing else ever removes entries from
-    // this HashMap: the 30s TTL only gates whether a hit is trusted, so
-    // without this sweep it grows for the lifetime of the process, one
-    // entry per distinct account_object_id ever looked up. Sweeping is a
-    // cheap in-memory `retain` (no I/O), so a 5-minute cadence against the
-    // 10-minute `DELEGATE_KEYS_CACHE_MAX_AGE` keeps the map bounded to
-    // recently-active accounts with headroom to spare.
+    // Spawn background task to bound the two in-memory delegate-key caches
+    // (see `storage/sui.rs`). Unlike the Postgres-backed eviction above,
+    // nothing else ever removes entries from these HashMaps: their TTLs only
```
ducnmm: **[nit]** The sweep-task comment still says “nothing else ever removes entries from these HashMaps.” That remains true for `DelegateKeysCache`, but `forget_verified_delegate` now deletes `VerifiedDelegateCache` rows on a definitive miss. The sweep is still needed (forget does not run on window expiry), but the claim is now wrong.

**Suggestion:** Split the two maps in the comment: `/agents` cache is sweep-only; the re-verify map is also pruned on `Evict`, and the sweep only bounds entries that age out without a miss.

ducnmm (commented):
## Summary

This adds a short in-memory re-verify window so Strategy 1 cache hits skip `GetObject` for up to 45s (hard-capped at 60s, `0` restores per-request verify). The security shape looks right: only a successful on-chain verify opens a window, the cache is keyed on the `(public_key, account_id)` pair, auth uses the owner from that verify rather than the mutable Postgres row, `RpcError` still 503s and keeps the row, and a definitive miss forgets the window alongside the DB eviction. Dominant residual risk is the documented staleness bound (revocation / deactivation / owner change can ride for one interval) plus the fact that `resolve_account` wiring is reviewed by inspection rather than an end-to-end mock; I did not find a fail-open or interval-bypass bug. No mandatory setup change: the new env var is optional and defaults to 45s.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 2

harrymove-ctrl:
@ducnmm ready for another look when you have a moment — head `0af4e8fa` addresses all three comments from the `a58b3591` review and has not been re-reviewed yet.

- **nit (hot-path allocations):** `recent_verified_owner` probes the map through a borrowed `dyn DelegatePairKey` built before the read guard, so the two `String`s per signed request are gone. `(String, String)` and `(&str, &str)` hash identically, so the borrowed key finds the owned one. The trait is `Send + Sync`: the probe crosses the cache `.await`, and without it every signed-request future stops being `Send` (which surfaced as an axum middleware trait-bound failure, not a nice error).
- **nit (stale sweep comment):** split the two claims. `DelegateKeysCache` is sweep-only; `VerifiedDelegateCache` is also pruned by `forget_verified_delegate` on a definitive miss, and the sweep still bounds keys that go idle.
- **suggestion (ticket narrative):** trimmed `sui.rs`, `CacheHitOutcome`, the Strategy 1 branch and `DEFAULT_AUTH_REVERIFY_INTERVAL_SECS` down to the invariants.

harrymove-ctrl:
Closing this as superseded rather than merging it.

PR #900 (WALM-618 / WALM-386) merged to `dev` on 14 Sep and shipped the same mechanism this PR proposes — a delegate-verify cache, `DELEGATE_VERIFY_CACHE_TTL` 30s plus `DELEGATE_VERIFY_STALE_GRACE` 600s at `services/server/src/storage/sui.rs:255-291`, consumed by `auth.rs:455,505` and `mcp_proxy.rs:348-380`. This PR's 45s re-verify window is the same idea arriving second, and GitHub now reports the branch `mergeable_state: dirty` against `dev`.

Worth recording for whoever picks up WALM-606: the cache has **not** made the 503s go away. Prod runs `0ffbacbb`, which contains #900, and four read-only `/metrics` scrapes spanning 78s on 21 Sep show 57% of live Sui `GetObject` calls failing `Unavailable` (+150 vs +112 succeeding), producing +125 user-visible 503s in that window. Caching reduces call volume; it does not help when the calls that are still made are being refused.

The remaining gap — no retry on a transient `Unavailable` — is now handled by #949 under WALM-614.

Issue #438, opened 2026-07-13
perf(relayer): production /health p50 ~9s — slower than SDK's hardcoded 15s recall abort, causing cascading AbortErrors


OkeyAmy:
## Measurement

```
$ time curl -s -o /dev/null -w '%{http_code}' https://relayer.memory.walrus.xyz/health
200
real    0m9.288s
```

Repeated runs on 2026-07-13 from Europe/Africa consistently showed 8–10 s for `/health` — an endpoint documented as a 'lightweight connectivity check' with no signing, search, or decryption.

## Why it compounds

- The TS SDK hardcodes a 15 s AbortController on `recall()` (see also #417/#407/#420). When `/health` alone costs ~9 s, a real recall (auth + vector search + Walrus fetch + SEAL decrypt) routinely blows the 15 s budget → `AbortError: This operation was aborted` with no hint that it's latency, not a bug. Our first three recall attempts of the day all failed this way, then succeeded once warm.
- `memwal_health` is what the MCP docs recommend agents call to decide whether memory is available; at 9 s it eats a large chunk of a tool-call budget just to say 'ok'.

## Suggested improvements

- Investigate cold-start / connection pooling on the production relayer (staging felt similar).
- Serve `/health` from the edge or in-process without touching downstream deps.
- Until latency is fixed, raise the SDK's default recall abort above worst-case relayer p95, or make it configurable (linking #420).

Environment: relayer version 0.1.0, apiVersion 1.0.0, commit 542071f; Node v20.19.5 client.

PR #839, opened 2026-08-31
fix(sdk): stop telling headless clients to call memwal_login (WALM-454)

ducnmm:
## Ticket

WALM-454 — https://linear.app/mysten-labs/issue/WALM-454
GH #829 — https://github.com/MystenLabs/MemWal/issues/829

## What changed?

- Empty HTTP 401s from the TypeScript SDK now use the same AUTH_REJECTED troubleshooting copy as non-empty 401s.
- That copy no longer tells callers to run `memwal_login` (an MCP tool). MCP keeps its own signed-out message.
- TypeScript SDK patch bump 0.1.5 → 0.1.6.

## Why is this needed?

Headless SDK users with a private key and account ID were told to call an MCP login tool they do not have.

## Scope

TS SDK `sanitizeServerError` 401 copy.

### Out of scope

- MCP signed-out instructions
- Relayer 401 vs 503 classification (WALM-429 / PR #811)

## How was this tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] End-to-end tests
- [ ] Manual testing
- [ ] Not applicable

Commands: `node --test packages/sdk/test/sanitize-server-error.test.mjs` (5 passed); `node scripts/verify-manual-sdk-release.mjs`

## How can the reviewer verify it?

1. Read `sanitizeServerError` in `packages/sdk/src/utils.ts`.
2. Run `node --test packages/sdk/test/sanitize-server-error.test.mjs` from the SDK package after `tsc`.
3. Confirm the message does not contain `memwal_login` and still links the 401 troubleshooting doc.

## Risks and dependencies

Conflicts with PR #811 on the TypeScript 0.1.6 version files. Merge #811 first, then rebase this and keep the Fixed bullet under `## 0.1.6` without bumping to 0.1.7. MCP clients still get `memwal_login` from the MCP package itself.

## Author checklist

- [x] This pull request maps to one ticket and one logical outcome.
- [x] I reviewed the complete diff myself.
- [x] I removed unrelated, debug, and temporary changes.
- [x] I ran the relevant tests.
- [ ] CI is green.
- [x] The branch is up to date with its target branch.
- [x] I added or updated tests where appropriate.
- [x] I documented any important risk, dependency, rollout, or follow-up.
- [x] I provided clear verification steps.
- [x] The pull request is ready for review and is no longer a Draft.

PR #827, opened 2026-08-28
fix(sdk): use typed tx.pure helpers in account and manual PTBs (WALM-442)

ducnmm:
## Summary
- `account.ts` and `manual.ts` built Move call args with legacy two-arg `tx.pure`.
- SDK peer-depends on `@mysten/sui` 2.5+, which documents `tx.pure.vector` / `tx.pure.string`.
- Switch the three call sites and pin them with a source regression test.

Fixes WALM-442 / GH #799.

## Test plan
- [x] `npm test` in `packages/sdk` (includes typed-pure-args regression)

PR #883, opened 2026-09-08
fix(sdk): stop preflights from consuming recall's abort budget (WALM-598)

harrymove-ctrl:
Fixes WALM-598 — https://linear.app/mysten-labs/issue/WALM-598
Source: https://github.com/MystenLabs/MemWal/issues/438

## The `/health` handler was not the cascade

The report reads as "prod `/health` p50 is ~8–10s, so the SDK's 15s recall abort trips." The handler cannot produce that on its own. `health()` (`services/server/src/routes/admin.rs`) makes exactly one I/O call — `sidecar_write_ready`, a `GET {SIDECAR_URL}/health` bounded by `.timeout(300ms)` and memoized behind a 2s TTL. No DB, Sui, Walrus, Seal or Redis. The handler body is hard-capped at ~300ms.

The cascade was on the client. `recall()` armed a single 15s `AbortController` **before** calling `signedRequest`, which then awaited two preflights that were both unabortable and untimed:

- `ensureCompatibleRelayer()` → `GET /version`, falling back to `GET /health` — no signal, no timeout
- `buildSealSession()` → `GET /config` (no signal, no timeout), dynamic imports of `@mysten/seal` / `@mysten/sui`, then `SessionKey.create()` against Sui RPC/gRPC

Only the final `/api/recall` fetch received the signal. So up to four round-trips ran inside the 15s budget before recall was dispatched, none of them abortable. A ~9s `/health` left ~6s for the actual recall, and the `AbortError` surfaced on the recall fetch even though recall itself was healthy. Both preflights are memoized, so this hit the **first** recall of a process and every recall after the 5-minute SEAL session expiry.

`15000` was also the SDK's only `AbortController` and was not configurable — not exposed on `RecallOptions` or `RecallParams`.

## SDK

**The budget now covers only the request it names.** The deadline moved into `signedRequest`, armed after `ensureCompatibleRelayer()` and `buildSealSession()` have resolved. Preflight latency is no longer charged to recall.

**Each preflight got its own independent deadline.** `GET /version`, `GET /health` and `GET /config` run through a new `fetchWithDeadline` helper with `preflightTimeoutMs` (default 5000, per round-trip). They previously carried no signal at all, so a wedged relayer could hang a caller forever. Worst-case total latency stays bounded.

**The recall timeout is configurable.** `MemWalConfig.recallTimeoutMs` (client-level) and `RecallOptions.timeoutMs` (per-call, wins). Default stays 15000, so existing callers are unaffected. `0` / `Infinity` opts out.

**Timeouts say which phase blew.** A blown budget now throws a `TimeoutError` carrying `phase` (`"preflight GET /version"`, `"POST /api/recall"`, `"GET /health"`) and `timeoutMs`, rather than an unlabeled `AbortError` on the recall fetch regardless of which round-trip actually stalled. `isTimeoutError()` and the `TimeoutError` type are exported.

**Also bounded:** the standalone `MemWal.health()` and the same `/version` → `/health` fallback in `packages/sdk/src/manual.ts`.

### Behaviour change worth a reviewer's eye

Recall used to reject with a DOM `AbortError`; it now rejects with a `TimeoutError`. Nothing in this repo catches recall's `AbortError` by name (checked `packages/`, `apps/`, `docs/`), but it is a visible change for external callers and is called out in the SDK changelog.

## Server

These support AC 1 (keep `/health` cheap and, crucially, *measurable*) rather than fixing the 9s directly — the handler was already 300ms-capped.

- **Single-flight.** The 2s TTL was checked under the lock but not held across the probe, so every request arriving during an in-flight probe started its own — a stampede at each TTL boundary, which the added CI `/health` polling (`4de09d4a`, `48671a57`, `3a79fafa`) made routine. The write guard is now held for the probe.
- **Poisoned-mutex fix.** The read side recovered with `unwrap_or_else(|e| e.into_inner())`; the write side was a bare `if let Ok(mut cache)`. Once poisoned, the timestamp froze, `at.elapsed()` stayed `> 2s` forever, and *every* `/health` fanned out to the sidecar for the life of the process.
- **Async-aware primitive.** `std::sync::Mutex` → `tokio::sync::RwLock`, following the `Arc<tokio::sync::RwLock<..>>` idiom in `services/server/src/storage/sui.rs`. It also cannot poison, which is what makes the previous point structurally impossible rather than merely patched.
- **Instrumentation.** The probe now calls `observe_external("sidecar", "health_probe", ..)` and `record_sidecar_failure`, matching `storage/walrus.rs` / `storage/seal.rs`, and failures log at `warn!` instead of `debug!` (filtered out in production). Probe latency, failure count and cache hit-rate were previously unmeasurable — which is why the actual p50 contributor could not be identified from telemetry.

The `/health` response shape (`services/server/src/types.rs`) and its public, unauthenticated status are unchanged.

## Follow-up bug found, not fixed here

The saturation monitor at `services/server/src/main.rs:1206-1237` reads `queuedWalrusUploads`, `activeWalrusUploads` and `walrusUploadLimits.globalCapacity` from the sidecar's `/health`. Those fields only exist on the sidecar's `/ready` (`services/server/scripts/sidecar/routes/health.ts:103-138`); `/health` returns `{status, uptimeMs}` only. All three `.as_u64().unwrap_or(0)` calls therefore always yield 0, `queued > saturation_threshold` is never true, and **the saturation alert can never fire**. Verified on this branch: `main.rs:726` builds `health_url` as `{sidecar_url}/health` and `main.rs:1206` reuses it directly. Deserves its own ticket.

## Tests

**SDK** — `packages/sdk/test/recall-timeout-budget.test.mjs`, 10 tests:

- a slow `/version` preflight does not abort a healthy recall (the regression: preflight 120ms, recall budget 40ms — this rejected before)
- a slow SEAL session build does not abort a healthy recall (the other preflight inside the old budget)
- per-call `timeoutMs`, client-level `recallTimeoutMs`, and per-call overriding client-level
- a recall timeout names `"POST /api/recall"` as the phase
- a preflight timeout names `"preflight GET /version"` and fires on its own budget rather than hanging
- `health()` is bounded by the preflight budget
- the default budget is still 15000
- the emitted `.d.ts` declares both config knobs

The stub honours `init.signal` (rejecting with a real `AbortError`) so an abort is actually observable; the existing `remember-idempotency.test.mjs` pattern was otherwise followed. Note that file's `/api/config` route is dead code — the SDK fetches `${serverUrl}/config` — so it was not copied.

**Server** — `write_ready_cache` module in `admin.rs`, 4 tests over a probe-generic `write_ready_cached` so the cache semantics are testable without a live `AppState`: fresh value served without re-probing, expired value re-probes, 16 concurrent callers share exactly one probe, and a panicking probe does not disable the cache (the poisoning regression).

### Results

```
pnpm --filter @mysten-incubation/memwal build   # tsc, clean
pnpm --filter @mysten-incubation/memwal test    # tests 121 | pass 121 | fail 0
pnpm --filter @mysten-incubation/memwal typecheck  # clean, exit 0

cargo test (services/server)  # 694 passed; 0 failed; 37 ignored — exit 0
```

### Pre-existing failures on `dev`, unrelated to this PR

`cargo clippy --all-targets -- -D warnings` fails with 612 diagnostics / "256 previous errors" on rustc 1.97.0. Verified pre-existing: reverting `admin.rs` to its `dev` content and re-running produced the **identical** 612-diagnostic, exit-101 result. None reference `admin.rs`. They cluster in `types.rs` (102), `jobs.rs` (63), `storage/security_delete_store.rs` (55), `observability.rs` (46), `alerts.rs` (45) and are mostly `unused import` / `constant is never used` / `too_many_arguments` under `--all-targets`, e.g.:

```
error: unused import: `plaintext::PlaintextEngine`
error: constant `ALERT_TO_SLACK_ENV` is never used
error: unnecessary use of `clone` to create a slice from a reference
   --> src/storage/db.rs:614:34
error: this assertion has a constant value
    --> src/types.rs:3024:9
```

`cargo fmt` likewise reports pre-existing drift in `services/server/src/routes/memory_read.rs` (untouched here). `cargo fmt` was run and then that file's reformatting was reverted to keep this diff scoped; `admin.rs` is clean under `cargo fmt --check`.

## Rebase note

This touches `packages/sdk/src/memwal.ts`, which the sibling WALM-597 and WALM-162 PRs also touch in different regions. Expect a possible rebase. The large `memwal.ts` line count is mostly reindentation from removing the `try`/`finally` that wrapped `recall()`'s old timer.

### DIFF
--- a/docs/sdk/api-reference.md
+++ b/docs/sdk/api-reference.md
@@ -56,7 +56,7 @@ Config:
 For the full config surface, see [Configuration](/reference/configuration).
 
 `recallTimeoutMs` covers the recall request alone. The preflights that run
-before it — the relayer compatibility check and the SEAL session build — are
+before it (the relayer compatibility check and the SEAL session build) are
 bounded separately by `preflightTimeoutMs`, so a slow relayer preflight cannot
 consume recall's budget.
 
@@ -125,7 +125,7 @@ Search for memories matching a natural language query, scoped to `owner + namesp
 - `scoringWeights` blends recency and importance into the ranking (see [Ordering](#ordering) below)
 - `timeoutMs` overrides the client's `recallTimeoutMs` for this call. On expiry
   recall throws a `TimeoutError` whose `phase` names the round-trip that
-  stalled — `"POST /api/recall"`, or `"preflight GET /version"` when it was the
+  stalled: `"POST /api/recall"`, or `"preflight GET /version"` when it was the
   compatibility check. Use the exported `isTimeoutError(err)` to detect it.
 
 **Returns:**
--- a/docs/sdk/api-reference.md
+++ b/docs/sdk/api-reference.md
@@ -50,8 +50,15 @@ Config:
 | `accountId` | `string` | Yes | — | MemWalAccount object ID on Sui |
 | `serverUrl` | `string` | No | `https://relayer.memory.walrus.xyz` | Relayer URL |
 | `namespace` | `string` | No | `"default"` | Default namespace for memory isolation |
+| `recallTimeoutMs` | `number` | No | `15000` | Deadline for the `/api/recall` request. Per-call `RecallOptions.timeoutMs` overrides it |
+| `preflightTimeoutMs` | `number` | No | `5000` | Deadline for each preflight round-trip (`GET /version`, `GET /health`, `GET /config`) |
 
 For the full config surface, see [Configuration](/reference/configuration).
+
+`recallTimeoutMs` covers the recall request alone. The preflights that run
+before it — the relayer compatibility check and the SEAL session build — are
+bounded separately by `preflightTimeoutMs`, so a slow relayer preflight cannot
+consume recall's budget.
 
 ## `MemWal` Methods
 
@@ -116,6 +123,10 @@ Search for memories matching a natural language query, scoped to `owner + namesp
 - `maxDistance` filters weak matches client-side by dropping results where `distance >= maxDistance`
 - `sort` picks the ordering: `"relevance"` (default) or `"recent"` for newest-among-matches
 - `scoringWeights` blends recency and importance into the ranking (see [Ordering](#ordering) below)
+- `timeoutMs` overrides the client's `recallTimeoutMs` for this call. On expiry
+  recall throws a `TimeoutError` whose `phase` names the round-trip that
+  stalled — `"POST /api/recall"`, or `"preflight GET /version"` when it was the
+  compatibility check. Use the exported `isTimeoutError(err)` to detect it.
 
 **Returns:**
 
--- a/packages/sdk/CHANGELOG.md
+++ b/packages/sdk/CHANGELOG.md
@@ -4,12 +4,17 @@
 
 ### Added
 
+- `MemWalConfig.recallTimeoutMs` and `RecallOptions.timeoutMs` make recall's abort budget configurable (default: 15000, unchanged). `MemWalConfig.preflightTimeoutMs` / `MemWalManualConfig.preflightTimeoutMs` bound each preflight round-trip (default: 5000).
+- `isTimeoutError()` and the `TimeoutError` type are exported. A blown budget now throws a `TimeoutError` whose `phase` names the round-trip that stalled (`"preflight GET /version"`, `"POST /api/recall"`) instead of an unlabeled `AbortError`.
 - `restore()` results include `failed` (required like `truncated`; SDK defaults omitted to `0`) for permanent decrypt/UTF-8 failures instead of folding them into `skipped` or dropping them silently.
 
 ### Fixed
 
 - Hash request bodies with `@noble/hashes` instead of WebCrypto-or-`node:crypto`, so the package no longer imports a Node builtin on a browser-reachable path. Vite externalises such an import without warning: the app builds clean and the browser crashes the first time the path runs. The fallback could never have helped a browser anyway — `crypto.subtle` is absent precisely when the page is not a secure context, where `node:crypto` is absent too — so it only served Node <19 while being the sole source of the exposure. `sha256hex` sits on the signed-request path, so every remember and recall reached it. (#322, WALM-136)
 - Declare `engines.node >= 20.0.0`, matching `memwal-mcp` and `openclaw-memory-memwal`. The SDK was the only published package without a floor. (WALM-599)
+- `recall()`'s timeout budget covers only the recall request. It previously started before `ensureCompatibleRelayer()` (`GET /version`, falling back to `GET /health`) and `buildSealSession()` (`GET /config` + Sui RPC), so a slow relayer preflight silently consumed the budget and recall raised an `AbortError` even when recall itself was healthy.
+- `GET /version`, `GET /health` and `GET /config` carry an `AbortSignal` and a deadline. They previously ran with neither and could hang indefinitely.
+- A request's deadline now covers the response body, not just its headers. `fetch()` resolves when headers arrive, so a server that then trickled the body could outlive the budget the caller named.
 - Empty-body 401s now use the same AUTH_REJECTED troubleshooting message as credential 401s instead of telling callers to run `memwal_login`. Headless SDK clients do not have that MCP tool.
 - `account.ts` and `manual.ts` PTBs use typed `tx.pure` helpers instead of the legacy untyped moveCall argument syntax that fails under modern `@mysten/sui`.
 
--- a/packages/sdk/node_modules
+++ b/packages/sdk/node_modules
@@ -1,1 +1,0 @@
-/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules
--- a/packages/sdk/node_modules
+++ b/packages/sdk/node_modules
@@ -1,0 +1,1 @@
+/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules
--- a/packages/sdk/src/types.ts
+++ b/packages/sdk/src/types.ts
@@ -21,6 +21,23 @@ export interface MemWalConfig {
     serverUrl?: string;
     /** Default namespace for memory isolation (default: "default") */
     namespace?: string;
+    /**
+     * Deadline for the `/api/recall` request itself, in milliseconds
+     * (default: 15000). Per-call `RecallOptions.timeoutMs` overrides it.
+     *
+     * This budget covers only the recall round-trip. The preflights that
+     * precede it (`/version`, `/config`, the SEAL session build) are bounded
+     * separately by `preflightTimeoutMs`, so a slow relayer preflight can no
+     * longer consume recall's budget and surface as a spurious abort.
+     */
+    recallTimeoutMs?: number;
+    /**
+     * Deadline for each unauthenticated preflight fetch — `GET /version`,
+     * `GET /health`, `GET /config` — in milliseconds (default: 5000).
+     *
+     * Applied per round-trip, not to the preflight phase as a whole.
+     */
+    preflightTimeoutMs?: number;
 }
 
 // ============================================================
@@ -181,6 +198,19 @@ export interface RecallOptions {
      * the window.
      */
     sort?: "relevance" | "recent";
+    /**
+     * Deadline for the `/api/recall` request, in milliseconds. Defaults to
+     * `MemWalConfig.recallTimeoutMs`, itself 15000.
+     *
+     * The clock starts once the preflights (`/version`, `/config`, the SEAL
+     * session build) have completed, so this is a budget for recall alone.
+     * Pass `0` (or `Infinity`) to opt out of the deadline entirely.
+     *
+     * On expiry recall throws a `TimeoutError` whose `phase` is
+     * `"POST /api/recall"`; a preflight that blows its own budget throws one
+     * naming that preflight instead.
+     */
+    timeoutMs?: number;
 }
 
 /** Recommended object-style recall input — preferred over positional args. */
@@ -560,6 +590,12 @@ export interface MemWalManualConfig {
     walrusPublisherUrl?: string;
     /** Default namespace for memory isolation (default: "default") */
     namespace?: string;
+    /**
+     * Deadline for the unauthenticated compatibility preflight (`GET /version`,
+     * falling back to `GET /health`), in milliseconds (default: 5000).
+     * Applied per round-trip.
+     */
+    preflightTimeoutMs?: number;
 }
 
 /**
--- a/packages/sdk/src/utils.ts
+++ b/packages/sdk/src/utils.ts
@@ -399,13 +399,9 @@ export function clockDriftErrorFromResponse(
  * (`GET /version`, `GET /health`, `GET /config`) the SDK makes before it can
  * dispatch a signed request.
  *
- * These used to run with no signal and no timeout at all. Because they are
- * awaited *inside* a caller's request budget, a relayer whose `/health` p50
- * had drifted to ~9s silently consumed most of `recall()`'s 15s
- * `AbortController` window, and the resulting `AbortError` surfaced on the
- * `/api/recall` fetch even though recall itself was healthy (GH #438).
- * Each preflight now carries its own independent budget so it can neither
- * hang forever nor eat someone else's.
+ * Each preflight carries its own budget: they are awaited inside a caller's
+ * request budget, so an unbounded one both hangs forever and spends someone
+ * else's deadline (WALM-598).
  */
 export const DEFAULT_PREFLIGHT_TIMEOUT_MS = 5_000;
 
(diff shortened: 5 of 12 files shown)


jessiemongeon1:
## Style Guide Audit

### `docs/sdk/api-reference.md` (4 violation(s))

_4 violation(s) (2 regex, 2 claude)_

- **Line 59** — No em dashes in prose
  - Current: `—`
- **Line 128** — No em dashes in prose
  - Current: `—`
- **Line 59** — no-em-dash
  - Current: `before it — the relayer compatibility check`
  - Fix: `before it (the relayer compatibility check`
- **Line 128** — no-em-dash
  - Current: `stalled — `"POST /api/recall"`,`
  - Fix: `stalled (`"POST /api/recall"`,`

---

packages/sdk/src/utils.ts
```diff
@@ -388,6 +388,105 @@ export function clockDriftErrorFromResponse(
...
+          }, timeoutMs)
+        : undefined;
+
+    try {
+        return await fetch(url, { ...init, signal: ac.signal });
+    } catch (err) {
+        if (timedOut) throw timeoutError(phase, timeoutMs, err);
+        throw err;
+    } finally {
+        if (tid !== undefined) clearTimeout(tid);
```
ducnmm, line 480: **[suggestion]** `fetchWithDeadline` clears its timer in `finally` as soon as `fetch()` resolves, which is when response headers arrive. `signedRequest` then calls `res.json()` / `res.text()` with no remaining deadline or signal (see `memwal.ts` after the `fetchWithDeadline` call). The previous recall `AbortController` stayed armed through `signedRequest`'s return, so a slow body used to abort; a server that sends headers promptly and trickles the body can now hang well past `recallTimeoutMs`. `AbortSignal.timeout()` would keep the same signal attached to the body stream.

**Suggestion:** Keep the abort signal alive until the caller finishes reading the body (pass `AbortSignal.timeout(timeoutMs)` through, or delay `clearTimeout` until after `res.json()` / `res.text()`), so the named budget covers the whole round-trip it documents.

docs/sdk/api-reference.md
```diff
@@ -50,9 +50,16 @@ Config:
 | `accountId` | `string` | Yes | — | MemWalAccount object ID on Sui |
 | `serverUrl` | `string` | No | `https://relayer.memory.walrus.xyz` | Relayer URL |
 | `namespace` | `string` | No | `"default"` | Default namespace for memory isolation |
+| `recallTimeoutMs` | `number` | No | `15000` | Deadline for the `/api/recall` request. Per-call `RecallOptions.timeoutMs` overrides it |
+| `preflightTimeoutMs` | `number` | No | `5000` | Deadline for each preflight round-trip (`GET /version`, `GET /health`, `GET /config`) |
 
 For the full config surface, see [Configuration](/reference/configuration).
```
ducnmm, line 57: **[suggestion]** The new `recallTimeoutMs` / `preflightTimeoutMs` rows and the paragraph under the table tell readers the full config surface lives at Configuration, but `docs/reference/configuration.md` was not updated — its `MemWalConfig` / `MemWalManualConfig` tables still stop at `namespace`. Anyone following that pointer will miss the timeout knobs.

**Suggestion:** Add `recallTimeoutMs`, `preflightTimeoutMs`, and `RecallOptions.timeoutMs` to the Configuration tables (and the MemWalManual preflight row), or drop the "full config surface" pointer if api-reference is now the source of truth.

packages/sdk/src/utils.ts
```diff
@@ -388,6 +388,105 @@ export function clockDriftErrorFromResponse(
...
+// Bounded fetch (WALM-598)
+// ============================================================
+
+/**
+ * Default deadline for the unauthenticated preflight round-trips
+ * (`GET /version`, `GET /health`, `GET /config`) the SDK makes before it can
+ * dispatch a signed request.
+ *
+ * These used to run with no signal and no timeout at all. Because they are
+ * awaited *inside* a caller's request budget, a relayer whose `/health` p50
```
ducnmm: **[suggestion]** Production comments retell the WALM-598 incident (prod `/health` p50 ~9s, GH #438, unlabeled `AbortError` on a healthy recall) on the constants, `SignedRequestOptions`, `recall()`, `fetchCompatibilityMetadata`, `index.ts` exports, and `manual.ts`. Ticket IDs are in-repo style; the incident recap is not a non-obvious constraint and will rot.

**Suggestion:** Shrink these to the actual contract (recall budget starts after preflights; each preflight has its own deadline). Leave the narrative in the changelog and the test file header.

services/server/src/routes/admin.rs
```diff
@@ -154,41 +154,126 @@ pub async fn health(State(state): State<Arc<AppState>>) -> Json<HealthResponse>
...
+/// Hard cap on the outbound sidecar probe. `/health` is a connectivity check,
+/// so the handler body must stay bounded well under any client's request
+/// budget (WALM-598 / GH #438).
+const WRITE_READY_PROBE_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(300);
+
+/// Memoized `write_ready` value plus the instant it was observed.
+///
+/// `tokio::sync::RwLock`, not `std::sync::Mutex`, for two reasons:
+///
+/// 1. This is taken on an async request path. A blocking mutex parked a
```
ducnmm: **[nit]** The `RwLock` justification says a blocking `std::sync::Mutex` "parked a Tokio worker thread for the duration of the outbound probe." The previous code dropped the mutex before `.send().await` and only re-acquired it to store the result, so workers were not parked for the 300ms probe. The real reasons to switch are: holding the lock across the probe without blocking a worker (single-flight), and avoiding poison.

**Suggestion:** Drop claim (1) or restate it as "do not hold `std::sync::Mutex` across `.await` now that the write guard spans the probe."

ducnmm (commented):
## Summary

The SDK fix matches WALM-598: `recall()` no longer arms a 15s `AbortController` before `ensureCompatibleRelayer()` / `buildSealSession()`, so a slow `/version`, `/health`, or `/config` cannot spend the recall budget and surface as an unlabeled `AbortError`. Independent `preflightTimeoutMs` (default 5000), configurable `recallTimeoutMs` / `RecallOptions.timeoutMs` (default 15000, unchanged), and `TimeoutError.phase` are the right shape, and the new tests reproduce the GH #438 regression. Server-side `/health` work (single-flight cache, tokio `RwLock`, probe metrics) is support instrumentation, not the 9s itself — the handler was already 300ms-capped. Package version correctly stays `0.1.6` (origin/dev is already ahead of origin/main and npm `0.1.5`); notes belong under Unreleased.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 1

packages/sdk/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # @mysten-incubation/memwal
 
+## Unreleased
```
ducnmm: **[suggestion]** `origin/dev` already dumped unreleased TS as `## 0.1.7` with `restore()` `failed` (#845), empty-body 401 AUTH_REJECTED (#839), and typed `tx.pure` (#827). This head still opens with `## Unreleased` and only the timeout notes. `git merge-tree` against current `origin/dev` conflicts at that heading; taking the PR side drops the landed 0.1.7 bullets. `package.json` is unmodified on this branch (stays 0.1.6 vs this PR's base), so a 3-way merge keeps `origin/dev`'s 0.1.7 — do not resolve the conflict by writing 0.1.8 or by keeping 0.1.6.

**Suggestion:** Rebase onto current `origin/dev`. Leave `packages/sdk/package.json` at `0.1.7`. Fold the timeout Added/Fixed bullets under the existing `## 0.1.7` next to the restore / 401 / `tx.pure` notes. Do not introduce a new version heading.

packages/sdk/test/recall-timeout-budget.test.mjs
```diff
@@ -0,0 +1,264 @@
...
+    );
+    await assert.rejects(res.json(), (err) => {
+        assert.ok(isTimeoutError(err));
+        assert.equal(err.phase, "POST /api/recall");
+        assert.equal(err.timeoutMs, 25);
+        return true;
+    });
+});
+
+test("a body read inside the deadline releases the timer", async () => {
```
ducnmm: **[nit]** `dde02c00` claims a test that a normal body read releases the timer. The test only asserts `res.json()` returns `{ ok: true }`. That passes whether or not `finally { release() }` ran; a regression that leaves the 5s/15s timer armed after a successful read would not fail CI.

**Suggestion:** After the successful `json()`, assert the deadline is gone (fake timers: the callback must not fire; or a later `res.text()` on the already-consumed body must stay a body-used error, not a `TimeoutError` after the original budget).

ducnmm (commented):
## Summary

Follow-up `dde02c00` addresses all four comments from the `445184c2` review. `fetchWithDeadline` no longer clears its timer when headers arrive: `armBodyDeadline` keeps the abort armed until `json` / `text` / `arrayBuffer` / `blob` / `formData` settle, and relabels a self-abort during the body as a phase-tagged `TimeoutError`. That closes the regression against the old recall `AbortController`, which stayed live through `signedRequest`'s `res.json()` / `res.text()`. `docs/reference/configuration.md` now lists `recallTimeoutMs`, `preflightTimeoutMs`, and `RecallOptions.timeoutMs`. Incident recap comments on the constants, `recall()`, the preflights, `manual.ts`, and `admin.rs` are down to the contract; the `RwLock` note no longer claims the old `std::sync::Mutex` parked a worker across the probe.

WALM-598 itself is still the right shape: the 15s budget starts after `ensureCompatibleRelayer()` / `buildSealSession()`, each preflight has its own 5s cap, defaults are unchanged, and the new helper tests cover a trickled body. Do not bump TS (`origin/dev` is already 0.1.7 > `origin/main` 0.1.6); this head does not edit `packages/sdk/package.json`, so a 3-way merge keeps 0.1.7. The style-guide audit on `docs/sdk/api-reference.md:59` and `:128` (em dashes) is still open and was not touched in `dde02c00`.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

packages/sdk/node_modules
```diff
@@ -0,0 +1 @@
+/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules
```
ducnmm: **[bug]** `be17511a` added `packages/sdk/node_modules` as a git symlink (mode 120000) pointing at `/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules`. `.gitignore`'s `node_modules/` does not match a symlink, so it was committable. Checkout on CI creates a dangling symlink; `pnpm install --frozen-lockfile` then fails with `ENOENT: no such file or directory, mkdir '.../packages/sdk/node_modules'` (SDK Unit Node 22/24 and the other JS install jobs on run 34432538514).

**Suggestion:** `git rm --cached packages/sdk/node_modules` and drop it from the commit. Do not track a machine-local path.

ducnmm (dismissed):
## Summary

`be17511a` addresses both open comments on `dde02c00`: timeout notes now sit under the existing `## 0.1.7` (with `package.json` still at origin/dev's 0.1.7), and the timer test spies live `setTimeout` ids so it fails unless the deadline outlives headers and is cleared by the body read. WALM-598 itself is still the right shape — recall's budget starts after the preflights, each HTTP preflight has its own 5s cap, and body reads stay inside the deadline.

The same follow-up staged `packages/sdk/node_modules` as a git symlink to a machine-local path. `.gitignore`'s `node_modules/` does not match a symlink, so it was committable; CI then fails at `pnpm install` with `ENOENT` creating `packages/sdk/node_modules`. Drop it before merge.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

harrymove-ctrl:
@ducnmm addressed both points from the `dde02c00` review, head is now `be17511a`.

**suggestion (changelog):** rebased onto `origin/dev` and folded the timeout notes into the existing `## 0.1.7`, keeping the restore / 401 / `tx.pure` bullets. `package.json` stays at 0.1.7 — no bump, no new heading. Mirrored into `docs/sdk/changelog.mdx`.

**nit (timer test):** you were right that it asserted nothing. It now spies `setTimeout`/`clearTimeout` and pins both halves of the fix: the deadline is still armed when `fetch()` resolves, and released once the body settles. A regression in either direction fails.

Also cleaned the style-guide em dashes on `docs/sdk/api-reference.md:59` and `:128`.

harrymove-ctrl:

```
```

packages/sdk/src/types.ts
```diff
@@ -21,6 +21,23 @@ export interface MemWalConfig {
     serverUrl?: string;
     /** Default namespace for memory isolation (default: "default") */
     namespace?: string;
+    /**
+     * Deadline for the `/api/recall` request itself, in milliseconds
+     * (default: 15000). Per-call `RecallOptions.timeoutMs` overrides it.
+     *
+     * This budget covers only the recall round-trip. The preflights that
+     * precede it (`/version`, `/config`, the SEAL session build) are bounded
```
ducnmm, line 29: **[suggestion]** This says the SEAL session build is bounded by `preflightTimeoutMs`. It is not. That budget is only applied to `GET /version`, `GET /health`, and `GET /config`. `buildSealSessionInner` still awaits `SessionKey.create()` against Sui with no signal and no deadline, and `signedRequest` starts the recall clock only after that returns (`memwal.ts:1422`). The comment just above that fetch (`memwal.ts:1425`) says everything above already has its own deadline. A hung Sui RPC hangs `recall()` with neither budget firing. `RecallOptions.timeoutMs` (`types.ts:205`) and `docs/reference/configuration.md` already describe the clock correctly. `docs/sdk/api-reference.md:59` repeats this overclaim.

**Suggestion:** Drop "the SEAL session build" from the set `preflightTimeoutMs` bounds, in this comment and in `docs/sdk/api-reference.md:59`, and fix `memwal.ts:1425`. Bound `SessionKey.create()` only if that was the intended guarantee — the slow-session test requires the recall budget not to cover it.

packages/sdk/src/utils.ts
```diff
@@ -390,6 +390,145 @@ export function clockDriftErrorFromResponse(
...
+        const read = (original as (...args: unknown[]) => Promise<unknown>).bind(res);
+        Object.defineProperty(target, name, {
+            configurable: true,
+            writable: true,
+            value: async (...args: unknown[]) => {
+                try {
+                    return await read(...args);
+                } catch (err) {
+                    throw relabel(err);
+                } finally {
```
ducnmm, line 523: **[suggestion]** `release()` runs only from these wrapped body methods (and from the `fetch` rejection path). Callers that throw on `res.status` without reading the body leave the referenced timer alive until `preflightTimeoutMs`. Once the socket is closed, that timer alone holds the event loop — a discarded 500 whose socket was destroyed kept the process up for the full timeout; draining the body and clearing the timer let it exit immediately. `res.body.cancel()` does not release either.

Hit sites: `MemWal.health()` on `!res.ok` (`memwal.ts:1048`); `fetchCompatibilityMetadata` (404/405 never reads `versionRes`; failed `/health` throws at `memwal.ts:1111`; other statuses at `memwal.ts:1119`); `fetchServerConfig` on `!res.ok` (`memwal.ts:1160`); the same fallback in `manual.ts:837` and `manual.ts:845`. `signedRequest` always reads the body, so recall's own deadline is fine.

**Suggestion:** Drain with `res.text()` before throwing (and drain `versionRes` before the `/health` fallback), or wrap `body.cancel` so it calls `release()` and cancel on those paths. Test that a non-OK `health()` leaves no live timer.

ducnmm (commented):
## Summary

WALM-598 holds on `a4f4fdcc`: `recall()`'s abort budget starts inside `signedRequest` after `ensureCompatibleRelayer()` and `buildSealSession()`, each of `GET /version`, `GET /health`, and `GET /config` has its own `preflightTimeoutMs` (default 5000), and `recallTimeoutMs` / `RecallOptions.timeoutMs` still default to 15000. A blown budget is a `TimeoutError` with `phase`. The deadline stays armed through the body read, and the relayer `/health` write-ready cache is single-flight under a non-poisoning `tokio::sync::RwLock`.

No correctness bug in that split. Two suggestions below.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

Issue #295, opened 2026-06-17
bug(sdk): SEAL SessionKey 5-minute TTL causes silent auth failures in long-running batch operations


anna-stolbovskaja:
## Description

The SDK builds a SEAL `SessionKey` with a 5-minute TTL (scoped to the server's `packageId`), cached in `sessionCache`. For single `remember()` or `recall()` calls this is fine, but batch operations that exceed 5 minutes hit silent failures:

```ts
// Processing 100+ match predictions sequentially
for (const match of matches) {
  for (const agent of agents) {
    const prediction = await generatePrediction(agent, match);
    // After ~5 minutes, this starts failing with opaque SEAL errors
    await memwal.rememberAndWait(prediction);
  }
}
```

The error surfaces as a SEAL decryption/verification failure deep in the request chain, not as a clear "session expired" message. The `sessionBuildPromise` single-flight guard suggests the SDK is aware of concurrent builds, but there is no proactive refresh before expiry.

## Expected behavior

1. The SDK should transparently refresh the SessionKey before it expires (e.g., at 4 min mark or on 401/SEAL error)
2. If refresh is not possible, surface a clear error: `MemWalSessionExpired: SEAL SessionKey expired after 300s. Call client.refreshSession() or use rememberBulk() for batch operations.`
3. Document the 5-min TTL limitation and recommend `rememberBulk()` (max 20 items) for batch workloads

## Workaround

We currently split batch operations into chunks of ~15 items with forced delays, and wrap each chunk in a try-catch that recreates the MemWal client on SEAL errors. This is brittle.

## Environment

SDK v0.0.7, `@mysten/seal` peer dependency, batch operations with 50+ sequential writes.

ducnmm:
Thanks — this is a good catch.

Proactively refreshing the SEAL `SessionKey` before its 5-minute TTL expires (rather than failing deep in a long batch with an opaque decryption/verification error) is the right behavior. We've taken this on board and will work it into the SDK's session handling. We'll update here when it lands.

PR #884, opened 2026-09-08
fix(sdk): refresh the SEAL SessionKey before it expires (WALM-162)

harrymove-ctrl:
Fixes WALM-162 — https://linear.app/mysten-labs/issue/WALM-162 (upstream: https://github.com/MystenLabs/MemWal/issues/295)

## What was already there, and what was actually missing

`SEAL_SESSION_SAFETY_MARGIN_MS = 30_000` already existed (`memwal.ts`), so the cache was never handed out inside SEAL's own expiry window. That is a **staleness guard**: it moves the cliff 30s earlier. It does not move the *work* off the request path.

The gap was that `buildSealSession()` was purely lazy. Consequences on `dev`:

1. Roughly every 4m30s, one unlucky user-facing request paid the whole rebuild inline — `GET /config`, dynamic imports of `@mysten/seal` and `@mysten/sui`, iterating the gRPC/JSON-RPC client candidates, `SessionKey.create()`, then `signPersonalMessage` + `setPersonalMessageSignature`. Several network round-trips inside a `remember`/`recall`.
2. A build failure escaped raw. It throws *before* any `fetch`, so it never passed through `sanitizeServerError` and reached callers with **no `.status` at all**.
3. Nothing invalidated `sessionCache` when the server rejected the session, so the client kept re-sending a credential it still believed was fresh.

The 5-minute TTL is **unchanged** — shortening it is security finding LOW-13, a separate ticket.

## Refresh strategy: refresh-ahead-on-use, no timers

On a cache hit inside the last `SEAL_SESSION_REFRESH_AHEAD_MS` (60s) of the session's usable life, the caller gets the still-valid cached bytes **immediately** and a rebuild is kicked off out of band. With the 30s margin that means the swap happens around T+3m30s, while the session stays valid until T+4m30s.

I deliberately did not use a timer. A `setInterval`/`setTimeout` refresher holds a Node event-loop handle for the client's whole lifetime — keeping short-lived processes alive and leaking if the client is dropped without an explicit disposal call — and it buys nothing, because the only moment a fresh session matters is when a request is about to be made. **No timer is created by this PR, so there is nothing to `unref()` and no new disposal contract.**

The pieces that keep it safe:

- **Single-flight preserved.** `startSealSessionBuild()` owns `sessionBuildPromise`. A blocking caller (expired cache) and a background refresh-ahead caller share one in-flight build; three concurrent calls produce exactly one `SessionKey.create()`. It also only clears the slot if it still points at its own build.
- **A failed background refresh is inert.** The rejection is swallowed at the call site, so it neither surfaces to the caller nor raises an unhandled rejection; the still-valid cached session is untouched and the next call re-attempts. Once the cache genuinely expires, the failure is surfaced synchronously instead.
- **Cannot fire after the client is gone.** `destroy()` already wiped `sessionCache`; a background rebuild in flight could have landed afterwards and written fresh session material straight back in. `destroy()` now sets a tombstone (and clears `sessionBuildPromise`) before wiping, and the cache write checks it. There is a test for exactly this race.
- The cache write moved from `buildSealSessionInner()` up into `startSealSessionBuild()` so the freshness bookkeeping sits next to the refresh policy — and so the policy is testable without the peer deps.

## Server-side expiry recovery

`signedRequest` now retries **at most once** on a rejected session: it drops the cache entry and rebuilds. Details:

- Invalidation is scoped to the exact bytes that were rejected, so a refresh that landed concurrently with the failing request is not thrown away.
- The retry re-derives the timestamp, nonce and signature — reusing the nonce would be rejected by MED-1 replay protection. There is an assertion on this.
- Bounded at 2 attempts. A persistently rejected session (a badly skewed clock, say) fails fast rather than looping.
- Detection is **body-driven**, not status-driven — matching the markers `@mysten/seal` actually emits (`ExpiredSessionKeyError`, `"Session key has expired"`, `InvalidCertificate`). A bare status match would turn every relayer 500 into a session rebuild plus retry.

### Honest limitation — please read

While implementing this I checked what an expired session actually looks like from the SDK's side on today's server, and **two claims in the ticket do not hold**:

- **The 400s at `seal.ts:307-313` are for a *missing* credential header, not an expired session.** An expired session throws inside `SessionKey.import()`, is caught by the route's outer `try`, and goes through `sendSealFailure` → sidecar **500**.
- **That 500 reaches the client fully redacted.** `services/server/src/storage/seal.rs` wraps it as `AppError::Internal`, and `AppError::Internal`'s `IntoResponse` (`services/server/src/types.rs`) deliberately replaces the message with `"Internal server error (traceId: …)"`. The string `"Session key has expired"` is logged server-side and never crosses the wire.
- Worse, on the **recall** path a SEAL decrypt failure is not even an error: `engine/walrus_seal.rs` logs it and returns `Ok(None)` / `DecryptOutcome::Missing`, so an expired session surfaces as **HTTP 200 with fewer or zero memories**. Same in `routes/admin.rs` restore.

So on a stock deployment the recovery path above is **dormant** — there is no signal for it to key off. It fires where the sidecar body does survive to the client, and is forward-compatible if the relayer ever stops redacting. **The proactive refresh is what actually fixes the reported symptom**, by ensuring a near-expired session is never sent in the first place. I did not widen detection to "any 500", because that would rebuild the session and re-issue the request on every unrelated internal error.

Making the relayer surface session expiry legibly (a distinct status/code instead of a redacted 500, and not silently dropping decrypt failures on recall) is a server-side change and out of scope here — worth its own ticket.

## Error legibility

New helpers in `utils.ts`, following `sanitizeServerError`'s house style (control-char stripping, `redactInternalUrls`, `serverCode`):

| Failure | `status` | `serverCode` | Why |
| --- | --- | --- | --- |
| Transient build failure (Sui RPC, key server, network) | `503` | `SEAL_SESSION_BUILD_FAILED` | The documented `withRetry` retries non-4xx — correct, this is worth retrying |
| Missing / too-old `@mysten/sui` or `@mysten/seal`, or a `/config` that can't describe a transport | `400` | `SEAL_SESSION_UNAVAILABLE` | Deterministic; `withRetry` stops on 4xx instead of burning its budget |
| Server rejected the session even after one rebuild | the server's real status | `SEAL_SESSION_EXPIRED` | Message says "SEAL session expired" and points at clock skew |

Note the previous behaviour was not merely unclassified but actively wrong for the deterministic case: `withRetry` reads `status === undefined` as falsy and **retries**, so a missing peer dependency was retried the full budget while telling the caller nothing about SEAL.

## Tests

`packages/sdk/test/seal-session-refresh.test.mjs` — 20 new tests, `node:test`, no new dependencies. Unlike every other suite in that directory, these do **not** monkey-patch `buildSealSession` away; that method is the unit under test.

Covered:
- Refresh policy (`sealSessionCacheState`) incl. both window boundaries, garbage deadlines, and `refreshAheadMs = 0` collapsing to the old lazy behaviour.
- Expiry classification against the exact body the sidecar composes, plus negatives (`NoAccessError`, a plain internal error, an empty body, a 200).
- Error tagging: transient vs. permanent, non-`Error` throws, loopback-URL redaction.
- Orchestration on the real `buildSealSession`: a fresh cache rebuilds nothing; inside the refresh window three concurrent callers all get the cached bytes and trigger **exactly one** rebuild; a failed background refresh leaves the cached session usable, raises **no unhandled rejection** (asserted via a `process.on("unhandledRejection")` probe) and re-attempts on next use; an expired cache blocks and surfaces a classified error.
- The `destroy()` race: a rebuild that resolves after `destroy()` cannot re-arm the wiped cache.
- Through `signedRequest`: an expiry retries exactly once with a genuinely rebuilt session and a fresh nonce; a persistently rejected session stops at 2 attempts with `SEAL_SESSION_EXPIRED` and leaves no cache behind; an unrelated 500 is not mistaken for an expiry and is not retried; Manual-mode routes send no session and never enter the retry path.

**Not covered, and why:** `buildSealSessionInner()` itself. It dynamically imports `@mysten/seal` + `@mysten/sui` and calls `SessionKey.create()`, which reads the package object over gRPC/JSON-RPC — it needs a live Sui endpoint, so it stays out of the unit suite (it was already untested before this PR, and this PR only removes the cache write from it). Everything WALM-162 actually changed is real code under test; only that network/dynamic-import leaf is stubbed. This is stated in a comment at the top of the test file so the next person does not mistake the stub for broader coverage.

## Verification

```
pnpm --filter @mysten-incubation/memwal build      # tsc, exit 0
pnpm --filter @mysten-incubation/memwal test       # tests 131 | pass 131 | fail 0
pnpm --filter @mysten-incubation/memwal typecheck  # tsc --noEmit, exit 0
```

Baseline on `dev` before this change was 111 passing / 0 failing, so this is 111 + 20 new. No pre-existing failures were observed in this package.

## Reviewer note

This touches `packages/sdk/src/memwal.ts`, which sibling PRs **WALM-597** and **WALM-598** also touch in different regions — expect a possible rebase. The changes here are confined to the SEAL-session block (`buildSealSession` and neighbours), the `signedRequest` send loop, the `destroy()` wipe, and appended helpers in `utils.ts`.


packages/sdk/src/utils.ts
```diff
@@ -436,3 +436,167 @@ export async function delegateKeyToPublicKey(privateKeyHex: string): Promise<Uin
...
+};
+
+/**
+ * Deterministic build failures — a missing or too-old `@mysten/sui` /
+ * `@mysten/seal`, or a relayer `/config` that cannot describe a Sui transport.
+ * These fail identically on every attempt, so they are tagged 400 to stop the
+ * documented `withRetry` helper (`docs/sdk/production-readiness.md`) from
+ * burning its budget on them.
+ */
+const SEAL_SESSION_PERMANENT_MARKERS = [
```
ducnmm, line 549: **[suggestion]** `SEAL_SESSION_PERMANENT_MARKERS` only matches the hand-written `buildSealSessionInner()` string (`"not found in @mysten/sui"` / `"Ensure @mysten/sui"`). The common missing-peer path is an uncaught `import("@mysten/seal")` / `import("@mysten/sui/keypairs/ed25519")` throw (`Cannot find package '@mysten/seal'`, `Cannot find module '@mysten/sui/...'`). That does not match, so `sealSessionBuildError` tags it `503` / `SEAL_SESSION_BUILD_FAILED`. The changelog and helper docs say a missing or too-old `@mysten/sui` / `@mysten/seal` is `400` so `withRetry` stops; for an actually-missing package it still burns the retry budget, which is the misclassification this change set out to fix.

**Suggestion:** Treat module-not-found / `ERR_MODULE_NOT_FOUND` (and similarly a missing `SessionKey` export) as permanent. A test that passes `new Error("Cannot find package '@mysten/seal'")` into `sealSessionBuildError` would lock it.

packages/sdk/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # @mysten-incubation/memwal
 
+## Unreleased
+
+### Fixed
+
+- The SEAL `SessionKey` is now refreshed proactively. Once a cached session enters the last minute of its usable life, a request still gets the cached credential immediately and the rebuild (`GET /config`, two dynamic imports, `SessionKey.create()`, a personal-message signature) happens out of band instead of inline. Long-running batches no longer stall roughly every 4m30s. The 5-minute TTL is unchanged.
+- A SEAL session the server rejects as expired is dropped from the cache and rebuilt once, instead of being reused on every subsequent request because the client still believed it was fresh. Bounded at one retry.
```
ducnmm: **[suggestion]** The Unreleased “server rejects as expired → drop cache and rebuild once” bullet describes SDK behavior that does not fire on a stock relayer. Sidecar `sendSealFailure` does emit `ExpiredSessionKeyError` / `"Session key has expired"`, but `services/server/src/storage/seal.rs` wraps that as `AppError::Internal` and `IntoResponse` redacts it to `"Internal server error (traceId: …)"` — no marker survives, so `isSealSessionExpiredResponse` stays false. On recall, decrypt failures are often `Ok(None)` / HTTP 200 with fewer hits rather than an error. The tests only feed sidecar-shaped bodies. Separately, when the path does fire and both attempts fail, `sealSessionExpiredError` keeps the server’s real status (typically 500), so the documented `withRetry` helper and `waitForRememberJob`’s `status >= 500` poll loop treat a terminal clock-skew failure as transient.

**Suggestion:** Qualify the changelog (recovery is best-effort and currently dormant behind relayer redaction; proactive refresh is the actual fix). For the tagged error, use a 4xx `status` (keep the wire status on `cause`) so `withRetry` / polling stop, or follow up so the relayer surfaces a stable session-expired code the client can key off.

ducnmm (commented):
## Summary

Refresh-ahead-on-use is the right fix for WALM-162: the last 60s of the already-margin-reduced cache window returns the still-valid bytes immediately, a single-flight rebuild runs out of band, a failed background refresh is swallowed, and `destroy()` tombstones the client so a late build cannot re-arm the cache. Error tagging, bounded expiry retry, and the new unit suite (policy, single-flight, destroy race, nonce-fresh retry) match the ticket; versioning is also fine (`origin/main` 0.1.5, `origin/dev` 0.1.6, PR stays 0.1.6 with an Unreleased changelog — no extra bump). Two non-blocking gaps: missing-peer `import()` failures are still classified as retryable 503s, and the recovery changelog/status choice overstates what a stock relayer will actually do.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

packages/sdk/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # @mysten-incubation/memwal
 
+## Unreleased
```
ducnmm: **[suggestion]** Follow-up `6840518b` rewrote the Unreleased bullets (honest recovery note, 400/`SEAL_SESSION_EXPIRED`) but left the heading as `## Unreleased` above `## 0.1.6`. `origin/dev` has since shipped unreleased TS `## 0.1.7` (`restore()` `failed`, empty-body 401, typed `tx.pure`) and `package.json` `0.1.7`. Merge-tree will conflict this file; taking the PR side drops the landed 0.1.7 notes, taking `origin/dev` drops WALM-162. This is not a new dump — do not introduce `## 0.1.8` or bump the package.

**Suggestion:** Rebase onto current `origin/dev`. Leave `packages/sdk/package.json` at `0.1.7`. Keep the existing 0.1.7 Added/Fixed bullets, delete `## Unreleased`, and append these three WALM-162 Fixed bullets under `## 0.1.7`.

ducnmm (commented):
## Summary

Follow-up `6840518b` addresses both COMMENTED threads on `f04b20b1`. Refresh-ahead-on-use, single-flight, swallowed background failure, `destroy()` tombstone, byte-scoped invalidate, and the one-retry expiry path are unchanged and still match WALM-162.

The two prior gaps are closed in the classifier:

- `SEAL_SESSION_PERMANENT_MARKERS` now matches the real `import()` throws (`Cannot find package '@mysten/seal'`, `Cannot find module '@mysten/sui/...'`). Node’s `error.message` for a missing peer is `Cannot find package '…' imported from …` (`code` is `ERR_MODULE_NOT_FOUND` and is not in `.message`); `"cannot find package"` / `"cannot find module"` are what actually fire. The new unit test locks those strings at 400 / `SEAL_SESSION_UNAVAILABLE`, so `withRetry` stops instead of burning its budget.
- `sealSessionExpiredError` is now `status: 400` with the wire `{ status, body }` on `.cause`. That stops `withRetry` and `waitForRememberJob`’s `status >= 500` poll. Changelog bullets now say the expiry-recovery path is best-effort and dormant behind relayer redaction; proactive refresh is the actual fix.

Versioning: this head does not touch `packages/sdk/package.json` (still 0.1.6 on the branch). `origin/dev` is already 0.1.7 vs `origin/main` 0.1.6 — do not bump. Changelog is still under `## Unreleased` on top of `## 0.1.6`; after rebase it belongs under the existing `## 0.1.7`. Python / MCP / OpenClaw are untouched.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

packages/sdk/src/utils.ts
```diff
@@ -436,3 +436,180 @@ export async function delegateKeyToPublicKey(privateKeyHex: string): Promise<Uin
...
+    "ensure @mysten/sui",
+    "get /config response",
+    "get /config requires",
+    // Node ESM / CJS resolution, bundlers, and a package that resolves but no
+    // longer exports what we import.
+    "cannot find package",
+    "cannot find module",
+    "err_module_not_found",
+    "failed to resolve module specifier",
+    "does not provide an export named",
```
ducnmm: **[suggestion]** The follow-up claims a package missing the `SessionKey` export is permanent, and the new test feeds `"The requested module '@mysten/seal' does not provide an export named 'SessionKey'"`. That string is a named-import failure. `buildSealSessionInner()` does a namespace import (`const sealMod = await import("@mysten/seal"); const SessionKey = sealMod.SessionKey`) and never `import { SessionKey }`. A present-but-wrong package yields `undefined`, then `SessionKey.create()` throws `TypeError: Cannot read properties of undefined (reading 'create')`, which still tags 503 / `SEAL_SESSION_BUILD_FAILED`. The missing-package `import()` case from the prior comment is fixed; this sub-case is not. Ed25519Keypair already has a `typeof … !== "function"` guard that produces the `"not found in @mysten/sui"` message; `SessionKey` does not.

**Suggestion:** Either add `typeof SessionKey !== "function"` next to the existing Ed25519Keypair check (so the hand-written permanent markers fire), or drop the named-export test/marker so the suite does not imply coverage this import style cannot produce.

packages/sdk/node_modules
```diff
@@ -0,0 +1 @@
+/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules
```
ducnmm: **[bug]** `61102492` adds `packages/sdk/node_modules` as a mode-`120000` symlink whose target is `/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules`. Root `.gitignore` is `node_modules/` (directories only), so a symlink is not ignored and `git add` picked it up. Checkout on any other machine (and CI) creates a dangling link; Node's resolver will ENOENT when it tries to read `packages/sdk/node_modules` as a directory; the blob also leaks the author's local path. `package.json` `"files"` is `["dist", "README.md"]`, so npm publish would not ship it, but the git tree is still broken.

**Suggestion:** `git rm -f packages/sdk/node_modules` and do not re-add it. Leave `packages/sdk/package.json` at `0.1.7`. Optional: add a non-directory `node_modules` ignore so a local symlink cannot be committed again.

ducnmm (dismissed):
## Summary

Follow-up `61102492` closes both COMMENTED threads on `6840518b`. The changelog is folded under the existing `## 0.1.7` (package stays `0.1.7`, landed restore / 401 / `tx.pure` bullets kept, mirrored into `docs/sdk/changelog.mdx`). `buildSealSessionInner()` now guards `typeof SessionKey !== "function"` next to Ed25519Keypair, drops the unreachable named-export marker, and the classifier test feeds the hand-written message this path actually produces. Refresh-ahead-on-use, single-flight, swallowed background failure, `destroy()` tombstone, byte-scoped invalidate, and the one-retry 400/`SEAL_SESSION_EXPIRED` path are unchanged and still match WALM-162.

The same commit also added `packages/sdk/node_modules` as a git symlink to the author's local absolute path. That is not part of WALM-162 and must be removed before merge.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

harrymove-ctrl:
@ducnmm addressed both suggestions from the `6840518b` review, head is now `61102492`.

**suggestion (changelog):** rebased onto `origin/dev`, folded the three WALM-162 bullets under the existing `## 0.1.7`, deleted `## Unreleased`, kept the landed restore / 401 / `tx.pure` notes. No `## 0.1.8`, no package bump. Mirrored into `docs/sdk/changelog.mdx`.

**suggestion (`SessionKey` export):** you were right and I had it backwards. `buildSealSessionInner` namespace-imports, so a present-but-old `@mysten/seal` never raises a named-export error — `SessionKey` is just `undefined` and `.create()` throws a `TypeError` that still classified 503. Added `typeof SessionKey !== "function"` next to the existing `Ed25519Keypair` guard so it produces the same permanent message, dropped the unreachable `does not provide an export named` marker, and repointed the test at the message this import style can actually produce. The missing-package `import()` cases stay covered.

harrymove-ctrl:

```
```

packages/sdk/src/memwal.ts
```diff
@@ -1261,33 +1295,103 @@ export class MemWal {
...
+     * request is about to be made.
+     *
+     * The rejection is swallowed here on purpose. The cached session is still
+     * valid, so a failed background refresh must not surface to the caller nor
+     * raise an unhandled rejection; the next call re-attempts, and once the
+     * cache truly expires the failure is surfaced synchronously instead.
+     */
+    private refreshSealSessionInBackground(): void {
+        if (this.destroyed) return;
+        this.startSealSessionBuild().catch(() => {
```
ducnmm, line 1362: **[bug]** This shares `startSealSessionBuild`'s `sessionBuildPromise` (the early return at line 1313) and only drops the slot when that promise settles. The comment says a failed refresh is swallowed and retried on the next use, and that once the cache expires the failure is surfaced on the blocking path. That holds only when the promise rejects.

`buildSealSessionInner` awaits `GET /config` with no `AbortSignal` and no deadline, then `SessionKey.create()` → `suiClient.core.getObject`. If that attempt stops settling, callers inside the refresh window still get the cached bytes, but the first caller after `expiresAt` joins the stuck attempt instead of opening a new one. A recovered Sui RPC or `/config` cannot be used until the original attempt settles, and the open socket keeps the process alive after the request that triggered the refresh has returned. `destroy()` nulls the slot and the tombstone skips the cache write, but it does not abort the RPC.

**Suggestion:** Deadline the background build, and abort that signal from `destroy()`. Timeout rejects, the existing `.catch` swallows it, and `.finally` clears the slot so the post-expiry `buildSealSession()` starts a new attempt. Test a `buildSealSessionInner` that never settles: expire the cache and assert the next call is not stuck on that promise.

ducnmm (changes requested):
## Summary

Head `8fee5c3db2184c6a437b1ecde472130259438820`. The dismissed `61102492` threads are fixed: import-resolution markers, the `typeof SessionKey !== "function"` guard, `400` / `SEAL_SESSION_EXPIRED` with the wire status on `.cause`, the dormant-recovery changelog note, and removal of the `packages/sdk/node_modules` symlink. Refresh-ahead, single-flight, the destroy tombstone, byte-scoped invalidation, and the one-retry cap match the tests.

Rebase onto current `dev` before merge. `dev` is unreleased 0.1.8 and `@mysten-incubation/memwal@0.1.7` is already tagged without this fix. Keep dev's `deadlineSignal` on each `signedRequest` attempt and the destroyed-client throw, and move the changelog bullets off `## 0.1.7`. That rebase does not bound the background SessionKey build; see the inline bug.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

harrymove-ctrl:
Rebased onto current `dev` at `50beaca5`:

- All merge conflicts in `packages/sdk/src/memwal.ts` and `utils.ts` resolved cleanly.
- Preserved `deadlineSignal` (WALM-648) and `destroyed` checks.
- Handled `isSealSessionExpiredResponse(res.status, raw)` and `sealSessionExpiredError` invalidation.
- All 21 tests in `packages/sdk/test/seal-session-refresh.test.mjs` pass.

Issue #591, opened 2026-08-11
[SDK/Bug] Race: memwal_recall returns stale/deleted facts for ~5–15s after forget or override (indexer lag)


Olympusxvn:
## Summary

After a memory mutation that should remove or replace a fact (`forget` / namespace forget / dashboard entry delete / logical override), an immediate `memwal_recall` in the **same agent turn** can still return the **stale** payload for roughly **5–15 seconds** while the relayer vector index catches up. At drafting speed this desynchronizes agent state (contradiction-guard reads “alive” after a “dead” supersede, or re-reads a fact the author just retired).

This is related to, but distinct from, [#433](https://github.com/MystenLabs/MemWal/issues/433) (optimistic buffer after **remember**). Here the failure mode is **post-mutation stale recall** — deleted or superseded vectors still winning semantic search until the index hard-syncs.

## Environment

- Package: `@mysten-incubation/memwal-mcp` (+ Continuity Keeper CLI using MemWal for `forget` / rewrite)
- Network: Walrus **Mainnet** (`https://relayer.memory.walrus.xyz`)
- Client: Cursor agent loop (creative drafting / Session 7 Prompt Evolution)
- Namespace under test: `session7-continuity` (plus per-entity namespaces from [continuity-keeper](https://github.com/yukitran03/continuity-keeper))

## Reproducible scenario (agent drafting speed)

1. Seed an active canon fact, e.g. `[canon:char] Minh — alive, Harbor Node regular (as of: Ch.0)`.
2. Wait until `memwal_recall` reliably returns it.
3. Mutate: supersede / forget that fact (CLI namespace forget + rewrite, or dashboard entry delete, or write a replacement state).
4. **Immediately** (same second / next tool call, no sleep) run:
   `memwal_recall(query="Minh alive OR dead", namespace=<same>)`.
5. **Actual:** stale “alive” (or other retired) hit still appears in results (often with a low cosine distance).
6. **Expected:** retired/deleted entry absent (or clearly marked non-current); only the new state surfaces.
7. After ~10–15s, re-recall → stale hit usually gone. Agents cannot afford that wait mid-scene.

## Conceptual reproduction

```typescript
// Turn N: retire obsolete canon (API surface varies: namespace forget, dashboard delete, etc.)
await memwal.forget({ namespace: "ca_voi_con::char::minh" });
// or: dashboard Security Delete for blob_id "…xyz…"

// Turn N+1: immediate recall in the same agent loop (ms later)
[...]

PR #885, opened 2026-09-08
fix(server): hide claimed-for-deletion memories from recall (WALM-592)

harrymove-ctrl:
Fixes WALM-592 — https://linear.app/mysten-labs/issue/WALM-592/bug-recall-returns-stale-facts-for-5-15s-after-forget-gh-591
Reported as https://github.com/MystenLabs/MemWal/issues/591

## `/api/forget` was not the stale-read source

The report blames "indexer lag", so I checked whether one exists. It doesn't.
`routes/admin.rs::forget` → `db.rs::delete_by_namespace` hard-`DELETE`s from
`vector_entries` and inserts the tombstones in a single CTE, clears
`remember_jobs.idempotency_key`, and commits — one transaction. Recall reads
that same table via `db.rs::search_similar`, on the same Postgres, through the
same pool. There is no read replica and no separate/async vector store: the
index *is* `vector_entries` (pgvector HNSW, `migrations/001_init.sql`). Forget
was already read-your-writes.

## The dashboard delete is

`routes/security_delete.rs::prepare_deletion` →
`security_delete_store.rs::create_batch_and_claim_*` only flips
`delete_blobs_tracking.state` to `deleting`. **`vector_entries` is untouched —
the row stays fully live and recallable.** It is deleted much later, by
`finalize_batch_deleted`, and only once the on-chain batch completes: inline
from `complete_submit` after the Sui tx confirms, or from the background
reconciler, which ticks on a **30-second** interval. Nothing anywhere filtered
`state = 'deleting'` rows out of recall — `delete_blobs_tracking` was never
referenced from `routes/recall.rs`, `search_similar`, or the engine.

That window is the reported 5–15s. It is also worse than a merely redundant
row: the Redis ciphertext cache (`engine/walrus_seal.rs`,
`memwal:blob:v1:`, 14-day default TTL) is keyed on `blob_id` and is **never
invalidated on delete**, so a surviving row can hydrate the deleted plaintext
long after the Walrus blob is gone.

## The fix

`search_similar` anti-joins `delete_blobs_tracking` and drops rows whose Walrus
Blob object is gone or going.

```sql
AND NOT EXISTS (
      SELECT 1 FROM delete_blobs_tracking t
      WHERE t.owner = v.owner
        AND t.blob_id = v.blob_id
        AND t.state IN ('deleting', 'deleted', 'deleted_external')
    )
```

The join predicate is the same `(owner, blob_id)` equality
`finalize_batch_deleted` already uses to find the rows it deletes, so the
filter is exactly as effective as the delete it front-runs.

**Which states, and why the other three are kept.** The line is "is the Blob
object gone, or going":

| state | filtered | why |
|---|---|---|
| `deleting` | yes | claimed by a live batch — the reported bug |
| `deleted` | yes | our own delete confirmed; `finalize_batch_deleted` removes the row in the same tx, so this is belt-and-braces |
| `deleted_external` | yes | `batch_get_objects` no longer returns the object at all |
| `deletable` | **no** | an ordinary live memory that merely *could* be deleted |
| `not_owner` | **no** | object is there and readable; only means the relayer can't delete it on chain |
| `expired` | **no** | see below |

`expired` is a *look-ahead*, not a lapse: `terminal_state` applies
`EXPIRY_MARGIN_EPOCHS` (default 1), so a blob is marked expired a full Walrus
epoch before its content actually goes. Recall applies no `expires_at` filter
today; adding one through this back door would silently retire live memories
early. That belongs with per-memory expiry, not with this fix.

**Why SQL and not a post-`LIMIT` filter.** Filtering after `LIMIT` would
silently shrink result sets and skew the ranker's candidate window. In SQL the
row never enters the candidate set — which also means hydration, and therefore
the un-invalidated Redis entry, is never reached for a pending-delete blob.

**No soft-delete column.** `migrations/020_read_api_followups.sql` records the
approved decision to keep deletion state out of `vector_entries` ("21
production queries read that table; a missed soft-delete filter would leak
content or over-count storage"). This respects that: it reads the
security-delete subsystem's own state, and duplicates nothing.

## Performance

**No new index needed.** `delete_blobs_tracking`'s primary key is
`(owner, blob_id)` — exactly the anti-join key. `EXPLAIN (ANALYZE, BUFFERS)`
against 4,000 rows in one namespace, 40 of them `deleting`:

```
BEFORE  Limit (actual time=0.103..0.132 rows=10)  Buffers: shared hit=215
          -> Index Scan using idx_vector_entries_embedding on vector_entries
        Execution Time: 0.148 ms

AFTER   Limit (actual time=0.104..0.143 rows=10)  Buffers: shared hit=246
          -> Nested Loop Anti Join
               -> Index Scan using idx_vector_entries_embedding on vector_entries v
               -> Index Scan using delete_blobs_tracking_pkey on delete_blobs_tracking t
                    Index Cond: ((owner = ...) AND (blob_id = v.blob_id))
        Execution Time: 0.150 ms
```

The HNSW scan still drives the plan; the anti-join adds one PK probe per
surviving candidate. No sequential scan.

## Deployment safety: why the query is selected at runtime

`delete_blobs_tracking` is defined in `migrations_legacy`, which `LegacyDb`
applies to `LEGACY_DB_URL` **only when a security-delete component is
enabled**. In a default deployment the table does not exist — confirmed
against a local stack, where the relayer database has no
`delete_blobs_tracking` at all. Referencing it unconditionally from the
hottest read in the service would turn *every* recall into an
`undefined_table` 500.

So `VectorDb` probes once (`SELECT to_regclass('delete_blobs_tracking')`) and
`search_similar` picks between two `&'static str` queries — one branch on the
hot path, no per-call `format!`. A failed probe leaves the filter off: the
failure mode is "behave exactly as today", never "break recall". `main`
re-probes right after `LegacyDb::new()`, because those legacy migrations are
what *create* the table and they run after `VectorDb::new()` — without the
second probe the first boot that enables security deletion would keep serving
pending-delete rows until the next restart.

## Residual risk and follow-ups

- **Redis blob cache is still not invalidated on delete.** This PR closes the
  path that reached it (a pending-delete row can no longer become a recall
  candidate, so its ciphertext is never fetched), but the entry itself
  survives up to its 14-day TTL. Any *future* path that leaves a live
  `vector_entries` row pointing at a dead blob would hit the same stale
  plaintext. A real invalidation hook on
  `finalize_batch_deleted` / `delete_by_namespace` is the durable fix; it needs
  a Redis handle inside `VectorDb`, which is a wider change than this bug
  warrants.
- **Override still does not supersede the old fact** (the other half of
  GH #591). `remember` writes are async and there is no UNIQUE on
  `(owner, namespace, blob_id)`, so a rewrite adds a row rather than replacing
  one and both stay recallable. Deliberately out of scope — it is a design
  problem, not a filter.
- The filter joins on exact `owner` equality, matching `finalize_batch_deleted`.
  `delete_blobs_tracking.owner` is canonicalized by the legacy insert trigger;
  a non-canonical `vector_entries.owner` would not match — but such a row is
  equally invisible to the delete itself, so the filter is never *less*
  effective than the deletion it fronts.

## Tests

Three new tests in `services/server/src/storage/db.rs`:

- `search_similar_hides_memories_claimed_by_a_pending_security_delete` —
  table-driven over all six tracking states, with two rows carrying identical
  embeddings so a dropped row can only have been dropped by the filter. Pins
  both directions: the three "gone or going" states are hidden, the three live
  ones still recall. **Verified it actually reproduces the bug**: with the
  filter forced off, it fails with the tracked blob still returned for
  `deleting` / `deleted` / `deleted_external`.
- `forget_then_immediate_recall_returns_nothing` — the ticket's acceptance
  criterion: forget, then `search_similar` with no sleep and no retry, plus a
  second namespace that must survive.
- `pending_delete_filter_matches_the_security_delete_store` — pure unit test;
  the SQL needs the state names as literals, so this pins them against
  `security_delete_store`'s constants, in both directions. A stray
  `'deletable'` in that list would make every tracked memory in a
  security-delete deployment silently unrecallable, and this catches it.

The DB-backed tests skip cleanly when `DATABASE_URL` is unset (repo pattern),
and were run for real against `pgvector/pgvector:pg17`:

```
cargo test          lib  457 passed; 0 failed; 23 ignored
                    bin  693 passed; 0 failed; 37 ignored
```

`cargo fmt` clean on both changed files (`memory_read.rs` has pre-existing
formatting drift on `dev`; deliberately left alone rather than dragged into
this diff). `cargo clippy --all-targets` exits 0. It cannot run with
`-D warnings` — `.github/workflows/test.yml` soft-fails clippy for exactly
this reason — so the delta: **bin target unchanged (25 → 25, bin test 28 → 28)**;
lib target 585 → 588, three `dead_code` notices for the new SQL constants and
field, which the lib target reports because `lib.rs` does not compile `routes`
and so cannot see `search_similar`'s caller. `db.rs` already emits 18 warnings
of that exact class, including `constant TOMBSTONE_RETENTION is never used`.

### DIFF
--- a/services/server/src/storage/db.rs
+++ b/services/server/src/storage/db.rs
@@ -22,45 +22,19 @@ const SEARCH_SIMILAR_SQL: &str =
              ORDER BY embedding <=> $1
              LIMIT $4";
 
-/// The same query with the memories the security-delete subsystem has
-/// already given up on filtered out (WALM-592 / GH #591).
+/// `SEARCH_SIMILAR_SQL` with the memories the security-delete subsystem has
+/// already claimed hidden (WALM-592).
 ///
-/// `POST /api/forget` hard-`DELETE`s its rows inside one transaction, so it
-/// never had a stale-read window. The dashboard delete does: `prepare_deletion`
-/// only flips `delete_blobs_tracking.state` to `deleting`, and the matching
-/// `vector_entries` rows survive until the Sui transaction confirms and
-/// `finalize_batch_deleted` runs — inline on submit, or up to one 30s
-/// reconciler tick later. That gap is the reported 5–15s of stale recall
-/// hits, and it is worse than a merely redundant row: the Redis ciphertext
-/// cache (`memwal:blob:v1:`, 14-day TTL) is never invalidated on delete, so
-/// a surviving row can hydrate the deleted plaintext long after the Walrus
-/// blob is gone. Filtering in SQL keeps the row out of the candidate set
-/// entirely, so hydration — and therefore that cache — is never reached for it.
+/// Anti-join rather than a soft-delete column: migration 020 keeps deletion
+/// state out of `vector_entries` on purpose. `delete_blobs_tracking`'s primary
+/// key is `(owner, blob_id)` — the anti-join key — so the HNSW scan still
+/// drives the plan and no new index is needed.
 ///
-/// This is an anti-join, not a soft-delete column: migration 020 records the
-/// deliberate decision to keep deletion state out of `vector_entries`
-/// ("21 production queries read that table; a missed soft-delete filter
-/// would leak content or over-count storage").
-///
-/// Cost: `delete_blobs_tracking`'s primary key is `(owner, blob_id)` — exactly
-/// the anti-join key — so each surviving candidate costs one index probe and
-/// the HNSW scan still drives the plan. No new index is required.
-///
-/// The line between the six tracking states is "is the Walrus Blob object
-/// gone, or going": `deleting` (claimed by a live batch — the reported bug),
-/// `deleted` (our own delete confirmed on chain; `finalize_batch_deleted`
-/// removes the row in the same transaction, so this is belt-and-braces) and
-/// `deleted_external` (`batch_get_objects` no longer returns the object at
-/// all). The other three are left alone because the object is still there and
-/// still readable: `deletable` is an ordinary live memory, `not_owner` only
-/// means the relayer cannot delete it on chain, and `expired` is a
-/// look-ahead — `terminal_state` applies `EXPIRY_MARGIN_EPOCHS` (default 1)
-/// so a blob is marked expired a full Walrus epoch before its content
-/// actually lapses. Recall applies no `expires_at` filter today, and adding
-/// one here would silently retire live memories early; that belongs with
-/// per-memory expiry, not with this fix.
-///
-/// Pinned against the store's own constants by
+/// Only the three states where the Walrus Blob object is gone or going are
+/// filtered. `expired` in particular must stay visible: `terminal_state`
+/// applies `EXPIRY_MARGIN_EPOCHS` ahead of the real lapse, so filtering it
+/// would retire live memories early. The state list is pinned against the
+/// store's own constants by
 /// `pending_delete_filter_matches_the_security_delete_store`.
 const SEARCH_SIMILAR_EXCLUDING_PENDING_DELETES_SQL: &str =
     "SELECT v.blob_id, (v.embedding <=> $1)::float8 AS distance, v.created_at, v.importance
@@ -1271,6 +1245,31 @@ mod tests {
         Some(db)
     }
 
+    /// A probe that errors must not turn off a filter a previous probe
+    /// enabled. `main` re-probes after `LegacyDb::new()`; a transient failure
+    /// there used to disable the filter for the life of the process.
+    #[tokio::test]
+    async fn a_failed_probe_keeps_the_pending_delete_filter_it_already_had() {
+        let Some(db) = security_delete_test_db().await else {
+            eprintln!("skipping DB integration test: DATABASE_URL is not configured");
+            return;
+        };
+        assert!(
+            db.pending_delete_filter_enabled(),
+            "security_delete_test_db must hand back a db with the filter on"
+        );
+
+        // Closing the pool makes the next probe fail the way a transient
+        // connection error would.
+        db.pool().close().await;
+
+        assert!(
+            db.refresh_pending_delete_filter().await,
+            "a failed probe must keep the filter that was already enabled"
+        );
+        assert!(db.pending_delete_filter_enabled());
+    }
+
     /// WALM-592 / GH #591. The dashboard delete claims a memory by flipping
     /// its tracking row to `deleting`; the `vector_entries` row then survives
     /// until the Sui transaction confirms and `finalize_batch_deleted` runs
@@ -1869,10 +1868,11 @@ impl VectorDb {
     /// first boot that turns security deletion on would keep serving
     /// pending-delete rows until the next restart.
     ///
-    /// A failed probe leaves the filter off. Recall is the hottest read path
-    /// in the service and referencing a missing table would turn every
-    /// recall into an `undefined_table` 500, so the failure mode here is
-    /// "behave exactly as before", never "break recall".
+    /// The flag starts `false`, so a table that is genuinely absent can never
+    /// turn recall into an `undefined_table` 500. A probe that *errors* is not
+    /// evidence of absence, though: it leaves the current setting alone, or a
+    /// transient failure on the `main` re-probe would disable a working filter
+    /// for the life of the process.
     pub async fn refresh_pending_delete_filter(&self) -> bool {
         let present: bool =
             match sqlx::query_scalar("SELECT to_regclass('delete_blobs_tracking') IS NOT NULL")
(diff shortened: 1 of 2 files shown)


services/server/src/storage/db.rs
```diff
@@ -1,21 +1,96 @@
...
+/// the `ORDER BY embedding <=> $1 LIMIT $4` shape is what lets the planner
+/// drive off the HNSW index (`idx_vector_entries_embedding`, migration 001).
+const SEARCH_SIMILAR_SQL: &str =
+    "SELECT blob_id, (embedding <=> $1)::float8 AS distance, created_at, importance
+             FROM vector_entries
+             WHERE owner = $2 AND namespace = $3
+             ORDER BY embedding <=> $1
+             LIMIT $4";
+
+/// The same query with the memories the security-delete subsystem has
```
ducnmm: **[suggestion]** `SEARCH_SIMILAR_EXCLUDING_PENDING_DELETES_SQL` is preceded by ~40 lines of ticket narrative (WALM-592 / GH #591, Redis `memwal:blob:v1:` TTL, migration 020 quote, EXPLAIN rationale, per-state essay). That belongs in the PR body; in-tree it will rot and is louder than the surrounding `db.rs` comments. The load-bearing bits (which states, why not `expired`, anti-join on the tracking PK, fail-open probe) are already repeated on `refresh_pending_delete_filter` and the tests.

**Suggestion:** Keep a short comment: anti-join `delete_blobs_tracking` for `deleting`/`deleted`/`deleted_external` so claimed dashboard deletes never enter the HNSW candidate set; `expired` is a look-ahead and must stay visible. Point at `pending_delete_filter_matches_the_security_delete_store` for the state list.

services/server/src/storage/db.rs
```diff
@@ -1491,7 +1827,60 @@ impl VectorDb {
...
+                .fetch_one(&self.pool)
+                .await
+            {
+                Ok(present) => present,
+                Err(e) => {
+                    tracing::warn!(
+                        "could not probe for delete_blobs_tracking; recall will not \
+                         filter pending security deletes: {}",
+                        e
+                    );
```
ducnmm, line 1890: **[suggestion]** `refresh_pending_delete_filter` maps any probe `Err` to `false` and `swap`s that into `pending_delete_filter`. `VectorDb::new()` already probes successfully on a subsequent boot (table present → filter on); `main` then re-probes after `LegacyDb::new()`. A transient error on that second `SELECT to_regclass(...)` disables a working filter for the life of the process — the opposite of the first-boot enablement the re-probe exists for. Fail-open on *first* detection is right; fail-open that clobbers a prior `true` is not.

**Suggestion:** On `Err`, leave the previous value (and warn). Only `swap(false)` when the probe succeeds and reports the relation missing. Optionally skip the `main` re-probe when the flag is already true.

ducnmm (commented):
## Summary

The change matches WALM-592: `/api/forget` was already a single-transaction hard delete (now pinned by `forget_then_immediate_recall_returns_nothing`), and dashboard delete is hidden from `search_similar` as soon as `delete_blobs_tracking` is `deleting` / `deleted` / `deleted_external`. Filtering in SQL (not post-`LIMIT`), leaving the three still-readable states alone, fail-open gating so a missing table cannot 500 recall, and the six-state plus store-constant tripwire tests are the right shape. Two non-blocking notes: a design-dump comment on the hot query, and the re-probe treating a failed `to_regclass` as “table absent,” which can turn a working filter off.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

Olympusxvn:
Thanks for the detailed fix, @harrymove-ctrl! The explanation of the root cause (WALM-592) and the filter approach are really clear. Appreciate the team's work on this. Looking forward to seeing it merged!

harrymove-ctrl:
@ducnmm ready for another look when you have a moment — head `c47a10eb` addresses both suggestions from the `5f0cfd0c` review and has not been re-reviewed yet.

- **suggestion (probe clobbers a working filter):** on `Err`, `refresh_pending_delete_filter` now keeps the current setting and warns; it only swaps in `false` when the probe succeeds and reports the relation missing. Fail-open on first detection still holds, since the flag starts `false`. Regression test closes the pool to force a probe error and asserts the filter stays on. Note it is a DB-integration test, so it skips locally without `DATABASE_URL` — CI is the real check.
- **suggestion (ticket narrative on the hot query):** trimmed to the anti-join key and why `expired` must stay visible, pointing at `pending_delete_filter_matches_the_security_delete_store` for the state list.

ducnmm (approved):
LGTM, probe error handling addressed; all checks pass.

PR #887, opened 2026-09-08
fix(server): return each blob once in recall (WALM-594)

### LINKED ISSUE #694: memwal_recall returns duplicate records for the same blob
### Description
When calling `memwal_recall` on a namespace that contains a recently written record, the same blob can appear multiple times in the results with identical scores.

### Reproduction
1. Write a record using `memwal_remember`:
   `memwal_remember(text='...', namespace='buildmem/my-project/state')`  
2. Immediately call`memwal_recall` on the same namespace.
3. The same record appears 2+ times in the result list with identical distance scores.

### Observed Output
`json
1. [score=0.453] { ...same record... }
2. [score=0.453] { ...same record... }
`  

### Expected Output
Each unique blob should appear at most once in recall results.

### Impact
Duplicates inflate the result list, consume the `limit` budget, and cause downstream logic (e.g. ''take the newest'') to behave incorrectly.

---
*Reported during Walrus Session 7 - Prompt Evolution*

HoangDucBach:
## Ticket

WALM-594 — https://linear.app/mysten-labs/issue/WALM-594
GH #694 — https://github.com/MystenLabs/MemWal/issues/694

## What changed?

- `search_similar` collapses hits that share a `blob_id`, keeping the closest row.
- `/api/recall`, `recall/manual`, and analyze all go through that query, so each blob appears at most once in a result set.
- Unit tests pin first-hit-wins; a DB test inserts two rows for one blob plus a distinct blob and asserts unique ids.

## Why is this needed?

`vector_entries` has no UNIQUE on blob_id. A remember retry after a successful Walrus upload, or restore re-indexing a blob that remember already wrote, inserts a second row with a new `id`. Recall then returned the same blob twice with identical scores and burned `limit` on one memory.

## Scope

Relayer vector search (`search_similar`) and its tests.

### Out of scope

- UNIQUE constraint / upsert on write (duplicate index rows can still exist; recall just will not return both)
- MCP text collapse (`collapseDuplicates`) — already folds identical plaintext

## Test plan

- [x] `cargo test --bins unique_hits_by_blob_id` (2 passed)
- [x] `cargo test --bins search_similar_returns_each_blob_id_once`
- [ ] Remember then immediate recall on a namespace: the new blob appears once

Fixes #694

Made with [Cursor](https://cursor.com)

### DIFF
--- a/services/server/src/routes/recall.rs
+++ b/services/server/src/routes/recall.rs
@@ -84,12 +84,10 @@ async fn generate_recall_embedding_cached(
 /// `text` (the ranker never reads text; manual recall never decrypts).
 ///
 /// The ranked output is mapped back to the original `SearchHit`s **by
-/// original index, not by `blob_id`**. `search_similar` already collapses
-/// duplicate blob_ids (WALM-594), but this remap still must not key on
-/// `blob_id`: the ranker only reads distance/recency/importance, and we
-/// stash each hit's input index in the throwaway `HydratedMemory`'s
-/// `blob_id` slot so a duplicate-bearing input cannot silently drop or
-/// reorder hits. Indices are unique, so `results.len() == hits.len()`.
+/// original index, not by `blob_id`**. The ranker treats `blob_id` as
+/// opaque carry-through (it scores only distance/recency/importance), so
+/// we stash each hit's input index in that slot. Indices are unique, so a
+/// blob_id-keyed round-trip cannot drop hits and `results.len() == hits.len()`.
 ///
 /// At default weights `rank()` short-circuits, so the input (cosine) order
 /// is returned unchanged.
@@ -529,11 +527,8 @@ mod tests {
         assert_eq!(ranked[0].created_at, t_now() - chrono::Duration::days(7));
     }
 
-    /// Ranker remap is by input index, not `blob_id`. `search_similar`
-    /// collapses duplicate blob_ids (WALM-594); this still pins that a
-    /// blob_id-keyed round-trip would drop/reorder if duplicates ever
-    /// reached the ranker. Tested at BOTH default (short-circuit) and
-    /// active weights.
+    /// Remap is by input index, not blob_id, so two hits sharing a blob_id
+    /// are both kept. Tested at default (short-circuit) and active weights.
     #[test]
     fn manual_ranking_keeps_duplicate_blob_ids() {
         use crate::services::extractor::{IMPORTANCE_TRIVIAL, IMPORTANCE_VITAL};
--- a/services/server/src/routes/recall.rs
+++ b/services/server/src/routes/recall.rs
@@ -84,18 +84,12 @@ async fn generate_recall_embedding_cached(
 /// `text` (the ranker never reads text; manual recall never decrypts).
 ///
 /// The ranked output is mapped back to the original `SearchHit`s **by
-/// original index, not by `blob_id`**. `blob_id` is not unique
-/// (`vector_entries` has no UNIQUE constraint on it and `search_similar`
-/// does not `SELECT DISTINCT`, so `restore` can produce — and a query can
-/// return — multiple hits with the same blob_id). Keying the round-trip on
-/// blob_id would collapse those duplicates, silently dropping hits and
-/// reordering them — re-introducing the very manual-vs-non-manual
-/// divergence this fix removes (the non-manual paths keep duplicates 1:1).
-/// So we stash each hit's input index in the throwaway `HydratedMemory`'s
-/// `blob_id` slot — the ranker treats that field as opaque carry-through
-/// (it scores only distance/recency/importance) — and reorder the original
-/// `Vec<SearchHit>` by the ranked index sequence. Indices are unique, so no
-/// hit is dropped and `results.len() == hits.len()` always.
+/// original index, not by `blob_id`**. `search_similar` already collapses
+/// duplicate blob_ids (WALM-594), but this remap still must not key on
+/// `blob_id`: the ranker only reads distance/recency/importance, and we
+/// stash each hit's input index in the throwaway `HydratedMemory`'s
+/// `blob_id` slot so a duplicate-bearing input cannot silently drop or
+/// reorder hits. Indices are unique, so `results.len() == hits.len()`.
 ///
 /// At default weights `rank()` short-circuits, so the input (cosine) order
 /// is returned unchanged.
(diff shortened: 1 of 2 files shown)


services/server/src/storage/db.rs
```diff
@@ -1744,7 +1866,7 @@ impl VectorDb {
             })
             .collect();
 
-        Ok(results)
+        Ok(unique_hits_by_blob_id(results))
```
ducnmm: **[suggestion]** `unique_hits_by_blob_id` runs after `ORDER BY embedding <=> $1 LIMIT $4`, so duplicate index rows still consume the SQL `limit` and are only dropped afterward. A namespace with two `vector_entries` rows for one blob (remember retry / restore) plus other distinct blobs can therefore return fewer than `limit` unique hits even when more unique blobs exist — the GH #694 “consume the limit budget” impact. The helper’s own comment frames that as the problem being solved, but the caller-visible `limit` is not refilled. A `DISTINCT ON (blob_id)` around the same inner `LIMIT` has the same hole and may also inhibit HNSW.

**Suggestion:** Keep first-wins unique (this is the right ANN-friendly place). If filling `limit` still matters, over-fetch (e.g. `LIMIT $limit` plus the count of dropped dups, or a small slack) then unique and truncate; add a DB test with `limit=2`, two rows for blob A and one for blob B, so crowding cannot silently regress. Do not require a UNIQUE/upsert write-path fix here (WALM-610).

services/server/src/routes/recall.rs
```diff
@@ -84,18 +84,12 @@ async fn generate_recall_embedding_cached(
...
-/// return — multiple hits with the same blob_id). Keying the round-trip on
-/// blob_id would collapse those duplicates, silently dropping hits and
-/// reordering them — re-introducing the very manual-vs-non-manual
-/// divergence this fix removes (the non-manual paths keep duplicates 1:1).
-/// So we stash each hit's input index in the throwaway `HydratedMemory`'s
-/// `blob_id` slot — the ranker treats that field as opaque carry-through
-/// (it scores only distance/recency/importance) — and reorder the original
-/// `Vec<SearchHit>` by the ranked index sequence. Indices are unique, so no
-/// hit is dropped and `results.len() == hits.len()` always.
+/// original index, not by `blob_id`**. `search_similar` already collapses
```
ducnmm: **[suggestion]** The `rank_search_hits` docs (and the sibling test comment at line 532) now embed WALM-594 and restate the index-stash implementation (“search_similar already collapses… stash each hit’s input index…”). Remap-by-index is an independent invariant of the throwaway `HydratedMemory`; tying it to this ticket makes the comment a change narrative rather than a concise why.

**Suggestion:** Trim to why the remap keys on input index (the ranker treats `blob_id` as opaque carry-through; indices stay unique so a blob_id-keyed round-trip cannot drop hits). Drop the WALM-594 recap from the ranker docs and from `manual_ranking_keeps_duplicate_blob_ids`.

ducnmm (commented):
## Summary

`search_similar` now collapses hits that share a `blob_id`, keeping the first row in cosine order. That matches WALM-594 / GH #694: recall, recall/manual, ask, analyze, and admin all go through this query, so each Walrus blob appears at most once. The helper and the two-row DB test pin the behavior; WALM-610 (override leaving both versions live) is correctly left alone. The remaining gaps are that unique-after-`LIMIT` can still return a short page, and the recall.rs comment edits still narrate ticket history.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

services/server/src/routes/recall.rs
HoangDucBach: fixed with sugg

services/server/src/storage/db.rs
HoangDucBach: fixed with sugg

services/server/src/storage/db.rs
```diff
@@ -1692,14 +1892,52 @@ impl VectorDb {
...
+            let mut unique = unique_hits_by_blob_id(rows);
+            if unique.len() >= limit {
+                unique.truncate(limit);
+                return Ok(unique);
+            }
+            if fetched < fetch_limit {
+                return Ok(unique);
+            }
+            let dropped = fetched - unique.len();
+            let next = limit.saturating_add(dropped);
```
ducnmm: **[suggestion]** `next = limit.saturating_add(dropped)` is `fetch_limit + (limit - unique.len())` on a full page, so each extra ANN query only asks for as many new rows as unique blobs still needed. That is enough for the new test (two rows of A crowding `limit=2` → one extra `LIMIT 3`). It crawls when the duplicate factor stays high: with uniform 2× duplication (the WALM-594 write shape) and `limit=100` (the HTTP cap on recall/ask), `fetch_limit` walks 100 → 150 → 175 → … and needs ~8 extra sequential HNSW queries to reach 200 rows. One heavily duplicated nearest blob is worse: `unique` stays 1, `fetch_limit` grows by `limit-1` per iteration with no cap, and `loop` only stops when Postgres returns a short page (whole namespace scanned). Analyze wraps this in a 300ms timeout, so a slow refill also drops pre-extraction context. `if next <= fetch_limit` is unreachable on a full page (next always grows by the shortfall), so it is not a bound.

**Suggestion:** Grow the window by the observed duplicate factor (e.g. `fetch_limit + dropped`, or double `fetch_limit`) and cap it (max iterations or `fetch_limit <= limit * k`). Keep first-wins unique after ANN `LIMIT`. The existing crowding test can stay; a second case with three copies of A at `limit=2` would pin that the first widen is not assumed to be enough.

ducnmm (commented):
## Summary

The follow-up correctly fixes the short-page hole from 69f1e600: `search_similar` now unique-by-`blob_id` (closest row wins) and widens the ANN `LIMIT` until `limit` distinct blobs are filled or the namespace is exhausted. That matches WALM-594 / GH #694, the new `limit=2` crowding test pins the requested case, and the `rank_search_hits` comments no longer narrate the ticket. Results look right for the typical “one blob, two index rows” shape; the remaining risk is the refill loop itself — `next = limit + dropped` only grows by the unique shortfall and has no iteration / `fetch_limit` cap.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 0

services/server/src/storage/db.rs
HoangDucBach: fixed with sugg

ducnmm (approved):
LGTM, capped 4x over-fetch addresses the refill loop suggestion; all checks pass.

PR #888, opened 2026-09-08
feat(app): paginate the dashboard namespaces list (WALM-607)

harrymove-ctrl:
Closes [WALM-607](https://linear.app/mysten-labs/issue/WALM-607/fe-paginate-the-dashboard-namespaces-list).

## Problem

The dashboard Namespaces card already talked to the cursor-paginated relayer endpoint, but the UI did not paginate. `fetchNamespaces` walked `has_more` / `next_cursor` internally — up to 20 pages of 100 — concatenated everything into one table, and past that cap only appended `(list truncated)` to the subtitle. Namespaces beyond ~2000 were unreachable.

## What changed

One page at a time, driven by a cursor stack:

- `namespacesCursors[i]` holds the `updated_after` that opens page *i*, so **Previous replays a stored cursor** instead of re-walking from the start.
- **Next is gated on `has_more`, never on `rows.length`.** The relayer clamps `limit` (default 100, max 500), so a full page is not evidence of more data.
- **Refresh and page-size changes truncate the cursor stack** back to `[undefined]`, so no stale cursor survives into a new walk.
- Dropped the 20-page cap and `(list truncated)`.
- Per-row `/playground?namespace=` link is unchanged.

### Why not numbered pages

The endpoint is *cursor*-paginated — there is no total and no offset — so the [Astryx](https://astryx.atmeta.com/components/Pagination) numbered-pages variant does not apply (it wants `totalItems` *when the total is known*). This uses the count-readout + Prev/Next shape with a page-size selector, which Astryx's own "Page Size Selector" example places in a footer below the table. The footer stays hidden when everything fits on one page.

The second commit fixes a trap I introduced in the first: hiding the footer whenever the list fit one page meant that raising the page size to fit the whole list removed the only control that could lower it again. The footer now also stays while the page size differs from the default.

## Verification

Walked the real dev relayer with 24 seeded namespaces at `limit=20`:

```
page 1: 20 rows  has_more=true   next_cursor=eyJ1cGRhdGVkX2F0…
page 2:  4 rows  has_more=false  next_cursor=eyJ1cGRhdGVkX2F0…   <- still non-null
no duplicates across pages
```

Worth noting: **on the last page the relayer returns `has_more=false` with a non-null `next_cursor`.** Keying Next off the cursor — the more intuitive read — would leave the button enabled forever and fetch empty pages. That is exactly what the acceptance criterion about `has_more` guards against.

- 9 new tests in `DashboardNamespaces.test.tsx` cover each acceptance criterion: cursor sent as `updated_after`, stopping at `has_more=false` despite a full page, Previous, Refresh resetting the cursor, page-size restart, the page-size lockout regression, empty/error states, playground links.
- Full app suite 100 passed (13 files); `tsc -b` and `eslint` clean.
- Rendered the card's markup against the live stylesheet and checked desktop + mobile layout.

## Notes for the reviewer

- The card hides the table while a page loads (pre-existing behaviour), so paging briefly blanks the rows. Buttons disable and the subtitle reads "Loading…", so there is feedback — I kept the card's existing convention rather than inventing a new one. Happy to change it if you'd rather the rows stayed put.
- The error string renders twice (card subtitle *and* the inline paragraph). Pre-existing; left alone as out of scope, but the test uses `findAllByText` because of it.
- Separately, `apps/app/.env.example` pairs `relayer.dev.memwal.ai` with a package ID that relayer rejects, so a fresh checkout cannot register a delegate key (`Transaction kind is not permitted for sponsorship`). Not touched here — tracked separately.

## Out of scope

Playground namespace picker, Python SDK / MCP `listNamespaces`, and WALM-185.

### DIFF
--- a/apps/app/src/index.css
+++ b/apps/app/src/index.css
@@ -8505,7 +8505,6 @@ h1, h2, h3 {
   border-bottom: 0;
 }
 
-/* Namespaces pagination footer — sits below the table it controls. */
 .dash-page .dashboard-pagination {
   display: flex;
   align-items: center;
--- a/apps/app/src/index.css
+++ b/apps/app/src/index.css
@@ -8505,6 +8505,80 @@ h1, h2, h3 {
   border-bottom: 0;
 }
 
+/* Namespaces pagination footer — sits below the table it controls. */
+.dash-page .dashboard-pagination {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  flex-wrap: wrap;
+  gap: 16px;
+  margin-top: 16px;
+}
+
+.dash-page .dashboard-pagination-size {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+}
+
+.dash-page .dashboard-pagination-size label,
+.dash-page .dashboard-pagination-status {
+  color: #b8bbbe;
+  font-size: 14px;
+  line-height: 1.35;
+}
+
+.dash-page .dashboard-pagination-select {
+  padding: 6px 10px;
+  border: 1px solid rgba(250, 248, 245, 0.12);
+  border-radius: 10px;
+  background: rgba(255, 255, 255, 0.06);
+  color: #faf8f5;
+  font-size: 14px;
+  line-height: 1.35;
+  cursor: pointer;
+}
+
+.dash-page .dashboard-pagination-select:disabled {
+  opacity: 0.35;
+  cursor: not-allowed;
+}
+
+.dash-page .dashboard-pagination-end {
+  display: flex;
+  align-items: center;
+  gap: 16px;
+}
+
+.dash-page .dashboard-pagination-controls {
+  display: flex;
+  gap: 8px;
+}
+
+.dash-page .dashboard-pagination-btn {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  width: 36px;
+  height: 36px;
+  padding: 0;
+  border: 1px solid rgba(250, 248, 245, 0.12);
+  border-radius: 10px;
+  background: transparent;
+  color: #b8bbbe;
+  cursor: pointer;
+}
+
+.dash-page .dashboard-pagination-btn:hover:not(:disabled) {
+  border-color: #faf8f5;
+  color: #faf8f5;
+}
+
+.dash-page .dashboard-pagination-btn:disabled {
+  opacity: 0.35;
+  cursor: not-allowed;
+}
+
 .dash-page .dashboard-key-table-select {
   width: 62px;
 }
@@ -9133,6 +9207,16 @@ h1, h2, h3 {
 
   .dash-page .dashboard-key-table tbody tr:last-child {
     margin-bottom: 0;
+  }
+
+  .dash-page .dashboard-pagination {
+    align-items: stretch;
+    justify-content: flex-start;
+  }
+
+  .dash-page .dashboard-pagination-end {
+    justify-content: space-between;
+    width: 100%;
   }
 
   .dash-page .dashboard-key-table td {
--- a/apps/app/src/pages/Dashboard.tsx
+++ b/apps/app/src/pages/Dashboard.tsx
@@ -14,7 +14,7 @@ import { useSponsoredTransaction } from '../hooks/useSponsoredTransaction'
 import { generateDelegateKey } from '@mysten-incubation/memwal/account'
 import type { WalletSigner } from '@mysten-incubation/memwal/manual'
 import { Link, useNavigate } from 'react-router-dom'
-import { TriangleAlert, Info, Copy, Eye, EyeOff, Trash2, RefreshCw, Plus, LogOut, Github, MessageCircle } from 'lucide-react'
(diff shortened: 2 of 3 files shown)


harrymove-ctrl:
Pushed `cf21adbe` — default page size **20 → 15**, per Henry's sign-off on Slack (20 rows made the card long enough to need scrolling before the footer came into view).

The default has to be present in the options list or the `<select>` renders with no matching option, so the choices are now **15 / 25 / 50 / 100**.

Re-verified against the dev relayer at `limit=15` (27 namespaces on that account):

```
page 1: 15 rows  has_more=true   next_cursor=eyJ1cGRhdGVkX2F0…
page 2: 12 rows  has_more=false  next_cursor=eyJ1cGRhdGVkX2F0…
no duplicates across pages
```

Tests updated for the new default (the range assertions are page-size dependent): 9 namespace tests pass, full app suite 100 passed, `tsc -b` and `eslint` clean.

apps/app/src/pages/Dashboard.tsx
```diff
@@ -1121,12 +1142,14 @@ const result = await generateText({
                                     ? namespacesError
                                     : namespaces.length === 0
                                         ? 'No indexed namespaces yet (or none the relayer can see for this account)'
```
ducnmm: **[suggestion]** The subtitle still treats any empty `namespaces` array as “no indexed namespaces yet”, and the table is gated on `namespaces.length > 0`. That is wrong once the user has already paged forward. The relayer documents that a continuation page can come back with zero rows when remaining namespaces race past `snapshot_at` (`has_more: false`, but this is not a fresh empty account). The footer already special-cases that UI (`Page ${namespacesPage + 1}` when `namespaces.length === 0`), so the empty-page path is expected, but the subtitle/table still show the first-load empty state. `SecurityDeleteSection` in this same app steps back when a continuation page is empty.

**Suggestion:** If `page > 0` and the response has no rows, either step back and land on the previous page (same pattern as the security-delete pager) or keep the pagination chrome and use copy like “this page is empty” instead of the first-load empty string. Do not take the `namespaces.length === 0` branch while `namespacesPage > 0`.

apps/app/src/pages/Dashboard.tsx
```diff
@@ -462,61 +467,77 @@ export default function Dashboard({
...
             console.error('Failed to list namespaces:', err)
             setNamespacesError('Could not load namespace counts from the relayer.')
         } finally {
             if (gen === namespacesFetchGen.current) setNamespacesLoading(false)
         }
-    }, [delegateKey, effectiveAccountObjectId, address])
+    }, [delegateKey, effectiveAccountObjectId, address, namespacesPageSize])
 
-    useEffect(() => {
+    // Restart the walk at the first page with no leftover cursor. Also used by Refresh.
```
ducnmm: **[suggestion]** These comments restate what the code already does: `refreshNamespaces` resets the cursor stack and is the Refresh handler; the `useEffect` on `refreshNamespaces` restarts when that callback’s identity changes (account, delegate key, page size). They narrate the change rather than a non-obvious invariant (unlike the `has_more` / clamp comment, which is worth keeping).

**Suggestion:** Drop the `refreshNamespaces` and `useEffect` comments, or replace them with a one-liner only if there is a constraint that is not obvious from the identifiers.

apps/app/src/index.css
```diff
@@ -8505,6 +8505,80 @@ h1, h2, h3 {
   border-bottom: 0;
 }
 
+/* Namespaces pagination footer — sits below the table it controls. */
```
ducnmm: **[suggestion]** `/* Namespaces pagination footer — sits below the table it controls. */` restates the markup/CSS placement (the nav is already a sibling after the table).

**Suggestion:** Remove the comment, or keep a short note only if there is a non-obvious layout constraint (e.g. why it must live outside `.dashboard-key-table-wrap`).

ducnmm (commented):
## Summary

The Namespaces card now fetches one cursor page at a time instead of concatenating up to 20×100 rows. Next is correctly gated on `has_more` (not page length or a possibly non-null terminal `next_cursor`), Previous replays a stored cursor, and Refresh / page-size changes drop the cursor stack. That matches the relayer contract, and the new tests pin the important cases. The remaining gaps are empty continuation pages falling through to the first-load empty copy, and a handful of comments that restate the code rather than a non-obvious invariant.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 0

harrymove-ctrl:
@ducnmm ready for another look when you have a moment — head `feebf298` addresses all three suggestions from the `cf21adbe` review and has not been re-reviewed yet.

- **suggestion (empty continuation page):** took the step-back option, matching `SecurityDeleteSection`. `fetchNamespacesPage` drops the cursor that opened an empty page and lands on a real one, and the subtitle no longer takes the first-load empty branch while `namespacesPage > 0`. New test covers it.
- **suggestions (narrating comments):** dropped the `refreshNamespaces` / `useEffect` comments and the CSS one. The `has_more` / clamp comment stays, since that constraint is not obvious from the identifiers.

ducnmm (approved):
LGTM, addresses empty continuation pages and suggestions from previous review; all checks pass.

PR #889, opened 2026-09-08
fix(relayer): Slack and write_ready when Postgres storage is exhausted (WALM-612)

ducnmm:
## Ticket

WALM-612 — https://linear.app/mysten-labs/issue/WALM-612/prod-writes-hit-neon-schema-only-3gb-cap-no-slack-alert-write-ready

## What changed?

- Slack `ALERT_TO_SLACK` fires (deduped 30m, one-per-network) when Postgres cannot extend files / hits project size limit — on `insert_vector` (`vector_entries` + HNSW), on `/api/remember` and `/api/remember/bulk` `remember_jobs` INSERT, and on Apalis wallet-job enqueue (`apalis.jobs`). Slack POST from `insert_vector` runs after the transaction drops.
- `GET /health` `write_ready` is sidecar health AND Neon cluster size. Prefer `public.pg_cluster_size()` vs `neon.max_cluster_size`. If that function is missing (no `neon` extension), fall back to `sum(pg_database_size)` against the same GUC (warn once). Self-host without the GUC is unchanged. Probe errors fail open at warn; timeouts fail open at warn with a 1s budget (cached 2s). HTTP 200 is unchanged; Slack send failure does not fail the write.

## Why is this needed?

Prod writes failed for days at the Neon 3GB cap. `/health` stayed `write_ready: true` (sidecar-only) and Slack only covered Walrus upload/wallet, so nothing alerted.

## Scope

Alert + health signal for Postgres storage exhaustion.

### Out of scope

DB cutover (already done). Alerting on every write failure (Sui 429, user quota, Walrus timeout). `CREATE EXTENSION neon` (breaks local/CI vanilla Postgres).

## How was this tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] End-to-end tests
- [ ] Manual testing
- [ ] Not applicable

Commands: `cargo test --bin memwal-server -- postgres_can_accept neon_max_cluster health_response_serializes postgres_storage missing_pg_cluster_size postgres_write_ready_probe_timeout`

1. Matcher tests include the prod error `could not extend file because project size limit (3072 MB) has been exceeded`, plus wrapped `remember_jobs` INSERT / enqueue messages.
2. `write_ready` probe uses `public.pg_cluster_size()` when it exists; SQLSTATE `42883` / missing function falls back to `sum(pg_database_size)` instead of fail-open.
3. Probe timeout is 1s and the timeout arm logs at `warn`.
4. Slack payload says this is a database disk/project size cap, not a user quota.

## Risks and dependencies

`write_ready` meaning is slightly tighter (false when Neon cluster is at cap). The `sum(pg_database_size)` fallback can undercount vs true cluster size (WAL/history). Probe errors and timeouts still fail open so CI `wait-for-relayer` is not blocked. `/health` still returns 200.

## Author checklist

- [ ] CI is green.


jessiemongeon1:
## Style Guide Audit

harrymove-ctrl (changes requested):
Sound approach and well-tested — all 6 tests pass locally. Two findings I'd like addressed before merge, plus some smaller ones.

**Verified in the PR's favour first:** using `pg_cluster_size()` rather than `pg_database_size` is exactly right. Neon's `neon_extend()` does `current_size >= (uint64) max_cluster_size * 1024 * 1024` against `GetNeonCurrentClusterSize()`, which is what `pg_cluster_size()` exposes — so the probe compares the same two quantities Neon does, and the 1MB headroom trips strictly before Neon's own error. The MB unit assumption is correct.

## Blocking

### 1. The alert hooks miss the write that actually failed in the incident

Per the ticket, `remember()` returned **202** and the job *then* went `failed`. The 202 is emitted after the `remember_jobs` INSERT — so the write instrumented at `services/server/src/routes/remember.rs:923` **succeeded** during the outage this PR is meant to catch.

The heavy, extension-prone write is `insert_vector` inside `insert_vector_and_mark_remember_done` (`services/server/src/jobs.rs:875`) — `vector_entries` ~2002MB + HNSW ~861MB per the ticket, versus `remember_jobs` at ~137MB. `storage_reservations` churns (insert then delete), so it can often reuse a free page and commit without an `smgrextend`, while `vector_entries` is append-heavy and extends constantly. That is very likely why the reservation probe succeeded during triage while writes were still failing.

That path has no hook. The message is already built there:

```rust
let msg = format!("insert_vector failed: {}", e);
```

so it's a one-liner once `maybe_alert_postgres_storage_exhausted` takes `&str` instead of `&AppError`.

### 2. `pg_cluster_size()` requires `CREATE EXTENSION neon`, and the probe silently no-ops without it

`pg_cluster_size()` is defined in Neon's `pgxn/neon/neon--1.0.sql` — it is not built in. `neon.max_cluster_size`, by contrast, is a `shared_preload_libraries` GUC and shows in `pg_settings` regardless. So the first gate passes, then `SELECT pg_cluster_size()` errors with *function does not exist*, which is not a storage error, so `postgres_write_ready` fails open (`services/server/src/routes/admin.rs:244`) and `write_ready` stays sidecar-only. At `tracing::debug!`, nobody would ever notice.

This is not hypothetical: per the ticket, prod and staging were `pg_dump`/`pg_restore`d into a **new `memwal` database** on the regular branches, and `services/server/migrations/001_init.sql` only does `CREATE EXTENSION IF NOT EXISTS vector`. Worth one query against prod before merging:

```sql
SELECT extname FROM pg_extension;
SELECT pg_cluster_size();
```

Two changes regardless of the result:

- Log the fail-open at `warn`, not `debug`, so a silent no-op is visible.
- Schema-qualify the call. We already hit an empty `search_path` through the PgBouncer pooler during this same cutover (`no schema has been selected to create in`). If `DATABASE_URL` ever points back at a pooler, the unqualified name fails to resolve and silently fails open too.

## Should fix

### 3. `write_ready` is a CI gate

`.github/actions/wait-for-relayer/action.yml:99` blocks on `data.get("write_ready") is True`, so tightening its meaning can hang the SDK deploy-wait while reads are perfectly fine. Not currently in the PR's Risks section.

### 4. The MCP health tool now misreports the cause

`services/server/scripts/mcp/tools/health.ts:29` hardcodes:

```
write_ready=false (relayer is up; encryption sidecar did not answer health)
```

After this PR that string will blame the sidecar for a Postgres disk outage — exactly the misdiagnosis the Slack copy goes out of its way to prevent.

### 5. A fully unreachable Postgres still reports `write_ready: true`

Every non-storage probe error fails open, including pool-acquire timeouts. Defensible as anti-flap, but it sits awkwardly against the acceptance criterion "`/health` `write_ready` is false when the DB cannot accept writes". Worth making that a deliberate choice rather than an incidental one.

## Minor

- **The `"53100"` arm is effectively dead.** sqlx 0.8's `Error::Database` is `#[error("error returned from database: {0}")]` and `PgDatabaseError::fmt` writes only `self.message()` — SQLSTATE never reaches the string. No functional loss (real disk-full messages hit `could not extend file` / `no space left on device`), but the doc comment claims coverage it does not have, and substring-matching a 5-digit number against arbitrary error text is a small false-positive surface.
- **Two DB round-trips share the sidecar's 300ms budget.** `neon.max_cluster_size` is static per compute — read it once into a `OnceCell` instead of re-querying `pg_settings` on every 2s probe. Removes a round-trip from the timeout.
- **Helper placement.** `maybe_alert_postgres_storage_exhausted` lives in `services/server/src/rate_limit.rs:833` but is a generic alert helper called from `remember.rs`. It belongs in `alerts.rs`.
- **The PR's verification command does not run** — `cargo test` takes a single TESTNAME. Working form:

```
cargo test --bin memwal-server -- postgres_can_accept neon_max_cluster health_response_serializes postgres_storage
```

- `postgres_storage_dedup_is_per_network_not_per_job` calls `AlertManager::from_env`, so it picks up a real `POSTGRES_STORAGE_ALERT_DEDUP_SECS` if one is set in the environment.

ducnmm:
Addressed review on `a04906ff`. `insert_vector` now alerts (single site on `VectorDb`), Neon probe is `public.pg_cluster_size()` with warn fail-open, MCP health no longer blames the sidecar. @harrymove-ctrl please re-review.

harrymove-ctrl (changes requested):
Re-review at `a04906ff` (6 commits since `243c2ffd`). CI green, including `Server / Clippy + Unit tests`.

## Fixed since the last review

- Alert moved to the write that actually extends — `insert_vector` (`services/server/src/storage/db.rs:1601`), covering `vector_entries` + HNSW.
- `pg_cluster_size` is schema-qualified (`public.pg_cluster_size()`) and the fail-open logs at `warn!`, not `debug!`.
- `services/server/scripts/mcp/tools/health.ts` no longer hardcodes the sidecar attribution — `write_ready=false (writes unavailable)`, plus a `writes=paused` note.
- `docs/relayer/api-reference.md` and the `HealthResponse` doc comment both describe the new meaning.

## Still blocking

**1. The prod existence check is still unrun, and nothing creates the extension.**

`grep -rni "create extension" services/server/migrations/` returns exactly one hit: `CREATE EXTENSION IF NOT EXISTS vector` (`services/server/migrations/001_init.sql:4`). Extensions are per-database, and the cutover `pg_restore`d into a **brand-new `memwal` database** on the regular branches. If `neon` is not installed there, `public.pg_cluster_size()` errors on every probe → fail-open → the `write_ready` acceptance criterion silently no-ops in prod, while emitting a `warn!` up to once every 2s (the `WRITE_READY_CACHE` TTL) indefinitely.

Same check I asked for on the previous review, still unanswered — direct host, not the pooler:

```sql
SELECT extname, extnamespace::regnamespace FROM pg_extension;
SELECT public.pg_cluster_size();
```

If it is missing: add `CREATE EXTENSION IF NOT EXISTS neon` as a migration (and confirm `neondb_owner` is allowed to create it), or drop the GUC approach in favour of a `sum(pg_database_size(datname))` fallback.

**2. The probe is least reliable exactly when it matters.**

`postgres_write_ready` (`services/server/src/routes/admin.rs:206`) returns `true` on both probe error *and* the 300ms timeout. At the cap the cluster is degraded, and `pg_cluster_size()` — which reads through the pageserver — is the query most likely to be slow. So `write_ready` can stay `true` right through the outage it was added to catch.

The timeout arm should log at `warn`, not `debug`. And 300ms is tight for a probe that is already cached for 2s; 1s costs nothing.

**3. Coverage regression in `a04906ff`, and the PR body no longer matches the code.**

The description still says the alert fires "on `/api/remember` and `/api/remember/bulk` job INSERT, and on storage-quota reservation". The slim commit removed both; only `insert_vector` is hooked now.

At the cap, the apalis enqueue (`apalis.jobs`, 897MB at incident, written on every remember) and the `remember_jobs` INSERT are equally plausible first failures. If either goes first, `remember()` 500s and nothing alerts. The classifier is a string match and dedup is one-per-network-per-30m, so restoring the enqueue hook is nearly free. Either restore it or correct the body.

## Non-blocking

**4. Slack POST runs inside the open transaction.**

`maybe_alert_storage_exhausted(e).await` sits between the failed INSERT and `result?` (`services/server/src/storage/db.rs:1601`), so `tx` is still alive. `send_payload` sets no per-request timeout and inherits the shared 30s client (`services/server/src/main.rs:722`), so a slow Slack pins an aborted transaction and one of only 10 pool connections (`services/server/src/storage/db.rs:1335`) for up to 30s.

`AlertDedup::should_suppress` stamps on check, so this is one send per window — bounded, not a pile-up. Still: classify inside the tx and send after it drops, or `tokio::spawn` the send.

**5. The acceptance criterion is narrower than its wording.**

With fail-open, `write_ready` goes false for the Neon cap only — pool exhaustion or DB down still report `true`. Deliberate given the CI concern, but worth writing into WALM-612 so nobody reads the AC as broader than it is.

**6.** `.github/actions/wait-for-relayer/action.yml:99` still gates deploys on `write_ready is True`. Safe now given fail-open, but a genuine cap incident would block the very deploy that fixes it. Worth a note in the ticket.

## What's right

`pg_cluster_size()` over `pg_database_size` matches Neon's own `neon_extend()` check. The 1MB headroom trips strictly before Neon's error (`used + 1MB < max`). SQLSTATE `53100` is matched via `DatabaseError::code()` rather than a message substring. Dedup is per-network, not per-job. And the payload test asserts the alert cannot be misread as a user quota (`do not tell users to send SUI/WAL`).

ducnmm:
Addressed round-2 review on `7cd06b36`.

- Missing `public.pg_cluster_size()` falls back to `sum(pg_database_size)` vs `neon.max_cluster_size` (warn once). No `CREATE EXTENSION neon`.
- Probe timeout 1s; timeout arm logs at `warn` (still fail-open on non-storage errors).
- Restored Slack on `remember_jobs` INSERT and Apalis enqueue, plus `insert_vector` (send after the tx drops).

@harrymove-ctrl please re-review.

harrymove-ctrl (approved):
Re-review at `7cd06b36` (2 commits since `a04906ff`). CI green on the head sha, including `Server / Clippy + Unit tests`. All three blocking findings are addressed — approving.

## Resolved

**1. Probe no longer silently no-ops without the `neon` extension.**

`cluster_used_bytes` (`services/server/src/routes/admin.rs:257`) takes the second option I offered: prefer `public.pg_cluster_size()`, and on SQLSTATE `42883` latch `PG_CLUSTER_SIZE_MISSING` and fall back to `SUM(pg_catalog.pg_database_size(datname))` against the same `neon.max_cluster_size` GUC. The GUC is `shared_preload_libraries`-backed, so it is there whether or not the extension is — the gate still holds and the acceptance criterion now produces a real number on either path. Classification is on `code()`, not a message substring, and `missing_pg_cluster_size_falls_back_instead_of_fail_open` pins `42883` apart from `08006` and `53100`. The warn fires once per process rather than every 2s.

**2. Timeout arm.** `WRITE_READY_PROBE_TIMEOUT` is 1s and both the error and timeout arms log at `warn!` (`admin.rs:158`, `:221`). `postgres_write_ready_probe_timeout_is_one_second` guards the constant against a silent revert.

**3. Coverage matches the description again.** The enqueue hook is back (`services/server/src/routes/mod.rs:75`, `apalis.jobs`) alongside the `remember_jobs` INSERT on both `/api/remember` (`remember.rs:906`) and `/api/remember/bulk` (`remember.rs:1309`), so a cap hit on any of the three plausible first failures alerts. The PR body now describes exactly these paths.

**4. (Was non-blocking) Slack POST is out of the transaction.** `insert_vector` drops `tx` before `maybe_alert_storage_exhausted` (`services/server/src/storage/db.rs:1589`), so a slow webhook no longer pins an aborted transaction and one of ten pool connections for up to 30s. Folding the classifier into `maybe_alert_sqlx_postgres_storage_exhausted` also removes the classify-then-send split that made this easy to get wrong at the next call site.

## Non-blocking, for the ticket rather than this PR

- `pg_database_size(datname)` needs `CONNECT` on each database (or `pg_read_all_stats`). On the Neon role set that holds — `postgres`, `template0`, `template1` all grant `CONNECT` to `PUBLIC` — but a single denied database errors the whole scalar and falls open. Worth knowing if the role set ever changes.
- `PG_CLUSTER_SIZE_MISSING` never resets, so installing the `neon` extension later needs a relayer restart to go back to the exact number. Fine as a latch; just not self-healing.
- Still open from round 2: fail-open keeps the AC narrower than its wording (pool exhaustion / DB down still report `true`), and `.github/actions/wait-for-relayer/action.yml:99` would block on `write_ready` during a genuine cap — the deploy that fixes it. Both belong in WALM-612, not here.

The prod `SELECT public.pg_cluster_size();` check is still worth running once after deploy to know which branch you are on — but it is no longer load-bearing for correctness, which is what made it blocking before.

harrymove-ctrl (approved):
Approving at `7cd06b36`. Note the ping named `a04906ff` — the branch has since moved two commits past it, and it is those two (`a0af5223`, `7cd06b36`) that resolve the round-2 review. All three blocking items are addressed. CI is green, and I ran the suite locally at this SHA: 9 passed, 0 failed.

## Blocking items, resolved

**1. Missing `neon` extension.** Taken the second option I offered rather than adding a migration: `42883` on `public.pg_cluster_size()` now latches `PG_CLUSTER_SIZE_MISSING` and falls back to `sum(pg_catalog.pg_database_size(datname))` against the same GUC, with a one-shot `warn!`. That is the right call — `CREATE EXTENSION neon` would have broken vanilla Postgres in local and CI, which the Out-of-scope section now says explicitly. Matching on `DatabaseError::code()` rather than the message is consistent with how `53100` is handled. The prod existence question is now moot: either branch works without an answer.

**2. Probe reliability under load.** Timeout is 1s and the timeout arm logs at `warn!`. The `NEON_MAX_CLUSTER_SIZE_BYTES` `OnceCell` also lands here, so the steady-state probe is one round-trip inside that 1s instead of two inside 300ms — that is a bigger real improvement than the timeout bump. `get_or_try_init` does not cache on error, so a probe during a blip re-reads rather than pinning `None` and disabling the check for the process lifetime.

**3. Coverage regression.** All three write paths are hooked: `remember_jobs` INSERT single (`remember.rs:920`) and bulk (`remember.rs:1321`), the Apalis enqueue (`routes/mod.rs:85`, which `remember.rs:1125` and `analyze.rs:874` both route through), and `insert_vector` (`db.rs:1591`). Whichever of `vector_entries`, `remember_jobs`, or `apalis.jobs` extends first at the cap, something alerts. Folding the classifier into `maybe_alert_sqlx_postgres_storage_exhausted` is what makes each of those call sites a few lines rather than a repeated `if`. The PR body now matches the code.

**4** (was non-blocking). `drop(tx)` before the Slack send at `db.rs:1591` fixes the send-inside-open-transaction problem — the pool connection is released before the POST, so a slow Slack can no longer hold one of ten connections for up to 30s during the exact incident this fires on.

## Non-blocking

- `PG_CLUSTER_SIZE_MISSING` never unlatches. If someone installs the extension on a live compute, or if it exists in a non-`public` schema (also `42883`), the server stays on the less accurate fallback until restart. Fine — the fallback works — but worth knowing it is a one-way door per process.
- The fallback undercounts: Neon's cluster size includes WAL and history, `sum(pg_database_size)` does not, so it trips later than the real cap and possibly not at all. The Risks section already says this. At the incident the database files alone (2002 + 861 + 897MB) blew past 3072MB, so it would have caught this one — just do not read it as equivalent to the `pg_cluster_size()` path.
- Fail-open still means pool exhaustion and a fully unreachable Postgres report `write_ready: true`, and `.github/actions/wait-for-relayer/action.yml:99` still gates deploys on it. Both are deliberate and documented now; they belong in WALM-612 so the acceptance criterion is not read more broadly than it is.

## Verified

`cargo test --bin memwal-server -- postgres_can_accept neon_max_cluster health_response_serializes postgres_storage missing_pg_cluster_size postgres_write_ready_probe_timeout` — 9 passed. `pg_cluster_size_unavailable` correctly separates `42883` from `08006` and from a `53100` disk-full, so a real storage error cannot be mistaken for a missing function and silently routed to the fallback.

PR #891, opened 2026-09-09
fix(researcher): use Sui gRPC in the browser so Enoki login survives JSON-RPC CORS (WALM-604)

nikola0x0:
Fixes WALM-604.

Researcher's Google/Enoki sign-in fails in dev, staging and production: Sui's public fullnodes no longer answer browser JSON-RPC preflights, so dapp-kit's ambient client dies on `getNormalizedMoveFunction`. Noter moved to gRPC for this in #684; researcher was left behind.

## Two departures from the ticket

- **Ports noter's merged #684, not the researcher half of `hotfix/noter-mainnet-rpc-cors`.** That branch needs a new `NEXT_PUBLIC_SUI_GRPC_URL` across three environments, and silently falls back to the broken path if one is missed.
- **Skips noter's `account-bcs.ts`**, reading via gRPC `include: { json: true }` instead. BCS schemas decode silently wrong once a struct grows a field, and `account.move` already has.

Also fixes `sui-providers.tsx`, which the ticket doesn't mention — `registerEnokiWallets` was getting the dead JSON-RPC client too.

## Verified

Typecheck, lint, `next build`. Plus live reads against both the testnet and mainnet registries: json and BCS agree, and 6 real owners round-trip to the correct account.

## Not verified

Google sign-in itself — the ticket's *How to verify* step. Blocked locally: the Enoki key is mainnet-only, and the Google OAuth client doesn't authorize `http://localhost:3000/login`. Needs the dev deploy to confirm. Login is already broken on dev today, so deploying this can't regress it.

### DIFF
--- a/apps/researcher/components/enoki-login-card.tsx
+++ b/apps/researcher/components/enoki-login-card.tsx
@@ -7,15 +7,20 @@ import {
   useCurrentAccount,
   useSignPersonalMessage,
   useSignTransaction,
-  useSuiClient,
 } from "@mysten/dapp-kit";
 import { isEnokiWallet } from "@mysten/enoki";
+import type { SuiGrpcClient } from "@mysten/sui/grpc";
 import { Transaction } from "@mysten/sui/transactions";
 import { createSponsorAuthorization } from "@mysten-incubation/memwal";
 import { Loader2 } from "lucide-react";
 import { useRouter } from "next/navigation";
 import { Button } from "@/components/ui/button";
 import { enokiConfig } from "@/lib/enoki/config";
+import { getSuiGrpcClient } from "@/lib/sui/grpc-client";
+import {
+  fetchAccountIdForOwner,
+  findCreatedObjectByType,
+} from "@/lib/sui/account-lookup";
 
 type Step =
   | "idle"
@@ -57,14 +62,14 @@ function uint8ArrayToBase64(bytes: Uint8Array): string {
 async function sponsoredSignAndExecute(
   transaction: Transaction,
   sender: string,
-  suiClient: ReturnType<typeof useSuiClient>,
+  suiClient: SuiGrpcClient,
   signTransaction: (args: {
-    transaction: Transaction;
+    transaction: Transaction | string;
   }) => Promise<{ signature: string }>,
   signPersonalMessage: (message: Uint8Array) => Promise<{ signature: string }>,
 ): Promise<{ digest: string }> {
   const kindBytes = await transaction.build({
-    client: suiClient as any,
+    client: suiClient,
     onlyTransactionKind: true,
   });
   const authorization = await createSponsorAuthorization(
@@ -93,7 +98,16 @@ async function sponsoredSignAndExecute(
 
   const sponsored = await sponsorRes.json();
   const sponsoredTx = Transaction.from(sponsored.bytes);
-  const { signature } = await signTransaction({ transaction: sponsoredTx });
+  // dapp-kit's useSignTransaction resolves move-call ABIs via the ambient
+  // client from SuiClientProvider, which is JSON-RPC (deprecated, no longer
+  // CORS-enabled for browser origins) — that resolution is what fails as
+  // `getNormalizedMoveFunction: Failed to fetch`. Pre-serializing with our
+  // gRPC client and handing off the resulting string short-circuits it:
+  // dapp-kit passes a string through as-is.
+  const sponsoredTxJson = await sponsoredTx.toJSON({ client: suiClient });
+  const { signature } = await signTransaction({
+    transaction: sponsoredTxJson,
+  });
 
   const execRes = await fetch(
     `${enokiConfig.memwalServerUrl}/sponsor/execute`,
@@ -127,7 +141,7 @@ export function EnokiLoginCard() {
   const wallets = useWallets();
   const { mutateAsync: connect } = useConnectWallet();
   const currentAccount = useCurrentAccount();
-  const suiClient = useSuiClient();
+  const suiClient = getSuiGrpcClient();
   const { mutateAsync: signTransaction } = useSignTransaction();
   const { mutateAsync: signPersonalMessage } = useSignPersonalMessage();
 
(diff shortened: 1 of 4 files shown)


apps/researcher/lib/sui/account-lookup.ts
```diff
@@ -0,0 +1,81 @@
...
+  digest: string,
+  objectType: string,
+): Promise<string | null> {
+  const response = await client.getTransaction({
+    digest,
+    include: { effects: true, objectTypes: true },
+  });
+
+  const transaction = response.Transaction ?? response.FailedTransaction;
+  const created = transaction.effects?.changedObjects.find(
```
ducnmm, line 75: **[suggestion]** First-time signup finds the new MemWalAccount only through gRPC `effects.changedObjects` + `objectTypes` with a substring match on `"MemWalAccount"`. Noter's production login (the claimed port source) reads the `AccountCreated` event BCS after `include: { events: true }`, and the main app later kept that event path as a fallback because object-change views can omit the account. This helper never requests events. That is the one on-chain read on the unverified Google sign-in path: if `objectTypes` is missing or `idOperation` is not `"Created"`, `findCreatedObjectByType` returns null after a successful sponsored `create_account` and the user sees "Account created on-chain but object ID not found".

**Suggestion:** Prefer noter's event lookup (or use it as a fallback): `getTransaction({ digest, include: { events: true, effects: true, objectTypes: true } })`, find `eventType.endsWith("::account::AccountCreated")`, and take `event.json.account_id` (or the 2-field `AccountCreated` BCS). That stays consistent with this PR's json-over-BCS choice and matches the production noter path.

apps/researcher/lib/sui/account-lookup.ts
```diff
@@ -0,0 +1,81 @@
+/**
+ * On-chain reads for the Enoki registration flow, over gRPC.
+ *
+ * These deliberately use gRPC's `include: { json: true }` rather than decoding
+ * BCS with hand-written struct schemas (the approach in
+ * apps/noter/lib/sui/account-bcs.ts). BCS schemas must list every field of a
+ * struct in declaration order; when the published package grows a field, a
```
ducnmm, line 7: **[suggestion]** The file header narrates the noter BCS approach, claims schemas "decode *silently wrong*" when a struct grows a field, and discusses `account.move` history. That is design/architecture commentary, not a constraint the following code needs. It is also stronger than noter's own decoder note: new `AccountRegistry` / `MemWalAccount` fields were appended, and noter's `account-bcs.unit.test.ts` documents that trailing bytes either fail closed or are ignored — only *inserted* fields would silently decode wrong. The json `Table` UID dual-shape handling is the only non-obvious bit.

**Suggestion:** Drop the noter/BCS history. Keep a short why on the UID ternary (gRPC json has rendered `Table.id` as both a string and `{ id }`).

apps/researcher/components/sui-providers.tsx
```diff
@@ -19,11 +19,20 @@ const { networkConfig } = createNetworkConfig({
 
 const queryClient = new QueryClient();
 
-/** Registers Enoki wallets (Google OAuth) with dapp-kit on mount. No-op if env vars are missing. */
+/**
+ * Registers Enoki wallets (Google OAuth) with dapp-kit on mount. No-op if env
+ * vars are missing.
+ *
+ * Uses a standalone SuiGrpcClient rather than SuiClientProvider's client:
```
ducnmm, line 26: **[suggestion]** `RegisterEnokiWallets` replaced a one-line doc with a multi-paragraph explanation of dapp-kit typing, JSON-RPC deprecation, and `ClientWithCoreApi`. `grpc-client.ts` similarly embeds #684 production history. None of that is required to understand the `getSuiGrpcClient()` call.

**Suggestion:** Keep a one-liner: Enoki must get the standalone gRPC client because `SuiClientProvider` is still JSON-RPC. Move the rest out of the component.

apps/researcher/components/enoki-login-card.tsx
```diff
@@ -93,7 +98,16 @@ async function sponsoredSignAndExecute(
...
   const sponsored = await sponsorRes.json();
   const sponsoredTx = Transaction.from(sponsored.bytes);
-  const { signature } = await signTransaction({ transaction: sponsoredTx });
+  // dapp-kit's useSignTransaction resolves move-call ABIs via the ambient
+  // client from SuiClientProvider, which is JSON-RPC (deprecated, no longer
+  // CORS-enabled for browser origins) — that resolution is what fails as
+  // `getNormalizedMoveFunction: Failed to fetch`. Pre-serializing with our
+  // gRPC client and handing off the resulting string short-circuits it:
+  // dapp-kit passes a string through as-is.
+  const sponsoredTxJson = await sponsoredTx.toJSON({ client: suiClient });
```
ducnmm, line 107: **[nit]** The `toJSON` + string handoff is the right CORS fix (`useSignTransaction` only skips ABI resolution when `transaction` is a string; Enoki then `Transaction.from(json).build({ client })` on the registered gRPC client). The seven-line comment restates the ticket error string and dapp-kit internals on top of that why.

**Suggestion:** Trim to: pre-serialize with the gRPC client and pass a string so dapp-kit does not resolve ABIs through `SuiClientProvider`'s JSON-RPC client.

ducnmm (commented):
## Summary

This is a sound port of noter's production gRPC login path for the actual CORS failure: `sponsoredSignAndExecute` now builds and `toJSON`s with `SuiGrpcClient`, then hands dapp-kit a string so `useSignTransaction` does not call JSON-RPC `getNormalizedMoveFunction`, and `registerEnokiWallets` gets that same gRPC client (Enoki's own `build({ client })` would otherwise still hit the ambient JSON-RPC client). Leftover JSON-RPC is only `SuiClientProvider`'s hard-typed `SuiJsonRpcClient` (same as noter); no remaining `useSuiClient` / `getObject` / `getTransactionBlock` call sits on the researcher login path. Registry lookup via `include: { json: true }` matches the already-shipped gRPC shape in researcher `delegate-account.ts` and `apps/app` (author also live-checked it). Dominant residual risk is first-time `create_account` object discovery, which this PR implements via gRPC `objectTypes` rather than noter's production `AccountCreated` events, and Google sign-in itself was not run.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 1

ducnmm (approved):
## Summary

`10d19e7f` is a merge of `origin/dev` only. `git diff da778fc1..10d19e7f -- apps/researcher` is empty, so the gRPC Enoki login path reviewed on `da778fc1` is unchanged. Approving this head.

Previous notes still stand and do not block: first-time `create_account` discovery via `objectTypes` rather than noter's `AccountCreated` event, and Google sign-in was not live-tested.

PR #890, opened 2026-09-09
fix(sdk): type waitForRememberJob poll timeout (WALM-600)

HoangDucBach:
## Ticket

WALM-600 — https://linear.app/mysten-labs/issue/WALM-600
GH #691 — https://github.com/MystenLabs/MemWal/issues/691

## What changed?

- `waitForRememberJob` / `rememberAndWait` throw exported `RememberJobTimeoutError` (`status` 504, `jobId`, `timeoutMs`) instead of an untyped `Error`.
- `waitForRememberJobs` still returns leftover items with `status: "timeout"` and does not throw.
- JSDoc and API docs call out the split and the default timeouts (60s single / 120s bulk).
- Python already raised `MemWalRememberJobTimeout`; docs/docstrings now match the TS contract. Bulk still does not raise.

## Why is this needed?

Callers who tested error handling against bulk shipped unhandled exceptions on the single API. Unifying both to a returned `{ status: "timeout" }` would hide empty `blob_id` as success and drop the idempotency key `rememberAndWait` keeps on throw. WALM-597 / #881 keeps the throw and attaches last-known blob state; this PR only types and documents the existing split.

## Scope

TypeScript SDK wait helpers plus Python docs/tests for the same contract.

### Out of scope

- Making single return `{ status: "timeout" }`
- Making bulk throw on poll timeout
- Last-known `blob_id` on timeout (WALM-597)

## Test plan

- [x] `pnpm --filter @mysten-incubation/memwal test` (112 pass)
- [x] `waitForRememberJob` with `timeoutMs: 1` throws `RememberJobTimeoutError`
- [x] `waitForRememberJobs` with the same timeout returns `status: "timeout"`
- [x] Python `test_wait_for_remember_timeout_contract`

Fixes #691

PR #892, opened 2026-09-09
fix(sdk): make a remember timeout a recoverable unknown outcome (WALM-595)

### LINKED ISSUE #658: [SDK/Relayer Bug & Friction] Synchronous Timeout Exception During Mainnet Relayer Congestion Without Idempotency Safety or Unknown-Outcome Recovery
## Overview
During mainnet integration testing of Walrus Memory (`MemWal`) in our production SaaS architecture (`Walrus Drop v2`), we encountered severe operational friction during network latency spikes and relayer congestion. Specifically, synchronous blocking calls waiting on `rememberAndWait()` threw hard timeout exceptions after 30,000ms, while the relayer backend had already accepted the transaction job queue.

## Verified Mainnet Evidence & Transcripts
We captured two distinct mainnet interaction states against `https://relayer.memory.walrus.xyz` using server-signed delegate credentials (`MEMWAL_ACCOUNT_ID: 0xe4d61ee7b4a9b4252426d7e045c52810749057ca1c6a745d37cfe8253e14895e` ):

1. **Uncertain Outcome State (Timeout):**
   - Job ID: `9b921ab2-e9a3-490d-8180-136fd98cfc2a`
   - Log Transcript: `[Error] Live round-trip encountered error: remember job timed out after 30000ms`
   - Impact: The client caller receives an unhandled error, leaving session state ambiguous and risking duplicate writes upon retry.

2. **Confirmed Stored State (Successful Round-Trip):**
   - Job ID: `dbcc8dc9-56e5-40b3-ac65-bff1e34f5825`
   - Blob ID: `6EPcpE73RpswQqMc52Flp47qm2TcGr_zdBaQZJKMNTE`
   - Owner Address: `0x12be5fae744808a64efc61401e2810b99254f0dd62c6c01cf42841f6e943ef61`
   - Status: `completed` / stored on mainnet.

## Proposed SDK & Relayer Enhancements
1. **Client-Supplied Idempotency Key:** Allow developers to pass an optional `idempotencyKey?: string` (UUID v4) in SDK `RememberOptions` to prevent duplicate writes during distributed retries.
2. **Asynchronous Non-Blocking Handles:** Expose asynchronous task queue handles or webhook polling primitives rather than enforcing strict synchronous blocking waits in client SDKs.

## Open-Source Reference Implementation
Complete reproduction scripts, evidence reports, and database-backed reconciliation state machines built to handle this friction are publicly documented in our open-source repository:
- **Repository:** https://github.com/sands786/walrus-drop-v2
- **Evidence Report:** `SUBMISSION_EVIDENCE_REPORT.md`

harrymove-ctrl:
Closes [WALM-595](https://linear.app/mysten-labs/issue/WALM-595/bug-remember-timeout-on-congestion-has-no-idempotency-recovery-gh-658). Fixes #658.

## Problem

`rememberAndWait()` threw a bare `Error` when its poll budget expired during relayer congestion — but the relayer had already **accepted** the job, and it usually finishes seconds later. Two things followed from that:

1. The caller could not tell *"the write may still land"* from *"the write failed"* without matching on the message text.
2. The idempotency key the SDK generated for the write never left the client instance (`pendingRememberKeys`, in memory). A service that restarts — the exact deployment shape in #658 — had nothing to retry *with*, so the obvious retry minted and paid for a second blob.

The relayer side is already correct: `POST /api/remember` collapses a replay onto the original job under `(owner, idempotency_key)`. What was missing was handing the caller the handle to replay with.

## Change

**TypeScript SDK**

- `RememberTimeoutError` (name `MemWalRememberTimeoutError`) and an exported `isRememberTimeoutError()`, so unknown-outcome is a branch rather than a string match.
- The error carries `jobId`, `idempotencyKey` and `namespace`. Its message names both recovery moves, since that string is what ends up in a caller's logs.
- `RememberAcceptedResult.idempotency_key` echoes the key back, so `remember()` / `rememberAsync()` callers can persist it next to `job_id`.

**Python SDK** — `MemWalRememberJobTimeout` gains `idempotency_key` / `namespace` and the same message, keeping the two SDKs aligned.

**Docs** — `production-readiness.md` drops the now-false *"the relayer does not deduplicate writes"* claim, documents `idempotencyKey`, and adds a recovery section.

## Recovery, from the caller's side

```ts
try {
  await memwal.rememberAndWait(text, namespace);
} catch (err) {
  if (!isRememberTimeoutError(err)) throw err;          // a real failure
  await memwal.waitForRememberJob(err.jobId);           // settle it; no second write
  // or, from another process:
  // await memwal.rememberAndWait(text, err.namespace, { idempotencyKey: err.idempotencyKey });
}
```

## Acceptance criteria

- [x] Timeout after accept does not require a second `remember` to recover status — `waitForRememberJob(err.jobId)`, pinned by a test asserting exactly one `POST /api/remember`.
- [x] Retry with the same idempotency key does not mint a second blob — pinned by a test that replays through a **fresh client** (standing in for a restarted process) and asserts one distinct key across both attempts, plus a companion test showing that losing the key is what produces two.

## Testing

- TS SDK: 8 new tests in `packages/sdk/test/remember-timeout-recovery.test.mjs`; full suite **125 passed**, `tsc` clean.
- Python SDK: 3 new tests in `TestRememberTimeoutRecovery`; full suite **162 passed, 16 skipped**.

## Notes for the reviewer

- No relayer change: the server-side idempotency path already handles this. This PR is entirely about what the client hands back.
- Purely additive to the public surface. The timeout still has `status === 504` and `jobId`, so existing catchers keep working; only the message text changed.
- Touches `waitForRememberJob`, which #881 (WALM-597) also edits. Expect a small conflict at the throw site if that lands first — the two changes are complementary (that PR keeps the *last observed* state, this one adds the *submission* handles).

### DIFF
--- a/docs/sdk/production-readiness.md
+++ b/docs/sdk/production-readiness.md
@@ -37,26 +37,64 @@ Several of these are patterns you implement around the client today. Where a cap
 
 ## Make writes idempotent
 
-The relayer does not deduplicate writes. A `remember` call that retries after a network blip, or fired twice by an at-least-once job queue, stores the same text twice and pollutes later recall. Until content-based deduplication is available natively, gate writes on a key your agent controls so a repeat is a no-op.
+Every write carries an idempotency key. Pass your own via `idempotencyKey` and a
+repeat submission returns the original job instead of minting — and charging for
+— a second blob. Omit it and the SDK generates one per call, which still covers
+its own internal retries but gives you nothing to replay with later.
 
 ```ts
 import { createHash } from "crypto";
 
-const written = new Set<string>(); // back this with Redis or a DB in production
+function writeKey(text: string, namespace = "default") {
+  return createHash("sha256").update(`${namespace}:${text}`).digest("hex");
+}
 
-async function rememberOnce(memwal: MemWal, text: string, namespace?: string) {
-  const id = createHash("sha256").update(`${namespace ?? "default"}:${text}`).digest("hex");
-  if (written.has(id)) return; // already stored this exact memory
+const job = await memwal.remember(text, namespace, { idempotencyKey: writeKey(text, namespace) });
+// job.idempotency_key is the handle to persist alongside job.job_id.
+```
 
-  const job = await memwal.remember(text, namespace);
-  await memwal.waitForRememberJob(job.job_id);
-  written.add(id);
+Reusing a key for *different* content is rejected with a `409`, so a key derived
+from the content itself is the safest choice.
+
+<Note>
+Persist the key outside the process (Redis, a database row, a Sui object), not in
+memory. An in-memory map resets on restart, which is exactly when a retry storm
+is most likely.
+</Note>
+
+## Recover an unknown outcome
+
+When `rememberAndWait()` exhausts its poll budget, the write is **unknown, not
+failed**. The relayer accepted the job before the budget expired and usually
+finishes it seconds later. Retrying that call blindly is what stores the memory
+twice.
+
+`isRememberTimeoutError()` separates the two cases, and the error carries both
+handles you need to settle the write:
+
+```ts
+import { isRememberTimeoutError } from "@mysten-incubation/memwal";
+
+try {
+  await memwal.rememberAndWait(text, namespace);
+} catch (err) {
+  if (!isRememberTimeoutError(err)) throw err; // a real failure
+
+  // Option A — settle the job that already exists. No second write.
+  const settled = await memwal.waitForRememberJob(err.jobId, { timeoutMs: 120_000 });
+  console.log(settled.blob_id);
+
+  // Option B — hand err.jobId and err.idempotencyKey to a durable queue and
+  // resume later, even from another process:
+  //   await memwal.rememberAndWait(text, err.namespace, {
+  //     idempotencyKey: err.idempotencyKey,
+  //   });
 }
 ```
 
-<Note>
-Persist the idempotency set outside the process (Redis, a database row, a Sui object), not in memory. An in-memory set resets on restart, which is exactly when a retry storm is most likely.
-</Note>
+Prefer option A while the process is still up: it is a read. Option B is for a
+worker that picks the write back up after a restart — the replay collapses onto
+the original job, so it settles the same blob rather than paying for a new one.
 
 ## Retry with backoff, but only retryable failures
 
--- a/docs/sdk/production-readiness.md
+++ b/docs/sdk/production-readiness.md
@@ -37,10 +37,10 @@ Several of these are patterns you implement around the client today. Where a cap
 
 ## Make writes idempotent
 
-Every write carries an idempotency key. Pass your own via `idempotencyKey` and a
-repeat submission returns the original job instead of minting — and charging for
-— a second blob. Omit it and the SDK generates one per call, which still covers
-its own internal retries but gives you nothing to replay with later.
+Every write carries an idempotency key. Pass your own through `idempotencyKey` and
+a repeat submission returns the original job instead of minting a second blob,
+which you would also pay for. Omit it and the SDK generates one per call, which
+still covers its own internal retries but gives you nothing to replay with later.
 
 ```ts
 import { createHash } from "crypto";
@@ -49,12 +49,20 @@ function writeKey(text: string, namespace = "default") {
   return createHash("sha256").update(`${namespace}:${text}`).digest("hex");
 }
 
+async function rememberOnce(memwal: MemWal, text: string, namespace?: string) {
+  return memwal.rememberAndWait(text, namespace, {
+    idempotencyKey: writeKey(text, namespace),
+  });
+}
+
+// job.idempotency_key echoes the key back, so a fire-and-forget caller can
+// persist it alongside job.job_id.
 const job = await memwal.remember(text, namespace, { idempotencyKey: writeKey(text, namespace) });
-// job.idempotency_key is the handle to persist alongside job.job_id.
 ```
 
-Reusing a key for *different* content is rejected with a `409`, so a key derived
-from the content itself is the safest choice.
+Derive the key from the content, as above. It is then stable across processes, so
+a replay after a restart lands on the original job. Reusing one key for
+*different* content is rejected with a `409`.
 
 <Note>
 Persist the key outside the process (Redis, a database row, a Sui object), not in
@@ -69,22 +77,22 @@ failed**. The relayer accepted the job before the budget expired and usually
 finishes it seconds later. Retrying that call blindly is what stores the memory
 twice.
 
-`isRememberTimeoutError()` separates the two cases, and the error carries both
+`isRememberJobTimeoutError()` separates the two cases, and the error carries both
 handles you need to settle the write:
 
 ```ts
-import { isRememberTimeoutError } from "@mysten-incubation/memwal";
+import { isRememberJobTimeoutError } from "@mysten-incubation/memwal";
 
 try {
   await memwal.rememberAndWait(text, namespace);
 } catch (err) {
-  if (!isRememberTimeoutError(err)) throw err; // a real failure
+  if (!isRememberJobTimeoutError(err)) throw err; // a real failure
 
-  // Option A — settle the job that already exists. No second write.
+  // Option A: settle the job that already exists. No second write.
   const settled = await memwal.waitForRememberJob(err.jobId, { timeoutMs: 120_000 });
   console.log(settled.blob_id);
 
-  // Option B — hand err.jobId and err.idempotencyKey to a durable queue and
+  // Option B: hand err.jobId and err.idempotencyKey to a durable queue and
   // resume later, even from another process:
   //   await memwal.rememberAndWait(text, err.namespace, {
   //     idempotencyKey: err.idempotencyKey,
@@ -93,12 +101,14 @@ try {
 ```
 
 Prefer option A while the process is still up: it is a read. Option B is for a
-worker that picks the write back up after a restart — the replay collapses onto
+worker that picks the write back up after a restart. The replay collapses onto
 the original job, so it settles the same blob rather than paying for a new one.
 
 ## Retry with backoff, but only retryable failures
 
-The client does not retry for you. Wrap calls in exponential backoff, and be careful to retry only failures that a retry can fix. A transient network error or relayer timeout is worth retrying. A `401 AUTH_REJECTED` is a configuration problem that fails identically on every attempt, so retrying it just delays the real fix.
+The client does not retry for you. Wrap calls in exponential backoff, and be careful to retry only failures that a retry can fix. A transient network error is worth retrying. A `401 AUTH_REJECTED` is a configuration problem that fails identically on every attempt, so retrying it just delays the real fix.
+
+A write that timed out while polling is the case to be careful with. It is not a failure, so re-running the write is not a retry: it is a second write. Treat `isRememberJobTimeoutError(err)` the way you treat a 4xx and stop, then recover through the job id or the key as shown above.
 
 ```ts
 async function withRetry<T>(fn: () => Promise<T>, attempts = 4): Promise<T> {
@@ -107,6 +117,9 @@ async function withRetry<T>(fn: () => Promise<T>, attempts = 4): Promise<T> {
     try {
       return await fn();
     } catch (err) {
+      // A poll timeout is an unknown outcome, not a failure. Retrying the
+      // write here is what stores the memory twice.
+      if (isRememberJobTimeoutError(err)) throw err;
       const status = (err as { status?: number }).status;
       // Do not retry auth or client errors; they will not succeed on a retry.
       if (status === 401 || (status && status >= 400 && status < 500)) throw err;
@@ -120,7 +133,7 @@ async function withRetry<T>(fn: () => Promise<T>, attempts = 4): Promise<T> {
 await withRetry(() => rememberOnce(memwal, "Observed: deploy at 14:03 UTC."));
 ```
 
-Pairing retry with the idempotency guard above is deliberate: a retry that fires after the write actually succeeded server-side stays safe, because `rememberOnce` short-circuits on the key.
+Pairing retry with the idempotency key above is deliberate: a retry that fires after the write actually succeeded server-side stays safe, because the relayer returns the original job for that key instead of minting a second blob.
 
 <Note>
 Built-in retry and backoff inside the client is on the roadmap. When it ships, you can drop the wrapper for the calls it covers and keep only the idempotency guard.
--- a/packages/python-sdk-memwal/memwal/client.py
+++ b/packages/python-sdk-memwal/memwal/client.py
@@ -498,11 +498,16 @@ class MemWal:
             )
         except MemWalRememberJobTimeout as timeout:
             # The in-memory key map only helps a caller retrying in this same
-            # process. Put the key on the exception too, so a service that
-            # restarts can still replay this exact write (WALM-595).
-            timeout.idempotency_key = resolved_key
-            timeout.namespace = resolved_namespace
-            raise
+            # process. Re-raise with the key so a service that restarts can
+            # replay this exact write, and so the key is in ``str(err)`` and not
+            # only in a field: after a restart the log line may be all that is
+            # left (WALM-595).
+            raise MemWalRememberJobTimeout(
+                job_id=timeout.job_id,
+                timeout_ms=timeout.timeout_ms,
+                idempotency_key=resolved_key,
+                namespace=resolved_namespace,
+            ) from timeout
         if generated_key:
             self._pending_remember_keys.pop(request_identity, None)
         return result
--- a/packages/python-sdk-memwal/memwal/client.py
+++ b/packages/python-sdk-memwal/memwal/client.py
@@ -395,6 +395,7 @@ class MemWal:
         return RememberAcceptedResult(
             job_id=data["job_id"],
             status=data.get("status", "pending"),
+            idempotency_key=resolved_key,
         )
 
     # Alias for parity with TS SDK ``rememberAsync``.
--- a/packages/python-sdk-memwal/memwal/client.py
+++ b/packages/python-sdk-memwal/memwal/client.py
@@ -489,11 +489,19 @@ class MemWal:
             self._pending_remember_keys[request_identity] = resolved_key
 
         accepted = await self.remember(text, resolved_namespace, resolved_key)
-        result = await self.wait_for_remember_job(
-            accepted.job_id,
-            poll_interval_ms=poll_interval_ms,
-            timeout_ms=timeout_ms,
-        )
+        try:
+            result = await self.wait_for_remember_job(
+                accepted.job_id,
+                poll_interval_ms=poll_interval_ms,
+                timeout_ms=timeout_ms,
+            )
+        except MemWalRememberJobTimeout as timeout:
+            # The in-memory key map only helps a caller retrying in this same
+            # process. Put the key on the exception too, so a service that
+            # restarts can still replay this exact write (WALM-595).
+            timeout.idempotency_key = resolved_key
+            timeout.namespace = resolved_namespace
+            raise
         if generated_key:
             self._pending_remember_keys.pop(request_identity, None)
         return result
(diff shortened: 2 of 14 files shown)


jessiemongeon1:
## Style Guide Audit

### `docs/sdk/production-readiness.md` (4 violation(s))

_4 violation(s) (4 regex, 0 claude)_

- **Line 40** — Use "through" not "via"
  - Current: `via`
  - Fix: `through`
- **Line 41** — No em dashes in prose
  - Current: `—`
- **Line 42** — No em dashes in prose
  - Current: `—`
- **Line 96** — No em dashes in prose
  - Current: `—`

---

docs/sdk/production-readiness.md
```diff
@@ -37,27 +37,65 @@ Several of these are patterns you implement around the client today. Where a cap
...
+}
+```
+
+Prefer option A while the process is still up: it is a read. Option B is for a
+worker that picks the write back up after a restart — the replay collapses onto
+the original job, so it settles the same blob rather than paying for a new one.
+
 ## Retry with backoff, but only retryable failures
 
 The client does not retry for you. Wrap calls in exponential backoff, and be careful to retry only failures that a retry can fix. A transient network error or relayer timeout is worth retrying. A `401 AUTH_REJECTED` is a configuration problem that fails identically on every attempt, so retrying it just delays the real fix.
```
ducnmm: **[bug]** The new recovery section correctly says a `rememberAndWait()` poll timeout is unknown and that retrying blindly stores the memory twice. The next paragraph, which this hunk re-homed, still says "A transient network error or relayer timeout is worth retrying," and `withRetry` only bails on 4xx — so a 504 `RememberTimeoutError` is retried. The same edit removed `rememberOnce` (old content-hash gate) but left `await withRetry(() => rememberOnce(...))` at line 120, the "rememberOnce short-circuits" sentence at line 123, and `budgetedRemember` at line 163. Copy-paste is now a ReferenceError; "fixing" it to `rememberAndWait` is the GH #658 double-mint.

**Suggestion:** In the retry section, treat `isRememberTimeoutError(err)` like 4xx (do not retry the write). Recover with `waitForRememberJob(err.jobId)` or replay under `err.idempotencyKey`. Replace the dangling `rememberOnce` call sites with `rememberAndWait(..., { idempotencyKey: writeKey(text) })` (or drop them). Rewrite the "relayer timeout is worth retrying" sentence so it does not contradict the section above.

packages/sdk/src/memwal.ts
```diff
@@ -366,10 +370,10 @@ export class MemWal {
...
         }
 
-        throw Object.assign(
-            new Error(`remember job timed out after ${timeoutMs}ms (job_id=${jobId})`),
-            { status: 504, jobId },
-        );
+        // Unknown outcome, not a failure: the relayer accepted this job and it
+        // may still finish. The error carries the handles needed to settle it
+        // without paying for a second write (WALM-595).
+        throw rememberTimeoutError(jobId, timeoutMs);
```
ducnmm: **[suggestion]** `waitForRememberJob` always throws `rememberTimeoutError(jobId, timeoutMs)` with empty context. `rememberAndWait` then mutates `err.idempotencyKey` / `err.namespace` after the fact (lines 405–407; Python does the same at `client.py:502–504`). The factory already knows how to put the key in the message (`utils.ts:489–491`), and the test at `remember-timeout-recovery.test.mjs:106` says the message is what lands in logs — but the rememberAndWait path, the one that actually has the key, never passes it at construction. `str(err)` / `err.message` therefore has `job_id` and recovery prose, not the key needed for Option B after a restart that only captured the log line.

**Suggestion:** Thread `{ idempotencyKey, namespace }` into `waitForRememberJob` (and Python `wait_for_remember_job`) and construct the timeout with them, or re-create the error in the `rememberAndWait` catch instead of mutating. Keep the object fields; also put the key in the message when the SDK owns the submission.

packages/python-sdk-memwal/memwal/client.py
```diff
@@ -489,11 +489,19 @@ async def remember_and_wait(
             self._pending_remember_keys[request_identity] = resolved_key
 
         accepted = await self.remember(text, resolved_namespace, resolved_key)
```
ducnmm, line 492: **[suggestion]** TS `rememberAsync` now returns `{ ...accepted, idempotency_key }` (`memwal.ts:285`, `types.ts:39`) so a fire-and-forget caller can persist the generated key next to `job_id`. Python `remember()` still returns `RememberAcceptedResult(job_id=..., status=...)` only (`client.py:395–398`; `types.py` has no field). After accept, `_pending_remember_keys` is popped, so the generated key never leaves the process unless `remember_and_wait` later times out. Split-API Python callers (`remember` + `wait_for_remember_job`) still cannot do Option B after a restart. The PR body claims the SDKs are aligned.

**Suggestion:** Add `idempotency_key` to Python `RememberAcceptedResult` and return `resolved_key` from `remember()` / `remember_async()`, matching TS. A one-liner test that `remember()` echoes a generated UUID (and an explicit key unchanged) would pin it.

packages/sdk/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # @mysten-incubation/memwal
 
+## Unreleased
```
ducnmm: **[suggestion]** Notes went under `## Unreleased` (Python the same at `packages/python-sdk-memwal/CHANGELOG.md:3`). This repo dumps into the already-unreleased version: `origin/dev` TS `0.1.7` > `main` `0.1.6`, Python `0.1.10` > `main` `0.1.9`. A later dump that promotes `Unreleased` would mint 0.1.8 / 0.1.11. `docs/sdk/changelog.mdx` and `docs/python-sdk/changelog.mdx` were not updated. Versions themselves are correctly unbumped.

**Suggestion:** Move the new bullets under existing `## 0.1.7` / `## 0.1.10` (`### Added`). Do not bump `package.json` / `pyproject.toml`. Mirror the same lines in the docs changelogs (FAQ `answer` + section intro) without dropping the landed restore / 401 / `tx.pure` notes. Leave MCP `0.0.13` and OpenClaw `0.0.6` alone.

packages/sdk/src/utils.ts
```diff
@@ -432,3 +432,74 @@ export async function delegateKeyToPublicKey(privateKeyHex: string): Promise<Uin
...
+ *
+ * Both handles needed to settle it are on the error, so a caller can recover
+ * without issuing — and paying for — a second write:
+ *
+ * - `jobId` → `waitForRememberJob(jobId)` polls the same job.
+ * - `idempotencyKey` → replaying `rememberAndWait(text, ns, { idempotencyKey })`
+ *   returns the original job instead of minting a second blob. This is the one
+ *   that survives a process restart, which the SDK's in-memory key map does not.
+ */
+export interface RememberTimeoutError extends Error {
```
ducnmm: **[suggestion]** This PR's public timeout is interface `RememberTimeoutError`, `err.name === "MemWalRememberTimeoutError"`, detected only via `isRememberTimeoutError()` (not `instanceof`). Open #890 (WALM-600) and #881 (WALM-597, same author) type the same 504 as class `RememberJobTimeoutError` (`name` `"RememberJobTimeoutError"`, `instanceof` works). If both land, callers get two names and two detectors for one poll timeout, and this type guard will not match the class (and vice versa).

**Suggestion:** Pick one exported class — likely `RememberJobTimeoutError` from #890/#881 — and put `idempotencyKey` / `namespace` (this PR) plus any `lastStatus` / `lastBlobId` (#881) on it. Keep `status === 504` and `jobId` so existing catchers still work.

docs/sdk/production-readiness.md
```diff
@@ -37,27 +37,65 @@ Several of these are patterns you implement around the client today. Where a cap
 
 ## Make writes idempotent
 
-The relayer does not deduplicate writes. A `remember` call that retries after a network blip, or fired twice by an at-least-once job queue, stores the same text twice and pollutes later recall. Until content-based deduplication is available natively, gate writes on a key your agent controls so a repeat is a no-op.
+Every write carries an idempotency key. Pass your own via `idempotencyKey` and a
```
ducnmm: **[nit]** Style-guide audit on this file: "via" on line 40; em dashes on lines 41–42, 83, 87, and 96. The bot already listed 40/41/42/96; 83 and 87 are in the new recovery snippet comments.

**Suggestion:** "through" instead of "via". Split the em-dash sentences with a comma, parentheses, or a period.

ducnmm (dismissed):
## Summary

The SDK change matches WALM-595 / GH #658: a poll timeout after accept is a typed unknown outcome, not a bare failure, and `rememberAndWait` / `remember_and_wait` now put `jobId` plus the generated idempotency key on that error so a restarted process can poll or replay without minting a second blob. TS also echoes `idempotency_key` on `RememberAcceptedResult`; tests cover both recovery moves, including a fresh client standing in for a restart.

Do not bump TS (`0.1.7`) or Python (`0.1.10`). Fold changelog bullets under those existing headings, not `## Unreleased`. Request changes for the production-readiness rewrite: it deletes `rememberOnce` but still tells callers to retry a relayer timeout, which is the duplicate-write path this PR exists to close.

## Issue counts by severity

- bugs: 1
- suggestions: 4
- nits: 1

packages/sdk/node_modules
```diff
@@ -0,0 +1 @@
+/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules
```
ducnmm: **[bug]** Follow-up adds `packages/sdk/node_modules` as a symlink whose target is `/Users/harryphan/Documents/2026 vault/DEV_PROJECTS/active/memwal/MemWal/packages/sdk/node_modules`. Root `.gitignore` is `node_modules/` (directories only), so this symlink is tracked. On any other machine it is a broken (or self-referential) link: install tools that see the name and skip, `stat`/`require` that follow it fail, and the author's local path is in the tree.

**Suggestion:** `git rm -f packages/sdk/node_modules` and keep it untracked. Do not commit a local `node_modules` path.

packages/python-sdk-memwal/memwal/client.py
```diff
@@ -489,11 +490,19 @@ async def remember_and_wait(
...
+            result = await self.wait_for_remember_job(
+                accepted.job_id,
+                poll_interval_ms=poll_interval_ms,
+                timeout_ms=timeout_ms,
+            )
+        except MemWalRememberJobTimeout as timeout:
+            # The in-memory key map only helps a caller retrying in this same
+            # process. Put the key on the exception too, so a service that
+            # restarts can still replay this exact write (WALM-595).
+            timeout.idempotency_key = resolved_key
```
ducnmm: **[suggestion]** TS `rememberAndWait` now reconstructs via `rememberJobTimeoutError(..., { idempotencyKey, namespace })`, and `remember-timeout-recovery.test.mjs:111` pins `err.message.includes(err.idempotencyKey)` because the log line may be all a restarted process still has. Python `remember_and_wait` still mutates `timeout.idempotency_key` / `timeout.namespace` after `MemWalRememberJobTimeout(job_id=..., timeout_ms=...)` built the message without them (`client.py:1464-1471`). Fields work for programmatic replay; `str(err)` still has no key. The follow-up claimed the SDKs aligned on this.

**Suggestion:** Re-raise `MemWalRememberJobTimeout(job_id=..., timeout_ms=..., idempotency_key=resolved_key, namespace=resolved_namespace)` (optionally `from timeout`) instead of mutating. Pin `resolved_key in str(err)` on the remember_and_wait timeout test, matching TS.

packages/sdk/src/types.ts
```diff
@@ -28,9 +28,40 @@ export interface MemWalConfig {
 // ============================================================
 
 /** Result from remember() / rememberAsync() */
```
ducnmm: **[nit]** The existing `/** Result from remember() / rememberAsync() */` was left in place when `RememberJobTimeoutError` was inserted above `RememberAcceptedResult`. The timeout type now sits under a leftover remember-result comment; `RememberAcceptedResult` lost that interface-level note.

**Suggestion:** Move that one-liner back onto `RememberAcceptedResult`, immediately above its `export interface`.

ducnmm (dismissed):
## Summary

Follow-up `926a87ff` addresses the prior CHANGES_REQUESTED on `0069cd32`: `rememberOnce` is restored on `idempotencyKey`, `withRetry` bails on `isRememberJobTimeoutError`, TS rebuilds the timeout so the key is in the message, Python `remember()` echoes `idempotency_key`, changelogs fold under `## 0.1.7` / `## 0.1.10`, and the production-readiness style-guide hits are gone. The recovery shape is still right.

Do not bump TS (`0.1.7`) or Python (`0.1.10`). Request changes for an accidental `packages/sdk/node_modules` symlink to the author's local absolute path, added in this follow-up.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 1

harrymove-ctrl:
@ducnmm addressed the bug and all four suggestions, head is now `926a87ff`.

**bug (production-readiness):** my rewrite deleted `rememberOnce` but left three call sites and the "relayer timeout is worth retrying" sentence, which contradicted the recovery section directly above it — and since `withRetry` only bails on 4xx, a 504 was retried. Exactly the double-mint this PR exists to close. `rememberOnce` is back, now built on `idempotencyKey`; `withRetry` stops on `isRememberJobTimeoutError(err)` alongside 4xx; the prose says a timed-out write is a second write, not a retry.

**suggestion (two names for one timeout):** converged on yours. The type is `RememberJobTimeoutError` in `types.ts`, matching #881/#890, and the guard keys off `status === 504` + `jobId` rather than `name`, so it matches whichever of the three lands first. `idempotencyKey` / `namespace` sit on that same type.

**suggestion (mutating the error):** the timeout is now constructed with its context instead of being patched afterwards, so the key is in the message and not only in a field.

**suggestion (Python parity):** `RememberAcceptedResult.idempotency_key` added and returned from `remember()` / `remember_async()`, with a test for both the generated and the explicit key. The "SDKs aligned" claim in the PR body holds now.

**suggestion (changelog) + nit (style):** folded under the existing `## 0.1.7` / `## 0.1.10`, mirrored into both docs changelogs, no version bumps. Style pass on `production-readiness.md`: no em dashes, "through" not "via", "might" not "may".

harrymove-ctrl:

```
```

docs/sdk/production-readiness.md
```diff
@@ -37,30 +37,78 @@ Several of these are patterns you implement around the client today. Where a cap
...
+a repeat submission returns the original job instead of minting a second blob,
+which you would also pay for. Omit it and the SDK generates one per call, which
+still covers its own internal retries but gives you nothing to replay with later.
 
 ```ts
 import { createHash } from "crypto";
 
-const written = new Set<string>(); // back this with Redis or a DB in production
+function writeKey(text: string, namespace = "default") {
+  return createHash("sha256").update(`${namespace}:${text}`).digest("hex");
```
ducnmm, line 49: **[bug]** `writeKey` defaults an omitted namespace to `"default"` and joins with `:`. `rememberAndWait` resolves an omitted namespace with `namespace ?? clientNamespace`, and the relayer scopes idempotency to `(owner, idempotency_key)` while fingerprinting `text || NUL || namespace`. A client created with `namespace: "drops"` that calls `rememberOnce(memwal, text)` hashes `default:text` but stores under `drops`; the same call with `"drops"` passed explicitly uses a different key and pays for a second blob. Two clients on one account that both omit the per-call namespace share the `default:` key, and the second request 409s so the write is dropped. `:` is legal in both fields (only empty, overlong, and NUL are rejected), so `("a:b", "c")` and `("a", "b:c")` collide the same way. The statement on line 60 also sits outside any scope that defines `memwal`, `text`, or `namespace`.

**Suggestion:** Resolve the namespace once, the same way the SDK does, and hash with a NUL or a length prefix, matching `request_fingerprint`. Do not default the hash input to `"default"` independently of the client.

packages/sdk/src/memwal.ts
```diff
@@ -389,7 +393,23 @@ export class MemWal {
...
+                throw rememberJobTimeoutError(err.jobId, err.timeoutMs, {
+                    idempotencyKey,
+                    namespace: resolvedNamespace,
+                });
+            }
+            throw err;
+        }
         // Clear only after terminal success. A polling timeout/transport failure
         // keeps the key so retrying the high-level operation reuses the same job.
         if (generatedKey) this.pendingRememberKeys.delete(requestIdentity);
```
ducnmm, line 415: **[suggestion]** This delete runs only when `generatedKey` is true on the call that finishes. The new timeout rethrow leaves `pendingRememberKeys` set, and replaying with `err.idempotencyKey` (the recovery this PR adds) has `generatedKey === false`, so a successful replay does not remove that entry. The next keyless `remember` / `rememberAndWait` of the same text reuses the old key, the relayer returns the original job, and the caller sees success for a write that was not stored. `waitForRememberJob(err.jobId)` never re-enters this function, so the poll recovery leaves the same pin. Keeping the entry so a keyless retry collapses is intentional; the new explicit-key success path never clears it.

**Suggestion:** On terminal success, delete the map entry when it holds the key this call just settled, including an explicit-key replay. Note in the recovery section that a keyless retry on the same client, before that success, is still a collapse.

packages/python-sdk-memwal/memwal/client.py
```diff
@@ -489,11 +490,24 @@ async def remember_and_wait(
...
+            # replay this exact write, and so the key is in ``str(err)`` and not
+            # only in a field: after a restart the log line may be all that is
+            # left (WALM-595).
+            raise MemWalRememberJobTimeout(
+                job_id=timeout.job_id,
+                timeout_ms=timeout.timeout_ms,
+                idempotency_key=resolved_key,
+                namespace=resolved_namespace,
+            ) from timeout
         if generated_key:
```
ducnmm, line 511: **[suggestion]** Same pin as the TypeScript client. The new `raise MemWalRememberJobTimeout(...)` skips this pop, and a later `remember_and_wait(..., idempotency_key=err.idempotency_key)` has `generated_key` false, so success does not pop the entry the timed-out call stored. The next keyless `remember` / `remember_and_wait` of that text resubmits the old key. `remember()` pops on accept, so that next call both collapses onto the original job and then drops the pin, which hides the miss behind a normal `RememberAcceptedResult`.

**Suggestion:** Pop `_pending_remember_keys[request_identity]` when it equals the key that just reached a terminal accept or a successful wait, not only when this call generated the key.

docs/sdk/production-readiness.md
```diff
@@ -69,6 +117,9 @@ async function withRetry<T>(fn: () => Promise<T>, attempts = 4): Promise<T> {
     try {
       return await fn();
     } catch (err) {
+      // A poll timeout is an unknown outcome, not a failure. Retrying the
+      // write here is what stores the memory twice.
```
ducnmm, line 121: **[suggestion]** This comment, and the paragraph above the snippet, say retrying a poll timeout stores the memory twice. The wrapper is applied to `rememberOnce`, which always sends a stable `idempotencyKey`, and the next paragraph says that retry is safe because the relayer returns the original job. Lines 42–43 also say an omitted key covers the SDK's own internal retries, but `signedRequest` is a single `fetch` and this page says the client does not retry. The map only reuses a key when the caller hits the same instance again before terminal success.

**Suggestion:** Say a timeout retry is a second write only when the key changes, and that a stable key collapses. Drop the internal-retry claim, or point it at the same-instance map.

packages/sdk/src/utils.ts
```diff
@@ -438,3 +441,62 @@ export async function delegateKeyToPublicKey(privateKeyHex: string): Promise<Uin
...
+ * Both handles needed to settle it are on the error, so a caller can recover
+ * without issuing — and paying for — a second write:
+ *
+ * - `jobId` → `waitForRememberJob(jobId)` polls the same job.
+ * - `idempotencyKey` → replaying `rememberAndWait(text, ns, { idempotencyKey })`
+ *   returns the original job instead of minting a second blob. This is the one
+ *   that survives a process restart, which the SDK's in-memory key map does not.
+ *
+ * The type lives in `types.ts` as `RememberJobTimeoutError` so this and the
+ * WALM-597 / WALM-600 work describe one error rather than three.
```
ducnmm, line 463: **[suggestion]** This comment says the type lives in `types.ts` so this change and the unmerged WALM-597 / WALM-600 work describe one error rather than three. That coordination note will be stale once those land and does not tell a reader how to handle the error.

**Suggestion:** Delete it. The following note, that detection uses `status === 504` plus `jobId` rather than `name`, is the constraint worth keeping if that stays intentional.

ducnmm (changes requested):
This PR turns a remember poll timeout into a typed unknown outcome on both SDKs, carrying the job id, idempotency key, and namespace so a caller can poll the accepted job or replay it instead of minting a second blob. That matches the relayer, which already collapses POST /api/remember on (owner, idempotency_key) and 409s on a fingerprint mismatch. The new production writeKey helper does not hash the namespace the write actually uses, and a successful explicit-key replay does not clear the in-memory pin left by the timed-out call.

**Issues:** 1 bug, 4 suggestions, 0 nits.

harrymove-ctrl:
Closing: Superseded by commit 1540349e which implemented 30-minute bucketed content-derived idempotency keys (derivedIdempotencyKey) on dev, and WALM-600 (#890) for typed timeout errors.

Issue #365, opened 2026-07-06
[Bug] Auth session expires silently between sessions with no re-auth prompt


zazadra:
## Problem

After authenticating via `memwal_login`, the session appears to expire between separate chat sessions. On the next session, tools like `memwal_recall` silently return empty results instead of returning an authentication error.

## Impact

Users (and agents) cannot distinguish between 'no memories stored' and 'not authenticated'. This leads to hours of debugging where the developer believes their memory writes failed, when in reality the session simply expired.

## Steps to Reproduce

1. Log in via `memwal_login` and write several memories
2. Close the chat session and start a new one
3. Call `memwal_recall` - it returns empty instead of an auth error

## Suggested Fix

Return an explicit error response such as:  
`{ 'error': 'UNAUTHENTICATED', 'message': 'Session expired. Please run memwal_login again.' }`  

This allows the agent (and the user) to handle re-authentication gracefully.

ducnmm:
Thanks for the detailed report — an empty recall result being indistinguishable from an expired session is a real footgun. The team has received this and is investigating the auth/session error surfacing.

nikola0x0:
PR up for the remaining half of this: #894.

There were two halves. The **empty results** — recall returning zero rows rather than an auth error — closed server-side in `@mysten-incubation/memwal-mcp@0.0.11` (`45b0ad87`). Before it, the MCP proxy would open a session for a delegate key that was not registered, and recall then honestly reported nothing found.

What remained is the **client half**, which #894 fixes. Once the relayer rejects the key, the bridge treated the 401 like any other connection failure and kept retrying, leaving your `memwal_recall` queued until a 240s deadline — then answering it with "the connection to the relayer dropped, please retry", which cannot work when the key itself is the problem. It now fails immediately, names the rejection, and points at `memwal_login`, which is close to the response shape you suggested.

One thing that would help confirm we have fully accounted for what you saw: do you recall which version you were on? Anything at or below 0.0.10 would fit both halves.

PR #894, opened 2026-09-10
fix(mcp): tell the client its credentials were rejected (WALM-602)

nikola0x0:
Closes WALM-602. Addresses the client half of [#365](https://github.com/MystenLabs/MemWal/issues/365).

## The bug

When the relayer rejects the saved delegate key, a `memwal_recall` today comes back — after up to **240 seconds** — with:

> ❌ Walrus Memory did not answer this call. The connection to the relayer dropped before the result came back. Please retry.

Retrying cannot work, and nothing in that message mentions auth or `memwal_login`.

`openSseStream` throws a well-worded 401 error, but `connectInBackground` catches it with the same generic `catch` as every other connect failure, backs off, and retries a key the relayer will never accept. The queued call sits in `pendingForward` until the orphan sweeper's deadline closes it out.

## Relation to #365

#365 reported this as *"session expires silently, recall returns empty instead of an auth error."*

The **empty-results half already closed server-side in 0.0.11**: `45b0ad87` made the MCP proxy require a registered delegate. Before it, `classify_and_resolve` returned `Passthrough` before even reading the `Authorization` header, so an unregistered — or entirely absent — key opened a session and `recall` then honestly returned zero rows.

That fix removed the empty results but replaced them with the hang above, so the ticket's first acceptance criterion ("expired session on recall is UNAUTHORIZED / signed-out, not empty results") was still not met from the user's side. This PR is that half.

## The change

- Carry the handshake 401 as `RelayerUnauthorizedError` so both connect paths can distinguish it from a retryable failure.
- On that error, fail everything queued immediately with a message naming the rejection and `memwal_login` — at first connect (`pendingForward`) and at reconnect (`inFlight`), so a key rejected at startup and one revoked mid-session read the same.
- While the key stays rejected, refuse later requests on arrival instead of parking them — the same shape as the existing signed-out refusal, and placed directly after it so `memwal_login` still returns locally and leaves a way back in.
- Any accepted handshake clears the flag and releases the server pump, whichever path published it. That covers a re-login, a re-registered key, and a transient WAF / rate-limit 401 — the last recovers unattended, since the pump keeps driving `reconnect()` on the dead stream.
- A 401 for a key a login has already replaced is ignored, so a stale handshake cannot refuse requests against a live session.
- Credentials are still never wiped automatically — that DoS guard is untouched.

## Tests

`packages/mcp/test/expired-credentials-recall.test.mjs`:

| Case | Expected |
| -- | -- |
| Relayer 401s the handshake | auth error naming `memwal_login`, **not** an empty result |
| Valid creds, namespace has no hits | ordinary empty result, **not** an auth error |
| Second call while still rejected | refused again on arrival, bridge still alive |
| `memwal_login` after a rejection | service restored, and promptly — not on the connect backoff |
| Key revoked with a call in flight | that call answered, **not** left to the orphan sweeper |
| Transient mid-session 401 | recovers with no client intervention |

Written test-first. Case 1 failed against `dev` at **5432 ms** with the wrong "please retry" text; it now answers in **94 ms**.

Each fix was ablated against the suite to confirm none of it is trivially green: reverting the `reconnect()` clear fails on the refusal text; reverting the `signalFirstConnect()` removal times out; reverting the `reconnect()` signal fails the 1500 ms bound at **3908 ms**; reverting the reconnect-401 handling gives back exactly the "please retry" text at **5.4 s**.

Full MCP suite: **66/66**, `tsc` clean.

## Not verified

No live relayer was exercised; all six cases run against stubs. The behaviour under a real revoked key on dev is unconfirmed.


packages/mcp/src/bridge.ts
```diff
@@ -1789,6 +1838,16 @@ export async function runBridge(
...
                 attempt += 1;
                 log.error("bridge.initial_connect_failed", { err: reason, attempt });
+                // A rejected key will not start working on the next attempt, so
+                // answer everything queued instead of leaving it to the orphan
+                // sweeper. Keep looping: `memwal_login` re-registers a key on
+                // this same relayer, and the successful connect above clears the
+                // flag and resumes normal buffering.
+                if (err instanceof RelayerUnauthorizedError) {
+                    credentialsRejected = true;
+                    signalFirstConnect();
```
ducnmm: **[bug]** On `RelayerUnauthorizedError`, `signalFirstConnect()` runs while `sse` is still null and `loggedOut` is false. `serverPump` then hits the "stdin closed before we ever connected" `break`, `runBridge` races that finished pump, and `markStdinClosed()` disables `reconnect()`. The `memwal_login` the error text points at cannot publish a new session. `failPendingForward` already writes stdout; it does not need `firstConnect` signaled.

**Suggestion:** Do not call `signalFirstConnect()` here. Leave `serverPump` parked on `firstConnect` until a real session exists (or stdin actually closes), matching every other retryable connect failure. Add a regression that, after the 401 recall error, a second `memwal_recall` still fail-fasts and a subsequent handshake / `memwal_login` can still be pumped.

packages/mcp/src/bridge.ts
```diff
@@ -1780,6 +1825,10 @@ export async function runBridge(
                 sessionEpoch += 1;
                 sse = candidate;
                 firstConnectDone = true;
+                // A key that was rejected earlier is evidently accepted now
+                // (re-registered, or the 401 was a transient WAF/rate-limit
+                // false positive), so stop failing requests fast.
+                credentialsRejected = false;
```
ducnmm, line 1868: **[bug]** `credentialsRejected` is cleared only on `connectInBackground`'s own publish path. `reconnect()` (line 1032) and `adoptCredentials` → `reconnect("login-credentials-updated")` never clear it, and the `if (sse)` shortcut at line 1798 returns without clearing. After a later login, every `tools/call` still hits the UNAUTHORIZED_FAILURE refusal. Issue 1 currently masks this by tearing the bridge down first.

**Suggestion:** Clear `credentialsRejected` wherever a handshake is accepted — at least next to `sse = candidate` in `reconnect()`, and keep this clear. Pin with a test: 401 recall error, then an accepted handshake / stubbed `memwal_login`, then `memwal_recall` is an ordinary result.

packages/mcp/test/expired-credentials-recall.test.mjs
```diff
@@ -0,0 +1,311 @@
...
+    };
+}
+
+function textOf(msg) {
+    const content = msg?.result?.content;
+    if (!Array.isArray(content)) return "";
+    return content.map((c) => c?.text ?? "").join("\n");
+}
+
+test("recall on rejected credentials reports an auth error, not empty results", async (t) => {
```
ducnmm, line 232: **[suggestion]** Case 1 only asserts that the first `memwal_recall` is an auth error. That stays green when `runBridge` shuts down immediately after `failPendingForward`, and when `credentialsRejected` can never be cleared. Nothing checks a second fail-fast call, that stdin/reconnect stay alive, or that `memwal_login` / a later 200 handshake resumes buffering — the behavior the changelog claims.

**Suggestion:** After the 401 assertion, send a second recall (still the auth error, well under `CALL_TIMEOUT_MS`) and then either flip the stub to 200 or complete a login handoff and assert a non-auth recall.

ducnmm (dismissed):
## Summary

The fail-fast 401 path is the right fix for WALM-602's reported symptom: a handshake rejection now answers `memwal_recall` with an auth error naming `memwal_login`, instead of parking until the orphan sweeper. The recovery story the PR and changelog document is not actually implemented. `signalFirstConnect()` on that 401 lets `serverPump` exit while `sse` is still null, so `runBridge` treats the session as never-connected, sets `stdinClosed`, and disables `reconnect()`. Independently, `credentialsRejected` is only cleared on `connectInBackground`'s own success path, so a later `memwal_login` (which publishes via `reconnect()`) would keep refusing every memory call. The new tests pin the first recall only and stay green through both defects.

## Issue counts by severity

- bugs: 2
- suggestions: 3
- nits: 0

## Issues outside the diff

- **[bug]** packages/mcp/src/bridge.ts:1032 — `reconnect()` publishes `sse` without clearing `credentialsRejected`. `adoptCredentials` then `await reconnect("login-credentials-updated")` so a later login still refuses every memory call. Clear the flag wherever a handshake is accepted (next to `sse = candidate` here, and keep the existing clear in `connectInBackground`).
  - **Suggestion:** Clear `credentialsRejected` in `reconnect()` when assigning `sse`, and in `adoptCredentials` when swapping in a newly registered key.
- **[suggestion]** packages/mcp/src/bridge.ts:1109 — `reconnect()`'s catch still treats handshake 401 as a generic `reconnect_failed`. Mid-session revocation still hangs until the orphan sweeper. On `RelayerUnauthorizedError` here, set `credentialsRejected` and fail in-flight with `UNAUTHORIZED_FAILURE`. If that is a follow-up, say so on the ticket.
- **[suggestion]** packages/mcp/src/bridge.ts:1412 — `initialize` arms `expectSuppressedReply` before the new `credentialsRejected` refusal, unlike logout which returns first. Mirror logout: `if (loggedOut || credentialsRejected) return;` before the suppress arm.

packages/mcp/src/bridge.ts
nikola0x0: Confirmed and fixed in 652024c5. Reproduced it first: the bridge logged `bridge.closed` immediately after the 401, so `reconnect()` was dead and `memwal_login` could never publish.

Dropped the `signalFirstConnect()`; `failPendingForward` writes stdout directly and never needed the pump. The pump now stays parked on `firstConnect` until a session actually exists.

One thing your fix alone would have left: `reconnect()` never signalled `firstConnect` either, so a login-established first session would sit undrained until the background backoff (up to 15s) happened to expire. Added the signal at `reconnect()`s publish site too, and the new test asserts the post-login recall lands inside 1500ms after the backoff has grown to ~4s.

packages/mcp/src/bridge.ts
nikola0x0: Confirmed and fixed in 652024c5 — `credentialsRejected = false` next to `sse = candidate` in `reconnect()`, keeping the existing clear.

Not clearing it in `adoptCredentials`: that clears on *adoption* rather than on *acceptance*, so a login whose reconnect fails would resume parking requests behind a connect that has not succeeded. Clearing at the publish sites covers both callers.

The `if (sse)` shortcut needed the opposite fix. Once both publishers clear the flag, `sse` non-null implies it is already false — but the catch could still latch it from a stale handshake that 401d after a login published a good session, which is how it got reached. Guarded the set with the same staleness test the publish path uses (`!sse && openingGeneration === credentialGeneration`).

packages/mcp/test/expired-credentials-recall.test.mjs
nikola0x0: Fixed in 652024c5. The test now sends a second recall (auth error again, asserted under `CALL_TIMEOUT_MS / 2`), then drives a real `memwal_login` handoff via the preflight/callback pattern from `live-login-credentials.test.mjs`, and asserts an ordinary result afterwards.

You were right that it was green through both defects — it was. Ablated each fix against the new tests: reverting the `reconnect()` clear fails on the refusal text, reverting the `signalFirstConnect()` removal times out entirely, reverting the `reconnect()` signal fails the 1500ms bound at 3908ms.

Also took your third point rather than deferring it: `reconnect()` now recognises `RelayerUnauthorizedError` and answers the replay set with `UNAUTHORIZED_FAILURE`. Two more tests cover it — the in-flight call at revocation time (652ms, against the 4000ms sweeper; reverting the fix gives back exactly the "please retry" text at 5.4s), and a transient 401 recovering unattended, since the server pump keeps driving `reconnect()` on the dead stream.

66/66, tsc clean. Not verified: still no live relayer, all five cases run against stubs.

packages/mcp/src/bridge.ts
```diff
@@ -1789,6 +1875,32 @@ export async function runBridge(
...
+                // key a login already replaced says nothing about the new one,
+                // and latching the flag on it would refuse every request against
+                // a session that is live and fine.
+                if (
+                    err instanceof RelayerUnauthorizedError &&
+                    !sse &&
+                    openingGeneration === credentialGeneration
+                ) {
+                    credentialsRejected = true;
+                    failPendingForward("credentials rejected", UNAUTHORIZED_FAILURE);
```
ducnmm, line 1902: **[suggestion]** `failPendingForward` on this 401 path still leaves the GH #415 suppress arm for an `initialize` that was queued before `credentialsRejected` was set and never POSTed. `failRequest` keeps the arm; after the now-working `memwal_login` recovery, a client that reuses the initialize id has the real reply consumed.

**Suggestion:** `suppressUpstreamReplies.delete` the buffered initialize id when the forward never left the process. The new recovery test uses id=5 after login, so it stays green.

ducnmm (dismissed):
## Summary

Follow-up `652024c5` fixes both previously requested recovery bugs. Handshake 401 no longer calls `signalFirstConnect()`, so `serverPump` stays parked until a real session exists; `reconnect()` clears `credentialsRejected`, signals `firstConnect` when it is the first session, and fail-fasts in-flight calls on a mid-session 401. Tests pin a second fail-fast recall, `memwal_login` restoring an ordinary result, in-flight revocation, and unattended transient recovery.

Dump is still blocking: `packages/mcp/CHANGELOG.md` has the WALM-602 bullet under unpublished `## 0.0.13` (no 0.0.14 bump — origin/dev is already 0.0.13). `docs/mcp/changelog.mdx` `## 0.0.13` and the frontmatter `answer:` still only describe WALM-480.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

## Issues outside the diff

- **[bug]** docs/mcp/changelog.mdx:34 — `## 0.0.13` and `answer:` never mention the 401 / `memwal_login` recovery. Copy the Fixed bullet from `packages/mcp/CHANGELOG.md` and mention it in the 0.0.13 intro and `answer:`. Do not bump `packages/mcp/package.json`.

packages/mcp/src/bridge.ts
nikola0x0: Fixed in de54c7fa: the arms of queued initializes are cleared before `failPendingForward`. It was a hang, not only a lost reply, because the pump also untracks the id and the sweeper never answers. The new test sends id 1 again after login and times out without the fix.

nikola0x0:
@ducnmm Changelog fixed in d7149c03, after merging dev: the WALM-602 bullet is now under `## 0.0.13` in `docs/mcp/changelog.mdx`, and the intro and `answer:` name the fix. No version bump.

ducnmm (commented):
## Summary

Follow-up `de54c7fa` / `d7149c03`: WALM-602 dump is under unpublished `## 0.0.13` in both changelogs (intro/`answer:` included). Queued `initialize` suppress arms are cleared before `failPendingForward` so a reused JSON-RPC id after `memwal_login` is not swallowed; the new test holds the 401 until both requests are queued. No open issues on this head.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

ducnmm (approved):

Issue #322, opened 2026-06-23
[Bug/DX] Fix browser build landmine with Vite and Node crypto dependency


Olayeniii:
### Description
The `memwal` package imports Node's native `crypto` module. When bundled using Vite, Vite externalizes this dependency without warning.

### Impact
The project builds cleanly but crashes at runtime in the browser during the claim flow with `crypto`/`Buffer` errors. Fixing this requires obscure `node-polyfill` configurations that are not obvious upfront.

### Proposed Solution
Either swap out Node-specific crypto dependencies for web-compatible crypto APIs (like `globalThis.crypto`) or clearly document the required Vite polyfill configuration in the README.

ducnmm:
Thanks for the report — this is a good catch. The team has received the info and we'll work on identifying the root cause (Node `crypto`/`Buffer` being externalized by Vite) soon. We'll update this thread once we have a fix or a documented polyfill path.

PR #895, opened 2026-09-10
fix(sdk): stop importing a Node builtin on a browser path (WALM-136)

nikola0x0:
Closes WALM-136 ([GH #322](https://github.com/MystenLabs/MemWal/issues/322)). Also settles WALM-599.

## The bug

`sha256hex` fell back to `await import("crypto")` when WebCrypto was missing. Vite replaces that with a stub: the app builds clean, then throws in the browser the first time the path runs.

It sat on the signed-request path (`memwal.ts:1347`, `manual.ts:854`) — every remember and recall — reached the browser through all four apps, and was the last bare Node-builtin import in `packages/sdk/src`.

The fallback could never have helped a browser: `crypto.subtle` is missing precisely when the page is not a secure context, and `node:crypto` is not there either. It only ever served Node <19 (EOL April 2025) while being the sole source of the exposure. So the branch is removed, not guarded — prefixing to `node:crypto` would just make a dead branch fail louder.

## Reproduced, before and after

Vite 5.4.21, standalone consumer project importing the built `dist`:

| | before | after |
| -- | -- | -- |
| build | succeeds, warns `Module "crypto" has been externalized` | clean |
| `__vite-browser-external` stubs | 1 | 0 |
| run with `crypto.subtle` undefined | `TypeError: crypto.createHash is not a function` | correct digest |

That `TypeError` is the stub being called — #322's crash reproduced end to end, then gone.

## The change

```ts
export async function sha256hex(data: string): Promise<string> {
    const { sha256 } = await import("@noble/hashes/sha2.js");
    return bytesToHex(sha256(new TextEncoder().encode(data)));
}
```

`@noble/hashes` is already a direct dependency here *and* of `@mysten/sui`, and blake2b two functions down already imports it. Nothing for a bundler to resolve, and it works in the non-secure contexts where both old paths failed. Costs 3.67 kB gzipped.

Measured rather than assumed: at a typical 512 B body noble is **faster** than WebCrypto (0.0034 ms vs 0.0101 ms — async overhead dominates below a few KB); +0.77 ms at 256 KB. Stays `async` because the signature is public API.

## Tests

`test/no-node-builtins.test.mjs` scans built `dist/` for any Node builtin specifier, so a future `fs` or `path` import fails too. Scanning output rather than source is the point — this bug builds cleanly.

Test-first: it failed on `dev` with `utils.js imports "crypto"`. SDK suite **119/119**, CI **18/18**.

## WALM-599

Does not reproduce. `globalThis.crypto` has been a default global since Node 19 (verified in both the main thread and a `worker_threads` worker), and `sponsor-auth.test.mjs` passes with no polyfill. What was real: the SDK had no `engines` field while both sibling packages declare `>=20.0.0`. Added here — suggest closing 599 on that basis.

## Follow-up

The SDK resolves `@noble/hashes@2.0.1` while `@mysten/sui` resolves `2.2.0`. Both satisfy `^2.0.0`, so it is a lockfile artifact rather than a range problem, but a bundle can carry two copies. Worth a dedupe on its own.

### DIFF
--- a/packages/sdk/CHANGELOG.md
+++ b/packages/sdk/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Fixed
 
+- Hash request bodies with `@noble/hashes` instead of WebCrypto-or-`node:crypto`, so the package no longer imports a Node builtin on a browser-reachable path. Vite externalises such an import without warning: the app builds clean and the browser crashes the first time the path runs. The fallback could never have helped a browser anyway — `crypto.subtle` is absent precisely when the page is not a secure context, where `node:crypto` is absent too — so it only served Node <19 while being the sole source of the exposure. `sha256hex` sits on the signed-request path, so every remember and recall reached it. (#322, WALM-136)
+- Declare `engines.node >= 20.0.0`, matching `memwal-mcp` and `openclaw-memory-memwal`. The SDK was the only published package without a floor. (WALM-599)
 - Empty-body 401s now use the same AUTH_REJECTED troubleshooting message as credential 401s instead of telling callers to run `memwal_login`. Headless SDK clients do not have that MCP tool.
 - `account.ts` and `manual.ts` PTBs use typed `tx.pure` helpers instead of the legacy untyped moveCall argument syntax that fails under modern `@mysten/sui`.
 
--- a/docs/sdk/changelog.mdx
+++ b/docs/sdk/changelog.mdx
@@ -28,12 +28,12 @@ questions:
   - When was bulk remember added to the Walrus Memory SDK?
   - What security improvements have been made to the MemWal SDK?
 answer: >-
-  The latest TypeScript SDK release is 0.1.7. `restore()` results include `failed` (required like `truncated`; SDK defaults omitted to `0`) for permanent decrypt/UTF-8 failures instead of folding them into `skipped`. Empty-body 401s use the AUTH_REJECTED troubleshooting message instead of telling callers to run `memwal_login`. Account and manual PTBs use typed `tx.pure` helpers so they work with modern `@mysten/sui`. 0.1.6 added optional `created_at` on `recall()` results, plus `sort` and `scoringWeights` on `RecallOptions`, and reports HTTP 503 as a retryable upstream outage instead of a sign-in failure.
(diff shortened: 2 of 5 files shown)


ducnmm (commented):
## Summary

`sha256hex` now hashes with `@noble/hashes/sha2.js` (same subpath style as `blake2b`) and drops `await import("crypto")`. Noble SHA-256 of UTF-8 bytes matches Node `createHash("sha256")` and WebCrypto `subtle.digest("SHA-256")` on `""` / `"hello"` / JSON-with-unicode; known-answer vectors in the new test are the FIPS empty and `"hello"` digests. `tsc` does not bundle, so dist keeps a static `import("@noble/hashes/sha2.js")` specifier — Vite has nothing Node-builtin to externalise. The dist scan would have failed on the old `import("crypto")`. Public `async` signature is unchanged.

Dump is the existing unreleased 0.1.7: `package.json` stays `0.1.7`, verify-script pin untouched, `packages/sdk/CHANGELOG.md` folds under `## 0.1.7` without dropping `restore()` / 401 / `tx.pure`. `docs/sdk/changelog.mdx` (and its FAQ `answer:`) was not updated.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 0

packages/sdk/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Fixed
 
+- Hash request bodies with `@noble/hashes` instead of WebCrypto-or-`node:crypto`, so the package no longer imports a Node builtin on a browser-reachable path. Vite externalises such an import without warning: the app builds clean and the browser crashes the first time the path runs. The fallback could never have helped a browser anyway — `crypto.subtle` is absent precisely when the page is not a secure context, where `node:crypto` is absent too — so it only served Node <19 while being the sole source of the exposure. `sha256hex` sits on the signed-request path, so every remember and recall reached it. (#322, WALM-136)
```
ducnmm, line 11: **[suggestion]** `origin/dev` is already ahead of `origin/main` (TS 0.1.7), so not bumping and folding these bullets under the existing `## 0.1.7` is correct, and `scripts/verify-manual-sdk-release.mjs` stays at `"0.1.7"` because it only asserts that heading exists. The Mintlify twin is not in the diff: `docs/sdk/changelog.mdx` 0.1.7 FAQ `answer:`, section intro, and `### Fixed` still only mention `restore()` `failed`, empty-body 401, and typed `tx.pure`. After merge the package changelog and the docs 0.1.7 notes diverge, so the published SDK changelog omits a browser crash on every signed remember/recall.

**Suggestion:** Copy the two new `### Fixed` bullets into `docs/sdk/changelog.mdx` under the existing `## 0.1.7`, and extend the FAQ `answer:` / section intro so they mention the noble-hash / no-Node-builtin fix (and the `engines` floor if you keep that bullet). Do not bump past 0.1.7.

harrymove-ctrl (approved):
approve

packages/sdk/CHANGELOG.md
nikola0x0: Fixed in 383de720. Copied both bullets into `docs/sdk/changelog.mdx` under the existing `## 0.1.7` verbatim, and extended the section intro and the frontmatter `answer:` so the AI-citation surface names the browser fix and the engines floor. Bullets now diff clean against `packages/sdk/CHANGELOG.md`.

Stayed on 0.1.7 as you said — no bump, `verify-manual-sdk-release.mjs` pin untouched.

Worth noting no CI check compares the two files: `check-docs-code-sync.mjs` and `check-docs-freshness.mjs` both passed with them diverged (I ran all three meta checks locally before pushing, all green). So this only gets caught by review.

Issue #419, opened 2026-07-13
feat: support passing custom suiRpcUrl and packageId in MemWal.create() options


sandman-sh:
### Summary
Currently, the SDK does not easily expose configuring a custom Sui RPC URL or on-chain `packageId` during initialization. The client defaults to public nodes which can be heavily rate-limited or slow, resulting in slow writes or read timeouts.

To switch to a custom low-latency RPC (like the official `https://fullnode.mainnet.sui.io:443`), developers have to bypass the type definitions and monkeypatch internal properties of the class:
```typescript
(memwal as any).serverConfig = {
  packageId: "0xcee7a6fd8de52ce645c38332bde23d4a30fd9426bc4681409733dd50958a24c6",
  network: "mainnet",
  suiRpcUrl: "https://fullnode.mainnet.sui.io:443"
};
```

### Proposed Solution
Add optional `suiRpcUrl` and `packageId` configuration arguments to `MemWal.create()`:
```typescript
const memwal = MemWal.create({
  key: delegateKey,
  accountId: accountId,
  serverUrl: "https://relayer.memory.walrus.xyz",
  suiRpcUrl: "https://my-custom-rpc-endpoint.com",
  packageId: "0x..."
});
```

Issue #302, opened 2026-06-19
@mysten/sui v2.6+/2.17 client rename breaks account ops (createAccount/addDelegateKey) in @mysten-incubation/memwal@0.0.7


EzraNahumury:
**Environment:** `@mysten-incubation/memwal` 0.0.7, `@mysten/sui` 2.17.0, Node 24, mainnet.

**What happens**
`createAccount()` / `addDelegateKey()` from `@mysten-incubation/memwal/account` throw:

```
SuiClient not found. For @mysten/sui v2.6.0+, pass suiClient in opts.
```

The SDK does `const { SuiClient } = await import("@mysten/sui/client")`, but in `@mysten/sui` v2.17 `@mysten/sui/client` no longer exports `SuiClient` — the JSON-RPC client was renamed/moved to `SuiJsonRpcClient` under `@mysten/sui/jsonRpc`. So the internal client construction fails.

**Repro**
1. `npm i @mysten-incubation/memwal@0.0.7 @mysten/sui@^2.17`
2. `import { createAccount } from "@mysten-incubation/memwal/account"`
3. `await createAccount({ packageId, registryId, suiPrivateKey, suiNetwork: "mainnet" })` → throws `SuiClient not found...`

**Workaround that works**
```ts
import { SuiJsonRpcClient } from "@mysten/sui/jsonRpc";
const suiClient = new SuiJsonRpcClient({ url: "https://fullnode.mainnet.sui.io:443", network: "mainnet" });
await createAccount({ packageId, registryId, suiPrivateKey, suiClient, suiNetwork: "mainnet" });
```

**Suggested fix**
- Internally support the v2.17 client (`SuiJsonRpcClient`) as a fallback, or
- pin/document the supported `@mysten/sui` range and show the `suiClient` injection for v2.6+ in the README.

_Reported while building "The Bitter Pundit" for Walrus Sessions 4._

ducnmm:
Hi @EzraNahumury, thanks for the detailed report! This is the same issue as #300 — same root cause (`@mysten/sui` v2.6+ moved the JSON-RPC client to `SuiJsonRpcClient` under `@mysten/sui/jsonRpc`). Closing as a duplicate so we can consolidate tracking in #300. 🙏

Issue #300, opened 2026-06-18
bug(sdk): createAccount/addDelegateKey throw "SuiClient not found" on @mysten/sui v2.6+ despite peer >=2.5.0


PugarHuda:
## Summary

`createAccount()` and `addDelegateKey()` throw `SuiClient not found. For @mysten/sui v2.6.0+, pass suiClient in opts.` on a fresh install, even though `peerDependencies` declares `@mysten/sui >=2.5.0` (which resolves to v2.18). This breaks the documented account-setup flow out of the box.

## Environment

- `@mysten-incubation/memwal@0.0.7`
- `@mysten/sui@2.18` (installed by the `>=2.5.0` peer range)
- Node 20, server-side (Next.js setup script)

## Root cause

In `dist/account.js`, when `opts.suiClient` is omitted the SDK falls back to:

```js
const mod = await import("@mysten/sui/client");
const SuiClient = mod.SuiClient;
if (typeof SuiClient !== "function") {
  throw new Error("SuiClient not found. For @mysten/sui v2.6.0+, pass suiClient in opts.");
}
```

In `@mysten/sui` v2.6+, the JSON-RPC client moved out of `@mysten/sui/client` (it's now `SuiJsonRpcClient` under `@mysten/sui/jsonRpc`), so `mod.SuiClient` is `undefined` and the call throws.

Two extra papercuts:
1. The `suiClient?` field docstring in `types.d.ts` says *"If omitted, the SDK will create one internally"* — which is misleading on v2.6+, where the internal fallback always throws.
2. Because the peer range is `>=2.5.0`, a clean `npm install` pulls v2.18 and hits this immediately — there's no version that satisfies the peer **and** works with the internal fallback.

## Workaround (works)

Construct the client yourself and pass it in:

```ts
import { SuiJsonRpcClient, getJsonRpcFullnodeUrl } from "@mysten/sui/jsonRpc";

const suiClient = new SuiJsonRpcClient({ url: getJsonRpcFullnodeUrl(network) });
await createAccount({ /* ...opts... */, suiClient });
await addDelegateKey({ /* ...opts... */, suiClient });
```

Repro + the exact fix we ship: https://github.com/PugarHuda/dendam (`scripts/setup-memwal.ts`).

## Suggested fix

[...]

PR #896, opened 2026-09-10
fix(sdk): construct JSON-RPC client on @mysten/sui 2.6+ (WALM-164)

HoangDucBach:
## Ticket

WALM-164 — https://linear.app/mysten-labs/issue/WALM-164
GH #300 — https://github.com/MystenLabs/MemWal/issues/300 (dup #302)

## What changed?

- Shared JSON-RPC client construction: `SuiClient` on `@mysten/sui` 2.5, `SuiJsonRpcClient` on 2.6+.
- `createAccount` / `addDelegateKey` / `MemWalManual` use it when the caller omits `suiClient`. Relayer-mode `MemWal` uses the same resolver for its JSON-RPC fallback.
- Passing an explicit `suiClient` still wins. The "pass suiClient for v2.6+" error is gone for the supported peer range.

## Why is this needed?

Peer `@mysten/sui >=2.5.0` installs 2.6+, which no longer exports `SuiClient` from `@mysten/sui/client`. Account setup and `/manual` still looked only there, so omitting `suiClient` always threw even though the docs say the SDK creates a client internally.

This is the JSON-RPC rename, not a gRPC migration. Relayer-mode still prefers gRPC when `GET /config` advertises it.

## Test plan

- [ ] SDK tests, including `packages/sdk/test/sui-jsonrpc-client.test.mjs`
- [ ] `createAccount` / `addDelegateKey` without `suiClient` on `@mysten/sui` >= 2.6
- [ ] `MemWalManual` default client is `SuiJsonRpcClient`
- [ ] Explicit `suiClient` is used as-is


nikola0x0:
**Blocker before merge:** the two endpoints this constructs against no longer serve JSON-RPC.

```
curl -s -X POST https://fullnode.mainnet.sui.io:443 \
  -d '{"jsonrpc":"2.0","id":1,"method":"sui_getChainIdentifier","params":[]}'

{"error":{"code":-32601,"message":"Method not found. JSON-RPC on public
fullnodes has been deprecated. Please migrate to gRPC or GraphQL endpoints."}}
```

Same on testnet. Those are the URLs in `sui-jsonrpc-client.ts`, so after this merges `createAccount()` without a `suiClient` constructs fine and then fails on the first call — WALM-164 AC1 ("succeed without a caller-supplied `suiClient`") isn't reachable this way. The new tests pass because they mock the loader and never make a call.

Not your miss: WALM-164 was filed 2026-07-07, right at the sunset, and specifies *"detect/build `SuiJsonRpcClient` the same way `memwal.ts` does."* You built exactly that — the spec went stale while the ticket sat in Backlog.

**What I'd change**

- Default to `SuiGrpcClient`. Same hostnames; `apps/noter/lib/sui/grpc-client.ts` is the in-repo pattern.
- Keep JSON-RPC behind a caller-supplied URL — self-hosted nodes still serve it, and prod uses `rpc-mainnet.suiscan.xyz`.
- That also closes #419 (custom `suiRpcUrl`) — same constructor block.
- One test that isn't fully mocked, so a client pointed at a dead endpoint can't go green again.

Your structure holds. All three call sites already take either generation (`account.ts:147` sends both option dialects, `manual.ts:681` documents the gRPC response shape), so this is the transport constant, not a rewrite.

Also needs a rebase — `docs/sdk/changelog.mdx` and `packages/sdk/CHANGELOG.md` conflict.

ducnmm (approved):
LGTM, cleanly resolves JSON-RPC client fallback on @mysten/sui 2.6+ with full test coverage.

nikola0x0 (changes requested):
Still json rpc is deprecated so please adjust according to my comment. TY!

PR #900, opened 2026-09-11
fix(mcp,relayer): stop refusing valid MCP credentials during a Sui throttle, and stop hiding it (WALM-618, WALM-386)

harrymove-ctrl:
Closes WALM-618. Supersedes #869, #901 and #903, whose commits are carried here unchanged.

Relayer and bridge in one PR because they are one failure: the relayer refuses a caller whose credentials are fine, and the bridge waits out the refusals without saying anything. Release still splits — deploy the relayer, publish the npm package — but the two halves only make sense read together.

Measured on production while writing this — `relayer.memory.walrus.xyz`, `GET /api/mcp/sse`, cumulative:

| status | count | share |
| --- | --- | --- |
| 200 | 50,958 | 4.6% |
| 401 | 784,627 | 70.1% |
| 429 | 128,062 | 11.4% |
| 503 | 155,874 | 13.9% |

Six handshakes with a valid, registered delegate key — one request each, 20s apart — **all six failed**, alternating 429 and 503.

## 1. Cache the verification (original commit)

`auth.rs::resolve_account` and `mcp_proxy.rs` both re-read the account object on every request, so one `memwal_remember` through MCP was ~10 `GetObject` calls. `GetObject` volume (1,147,414) tracks total SSE requests (1,119,744) almost 1:1. Caching the positive result for 30s collapses a burst carrying the same credentials into one read.

## 2. Serve a known-good key while Sui is unreachable

The 30s TTL only helps a caller who repeats inside 30s, and real MCP usage is not that shape — a tool call every few minutes misses every time. So during a sustained throttle the cache did nothing for exactly the users it was meant to protect.

`VerifyCacheMissAction::Keep` already survived that case, but nothing could read the entry it kept: the lookup filters on `is_fresh()`, so by the time `Keep` ran the entry was already unservable, and the sweeper deleted it at the TTL anyway. Keeping it was a no-op.

This gives it a reader. On an **unavailable** error only, a cached success within TTL + `DELEGATE_VERIFY_STALE_GRACE` (10 min) is served and logged at `warn`. The sweeper retains on the same bound.

Revocation latency is unchanged at 30s whenever the chain answers; it stretches to 10 minutes only while the chain cannot be read at all — a window in which the relayer could not have observed a revoke anyway. A definitive rejection still returns and still evicts.

## 3. Stop a refused key costing a read per retry

Caching only successes left the larger half uncached. The 784,627 × 401 on this route are not 784,627 people: a bridge holding a key the relayer will never accept treats the 401 as a transient connect failure and retries on a backoff capped at 15s, forever — each retry one more `GetObject` against the fullnode that is already throttling. That throttle is what produces the 503s **everyone else** sees, valid credentials included.

A definitive rejection is now remembered for 10s. #894 is the client half of the same loop; this half needs no client change, so it also covers the versions already installed on users' machines.

Bounded deliberately, because unlike the positive cache this map is keyed by what callers send rather than by what exists on chain:

- **10s**, an order of magnitude under the 30s success TTL. A stale positive authenticates a revoked key; a stale negative only delays one that just became valid.
- **The positive lookup runs first**, so a key registered after being refused works on its next successful verify instead of waiting out the TTL. A successful verify also drops any rejection for the pair.
- **Hard cap of 4,096 entries.** At the cap a new pair is not recorded and falls through to the live read — today's behaviour — rather than trading a throttle for unbounded memory. The sweeper expires on the same TTL so ordinary churn never reaches it.

An ordinary `memwal_login` is unaffected: it registers a freshly generated key, so the pair has never been rejected and has no entry.

## 3b. Not here: the `ip_active_cap` 429s

Real — 128,062 on this route, and six consecutive handshakes with a valid registered key failed while writing this — but the fix is an ops change, and the environment map matters more than it looks. Measured 2026-09-12:

| hostname | front | commit | client default? |
| --- | --- | --- | --- |
| `relayer.memory.walrus.xyz` | Railway AS400940, no CDN | `01aa559b` | **yes — prod** |
| `relayer.memwal.ai` | Cloudflare AS13335 | `01aa559b` | no (alias) |
| `relayer-staging.memory.walrus.xyz` | Railway AS400940, no CDN | `32c22759` | yes — staging |
| `relayer.staging.memwal.ai` | Cloudflare AS13335 | `32c22759` | no (alias) |
| `relayer.dev.memwal.ai` | Cloudflare AS13335 | `493c9e66` | yes — dev |

**Production is not behind Cloudflare**, so WALM-618's fix #1 — set `TRUSTED_PROXY_HOPS` on the prod relayer, still unset — is the right lever for the hostname Dio and Teo actually use ([`index.ts:39`](packages/mcp/src/index.ts)). The "no hop count recovers the client" result recorded during this investigation was measured against **dev**, which is Cloudflare-fronted, and does not transfer. WALM-618 also records staging as Cloudflare-fronted; it resolves to Railway today.

**Each deployment answers on two hostnames with different chain shapes.** A single global hop count cannot be right for both, and — the reason a `CF-Connecting-IP` commit was written and then dropped from this PR — a trusted-header approach would be outright forgeable here, because the Railway hostname stays reachable without passing through Cloudflare. That is GH #360 / WALM-289 reopened. Closing the direct hostname is a precondition for any header-based approach; until then `TRUSTED_PROXY_HOPS`, measured on prod, is both safer and sufficient.

## 4. Log a rejected delegate key

401 is 70% of traffic on this route and 73% of everything that is not a 200, and its only trace was `tracing::debug!` — dropped by the default `memwal_server=info` filter. The route records nothing in `memwal_errors_total` either. Raised to `warn!` with `account_id`; the sibling `Unavailable` arm already logged at `warn`, so the two failure modes are now symmetric. The key itself is never logged.

## 5. Tell the user, instead of waiting in silence (was #869, #901)

The relayer fixes above remove causes. These two remove the part Dio and Teo actually experienced.

**A refused handshake is a rate limit, and is treated as one.** The relayer answers 429 when the per-IP concurrent-session cap is full, and — measured — sends **no `Retry-After`** with it, because that cap clears when someone else's session closes rather than on a timer. The bridge retried ~500ms later, forever, and said nothing, so a rate limit read as a broken config or bad credentials. It now honours `Retry-After` when present (clamped to 60s), falls back to a 5s floor when it is not, carries the deadline across the separate `reconnect()` calls that would each otherwise restart from a fresh sub-second backoff, and prints one stderr line per throttle episode naming it as a rate limit.

**A request buffered while the handshake keeps failing gets answered.** It used to sit in `pendingForward` for the full 240s call timeout and come back as *"the connection to the relayer dropped, please retry"* — a message about a lost reply, for a request that had never been sent. After 90s of consecutive handshake failures the bridge now answers it with the reason the handshake actually gave. A request that was never sent cannot have executed, so failing it early is provably a no-op; the full deadline still applies whenever the handshake is healthy. Override with `MEMWAL_MCP_STALLED_HANDSHAKE_MS`.

## 6. Make the next one findable

This took two days to attribute, and the reason is worth fixing on its own: **the relayer's telemetry cannot represent a slow connect.** There is no slow request. A 3–8 minute wait is N handshakes of 10–30ms each, separated by client-side backoff the server never sees. Three concrete gaps, all closed here:

| gap | before | now |
| --- | --- | --- |
| nothing tied retries together | request id minted per request — 40 retries, 40 unrelated lines | bridge sends one `x-memwal-connect-id` per episode across every retry; `grep` returns the whole episode |
| no measure of the wait | per-request latency histogram, which is healthy by construction | `memwal_mcp_time_to_session_seconds`, observed on the attempt that succeeds |
| refusals were anonymous | 4 of 5 refusal paths logged nothing; none touched a metric | one path logs account, connect id, client, bridge version, stable reason code; `memwal_mcp_handshake_total{outcome,reason}` |
| `/api/mcp/*` missing from errors | `record_app_error` fires from `AppError::into_response`; this proxy returns raw tuples, so 784,627 errors incremented nothing | refusal and unavailable paths record |

`x-memwal-bridge-version` is new on the handshake because `x-memwal-client` comes from `initialize`, which a failed handshake never reaches — and a first connect happens before stdin is wired, so on the attempt that matters most the relayer had no idea who was calling. Which bridge build is looping is the actionable half, and the bridge always knows that.

Episode tracking is bounded like the caches beside it — keyed by a caller-supplied value, so 4,096 max, swept at 15 minutes, and at the cap new episodes are simply not timed.

## What this does not fix

The 784,627 × 401 are a client retry loop against credentials the relayer will never accept, and each one still costs a `GetObject` — that is what drives the fullnode into throttling in the first place. #894 is the fix for that loop and is the higher-leverage change. Nothing here is a substitute for it.

Nothing here helps production before it is deployed, either. The only lever available today is raising `MCP_MAX_SESSIONS_PER_IP` on the running service, which does not depend on any of this.

## Tests

`cargo test -p memwal-server` → **1,197 passed, 0 failed**.
`pnpm --filter @mysten-incubation/memwal-mcp test` → **81 passed, 0 failed**, including `sse-handshake-429.test.mjs` and `pending-forward-stalled.test.mjs` side by side.

New relayer coverage:

| | |
| --- | --- |
| `verify_delegate_key_cached_serves_stale_entry_while_chain_unavailable` | a valid key inside the grace survives a throttle |
| `verify_delegate_key_cached_refuses_stale_entry_past_the_grace` | the grace is bounded |
| `verify_delegate_key_cached_reverifies_after_ttl_expiry` | a failed re-verify never refreshes the timestamp |
| `stale_grace_is_only_reachable_through_the_unavailable_branch` | a definitive rejection can never ride the grace |
| `stale_grace_outlives_the_ttl_so_the_sweeper_has_something_to_serve` | pins the sweeper bound against regression |
| `a_remembered_rejection_is_reused_without_touching_the_chain` | the retry loop stops costing fullnode reads |
| `an_expired_rejection_goes_back_to_the_chain` | the negative window is bounded |
| `a_registered_key_is_never_held_back_by_an_older_rejection` | positive lookup ordering |
| `the_rejection_cache_cap_bounds_what_callers_can_grow` | the map cannot be grown by arbitrary callers |
| `a_rejection_is_forgotten_sooner_than_a_success_is_trusted` | pins the TTL asymmetry |


services/server/src/mcp_proxy.rs
```diff
@@ -191,7 +195,7 @@ async fn legacy_delegate_registered(
             McpAuthOutcome::Unavailable
         }
         Err(err) => {
-            tracing::debug!("mcp delegate rejected: {err}");
+            tracing::warn!(account_id = %account_id, error = %err, "mcp delegate rejected");
```
ducnmm: **[suggestion]** This arm is now `warn!` on every definitive MCP reject, with `account_id` taken from the unauthenticated `x-memwal-account-id` header. `/api/mcp/*` has no rate limit ahead of `legacy_delegate_registered`, rejections are intentionally not cached, and the commit cites 784,627 `/api/mcp/sse` 401s vs 50,958 200s — all of which the commit says already exited this arm. Unlike the sibling `Unavailable` warn (which this cache should make rare), a stuck or probing client with a parseable hex bearer and any account id will emit one warn per SSE open and per JSON-RPC envelope, forever. `memwal_errors_total` still does not count these 401s, so the log line is the only new signal and it is unsampled.

**Suggestion:** Keep a metric (`mcp_delegate_rejected_total`) as the always-on counter, and either leave the per-request line at `debug!` or rate-limit the warn (once per `account_id` per N seconds). Do not log the raw header on every 401 of an unauthenticated route.

services/server/src/storage/sui.rs
```diff
@@ -251,6 +251,150 @@ pub async fn list_delegate_keys_cached(
...
+        rpc_url,
+        grpc_client,
+        account_object_id,
+        public_key_bytes,
+        expected_type_origin_package_id,
+    )
+    .await
+    {
+        Ok(owner) => {
+            cache.write().await.insert(
```
ducnmm: **[suggestion]** A live `Ok` always `insert`s with `verified_at = Instant::now()`, and a live definitive error `remove`s the pair, with no generation / start-time check. The wrapper's own docs (lines 329–331) say a definitive rejection evicts so "an observed revoke cannot be overtaken by a positive still inside its window." That does not hold for the concurrent cold-misses the PR explicitly leaves un-single-flighted: request A starts `GetObject`, request B later observes `KeyNotFound` and evicts, then A's in-flight success completes and re-opens a 30s window. The same race in the other direction (in-flight `KeyNotFound` after a concurrent success) can also drop a still-valid fresh entry. Sequential tool-call bursts are fine; overlapping MCP envelopes, remember + poll, or retries during a slow fullnode are not.

**Suggestion:** Single-flight per `(account, key)`, or stamp a generation that `Evict` bumps and `Ok` insert refuses if stale. At minimum, set `verified_at` to the verify *start* so a slow `GetObject` cannot extend the stated 30s revocation bound by the RPC duration.

services/server/src/storage/sui.rs
```diff
@@ -251,6 +251,150 @@ pub async fn list_delegate_keys_cached(
...
+/// What a failed live verification means for any cached entry on the same
+/// `(account, key)` pair. Pure so the policy is testable without a chain.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum VerifyCacheMissAction {
+    /// The rejection is definitive (revoked, deactivated, wrong object) —
+    /// drop the pair so a positive still inside its TTL cannot outlive the
+    /// revoke we just observed.
+    Evict,
+    /// An unavailable RPC proves nothing about the key, so leave the entry
+    /// alone. It is expired anyway — a live entry would have been served
```
ducnmm: **[suggestion]** `Keep`'s doc says an unavailable RPC can leave the entry alone because "It is expired anyway — a live entry would have been served before the call was made." That is true only for the thread that just missed. `Keep` vs `Evict` is what stops a concurrent *fresh* insert (another request that verified successfully after this miss check) from being deleted when this request's live read returns `RpcError`. The accompanying test seeds a *stale* entry and asserts it is still present, which does not affect serving (`is_fresh` is already false). Policy (`is_unavailable` → Keep, else Evict) is the right WALM-429 split; the comment and test do not pin the case that actually needs it.

**Suggestion:** Rewrite the `Keep` doc around the concurrent-success case, and add a test that inserts a fresh entry between miss and `RpcError` (or documents that `Keep` is load-bearing for that race). The stale-entry keep test can stay as a memory/sweep assertion, not as a security one.

services/server/src/auth.rs
```diff
@@ -446,7 +446,8 @@ async fn resolve_account(
         // fail closed with 503 so a revoked key cannot ride a 24h cache
```
ducnmm: **[suggestion]** Strategy 1 still says to re-verify on-chain when Sui is reachable and to "fail closed with 503 so a revoked key cannot ride a 24h cache through a Sui outage." After this change a fresh `delegate_verify_cache` hit returns `Ok(owner)` without touching Sui, so `cache_reverify_action` authenticates through an outage for up to 30s. That is the documented WALM-618 trade, and Postgres eviction on a later live `KeyNotFound` still matches `cache_reverify_action`, but the comment now describes the pre-change behavior. A reader of `resolve_account` would think WALM-429 fail-closed still applies on every cache hit.

**Suggestion:** Update the Strategy 1 comment to: trust a fresh in-memory verify for 30s (including during an outage); on a live miss, 503-and-keep-Postgres on `is_unavailable`, evict both caches on a definitive miss.

services/server/src/storage/sui.rs
```diff
@@ -251,6 +251,150 @@ pub async fn list_delegate_keys_cached(
     Ok(keys)
 }
 
+// ============================================================
+// Delegate key verification — short-TTL in-memory result cache
+// ============================================================
+//
+// Every authenticated path re-verified the delegate key on-chain on every
```
ducnmm: **[suggestion]** The new module comment (and the MCP call-site comment at `mcp_proxy.rs:178`) narrates WALM-618 history: previous per-request `GetObject`, the ~10-read tool-call breakdown, the 503 retry amplifier, the 200 verifies/min measurement. That is ticket context, not an invariant the next reader needs. Same class of comment #882 was asked to trim. `is_fresh`, the pair key, "successes only", and the 30s revocation bound are the useful bits and are already on the const / type.

**Suggestion:** Drop the incident write-up. Keep one or two sentences: positive verifies are trusted for `DELEGATE_VERIFY_CACHE_TTL`; only `Ok` is stored; `is_unavailable` does not evict.

services/server/src/storage/sui.rs
```diff
@@ -251,6 +251,150 @@ pub async fn list_delegate_keys_cached(
...
+pub async fn verify_delegate_key_cached(
+    cache: &DelegateVerifyCache,
+    http_client: &reqwest::Client,
+    rpc_url: &str,
+    grpc_client: Option<&sui_rpc::Client>,
+    account_object_id: &str,
+    public_key_bytes: &[u8],
+    expected_type_origin_package_id: &str,
+) -> Result<String, OnchainVerifyError> {
+    let key = (account_object_id.to_string(), public_key_bytes.to_vec());
```
ducnmm: **[nit]** Every call, including a hit, allocates `(account_object_id.to_string(), public_key_bytes.to_vec())` just to probe the map. This is now on both the signed-request hot path and every MCP envelope. #882 hit the same pattern and switched the lookup to a borrowed key.

**Suggestion:** Probe with a borrowed `(&str, &[u8])` (or the same `dyn` key trick as #882) so a hit does not allocate; keep the owned key only for insert/remove.

ducnmm (commented):
## Summary

This is a sound 30s success-only verify-result cache, keyed by `(account object id, public key)`, shared by signed-request `resolve_account` and the MCP proxy so a burst of the same credentials collapses to one `GetObject`. `verify_cache_miss_action` matches WALM-429 / `cache_reverify_action` (`is_unavailable` → Keep, definitive miss → Evict), only `Ok` is stored so the unauthenticated MCP header cannot grow the map, and `AppState` is constructed only in `main.rs` so the new field is wired. The 30s revocation bound and Postgres interaction are correct on the sequential path (Strategy 1 still ignores `_cached_owner` and uses the verify-cache owner; a live `KeyNotFound` evicts both maps). Dominant leftovers: the new per-reject `warn!` on an unrate-limited MCP path that already produced hundreds of thousands of 401s, and an unconditional `Ok` insert that can resurrect a pair a concurrent request just evicted.

## Issue counts by severity

- bugs: 0
- suggestions: 5
- nits: 1

services/server/src/mcp_proxy.rs
harrymove-ctrl: Agreed, and this was the right call to flag — left as it was, it trades a latency problem for a log-volume one on an unauthenticated route.

Fixed in 73cee0ec:

- **The counter is now the always-on signal**, as you suggested: `memwal_mcp_handshake_total{outcome,reason}`, plus `memwal_errors_total`. You were also right that this route reached neither — `record_app_error` only fires from `AppError::into_response` and this proxy returns raw status tuples, so 784,627 errors incremented nothing. Both now record unconditionally, before any sampling decision.
- **The line is sampled** at most once per `account_id` per 60s, bounded at 4,096 accounts with expiry-on-insert so the map cannot be grown by the same unauthenticated header it is keyed on.
- **The raw on-chain error dropped back to `debug!`.**

While here I also routed the other four refusal paths through the same place. Only this arm logged anything before; `no_bearer`, `no_account_header`, `malformed_delegate_key` and the OAuth refusals all returned `Unauthorized(None)` silently, so the reason a 401 happened was never recorded at all.

Pinned by `one_account_is_logged_once_per_window_however_hard_it_retries`, `the_sampling_map_cannot_be_grown_without_bound`, `expired_accounts_free_their_slot_for_a_new_one`, `accounts_are_sampled_independently`.

services/server/src/storage/sui.rs
harrymove-ctrl: This one was a real bug, not a style point — thank you. Both halves fixed in 73cee0ec.

**The stamp.** `verified_at` now comes from an `Instant` taken *before* the read, so a slow `GetObject` can no longer extend the stated 30s revocation bound by its own duration. That was silently breaking the guarantee the doc makes two lines above it.

**The resurrect.** The cache now carries an eviction generation (`AtomicU64`, bumped by every definitive evict). The generation is read before the chain call and compared after; an `Ok` insert is skipped when it moved. The caller still gets its answer — the read did succeed — we just do not cache a result the chain has since contradicted.

I did not single-flight. The generation guard is cheaper, needs no per-key map of in-flight futures, and covers the direction that actually breaks the security property (a stale success re-opening a window on a revoked key). The opposite direction you mention — an in-flight `KeyNotFound` dropping a fresh entry — cannot happen, because `Evict` only runs on a definitive error and that error *is* the observation; the case where it would be wrong is an unavailable RPC, which is `Keep`. See the reply on that thread.

Pinned by `may_store_verification` and `a_verification_overtaken_by_an_eviction_is_not_stored`.

services/server/src/storage/sui.rs
harrymove-ctrl: Right on both counts, and the test criticism was fair — seeding a stale entry asserts nothing about serving.

**Doc** rewritten in 73cee0ec. `Keep` is now documented as load-bearing in two distinct ways: the concurrent-fresh-insert case you describe, and the stale-grace path this PR adds (an entry past its TTL is what gets served while the chain is unreachable, so evicting here would delete exactly what that path exists to use).

**Test** added in 75803ef9: `an_unavailable_error_never_evicts_so_a_concurrent_success_survives`. I pinned it at the policy level rather than by interleaving — `Keep` is unconditional, so the policy *is* the guarantee, and a timing-dependent test would be slower and flakier without proving anything the assertion does not. Say the word if you would rather see the real interleaving and I will add a seam for it.

The stale-entry test stays, reframed as the memory/sweep assertion you suggested.

services/server/src/auth.rs
harrymove-ctrl: Fixed in 73cee0ec, close to your wording: a fresh verify-cache hit answers without touching Sui — including during an outage — so the fail-closed rule now applies to a live read rather than to every request, with the 30s window named as the revocation bound. The `is_unavailable` → 503-and-keep-Postgres / definitive → evict-both split is spelled out.

services/server/src/storage/sui.rs
harrymove-ctrl: Done in 73cee0ec — both the module comment and the `mcp_proxy.rs` call site, and I trimmed the rejection-cache comment I added in the same pass before it grew the same way. What is left is the invariants: positive verifies are trusted for `DELEGATE_VERIFY_CACHE_TTL`, only `Ok` is stored, keyed by `(account_object_id, public_key_bytes)`, and `is_unavailable` does not evict.

The measurements moved to the PR description, where they are context for the review rather than something the next reader of this file has to scroll past.

services/server/src/storage/sui.rs
harrymove-ctrl: Deferring this one, and flagging it rather than quietly skipping it.

Probing without the allocation needs either `raw_entry` (still unstable) or the `dyn` key trick #882 is introducing. Landing a second spelling of that here means two different idioms for the same problem in the same crate, and whichever merges second gets rewritten anyway. I would rather it go in once, alongside #882's version, on both maps at the same time — this PR adds a second cache with exactly the same key shape, so they should be fixed together.

Happy to do it here instead if you would rather it not wait on #882.

harrymove-ctrl:
@ducnmm all six addressed — replies are on the individual threads. Two of them changed behaviour rather than wording: the refusal line is now sampled with the counter as the always-on signal, and the verify cache carries an eviction generation so a completed read cannot resurrect a pair a concurrent request evicted.

**Heads-up: the PR grew after you reviewed it, and not by a little.** You reviewed the verify cache alone. It now also carries #869, #901 and #903 (closed into this one), plus the relayer-side fixes that came out of measuring production. Re-reviewing from scratch is fair to ask for; a short map of what is new:

| | |
| --- | --- |
| serve a stale verification while Sui is unreachable | the 30s TTL only helps callers who repeat inside 30s, which MCP clients do not — so during a throttle a valid key took a 503 per attempt |
| remember definitive rejections for 10s | 784,627 × 401 were costing one `GetObject` each; that load is what throttles the fullnode and produces everyone else's 503s |
| bridge: honour `Retry-After`, 5s floor when absent | production sends 429 with no `Retry-After` — measured |
| bridge: answer a request buffered behind a failing handshake after 90s | it used to wait the full 240s and then blame a dropped reply |
| connect-episode id, reason codes, handshake metrics | a 3–8 minute wait is N fast requests with nothing linking them; nothing in the relayer could represent it |

**Measured end-to-end**, same stub relayer refusing handshakes the way production does (429 without `Retry-After` / 503 with `Retry-After: 5`), same MCP client, only the bridge build differs:

| | before | after |
| --- | --- | --- |
| time until the agent got an answer | 240.2s | **95.9s** |
| handshake attempts the relayer absorbed | 20 | **10** |
| answer | *"the connection to the relayer dropped… please retry"* | names the throttle, says nothing was stored, points at the next step |
| time until the user saw a readable explanation | never | **0.9s** |

The halved attempt count is the part that matters beyond one user: today a failing bridge retries harder the longer it fails, which feeds the cap and the Sui quota that caused the failure.

Tests: 1,208 Rust, 81 MCP, 0 failures.

**Not in this PR, and still the largest single lever:** `TRUSTED_PROXY_HOPS` is unset on the prod relayer, which is what puts every user in a handful of ingress buckets against `MCP_MAX_SESSIONS_PER_IP=16`. Worth measuring on prod specifically — prod and staging answer straight from Railway with no CDN, while dev sits behind Cloudflare, so a value measured on dev does not transfer.

services/server/src/storage/sui.rs
harrymove-ctrl: Done here rather than deferred — `7979132d`.

Keyed both maps through a `dyn DelegateAccountKey` borrow, the same shape #882 uses for its `(String, String)` pair, adapted to `(String, Vec<u8>)`. A hit allocates nothing; the owned key is built only on the paths that write. The rejection cache added in this PR shares the key shape, so one trait covers both and there is no second spelling to reconcile when #882 lands.

Added `a_borrowed_probe_finds_the_key_an_owned_insert_wrote`, covering lookup, both near-miss cases, and removal. The reason it is worth an explicit test: if `(String, Vec<u8>)` and `(&str, &[u8])` ever hashed differently the failure is silent — no error, no panic, just a cache that never hits and a `GetObject` per request, which is precisely the state this PR exists to get out of.

packages/mcp/CHANGELOG.md
```diff
@@ -18,6 +18,8 @@
 
 ### Fixed
 
+- Back off when the relayer refuses the SSE handshake with HTTP 429 instead of retrying ~500ms later. The bridge now honours a `Retry-After` (clamped to 60s) and falls back to a 5s floor when the header is absent — the `ip_active_cap` shape — and carries the deadline across reconnect attempts. It also prints one stderr line saying this is a rate limit rather than a bad config or bad credentials, so a throttled bridge no longer reads as a broken one. (WALM-386)
```
ducnmm: **[bug]** WALM-386 and WALM-618 were inserted under already-published `## 0.0.12` (`main` / npm is 0.0.12). `origin/dev` already has unpublished `## 0.0.13` and `packages/mcp/package.json` is already 0.0.13 — this PR correctly does not bump, but the new notes will be attributed to a released version and omitted from 0.0.13.

The same two bullets are under `## 0.0.12` in `docs/mcp/changelog.mdx`. This rebase also dropped the four #633 bullets already on `origin/dev` under 0.0.13 (`CHANGELOG.md` lines 11–14 and the matching mdx Fixed list / answer / release blurb).

**Suggestion:** Move both new bullets into the existing unpublished `## 0.0.13` Fixed sections in `packages/mcp/CHANGELOG.md` and `docs/mcp/changelog.mdx`. Restore the #633 bullets and the 0.0.13 sign-in wording in the mdx `answer:` / release blurb. Leave 0.0.12 and the package version untouched.

docs/mcp/changelog.mdx
```diff
@@ -53,6 +53,8 @@ This release forwards the MCP client's identity to the relayer so sidecar logs c
 
 ### Fixed
 
+- Back off when the relayer refuses the SSE handshake with HTTP 429 instead of retrying ~500ms later. The bridge now honours a `Retry-After` (clamped to 60s) and falls back to a 5s floor when the header is absent — the `ip_active_cap` shape — and carries the deadline across reconnect attempts. It also prints one stderr line saying this is a rate limit rather than a bad config or bad credentials, so a throttled bridge no longer reads as a broken one. (WALM-386)
```
ducnmm: **[bug]** Same dump as `packages/mcp/CHANGELOG.md`: WALM-386 / WALM-618 sit under published 0.0.12. `## 0.0.13` above also lost the four #633 bullets (and the sign-in sentence in `answer:` / the release blurb) that are already on `origin/dev`.

**Suggestion:** Move these two bullets to 0.0.13 Fixed, restore the #633 entries and blurb from `origin/dev`, leave 0.0.12 alone.

services/server/src/storage/sui.rs
```diff
@@ -1600,6 +1972,577 @@ mod tests {
...
+            DELEGATE_VERIFY_CACHE_TTL + std::time::Duration::from_secs(1),
+        )
+        .await;
+        seed_verify_cache(&cache, "0xfresh", &sample_pk(), std::time::Duration::ZERO).await;
+
+        // Mirrors main.rs's sweep task body verbatim. The sweep threshold is
+        // the TTL itself, not a second longer one: an entry past the TTL can
+        // never be served again (`is_fresh` is the same predicate the lookup
+        // uses), so holding it would cost memory for nothing.
+        cache.entries.write().await.retain(|_, v| v.is_fresh());
```
ducnmm: **[suggestion]** `delegate_verify_cache_sweep_drops_everything_past_its_ttl` retains with `v.is_fresh()` and claims to mirror `main.rs` verbatim, including that an entry past the TTL can never be served again. Production sweep uses `is_servable_while_unavailable()` (TTL + 10min grace) on purpose; a later "align the test with main" copy of `is_fresh()` into the sweeper would evict exactly the entries the outage path exists to serve.

**Suggestion:** Change this test to `retain(|_, v| v.is_servable_while_unavailable())`, seed a TTL-expired-but-in-grace entry that must survive, and drop the "mirrors main.rs verbatim / past TTL can never be served" comment.

services/server/src/mcp_proxy.rs
```diff
@@ -185,37 +329,53 @@ async fn legacy_delegate_registered(
     )
     .await
     {
-        Ok(_) => McpAuthOutcome::Passthrough,
+        Ok(_) => {
+            crate::observability::record_mcp_handshake("ok", "none");
```
ducnmm: **[suggestion]** `record_mcp_handshake("ok", "none")` lives in `legacy_delegate_registered` / `classify_and_resolve`, which `messages_proxy` and `streamable_proxy` also call. Every authenticated JSON-RPC POST therefore increments `memwal_mcp_handshake_total{outcome="ok"}`. The SSE 401/503 ratio this metric was added to expose will be diluted by per-envelope oks on a live session.

**Suggestion:** Record handshake outcomes only on `sse_proxy` (and streamable, which has no separate GET), or add a low-cardinality `route`/`kind` label so SSE can be queried alone.

ducnmm (changes requested):
## Summary

The relayer cache and MCP bridge changes are correct for the WALM-618 / WALM-386 failure modes: a revoked key can ride the 30s positive TTL while Sui answers, and the 10min stale-grace only after `is_unavailable()`; ordinary `memwal_login` is not blocked by the 10s reject cache (positive lookup first, new key has no reject entry); eviction generation prevents an in-flight success from re-caching a pair another request just revoked; borrowed `dyn DelegateAccountKey` hashing matches owned `(String, Vec<u8>)` (pinned by test); the 90s stalled deadline applies only to objects still in `pendingForward`, so a request that was actually POSTed keeps the full call timeout, and `memwal_login` stays local with `immediate` reconnect bypassing the 429 floor. Prior review notes (sampled 401 warn vs always-on counter, `verified_at` at read start, Keep on unavailable, eviction generation, borrowed-key lookup) are addressed. The one landing defect is the changelog: both MCP changelogs put WALM-386 / WALM-618 under already-published 0.0.12 instead of unpublished 0.0.13, and they drop the four #633 bullets already on `origin/dev` under 0.0.13.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 0

Cannot approve until the dump is under unpublished 0.0.13 and the #633 bullets are restored. Runtime looks good for a `dev` MCP test after that.

packages/mcp/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
...
+- Answer a request that is still buffered while the handshake keeps failing, instead of holding it for the full call timeout and then blaming a dropped reply. A request that was never sent cannot have executed, so after 90s of consecutive handshake failures the bridge fails it with the reason the handshake actually gave — rather than parking it four minutes and returning "the connection to the relayer dropped, please retry", which named the wrong layer and invited a retry of a `remember` that had never left the process. The full deadline still applies while the handshake is healthy. Override with `MEMWAL_MCP_STALLED_HANDSHAKE_MS`. (WALM-618)
```
ducnmm, line 9: **[bug]** WALM-386 / WALM-618 are now under unpublished `## 0.0.13`, but the four #633 bullets already on `origin/dev` under that same heading are gone (keep reading stdin after `memwal_login`; confirm a completed sign-in; stop claiming already signed in; serve one login prompt). `origin/main` is 0.0.12; this PR must not drop unreleased 0.0.13 notes.

**Suggestion:** Copy those four bullets from `origin/dev` back under `## 0.0.13` (after the `memwal_health` line). Keep WALM-386 / WALM-618 here. Do not bump the package.

docs/mcp/changelog.mdx
```diff
@@ -37,6 +37,8 @@ This release warns on unrecognised command-line options instead of ignoring them
...
+- Answer a request that is still buffered while the handshake keeps failing, instead of holding it for the full call timeout and then blaming a dropped reply. A request that was never sent cannot have executed, so after 90s of consecutive handshake failures the bridge fails it with the reason the handshake actually gave — rather than parking it four minutes and returning "the connection to the relayer dropped, please retry", which named the wrong layer and invited a retry of a `remember` that had never left the process. The full deadline still applies while the handshake is healthy. Override with `MEMWAL_MCP_STALLED_HANDSHAKE_MS`. (WALM-618)
```
ducnmm, line 42: **[bug]** Same dump as `packages/mcp/CHANGELOG.md`: the four #633 bullets are missing from unpublished `## 0.0.13`. The mdx intro/`answer:` also still describe only the #630 / WALM-480 slice, not the handshake fixes or the #633 sign-in work.

**Suggestion:** Restore the four #633 bullets under 0.0.13 to match the package changelog, and mention them in the 0.0.13 intro + frontmatter `answer:` so the two files stay in sync.

ducnmm (dismissed):
## Summary

Follow-up `91ab2823` moves WALM-386 / WALM-618 under unpublished `## 0.0.13` (good), labels handshake metrics by route (`sse` / `messages` / `streamable`), and pins the verify-cache sweep on `is_servable_while_unavailable()` so the 10-minute stale-grace is not deleted at 30s. Those two suggestions are addressed.

The landing defect is still the dump: `origin/dev`'s unpublished 0.0.13 already has the four #633 bullets (stdin after login, confirm sign-in, stop claiming already signed in, one login prompt). This head drops them from both `packages/mcp/CHANGELOG.md` and `docs/mcp/changelog.mdx`. Merging would rewrite the unreleased 0.0.13 section and lose WALM-394. Restore those four bullets under 0.0.13 (package + mdx, including the mdx intro/`answer:` if they belong there). Do not bump past 0.0.13.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

packages/mcp/CHANGELOG.md
harrymove-ctrl: Placement fixed in `22f660d7` — both bullets are under `## 0.0.13` now, `0.0.12` untouched, and no version bump: `packages/mcp/package.json` was already `0.0.13`, so there was nothing to move but the notes.

On the #633 bullets though, I think this one is a false alarm — the diff is misleading:

```
$ git show 1439e97d:packages/mcp/CHANGELOG.md | grep -c '(#633'
0
```

`1439e97d` is this branch's merge base. The #633 bullets were never here to drop — they landed on `dev` afterwards in `493c9e66` ("confirm a completed sign-in, and keep serving stdin after it"), and `dev` is 2 commits ahead of us. Diffing the branch against `origin/dev` shows them missing, which reads exactly like a deletion; diffing against the merge base shows there was never anything there.

So they arrive on merge, along with `messages.ts` and the three login test files from the same PR. I deliberately did **not** hand-restore them — that would put release notes on this branch for code that isn't on it, and both sides would then be adding identical bullets at the same insertion point. I tried it, and `git merge-tree origin/dev <head>` went from clean to conflicting in both files.

As pushed it merges clean. Happy to rebase onto `dev` if you'd rather see the whole 0.0.13 section assembled in the PR rather than at merge — it's only 2 commits.

docs/mcp/changelog.mdx
harrymove-ctrl: Same fix in `22f660d7` — both bullets moved to `## 0.0.13`, `0.0.12` untouched.

The `answer:` frontmatter and the 0.0.13 release blurb are the same story as the #633 bullets (detail on the CHANGELOG.md thread): `git show 1439e97d:docs/mcp/changelog.mdx | grep -c 'confirmed with a notification'` is `0` at the merge base and `1` on `dev`. That wording came from `493c9e66`, not from anything this branch removed, so it returns on merge. I left it rather than hand-restoring prose describing code that isn't on this branch.

services/server/src/storage/sui.rs
harrymove-ctrl: Fixed in `91ab2823`, and you were right that the old one guarded nothing.

It now sweeps with `is_servable_while_unavailable()` — what `main.rs` actually uses — over three seeded entries: fresh survives, past-TTL-but-inside-the-grace survives, past-the-grace is evicted. The middle one is the case that matters. The "mirrors main.rs verbatim / past TTL can never be served" comment is gone; it was wrong on all three counts, and `TimedVerifiedOwner::is_fresh`'s own doc already said the sweeper uses the other predicate.

Renamed to `delegate_verify_cache_sweep_keeps_what_the_outage_path_can_serve`, since what needs pinning is what the sweep *keeps*.

I checked it actually bites: reverting `main.rs` to `retain(|_, v| v.is_fresh())` makes it fail on the in-grace assertion. The old test passed either way, which is exactly what made it a false guard.

services/server/src/mcp_proxy.rs
harrymove-ctrl: Fixed in `0c635577` — went with the `route` label rather than restricting to `sse_proxy`.

`memwal_mcp_handshake_total` is now `{route, outcome, reason}`, `route` being `sse` / `messages` / `streamable`, threaded through `refuse`, `legacy_delegate_registered` and `classify_and_resolve`. SSE queries alone as `memwal_mcp_handshake_total{route="sse"}`.

I picked the label over your first option because `streamable_proxy` has no separate GET — recording only on `sse_proxy` would have stopped counting that transport's handshakes entirely. Cardinality is 3 routes over the existing outcome/reason sets.

While in there I also moved `record_mcp_handshake`'s doc comment back onto the function; an earlier fixup had left it stranded on `MCP_REFUSAL_LOG_INTERVAL`, 84 lines above.

packages/mcp/CHANGELOG.md
ducnmm: Confirmed on a merge-tree of `origin/dev` + `91ab2823`: both changelogs auto-merge and unpublished `0.0.13` keeps all four #633 bullets plus WALM-386/618. Dump request dismissed.

The mdx intro/`answer:` still omit the handshake notes until a rebase onto `dev` — fine as a follow-up, not a dump blocker.

services/server/src/mcp_proxy.rs
```diff
@@ -390,7 +574,16 @@ pub async fn sse_proxy(
...
+            // `retry-after` is load-bearing: the sidecar sets it on an
+            // `ip_burst_cap` 429 and the bridge honours it (WALM-386).
+            // Dropping it here left the client with no ETA and the wrong
+            // remediation, since it could not tell a timed cap from a
+            // concurrency one.
+            "content-type"
+                | "cache-control"
+                | "www-authenticate"
+                | "connection"
+                | "retry-after"
```
ducnmm, line 602: **[suggestion]** `sse_proxy` and `streamable_proxy` now forward `retry-after`. `messages_proxy` still does not — a 429 on `POST /api/mcp/messages` (JSON-RPC after the SSE session exists) will still drop the header, so the bridge cannot tell a timed cap from a concurrency cap on that path.

**Suggestion:** Add `retry-after` to the `messages_proxy` allowlist too, same as SSE.

packages/mcp/src/bridge.ts
```diff
@@ -274,6 +377,14 @@ interface RpcMessage {
...
     msg: RpcMessage;
     startedAt: number;
+    /** Set once a POST has been issued for this request.
+     *
+     * This is what separates "cannot have executed" from "might have
+     * executed", and it has to live on the entry: `pendingForward` only holds
+     * requests buffered before the first successful connect, so in a
+     * mid-session outage — the ordinary case — a request that never left the
+     * process was indistinguishable from one already sent. */
+    sent?: boolean;
```
ducnmm, line 408: **[suggestion]** `sent` on the in-flight entry is the right place for the mid-session case Dio/Teo hit. The new tests still fail the handshake from cold start (`pending-forward-stalled` / 429 suite). Nothing pins that a request issued after `firstConnectDone`, while `sse` is null, expires at 90s instead of 240s.

**Suggestion:** One test that connects successfully, drops SSE, sends `tools/call`, and asserts the stalled-handshake wording before the full call timeout.

ducnmm (commented):
## Summary

Follow-up `bc531e55` addresses the three runtime notes from the previous round: eviction generation bumps under the `entries` write lock; never-sent is tracked on the in-flight entry (`sent`) so a mid-session outage can hit the 90s stalled deadline; `retry-after` is forwarded on the SSE and streamable proxies; `Retry-After: 0` / past HTTP-date fall back to the floor. Dump #633 remains a merge-tree non-issue. No blocking bug on this head.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

services/server/src/mcp_proxy.rs
harrymove-ctrl: Done in `806ff2e7`. You're right that it was the odd one out, and for a dull reason: `messages_proxy` doesn't use the allowlist loop the other two share — it builds its response from a fixed `[(CONTENT_TYPE, ..)]` array — so adding a match arm wouldn't have reached it. It now captures `retry-after` before `bytes()` consumes the response and adds it when present.

Worth noting for whoever picks up the client half: the bridge only constructs `RelayerThrottledError` in `openSseStream`, so a 429 on the POST path has no consumer today. Forwarding it is still right — the proxy shouldn't destroy information the sidecar deliberately sent — but the header won't change behaviour until the POST path parses it too.

packages/mcp/src/bridge.ts
harrymove-ctrl: Fair, and this was the gap I'd have most wanted caught. Fixed in `806ff2e7`.

It needed its own mock rather than a new case on the existing one, which is the point you're making: `startUnavailableRelayer` fails from the very first handshake, so the request lands in `pendingForward` and the test passes *even with the bug present*. The new `startHealthyThenDeadRelayer` serves exactly one healthy session, kills it, and 503s every reconnect after that — so `firstConnectDone` is set and the call goes into `inFlight` only, which is the path `sent` exists for.

It asserts the reply names the failing connection and arrives inside the call timeout (3s stalled deadline vs 60s call timeout, so a regression shows up as a 60s wait rather than a wording diff).

One caveat I'd rather state than have you find: I could not run the mjs suite locally — this sandbox blocks `listen`, so all 27 tests fail at `EPERM` before any assertion. `tsc` and `node --check` are clean and the escaping is byte-identical to the sibling test, but CI is the first real execution. If it's flaky on timing I'll tune the deadline rather than the assertion.

ducnmm (commented):
## Summary

Follow-up `170e0ce9` / `806ff2e7` addresses both remaining suggestions: `messages_proxy` now forwards `Retry-After` (captured before `bytes()`), and `pending-forward-stalled.test.mjs` covers the mid-session path with a healthy-then-dead relayer so the 90s deadline is pinned after `firstConnectDone`. Stale-serve / refusal log sampling and connect-episode expiry at the cap look correct. No open issues on this head.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

ducnmm (commented):
## Summary

`82e44d0b` drops the mid-session stalled-deadline test after two different CI failures (ordering race, then sweeper never firing). The product change stays: `InFlightEntry.sent` still distinguishes never-left-the-process from maybe-ran. Cold-start coverage in `pending-forward-stalled.test.mjs` remains. That is an honest unpin, not a revert of the runtime fix. No open issues on this head.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

harrymove-ctrl:
@ducnmm ready for another look — CI is green on `82e44d0b` (18/18).

## Since your last review

Your two suggestions:

- **`messages_proxy` `retry-after`** — `806ff2e7`. It was missed because that handler builds its response from a fixed header array rather than the allowlist loop the other two share, so a match arm wouldn't have reached it. Worth knowing: the bridge only builds `RelayerThrottledError` in `openSseStream`, so a 429 on the POST path has no consumer yet — correct to forward, but it changes nothing until that path parses it.
- **Mid-session test** — attempted, then withdrawn. Details below, because the outcome isn't what you asked for.

Three notes from a cloud review of the same head, all the same shape — state or logging this PR adds, bounded everywhere except one place (`170e0ce9`):

| | |
| -- | -- |
| the stale-grace `warn!` was unsampled | now on the same per-account 60s policy as the refusal line — but its **own map**, so a refusal and a stale-serve for one account can't silence each other |
| neither sampler was swept | `sweep_log_samplers()` in the existing 300s loop; they were the only per-account maps here it didn't touch |
| the connect-episode cap didn't expire first | expires before the cap now, so it self-heals — it does **not** stop a caller sustaining the spam, and the comment says so |

## The test I owe you, and don't have

I wrote the mid-session test, and it failed CI twice.

Run 1 was my bug: the call went out in the same millisecond the session died, so `sse` was still set, the call took the POST path, `postIfCurrent` marked it `sent`, and it kept the full call timeout — the opposite of the point. That ordering is fixable and I fixed it.

Run 2 still timed out with the call unanswered, and stderr has no `bridge.call_orphaned` at all — the sweeper never decided it had expired, though both elapsed and handshake-failing time pass the 3s deadline by the first 5s tick. That contradicts my reading of the sweeper, which means my reading is wrong somewhere I haven't found. My suspect is the mock (it answers POSTs 202 regardless of session state, where the real relayer 404s a dead session), but that's a hypothesis.

So I removed it rather than land it flaky or leave CI red. **The mid-session path is covered by no test.** The product change stands on reasoning alone: `InFlightEntry.sent` is what separates "never left the process" from "might have run", and `pendingForward` membership structurally could not see the mid-session case. If you'd rather this not merge without coverage, that's a fair call and I'll take another run at it — I think it needs a seam the mock can drive rather than a timing race against a state transition it can't observe.

## Also still open, deliberately

Four findings I'm not fixing here, with reasons:

- **The eviction generation is still process-global**, so an unrelated rejection can still drop a good verification. I built the per-key version and reverted it: it relies on the rejection record, and `main.rs` sweeps that map holding no `entries` guard, so a third party can delete the record outside the order the proof needs. The right fix is a tombstone in the entries map — value type, sweeper, stale-serve lookup, ~10 tests. Not on a PR at approval.
- **`verified_at` is stamped before an unbounded read.** Correction to what I said earlier: production takes gRPC and `sui-rpc` sets no per-call deadline, so the reqwest 30s never applies here. I tried `tokio::time::timeout` and reverted it — it turns our own deadline into `RpcError`, which `is_unavailable()` reads as "chain unreachable", so a fullnode answering `KeyNotFound` at 6s gets cut at 5s and the revoked key rides the stale grace for up to 630s. Strictly worse than the defect.
- **The cache key is the raw account id.** `hex_address_bytes` strips an optional `0x` and right-aligns any length ≤ 64, so `0xabc` / `abc` / `0x0abc` are one address and three keys — and the positive cache has no size cap at all.
- The four copy-pasted sweep blocks in `main.rs`.

Happy to file these as follow-ups, or fold any one in if you'd rather it not merge without it.

harrymove-ctrl:
@ducnmm correction to my last comment, and you were right to ask for that test.

I said the mid-session path was unpinned and that `InFlightEntry.sent` fixed it. The first half was true; **the second half was wrong**, and the test proved it on the third attempt. CI is green on `43109b5f`.

## What the test found

`sse` is **not** cleared when the server pump hits EOF — it keeps pointing at the dead session until a reconnect succeeds. So a call arriving during a mid-session outage never reaches the buffering path at all. It takes the POST path, posts to the stale session URL, and is 404'd:

```
bridge.post_non_ok   status:404
bridge.session_stale sessionUrl=...?sessionId=session-1
```

`postIfCurrent` had already marked it `sent`, so it kept the full 240s call timeout and came back blaming a dropped connection — the exact symptom the stalled deadline exists to remove. So `b8a5c5ff` on its own did not fix the case it was added for.

## The fix

A 404 there is the relayer saying that session does not exist, so the message was discarded rather than routed — it provably did not run. The request goes back to never-sent and takes the short deadline (`43109b5f`).

The conservative default is unchanged everywhere else: marking still happens before the await, because a network error mid-flight is genuinely ambiguous and has to keep the full timeout. Only a definitive 404 reverses it.

## On the test

Three attempts. The first two raced an internal state transition the mock could not observe, and I removed it rather than leave CI red. The third is synchronised on the bridge's own stderr (`bridge.connected` before the kill, `bridge.reconnect_failed` before the call), every wait names what it was waiting for, and the mock 404s a POST against a dead session the way the relayer does — which is what surfaced the bug instead of hiding it behind a 202.

Two things I'd flag rather than leave implicit:

- **This was a real defect reaching `dev`** had you not pushed for coverage. Worth weighing when you decide whether the remaining open items need pinning before merge rather than after.
- I cannot run the mjs suite locally at all (sandbox blocks `listen`), so CI is my only execution path for the bridge. That is why these took three rounds, and why I would rather you treat my "tested" claims on that package as "CI-tested" specifically.

packages/mcp/src/bridge.ts
```diff
@@ -829,7 +1002,32 @@ export async function runBridge(
...
+        if (msg.id !== undefined && msg.id !== null) {
+            const tracked = inFlight.get(msg.id);
+            if (tracked) tracked.sent = true;
+        }
+        return postMessage(postUrl, msg, postCreds, extraHeaders).then((status) => {
+            // 404 is the relayer saying that session does not exist, so the
+            // message was discarded rather than routed: it provably did not
+            // run, and the request goes back to being never-sent.
+            //
+            // This is not a corner case. `sse` is not cleared when the server
```
ducnmm, line 1113: **[suggestion]** The new `postIfCurrent` 404 block is a ten-line design narrative (`sse` is not cleared on EOF, mid-session POST path, WALM-618 symptom) on top of the three lines that reset `sent`. The useful WHY is the first sentence: a 404 means the session did not exist, so the envelope was discarded. The rest restates the commit message and will rot the next time the pump's `sse` lifetime changes.

**Suggestion:** Keep a short WHY (`404` = session missing, envelope not routed, so `sent` can be cleared) and drop the architecture history. The mid-session test already documents the EOF/`sse` shape.

packages/mcp/test/pending-forward-stalled.test.mjs
```diff
@@ -0,0 +1,605 @@
...
+    // 2. the first session is genuinely up, so `firstConnectDone` is set and
+    //    this cannot degenerate into the cold-start case.
+    await until(() => /"event":"bridge\.connected"/.test(stderrBuf), 15_000, "bridge.connected");
+
+    // 3. the relayer goes away and refuses every reconnect.
+    mock.killSession();
+    await until(
+        () => /"event":"bridge\.reconnect_failed"/.test(stderrBuf),
+        20_000,
+        "bridge.reconnect_failed (so sse is null and the handshake is on record as failing)",
```
ducnmm, line 566: **[suggestion]** The `reconnect_failed` wait still claims `sse` is null. That is the mental model the 404 fix disproved: `reconnect()` aborts the handle but does not null it, which is why this call POSTs and 404s. A later flake that dumps this string will send someone looking at the buffering path again.

**Suggestion:** Describe what the log actually guarantees — `handshakeFailingSince` is set, so the stalled deadline can apply after the 404 reset — and leave `sse` out of it.

packages/mcp/test/pending-forward-stalled.test.mjs
```diff
@@ -0,0 +1,605 @@
...
+        "the tool-call answer on the stalled-handshake deadline",
+    );
+    const waitedMs = Date.now() - sentAt;
+
+    // `sse` is NOT cleared on server-pump EOF — it keeps pointing at the dead
+    // session until a reconnect succeeds — so the call does take the POST
+    // path and is 404'd by the relayer. That 404 is what returns it to
+    // never-sent: the session did not exist, so the message was discarded
+    // rather than routed, and it provably did not run.
+    assert.match(
```
ducnmm, line 590: **[suggestion]** `bridge.connected` is logged before `flushPendingForward()` returns, and the test kills the session immediately after that log. The buffered `initialize` POST can 404 first and already write `bridge.session_stale`. The later match then does not prove the *tool call* took the stale-session path. `waitedMs < CALL_TIMEOUT_MS` is also implied by the 36s `waitFor`, so an epoch-skip / never-posted tool call can still pass if initialize already 404'd and the stalled deadline fires for `sent !== true`.

**Suggestion:** Snapshot stderr (or `getPostCount()`) after `reconnect_failed` and before sending id=2, then assert a *new* `bridge.session_stale` (or a post-count bump) caused by that call. Optionally wait until initialize has been POSTed before `killSession()`, and assert `waitedMs >= STALLED_HANDSHAKE_MS` the way the cold-start sibling does.

ducnmm (commented):
## Summary

The 404 → `sent = false` path is correct for the mid-session case this follow-up is pinning. `postMessage` returns the status (it does not throw on HTTP 404), the orphan sweeper re-reads `entry.sent` on every tick rather than latching it at POST time, and a sidecar 404 is decided before the envelope is routed, so this reset does not turn a delivered `remember` into a safe retry. Existing `handleClientLine` / flush 404 handling still reconnects and replays; the new flag only shortens the wait while that reconnect keeps failing. Remaining notes are comment accuracy and how tightly the new test pins *this* call's 404.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 0

ducnmm:
Looked at `43109b5f`. The 404 → never-sent reset is the right fix: `postMessage` returns the status, the sweeper re-reads `sent` each tick, and a sidecar 404 is decided before routing so this does not invite a duplicate remember. No bugs on this head.

Three suggestions on the new block/test (inline) — comment trim, the `reconnect_failed` wait still saying `sse` is null, and pinning `session_stale` to the tool call rather than a log initialize can already have written. None of them block merge if Henry wants it on `dev` as-is.

ducnmm (approved):

ducnmm (approved):

PR #901, opened 2026-09-11
fix(mcp): say the handshake failed instead of blaming a dropped reply (WALM-618)

harrymove-ctrl:
## The bug

A tool call that arrives before the relayer session exists waits in `pendingForward`. Two things were wrong with how that wait ended.

**1. It barely ended.** The orphan sweeper bounded it at `callTimeoutMs` — 240s by default (`SLOWEST_SERVER_TOOL_MS` 180s + 60s). Four minutes of silence, plus one agent-level retry, is exactly the "3–8 min" the ticket reports.

**2. The answer blamed the wrong thing.** Every expiry was reported the same way:

> ❌ Walrus Memory did not answer this call. The connection to the relayer dropped before the result came back. Please retry.

For a buffered call that is wrong twice over. Nothing was sent, so no connection dropped and no reply was lost; and *"before the result came back"* implies a write that may or may not have landed, when none was attempted. Meanwhile the one thing the bridge did know — that the MCP handshake had been failing, and with what error — it kept to its own log.

In Dio's trace: 15 session opens for 6 calls that ever reached the relayer, every call that arrived finishing in 19–34s, and from his side a `memwal_remember` that was simply slow for minutes with nothing reported.

## The fix

**A shorter deadline for a call that provably never left the process.** Once no working connection has existed for `stalledHandshakeMs` (90s, overridable via `MEMWAL_MCP_STALLED_HANDSHAKE_MS`), a still-buffered request is answered on that bound instead of the full 240s.

The two cases can have different deadlines because they carry different risk. A request that was **sent** might have executed, so failing it early invites the agent to retry a `remember` that already landed — that one keeps the full deadline. A request that was **never sent** cannot have executed: answering it is provably a no-op and the retry costs one round trip.

90s is past a relayer cold start and past six reconnect attempts at the capped 15s backoff, so it does not fire on a slow-but-recovering relayer. The login flow is unaffected — the browser wait happens in `runAuthRequiredServer`, *before* `runBridge` is called with the handed-off lines, so no request is buffered across it. `coldstart-timeout.test.mjs`'s property still holds: a call the next attempt could serve is not eager-failed.

**Three cases in the answer, not one.** Track the last handshake failure and how long the run has lasted, cleared on every successful connect:

- never sent, handshake failing → *"could not reach the relayer — the MCP connection has been failing for 94s, so this call never ran and nothing was stored. Last handshake error: … Check the relayer, or run `memwal-mcp login` if the delegate key was revoked, then retry."*
- never sent, **connection healthy** (post-connect, `handleClientLine` also buffers behind an in-progress flush) → *"never sent this call — it was still queued when the call timed out, so nothing was stored."* No invented outage, and this case keeps the full deadline.
- actually sent, no reply → the original wording, unchanged.

Answering a buffered request also means **removing it from the buffer**. Left in, the flush after the next successful connect would run the call we just reported as never having run — with the agent, having been told nothing was stored, likely to have retried by then. `initialize` is exempt, as everywhere else in this file: it is answered locally and buffered only so the session can still negotiate capabilities.

## Tests

`test/pending-forward-stalled.test.mjs` drives the real shape: a relayer that 503s every handshake the way the relayer does when the on-chain delegate verify cannot reach a throttled fullnode, then recovers. It runs a 60s call timeout against a 3s stalled-handshake bound, so an answer near 3s proves the **new** deadline fired and not the ordinary timeout. It also asserts the call is not eager-failed mid-deadline, is answered exactly once, carries the handshake error, never reaches the relayer afterwards, that a call sent after recovery does, and that `initialize` is still forwarded once the session comes up.

Full suite: 79 passed, 0 failed. `tsc --noEmit` clean.

Relayer-side companion (the load that produces those 503s): #900

Linear: [WALM-618](https://linear.app/mysten-labs/issue/WALM-618/investigate-memwal-remember-takes-3-8-min-under-load-handshake-retries)

### DIFF
--- a/packages/mcp/src/bridge.ts
+++ b/packages/mcp/src/bridge.ts
@@ -809,6 +809,21 @@ export async function runBridge(
     let reconnectAttempt = 0;
     let reconnectPromise: Promise<void> | null = null;
     let firstConnectDone = false;
+    /** Why the last handshake attempt failed, and when the run of failures
+     * started. Kept so a request that ages out while buffered can say what
+     * it was actually waiting on instead of a generic "unavailable" — the
+     * user-visible half of WALM-618, where the bridge retried in silence and
+     * a `remember` looked like it was just slow. Cleared on every success. */
+    let lastHandshakeError: string | null = null;
+    let handshakeFailingSince: number | null = null;
+    const noteHandshakeFailure = (reason: string): void => {
+        lastHandshakeError = reason;
+        handshakeFailingSince ??= Date.now();
+    };
+    const clearHandshakeFailure = (): void => {
+        lastHandshakeError = null;
+        handshakeFailingSince = null;
+    };
     /** Bumped when the live SSE session is aborted or replaced so queued
      * POSTs captured against a stale URL are skipped (reconnect replays). */
     let sessionEpoch = 0;
@@ -1039,6 +1054,7 @@ export async function runBridge(
                     firstConnectDone = true;
                     activeCredentialGeneration = openingGeneration;
                     reconnectAttempt = 0;
+                    clearHandshakeFailure();
                     log.info("bridge.reconnected", {
                         relayer: openingCreds.relayerUrl,
                         replayCount: inFlight.size,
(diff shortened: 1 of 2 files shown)


harrymove-ctrl:
Superseded by #903, which carries this branch's commits unchanged plus the other half.

This PR and the other one both rewrote `packages/mcp/src/bridge.ts` and conflicted with each other, so merging either would have forced the other to rebase. They are also two halves of the same user-visible failure — a bridge that waits in silence and then names the wrong layer — so #903 reviews them together. All tests from both branches pass side by side there (90 passed, 0 failed).

Not deleting the branch, in case the history is wanted.

PR #902, opened 2026-09-11
fix(sdk): cap remember-job poll backoff (WALM-623)

ducnmm:
## Ticket

WALM-623 — https://linear.app/mysten-labs/issue/WALM-623/sdk-cap-waitforrememberjob-poll-backoff-no-public-api-change

## What changed?

- `waitForRememberJob` / `waitForRememberJobs` (and Python `wait_for_remember_job` / `wait_for_remember_jobs`) no longer sleep 5–10s after a ~20s write is already `done`.
- First poll is immediate. Later polls cap at 1.5s with ±25% jitter. No exponential climb to 10s.
- `pollIntervalMs: 0` / `poll_interval_ms: 0` still means no wait.
- No new SDK methods or request fields.

## Why is this needed?

Quiet-prod traces (WALM-618 / WALM-621): Walrus+index finished in ~21s; the client polled again 5–8s after `done` because backoff had climbed toward 10s. That idle is entirely client-side. MCP SSE push / webhooks are out of scope — recall is already ~1.6–2.6s once the request arrives.

## Scope

Cap remember-job poll delay in the TypeScript and Python SDKs.

### Out of scope

- MCP handshake / SSE 401 (WALM-618, #900 / #901)
- Relayer upload queue / Walrus duration
- Webhook or SSE job-completion API (WALM-152)

## How was this tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] End-to-end tests
- [ ] Manual testing
- [ ] Not applicable

Commands:
- `node scripts/verify-manual-sdk-release.mjs` (0.1.7 / 0.1.10 / 0.0.13 / 0.0.6)
- `packages/sdk`: `pnpm test` — 122 passed
- `packages/python-sdk-memwal`: `pytest tests/ -m 'not integration'` — 155 passed, 23 deselected

1. Read `packages/sdk/src/polling-delay.ts` and `packages/python-sdk-memwal/memwal/client.py` `_polling_delay_ms`.
2. Run the commands above.
3. Confirm `packages/sdk/src/index.ts` does not export `pollingDelayMs`.
4. Confirm changelogs sit under existing 0.1.7 / 0.1.10 (no bump; `dev` already ahead of `main`).

## Risks and dependencies

None. Callers that passed a large `pollIntervalMs` now wait at most ~1.5s between polls (the intended cap).

## Author checklist

- [ ] CI is green.

### DIFF
--- a/docs/python-sdk/usage.md
+++ b/docs/python-sdk/usage.md
@@ -100,4 +100,4 @@ print(done.blob_id)
 done = await memwal.wait_for_remember_job(accepted.job_id)
 ```
 
-Bulk (up to 20 items per call) follows the same pattern with `remember_bulk_async`, `wait_for_remember_jobs`, and `remember_bulk_and_wait`. Polling uses jittered exponential backoff (1.5× capped at 10s, ±25%) to stay relayer-friendly at scale.
+Bulk (up to 20 items per call) follows the same pattern with `remember_bulk_async`, `wait_for_remember_jobs`, and `remember_bulk_and_wait`. Polling starts immediately, then repeats at most ~1.5s with ±25% jitter; `poll_interval_ms: 0` means no wait.
--- a/packages/python-sdk-memwal/memwal/client.py
+++ b/packages/python-sdk-memwal/memwal/client.py
@@ -168,14 +168,18 @@ async def _sleep_ms(ms: int) -> None:
 
 
 def _polling_delay_ms(base_ms: int, attempt: int) -> int:
-    """Jittered exponential backoff matching TS ``pollingDelayMs``.
+    """Jittered poll delay matching TS ``pollingDelayMs``.
 
-    base * 1.5^min(attempt, 6), capped at 10s, with ±25% jitter so
-    concurrent clients don't synchronise.
+    ``base_ms <= 0`` means no wait (``poll_interval_ms: 0``). Attempt 0 is
+    immediate so the first GET is not delayed. Later polls cap at 1500ms
+    with ±25% jitter.
     """
 
-    base = max(100, base_ms)
-    capped = min(10_000, base * (1.5 ** min(attempt, 6)))
+    if base_ms <= 0:
+        return 0
+    if attempt == 0:
+        return 0
+    capped = min(1500, max(100, base_ms))
     jitter = 0.75 + random.random() * 0.5
     return int(capped * jitter)
 
@@ -420,8 +424,7 @@ class MemWal:
           ``status`` field (404 / ``status == "not_found"`` raises).
         - Transient HTTP errors (429, 5xx, network drop) are retried until
           the timeout, not surfaced as polling failures.
-        - Backoff is jittered exponential (1.5x cap 10s, ±25%) to avoid
-          thundering-herd at scale.
+        - First poll is immediate; later polls cap at 1.5s with ±25% jitter.
         """
 
         deadline_ms = _now_ms() + timeout_ms
--- a/packages/python-sdk-memwal/memwal/client.py
+++ b/packages/python-sdk-memwal/memwal/client.py
@@ -171,8 +171,7 @@ def _polling_delay_ms(base_ms: int, attempt: int) -> int:
     """Jittered poll delay matching TS ``pollingDelayMs``.
 
     ``base_ms <= 0`` means no wait (``poll_interval_ms: 0``). Attempt 0 is
-    immediate so the first GET is not delayed. Later polls cap at 1500ms
-    with ±25% jitter.
+    immediate so the first GET is not delayed.
     """
 
     if base_ms <= 0:
--- a/packages/sdk/src/polling-delay.ts
+++ b/packages/sdk/src/polling-delay.ts
@@ -1,0 +1,13 @@
+/**
+ * Delay before a remember-job poll.
+ *
+ * `baseMs <= 0` means no wait (`pollIntervalMs: 0`).
+ * Attempt 0 is immediate so the first GET is not delayed.
+ */
+export function pollingDelayMs(baseMs: number, attempt: number): number {
+    if (baseMs <= 0) return 0;
+    if (attempt === 0) return 0;
+    const capped = Math.min(1500, Math.max(100, baseMs));
+    const jitter = 0.75 + Math.random() * 0.5;
+    return Math.floor(capped * jitter);
+}
--- a/docs/python-sdk/changelog.mdx
+++ b/docs/python-sdk/changelog.mdx
@@ -29,7 +29,7 @@ questions:
   - What changes were made in memwal 0.1.4?
   - Where can I find the release history for the Walrus Memory Python SDK?
 answer: >-
-  The latest Python SDK release is 0.1.10. `restore()` results include `failed` (default `0`) for permanent decrypt/UTF-8 failures instead of folding them into `skipped`. 0.1.9 reports HTTP 503 as a retryable upstream outage instead of a credential failure, rejects empty `remember_bulk_async` batches and misaligned relayer `job_ids`, aligns restore `truncated` docs with WALM-431 retryable semantics, and warns when `server_url` uses plaintext HTTP on a non-localhost host without logging URL credentials.
+  The latest Python SDK release is 0.1.10. `restore()` results include `failed` (default `0`) for permanent decrypt/UTF-8 failures instead of folding them into `skipped`. Remember-job polling starts immediately and caps backoff at ~1.5s instead of climbing to 10s. 0.1.9 reports HTTP 503 as a retryable upstream outage instead of a credential failure, rejects empty `remember_bulk_async` batches and misaligned relayer `job_ids`, aligns restore `truncated` docs with WALM-431 retryable semantics, and warns when `server_url` uses plaintext HTTP on a non-localhost host without logging URL credentials.
 ---
 
 Track what's new, changed, and fixed in `memwal` (Python).
@@ -38,11 +38,15 @@ For the latest version, see the [PyPI project page](https://pypi.org/project/mem
 
 ## 0.1.10
 
-This release adds `failed` on `restore()` results for permanent decrypt and UTF-8 failures.
+This release adds `failed` on `restore()` results for permanent decrypt and UTF-8 failures, and caps remember-job poll backoff at ~1.5s with an immediate first poll.
 
 ### Added
 
 - `restore()` results include `failed` (default `0`) for permanent decrypt/UTF-8 failures instead of folding them into `skipped` or dropping them silently.
+
+### Fixed
+
+- Cap `wait_for_remember_job` / `wait_for_remember_jobs` poll delay at ~1.5s with an immediate first poll, instead of exponential backoff that climbed to 10s. `poll_interval_ms: 0` still means no wait. (WALM-623)
 
 ## 0.1.9
 
--- a/docs/sdk/changelog.mdx
+++ b/docs/sdk/changelog.mdx
@@ -28,12 +28,12 @@ questions:
   - When was bulk remember added to the Walrus Memory SDK?
   - What security improvements have been made to the MemWal SDK?
 answer: >-
-  The latest TypeScript SDK release is 0.1.7. Request bodies are hashed with `@noble/hashes` rather than WebCrypto-or-`node:crypto`, so the SDK no longer imports a Node builtin that Vite silently externalises into a runtime crash in the browser, and it declares a Node 20 floor. `restore()` results include `failed` (required like `truncated`; SDK defaults omitted to `0`) for permanent decrypt/UTF-8 failures instead of folding them into `skipped`. Empty-body 401s use the AUTH_REJECTED troubleshooting message instead of telling callers to run `memwal_login`. Account and manual PTBs use typed `tx.pure` helpers so they work with modern `@mysten/sui`. 0.1.6 added optional `created_at` on `recall()` results, plus `sort` and `scoringWeights` on `RecallOptions`, and reports HTTP 503 as a retryable upstream outage instead of a sign-in failure.
+  The latest TypeScript SDK release is 0.1.7. Request bodies are hashed with `@noble/hashes` rather than WebCrypto-or-`node:crypto`, so the SDK no longer imports a Node builtin that Vite silently externalises into a runtime crash in the browser, and it declares a Node 20 floor. `restore()` results include `failed` (required like `truncated`; SDK defaults omitted to `0`) for permanent decrypt/UTF-8 failures instead of folding them into `skipped`. Empty-body 401s use the AUTH_REJECTED troubleshooting message instead of telling callers to run `memwal_login`. Account and manual PTBs use typed `tx.pure` helpers so they work with modern `@mysten/sui`. Remember-job polling starts immediately and caps backoff at ~1.5s instead of climbing to 10s. 0.1.6 added optional `created_at` on `recall()` results, plus `sort` and `scoringWeights` on `RecallOptions`, and reports HTTP 503 as a retryable upstream outage instead of a sign-in failure.
 ---
 
 ## 0.1.7
 
-This release removes the Node `crypto` import that crashed bundled browser builds at runtime, adds `failed` on `restore()` results, declares a Node 20 floor, stops telling headless SDK clients to call `memwal_login` on empty-body 401s, and switches account and manual PTBs to typed `tx.pure` helpers.
+This release removes the Node `crypto` import that crashed bundled browser builds at runtime, adds `failed` on `restore()` results, declares a Node 20 floor, stops telling headless SDK clients to call `memwal_login` on empty-body 401s, switches account and manual PTBs to typed `tx.pure` helpers, and caps remember-job poll backoff at ~1.5s with an immediate first poll.
 
 ### Added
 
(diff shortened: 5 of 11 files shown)


ducnmm:
@nikola0x0 gentle ping — still waiting on review of fea6621d.

packages/sdk/src/polling-delay.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Delay before a remember-job poll.
+ *
+ * `baseMs <= 0` means no wait (`pollIntervalMs: 0`).
+ * Attempt 0 is immediate so the first GET is not delayed.
+ */
+export function pollingDelayMs(baseMs: number, attempt: number): number {
+    if (baseMs <= 0) return 0;
+    if (attempt === 0) return 0;
+    const capped = Math.min(1500, Math.max(100, baseMs));
```
nikola0x0: [bug] A flat 1.5s poll (~40 req/min) is more than the relayer's per-delegate-key budget. Status polls go through `rate_limit_middleware` at weight 1 (`services/server/src/main.rs:1656`, `rate_limit.rs:166`), the default limit is 30 weighted req/min, and `POST /api/remember` costs 5 of that.

Simulated against the 60s sliding window (mean jitter, 150ms RTT):

| scenario | before | this PR | 1.5× growth, 3s cap |
|---|---|---|---|
| one 21s write | 6 polls, 9.7s idle | 14 polls, 0.6s idle | 9 polls, 2.1s idle |
| two 21s writes back to back | 0 × 429 | 18 × 429; 2nd write 18.8s idle | 0 × 429 |
| one 60s write | 0 × 429 | 12 × 429 | 0 × 429 |

The waiter retries its own 429s, but `recall`/`remember` from the same key in that window throws. A 3s cap still meets both WALM-623 criteria. `ceiling` also stops clamping an explicit interval above the cap, and dropping line 8 brings back the 100ms floor for 0.

```diff
+const MIN_POLL_MS = 100;
+const MAX_BACKOFF_MS = 3000;
+
 /**
  * Delay before a remember-job poll.
  *
- * `baseMs <= 0` means no wait (`pollIntervalMs: 0`).
- * Attempt 0 is immediate so the first GET is not delayed.
+ * Attempt 0 is immediate. Later polls grow 1.5x from `baseMs` (floor 100ms)
+ * to 3s, or stay at `baseMs` if the caller set it higher. Each poll costs 1
+ * of the relayer's 30/min per-delegate-key budget, so a flat short interval
+ * rate-limits the caller's own writes.
  */
 export function pollingDelayMs(baseMs: number, attempt: number): number {
-    if (baseMs <= 0) return 0;
     if (attempt === 0) return 0;
-    const capped = Math.min(1500, Math.max(100, baseMs));
+    const base = Math.max(MIN_POLL_MS, baseMs);
+    const ceiling = Math.max(MAX_BACKOFF_MS, base);
+    const capped = Math.min(ceiling, base * 1.5 ** Math.min(attempt - 1, 6));
     const jitter = 0.75 + Math.random() * 0.5;
     return Math.floor(capped * jitter);
 }
```

`polling-delay.test.mjs`, `test_polling_delay.py` and the changelog/doc numbers need updating to match.

packages/sdk/src/polling-delay.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Delay before a remember-job poll.
+ *
+ * `baseMs <= 0` means no wait (`pollIntervalMs: 0`).
+ * Attempt 0 is immediate so the first GET is not delayed.
+ */
+export function pollingDelayMs(baseMs: number, attempt: number): number {
+    if (baseMs <= 0) return 0;
```
nikola0x0: [bug] `pollIntervalMs: 0` does not behave as before. Before this PR, `Math.max(100, baseMs)` set a 100ms floor that grew to ~1.1s. WALM-623 says callers passing 0 stay unchanged.

With no sleep, the loop spins. If the relayer is unreachable, `fetch` throws with no `status`, so `isTransientPollingStatus(0)` hits `continue` (`memwal.ts:332`). Each pass signs, builds a SEAL session and fetches again, until `timeoutMs` (60s, or 120s for bulk). A 429 spins the same way. The diff on line 10 drops this line. The "`pollIntervalMs: 0` still means no wait" wording in both CHANGELOGs, both docs changelogs and `docs/python-sdk/usage.md:103` should go too. The `remember-idempotency.test.mjs` edit still passes with the floor back.

packages/python-sdk-memwal/memwal/client.py
```diff
@@ -168,14 +168,17 @@ async def _sleep_ms(ms: int) -> None:
...
+    immediate so the first GET is not delayed.
     """
 
-    base = max(100, base_ms)
-    capped = min(10_000, base * (1.5 ** min(attempt, 6)))
+    if base_ms <= 0:
+        return 0
+    if attempt == 0:
+        return 0
+    capped = min(1500, max(100, base_ms))
```
nikola0x0: [bug] Python mirror of `polling-delay.ts:8` and `:10`: 429s from the flat 1.5s cadence, and a zero-sleep loop for `poll_interval_ms=0`.

```diff
-    ``base_ms <= 0`` means no wait (``poll_interval_ms: 0``). Attempt 0 is
-    immediate so the first GET is not delayed.
+    Attempt 0 is immediate. Later polls grow 1.5x from ``base_ms`` (floor
+    100ms) to 3s, or stay at ``base_ms`` if the caller set it higher. Each
+    poll costs 1 of the relayer's 30/min per-delegate-key budget.
     """
 
-    if base_ms <= 0:
-        return 0
     if attempt == 0:
         return 0
-    capped = min(1500, max(100, base_ms))
+    base = max(100, base_ms)
+    ceiling = max(3000, base)
+    capped = min(ceiling, base * (1.5 ** min(attempt - 1, 6)))
     jitter = 0.75 + random.random() * 0.5
     return int(capped * jitter)
```

docs/python-sdk/usage.md
```diff
@@ -100,4 +100,4 @@ print(done.blob_id)
 done = await memwal.wait_for_remember_job(accepted.job_id)
 ```
 
-Bulk (up to 20 items per call) follows the same pattern with `remember_bulk_async`, `wait_for_remember_jobs`, and `remember_bulk_and_wait`. Polling uses jittered exponential backoff (1.5× capped at 10s, ±25%) to stay relayer-friendly at scale.
+Bulk (up to 20 items per call) follows the same pattern with `remember_bulk_async`, `wait_for_remember_jobs`, and `remember_bulk_and_wait`. Polling starts immediately, then repeats at most ~1.5s with ±25% jitter; `poll_interval_ms: 0` means no wait.
```
nikola0x0: [suggestion] An explicit `poll_interval_ms` above 1.5s is silently clamped. `types.py:396` calls it the "base poll cadence", `types.ts:235` says "How often to poll", and `examples/async_remember_demo.py:137` and `interactive_demo.py:189` pass 2000. The `ceiling` in the suggested delay function keeps a larger value. With that change:

```diff
-Polling starts immediately, then repeats at most ~1.5s with ±25% jitter; `poll_interval_ms: 0` means no wait.
+Polling starts immediately, then backs off 1.5× from `poll_interval_ms` to 3s (or to `poll_interval_ms`, if higher), with ±25% jitter.
```

nikola0x0 (commented):
**Blocking on the rate-limit comment.** bugs: 2 · suggestions: 1

**Blocking**

The immediate first poll and dropping the 10s climb are right: a 21s write goes from ~9.7s idle to ~0.6s. But a flat 1.5s poll is ~40 req/min, and each poll costs 1 of the relayer's default 30/min per-delegate-key budget. A write that runs past ~45s, or a second write in the same minute, now gets 429s from its own polls, and a `recall` from that key in the window gets the 429. Growing 1.5× to a 3s cap keeps most of the gain (2.1s idle) with no 429s. Fix inline on `polling-delay.ts:10`, Python mirror on `client.py:181`.

**Majors (details inline)**

* `polling-delay.ts:8`: `pollIntervalMs: 0` used to floor at 100ms. It is now a zero-sleep loop that spins on network errors and 429s, and WALM-623 says 0 stays unchanged.

<details>
<summary><b>Minor / info (1) — details inline</b></summary>

* `docs/python-sdk/usage.md:103`: an explicit interval above 1.5s is silently clamped, and the Python examples pass 2000.

</details>

<details>
<summary><b>Verified</b></summary>

* `GET /api/remember/{job_id}` and `POST /api/remember/bulk/status` go through `rate_limit_middleware` (`services/server/src/main.rs:1656`) at weight 1 (`rate_limit.rs:166`). `RATE_LIMIT_DELEGATE_KEY_PER_MINUTE` defaults to 30, and `POST /api/remember` costs 5.
* With the suggested `pollingDelayMs`, `tsc` is clean and `remember-idempotency.test.mjs` passes 5/5. `polling-delay.test.mjs` needs new numbers.
* Changelog placement is right: npm is at 0.1.6 and PyPI at 0.1.9, so 0.1.7 and 0.1.10 are unreleased.
* `pollingDelayMs` is not in `index.ts` or `package.json` `exports`.

</details>

<details>
<summary><b>Follow-up, not this PR</b></summary>

* Job-status polls probably belong outside the write budget, the way `/v1/owners/*` reads were split out (`main.rs:1708`). That is a relayer ticket.

</details>

ducnmm:
Followed the inline patch on `4109e150`.

- Later polls grow 1.5× from `pollIntervalMs` (100ms floor) to a 3s cap, or stay at the caller value if it is already higher.
- `pollIntervalMs: 0` / `poll_interval_ms: 0` keeps the pre-PR 100ms floor (not a zero-sleep spin).
- Attempt 0 is still immediate. TS and Python match. Changelogs stay under unpublished 0.1.7 / 0.1.10; no bump.

Ready for another look.

ducnmm:
@harrymove-ctrl gentle ping — still waiting on review of 4109e150.

harrymove-ctrl (commented):
## Review — WALM-623

Direction is right and the implementation is clean, but the claim the cap rests on doesn't hold: **the new schedule overruns the per-delegate-key quota for bulk in the average case**, and for single remember under unlucky jitter. Worth a cap adjustment before merge.

### What's good

- Extracting `pollingDelayMs` into `packages/sdk/src/polling-delay.ts` makes it testable — it wasn't before.
- **TS/Python parity is exact.** I swept both implementations across base ∈ {0, 50, 100, 400, 1500, 2999, 3000, 3001, 5000, 10000} × attempt 0–11 × five jitter values; no divergence (`Math.floor` and `int()` agree on positives). All 9 assertions in both test files verify.
- No stale "capped at 10s" text left anywhere on the branch; changelog headings (`0.1.7` TS, `0.1.10` Python) match `package.json` / `pyproject.toml`.
- The `remember-idempotency.test.mjs` rework is necessary, not incidental: with an immediate first poll, the old `jobPolls < 2` gate could let the *first* `rememberAndWait` reach `done` and resolve, breaking the `assert.rejects`. Keying on `rememberPosts` fixes it deterministically.
- Bulk polls through a single batched `/api/remember/bulk/status`, so there's no N× fan-out per iteration.

### 1. The 3s cap overruns the delegate-key quota

`polling-delay.ts` justifies the cap with "so status GETs stay under the 30/min quota". That budget is tighter than it looks, because the quota is **weighted** (`endpoint_weight`, `services/server/src/rate_limit.rs:142`):

| endpoint | weight |
|---|---|
| `POST /api/remember` | 5 |
| `POST /api/remember/bulk` | 10 |
| `GET /api/remember/{job_id}` | 1 |
| `POST /api/remember/bulk/status` | 1 |

…against `max_requests_per_delegate_key: 30` per **sliding** 60s window (`rate_limit.rs:68`; documented default in `docs/relayer/self-hosting.md:126`, and nothing in the repo overrides it). The POST is paid out of the same bucket, so polling only ever gets 25 units (single) or 20 (bulk).

Simulating the new schedule at the default `pollIntervalMs: 1500` over the first 60s:

| scenario | polls | + POST | vs 30 |
|---|---|---|---|
| single, mean jitter | 21 | 26 | ok |
| single, worst jitter | 28 | 33 | **over** |
| bulk, mean jitter | 21 | 31 | **over** |
| bulk, worst jitter | 28 | 38 | **over** |
| single, old 10s cap | 9 | 14 | ok |

Growth *does* help versus flat polling (1500ms flat would be 40/min), so the mechanism is sound — but 3000ms leaves essentially no headroom, and the weight-10 bulk POST puts bulk over on its own. Any concurrent `recall` on the same delegate key tips single over too.

**Suggestion:** cap at **5s** rather than 3s. Worst case ~12 polls/min → 17/30 single, 22/30 bulk, both with real headroom, and you still halve the old worst-case latency-to-report (10s → 5s), which is the UX win the ticket is after. Either way the comment should be corrected — as written it asserts something the numbers don't support.

### 2. The resulting 429 surfaces as a timeout

Pre-existing, but this PR makes it likely. On rejection `isTransientPollingStatus(429)` is true, so both SDKs `continue` (`memwal.ts:332`, `client.py:440`), keep polling into the closed window, burn the remaining timeout, and throw `remember job timed out after 60000ms` / `MemWalRememberJobTimeout`. A quota rejection reaches the caller as a timeout.

Minimum: track whether any poll saw a 429 and say so in the timeout error. Better: honour `Retry-After` — `rate_limit_response` (`rate_limit.rs:463`) already sends the window.

### 3. Callers above 3s lose backoff entirely (minor)

`ceiling = Math.max(MAX_BACKOFF_MS, base)` combined with `capped = Math.min(ceiling, base * 1.5 ** …)` means that for `base > 3000` the result is flat `base` forever — `pollingDelayMs(5000, 1)` and `(5000, 15)` both return 5000. The changelog's "keep that interval" undersells it: those callers previously decayed toward 10s and now never back off at all. Someone who deliberately picked 5s for a long job sustains 12/min instead of decaying to 6/min. Worth wording as "backoff is disabled above the cap".

### 4. Python test patches stdlib `random` globally (nit)

`monkeypatch.setattr("memwal.client.random.random", …)` resolves `memwal.client.random` to the stdlib module object (`client.py` does a plain `import random`), so it patches `random.random` process-wide. `monkeypatch` unwinds it and pytest is single-threaded, so it works — it just reads as module-local when it isn't.

ducnmm:
@harrymove-ctrl gentle ping — still waiting on review of 451ecdc5024096aa6008089e4e4a7552b2a9991a.

harrymove-ctrl (dismissed):
## Review — WALM-623

**Depth:** targeted (adaptive). Cross-language parity would normally be a full seven-lens pass; I checked the TS/Python formula and the wait-loop conflict directly instead.

**Verdict: request changes.** The cadence is right. This head is not OK to merge.

### What is OK

- Immediate first poll, and dropping the climb to 10s. That is the idle WALM-623 measured (next poll 5–8s after a ~20s write was already `done`).
- Cap at **5s**, not the ticket's suggested 1–1.5s. A flat 1.5s is ~40 req/min, and status GETs share the default 30 weighted req/min delegate-key budget (`rate_limit.rs`, weight 1; `POST /api/remember` is 5). 5s keeps headroom and still cuts the old worst case in half. Backoff disabled above the cap is the right wording.
- `pollIntervalMs: 0` keeps the old 100ms floor after the first poll. No new `MemWal` methods or request fields. `pollingDelayMs` is not in package exports.
- Naming a 429 on the timeout is worth keeping.

Acceptance: the measured ~20s write no longer waits 5–10s. Worst jitter at the cap is 6.25s, not 10–12.5s. Formula tests pin attempt 20 at 5000 before jitter. That is enough for the unit-test criterion.

### What blocks merge

1. **`451ecdc` conflicts with `dev` (`f248bd37`).** `packages/sdk/src/memwal.ts` is a real logic conflict, not a changelog overlap. Taking this head's wait loop drops two fixes that landed after the branch point:
   - `pollDeadlineMs` on each status read. Without it, one stalled GET outlives `timeoutMs` again.
   - `retryAfterDelayMs` (`723baece`). Dev clamps `Retry-After` to the remaining budget and still does one status read. This PR returns null and skips that read when `Retry-After` does not fit.
2. **Changelog notes are under shipped versions.** TS notes sit in 0.1.7; Python in 0.1.10. Published heads are 0.1.8 and 0.1.11. Put them in unreleased 0.1.9 / 0.1.12. Do not rewrite the "latest release" blurbs, and do not bump `package.json` / `pyproject.toml` in this PR.
3. Python `client.py` should mirror the same combined policy and must keep the WALM-660 rule: a job missing from a bulk status stays pending.

### How to land it

Keep this PR's `pollingDelayMs` (attempt 0 is 0; later polls floor 100ms, grow 1.5× toward 5s, ±25% jitter). Keep dev's sleep shape: a clamped `Retry-After` wins and does not advance the exponent; otherwise use `pollingDelayMs`. Keep `{ timeoutMs: this.pollDeadlineMs(deadline) }` on both status calls. Append `; wait hit a rate limit (429)` only when the wait actually times out.

I am not approving `451ecdc`. Rebase onto `dev` with that resolution and this becomes an approve.

harrymove-ctrl:
Rebased onto current `dev` so this can be reviewed as a mergeable head: https://github.com/MystenLabs/MemWal/pull/1005

`451ecdc` still conflicts, and taking its wait loop would drop `pollDeadlineMs` and the clamped Retry-After read. #1005 keeps both, plus the 5s cap and immediate first poll. Please review that PR; this head should not be approved.

harrymove-ctrl:
Correction: ignore the note that pointed at #1005. That PR is closed.

The fix is on this branch now. Head is `5cddf3c5`, rebased onto current `dev`. It keeps the 5s cap and immediate first poll, and it also keeps `pollDeadlineMs` and the clamped Retry-After read. `451ecdc` is no longer the head.

This is the PR to review. I dismissed the changes-requested review, which was against `451ecdc`.

harrymove-ctrl (approved):
Approved. Head `5cddf3c5` is the rebase this review asked for.

Immediate first poll, 1.5× growth to a 5s cap, 100ms floor kept. Status reads still pass `pollDeadlineMs`. `Retry-After` is still clamped to the remaining budget and that remainder still buys one read. A 429 is named only on the timeout. Python matches, and a missing bulk status stays pending. No new public method, no version bump. Notes sit in unreleased 0.1.9 / 0.1.12.

PR #903, opened 2026-09-12
fix(mcp): stop the bridge waiting in silence on a throttled or failing handshake (WALM-386, WALM-618)

harrymove-ctrl:
Replaces #869 and #901. Same code, one branch — those two both rewrote `packages/mcp/src/bridge.ts` and **conflict with each other**, so whichever merged first would have forced the other to rebase. They also describe two halves of one user-visible failure, so reviewing them apart means reviewing half the story.

## Why they belong together

A throttled bridge and a stalled bridge look identical from the client: nothing happens for minutes, then a message blaming a layer that was never involved. Both fixes end that, at the two points where it starts.

**The handshake is refused (WALM-386, was #869).** The relayer answers `GET /api/mcp/sse` with 429 when the per-IP concurrent-session cap is full. The bridge retried ~500ms later, forever, and said nothing — so a rate limit read as a broken config or bad credentials. It now honours `Retry-After` (clamped to 60s), falls back to a 5s floor when the header is absent, carries the deadline across `reconnect()` calls that would each otherwise restart from a fresh sub-second backoff, and prints one stderr line per throttle episode naming it as a rate limit.

**The handshake keeps failing while a request waits (WALM-618, was #901).** A request buffered before any connection existed sat in `pendingForward` for the full 240s call timeout and then came back as *"the connection to the relayer dropped, please retry"* — a message about a reply that was lost, for a request that had never been sent. After 90s of consecutive handshake failures the bridge now answers it with the reason the handshake actually gave. A request that was never sent cannot have executed, so failing it early is provably a no-op; the full deadline still applies whenever the handshake is healthy.

## Measured against production

`relayer.memory.walrus.xyz`, sampled today over 125s:

| `GET /api/mcp/sse` | cumulative | share |
| --- | --- | --- |
| 200 | 50,958 | 4.6% |
| 401 | 784,627 | 70.1% |
| 429 | 128,062 | 11.4% |
| 503 | 155,874 | 13.9% |

Six handshakes with a valid, registered delegate key, one request each, 20s apart: **0 succeeded** — alternating 429 and 503. The 429s carry **no `Retry-After`**, which is the case this branch's 5s floor exists for; the 503s carry `Retry-After: 5`.

## What this does not fix

The 401s above are the largest share and this branch does not touch them. They are a client retry loop against credentials the relayer will never accept, and each one costs a Sui `GetObject` — `GetObject` volume (1,147,414) tracks total SSE requests (1,119,744) almost 1:1, which is what drives the fullnode into throttling and produces the 503s everyone else sees. #894 is the fix for that loop and is the higher-leverage change; it is left separate because it is @nikola0x0's. #900 caches the verify on the relayer side.

## Conflict resolution

Four conflict sites in `bridge.ts`, all additive — both sides adding complementary bookkeeping at the same place — except the `reconnect()` catch block, where the two error paths were folded into one `reason` binding and a single `log.error`. Every added line from both original branches is present verbatim in the merged file.

## Verification

- `pnpm --filter @mysten-incubation/memwal-mcp test` → **90 passed, 0 failed**, both suites together: `sse-handshake-429.test.mjs` (5 cases) and `pending-forward-stalled.test.mjs` (5 cases)
- `pnpm --filter @mysten-incubation/memwal-mcp typecheck` → clean

`MEMWAL_MCP_STALLED_HANDSHAKE_MS` was undocumented on #901; this branch adds its changelog entry and environment-variables row alongside the ones #869 already shipped.


harrymove-ctrl:
Folded into #900. Same commits, cherry-picked clean onto that branch — relayer and bridge now review as one change because they are one failure: the relayer refuses a valid caller during a Sui throttle, and the bridge waits out the refusals in silence.

Release still splits (deploy the relayer, publish the npm package), but the review does not have to.

Tests on the combined branch: 1,194 Rust, 81 MCP, 0 failures.

PR #910, opened 2026-09-14
fix(relayer): point the upload-queue saturation alert at real counters (WALM-608)

nikola0x0:
Closes WALM-608.

## The bug

The upload-queue saturation monitor in `main.rs` read `queuedWalrusUploads`, `activeWalrusUploads` and `walrusUploadLimits.globalCapacity` from the sidecar's `/health`. Since `cef97282` (v1_new port, 31 Jul) `/health` returns only `{ status, uptimeMs }`, so every field fell through to `unwrap_or(0)` and the alert could never fire. The monitor was added after the 2026-06-10 congestion incident, and has been blind to exactly that incident since 31 Jul.

## The change

**Sidecar:** `GET /metrics/uploads` returns the limiter counters and limits from memory. No auth, registered in both route modes next to `/metrics/wallet`. `/health` and `/ready` are unchanged.

**Relayer:** the monitor polls it. Parsing and the consecutive-check state moved to `src/sidecar_saturation.rs` so the alert decision is testable. A missing or non-integer field, a non-JSON body, or a non-2xx status now logs an `error` ("saturation alert is blind") instead of reading as an empty queue. An unreachable sidecar is still left to the watchdog.

**Why not poll `/ready`:** it has the counters, but waits on Sui gRPC, Walrus and an uncached archival GraphQL query (5s timeouts each) against the monitor's 2s budget, and the monitor skips silently on failure. Backlogs arrive with Sui RPC pressure (the June incident had 429s), so it would go blind exactly when needed.

## Tests

Test-first; each failed before the change.

- `sidecar-upload-metrics.test.ts` (full and writer mode): holds one running and one queued upload, and asserts the endpoint returns exactly those counts with no token and no Sui or Walrus call. Failed with 401 before the route existed.
- `sidecar_saturation::tests` (7): parsing, including the regression itself (a `/health` body is `Err("/queuedWalrusUploads")`, not an empty queue), plus the alert streak, the threshold boundary and drain.

Sidecar **257/257**, typecheck clean. `cargo test --bins` **715 passed, 0 failed** against pgvector and Redis as in CI. No new clippy warnings in the changed code.

## Not covered

A real relayer raising the Slack alert from a held backlog: a local boot needs contract IDs, Seal config and server wallets. Worth checking on staging; WALM-112's acceptance criteria already ask for that and for the threshold to be recorded.


ducnmm (commented):
## Summary

Approve with comment nits. WALM-608 is actually fixed: the saturation loop no longer treats a bare `/health` body as an empty queue. The monitor polls `/metrics/uploads`, `unwrap_or(0)` is gone, both sidecar modes register the route, and tests pin a `/health` body as an error rather than an empty queue. Unauthenticated `/metrics/uploads` matches existing `/metrics/wallet`; sidecar defaults to loopback.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 0

services/server/scripts/sidecar/routes/health.ts
```diff
@@ -138,6 +139,27 @@ export function registerHealthRoute(app: Express, requireProvenance = true): voi
     });
 }
 
+// Upload-queue metrics for the relayer's saturation monitor (main.rs).
+//
+// In-memory counters only, deliberately: /ready carries the same numbers but
```
ducnmm: **[suggestion]** The new route comment restates the PR’s design history (`/ready` waits on Sui/Walrus, backlogs arrive with RPC pressure). The handler is `getUploadCounts()` plus config constants; a future reader does not need the incident write-up in the source.

**Suggestion:** Keep a one-liner that it is in-memory limiter counters for the relayer probe, no I/O. Move the `/ready` comparison to the PR/commit body (already there).

services/server/scripts/__tests__/sidecar-upload-metrics.test.ts
```diff
@@ -0,0 +1,73 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import type { Server } from "node:http";
+
+// The relayer's upload-queue saturation monitor polls this endpoint every
```
ducnmm: **[suggestion]** Same architecture essay sits above the env setup. The test name, no-token `fetch`, and `dependencyCalls === 0` already pin “unauthenticated, no Sui/Walrus.”

**Suggestion:** Drop the block, or one line: live limiter counters, no sidecar token, no chain I/O.

services/server/src/sidecar_saturation.rs
```diff
@@ -0,0 +1,179 @@
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct UploadQueueSample {
+    pub queued: u64,
+    pub active: u64,
+    pub global_capacity: u64,
+}
+
+/// Reads the counters from a `/metrics/uploads` body. The error names the
+/// first missing or non-integer field; reading it as 0 instead is how this
```
ducnmm: **[suggestion]** `parse_upload_metrics` rustdoc embeds the WALM-608 failure mode (“reading it as 0 instead is how this alert went blind”). That is ticket narrative, not an invariant of the function. The const docs at lines 8–10 similarly list what the path is *not* (`/health`, `/ready`).

**Suggestion:** Document that missing/non-integer fields are `Err(json-pointer)` and must not be read as 0. Point `UPLOAD_METRICS_PATH` at the sidecar route only.

services/server/src/sidecar_saturation.rs
```diff
@@ -0,0 +1,179 @@
...
+            parse_upload_metrics(&body),
+            Ok(UploadQueueSample {
+                queued: 118,
+                active: 5,
+                global_capacity: 5,
+            })
+        );
+    }
+
+    // WALM-608: the monitor polled /health, whose body has none of these
```
ducnmm: **[suggestion]** The test comment retells that the monitor used to poll `/health` and coerce missing fields to 0. The test name `a_liveness_body_is_an_error_not_an_empty_queue` already states the contract.

**Suggestion:** Delete the comment (or leave the ticket id in the test name if you want a grep handle).

services/server/scripts/sidecar/routes/health.ts
nikola0x0: Trimmed to a one-liner in 874f9573; the /ready comparison stays in the PR body.

services/server/scripts/__tests__/sidecar-upload-metrics.test.ts
nikola0x0: Dropped in 874f9573.

services/server/src/sidecar_saturation.rs
nikola0x0: Now documents the contract (Err with the field's JSON pointer, never 0), and the const only points at the sidecar route. 874f9573.

services/server/src/sidecar_saturation.rs
nikola0x0: Deleted in 874f9573.

ducnmm (commented):
## Summary

Re-review of `874f9573`. Follow-up trim closes the prior comment suggestions. WALM-608 still holds (`/metrics/uploads`, no `unwrap_or(0)`). 0 issues.

ducnmm (approved):
## Summary

Re-review of `874f9573`. Comment-trim follow-up closed the prior suggestions. WALM-608 still holds (`/metrics/uploads`, no `unwrap_or(0)`). Approving.

PR #911, opened 2026-09-14
fix(server): recall sort precedence and the embedding size check (WALM-470)

nikola0x0:
Closes WALM-470. Two independent fixes, one commit each.

## 1. An explicit `sort` is no longer reordered by `scoring_weights`

`/api/recall` with both `sort` and `scoring_weights` returned neither order. `select_hits_for_sort` ordered the rows, then the ranker reordered them by composite score, so `sort=recent` stopped meaning newest-first as soon as the request also carried recency weights. The TS SDK's `recall()` sends both when a caller sets both. This is a bug in the WALM-383 feature, not a regression: callers that set only one of the two are unaffected.

Follows the decision on WALM-460 (2026-09-03): an explicit `sort` is the order, and weights apply only when `sort` is omitted.

- `RecallRequest.sort` is now `Option<RecallSort>`, so an omitted `sort` and `"relevance"` differ.
- `resolve_scoring_weights` validates the weights, then returns the defaults when `sort` is set. Malformed weights are still a 400.
- The recall request log gets a `scoring_weights_ignored` field.
- The SDK JSDoc, SDK API reference and relayer API reference state the rule. The relayer reference didn't mention `sort` before.

## 2. The WALM-423 embedding size check is narrowed (regression)

- **Size check:** it ran before the embedder looked for an API key, so deployments without one (local dev, CI, self-hosted) rejected text over 16 KiB, though the key-less fallback hashes locally with no limit. It now runs only when a key is set.
- **400 mapping:** every provider 400 was reported as "input exceeds the model context limit". Now only a 400 whose body names a context-length problem is `BadRequest`; anything else, such as an unknown model id, stays `Internal`, so it isn't blamed on the caller and still alerts.

Production's normal path has a key and summarises long `remember` text first, so it wasn't affected by this half.

The embed call moved into `embed_text` (key, base URL, text) so tests can run it against a local stand-in provider. No behaviour change from the move itself.

## Tests

Test-first. Each bug test failed on `dev`:

- Explicit `recent` and explicit `relevance` are not reordered by weights.
- An omitted `sort` and `"relevance"` deserialize differently (did not compile on `dev`).
- Without a key, 100 KiB of text embeds.
- A provider 400 saying "model does not exist" (and three other non-length bodies) is `Internal`.

Guards that already passed and must keep passing: an omitted `sort` still re-ranks; malformed weights are rejected alongside an explicit `sort`; with a key, oversized input is rejected before calling the provider; length-complaint 400s stay `BadRequest`.

`cargo test --bins` against pgvector and Redis: **717 passed, 0 failed**. SDK typecheck clean.


services/server/src/routes/mod.rs
```diff
@@ -461,6 +483,132 @@ mod recall_sort_tests {
     }
 }
 
+/// WALM-470: an explicit `sort` is the order, and `scoring_weights` apply only
```
ducnmm: **[suggestion]** New comments narrate the ticket and the design rather than a non-obvious constraint. The test-module rustdoc records “decided 2026-09-03 on WALM-460”. `resolve_scoring_weights` retells WALM-470 and why default weights were chosen as the suppression mechanism. `embed_text` restates the const comment and stamps WALM-470 on the call site.

**Suggestion:** Keep a one-line why (explicit `sort` must not be re-ranked; the byte limit is the model window, so it guards only the keyed branch). Drop dates, ticket IDs, and the “one code path” history.

ducnmm (dismissed):
## Summary

The two WALM-470 server fixes are correct and well covered. `RecallRequest.sort` is now `Option<RecallSort>`, so omitted vs explicit `"relevance"` stay distinct. `resolve_scoring_weights` validates first, then substitutes `ScoringWeights::default()` when `sort` is set. The embedding size check now runs only on the provider-key branch; provider 400s become `BadRequest` only when the body names a context-length problem.

Blocking: `packages/sdk/src/types.ts` JSDoc changed, but the unpublished 0.1.7 changelog dump is missing. Do not bump the version.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

## Issues outside the diff

- **[bug]** packages/sdk/CHANGELOG.md:3 — This PR edits `packages/sdk/src/types.ts` (`RecallOptions.scoringWeights` / `sort` JSDoc: an explicit `sort`, including `"relevance"`, makes the relayer ignore `scoringWeights`). Origin/dev is already unpublished 0.1.7, so that contract change must be dumped under `## 0.1.7` in both `packages/sdk/CHANGELOG.md` and `docs/sdk/changelog.mdx` (including the frontmatter `answer`). Neither file changed.
  - **Suggestion:** Do not bump the version. Add a `## 0.1.7` Fixed (or Changed) bullet in `packages/sdk/CHANGELOG.md` and the same under `docs/sdk/changelog.mdx` `## 0.1.7`, and fold it into the `answer:` blurb. Example: an explicit `sort` (including `"relevance"`) makes the relayer ignore `scoringWeights`; weights re-rank only when `sort` is omitted (WALM-470).

services/server/src/routes/mod.rs
nikola0x0: Trimmed in 5e513449: one-line whys on resolve_scoring_weights and RecallRequest.sort, no ticket ids or dates, and the embed_text call-site comment is gone (the const doc covers it).

nikola0x0:
@ducnmm changelog: added under the unreleased `## 0.1.7` in both `packages/sdk/CHANGELOG.md` and `docs/sdk/changelog.mdx` (Fixed bullet, the release intro and the `answer:` blurb), no version bump. Also merged latest `dev`. 5e513449.

ducnmm (commented):
## Summary

Re-review of `5e513449`. Previous CHANGES_REQUESTED (dump SDK 0.1.7 missing on `e7d29855`) is addressed: unpublished `## 0.1.7` in `packages/sdk/CHANGELOG.md` and `docs/sdk/changelog.mdx` (Fixed bullet + intro + `answer:`). No bump (`dev` already 0.1.7). Sort-precedence + embed-size check still look correct. 0 new issues.

ducnmm (approved):
## Summary

Re-review of `5e513449`. Previous dump REQUEST_CHANGES is addressed (unpublished 0.1.7 in package + mdx changelogs, no bump). Sort-precedence and the narrowed embed-size check still look correct. Approving.

PR #912, opened 2026-09-14
fix(mcp): make dev usable again — dial loopback instead of the deployment own public edge

harrymove-ctrl:
Makes the **dev** deployment usable again, and folds in the two client-side fixes from the field report that ship with the same package. One PR — #914 was opened separately by mistake and is closed in favour of this.

## Dev is not slow, it is non-functional

Ran the reported scenario against dev, which carries both halves of #900:

| tool | result |
| --- | --- |
| `memwal_recall` | ❌ **504 after 94.9s** |
| `memwal_remember_bulk` (5 facts, 4.1 KB) | ❌ **504 after 61.3s** |
| `memwal_recall` (verify) | ❌ **504 after 60.8s** |
| `memwal_remember` (single) | ❌ **504 after 60.8s** |
| `memwal_health` | ✅ ok — but **60.9s** |

Only `memwal_health` returns, because it is an unsigned `GET /health` with no SEAL preamble. **Everything that reads or writes memory fails.**

## The requests never reach the relayer

Relayer counters across that run (2 recalls + 1 bulk + 1 remember):

| route | before | after | delta |
| --- | --- | --- | --- |
| `/api/recall` | 46 | 46 | **+0** |
| `/api/remember` | 19 | 19 | **+0** |
| `/config` | 91 | 96 | +5 ← SDK preamble does get through |

And the same endpoint, called from the internet:

```
POST https://relayer.dev.memwal.ai/api/recall  →  401 in 0.29s
```

**0.29s from outside. 60s and a 504 from inside the same container.** The requests die between the sidecar and the relayer.

## Cause

Read from Railway, per environment:

| env | `MEMWAL_RELAYER_URL` — what the sidecar **dials** | `memwal_health` |
| --- | --- | --- |
| prod | `https://relayer.memory.walrus.xyz` | 0.3s |
| staging | `https://relayer.staging.memwal.ai` | 0.3s |
| **dev** | `https://relayer.dev.memwal.ai` | **61s, and 504 on every real call** |

All three dial their own public hostname; none sets `MEMWAL_PUBLIC_RELAYER_URL`. Every `memwal_*` call leaves the container and comes back to the process it started from. Two of those round trips are cheap and dev's is fatal. Why dev's specifically is fatal is not settled here — this PR removes the round trip rather than explaining it.

## The trap in the obvious fix

"Just stop setting `MEMWAL_RELAYER_URL`" **would take every Claude custom connector offline**, and the first revision of this PR shipped a startup warning saying exactly that. It is also the MCP OAuth issuer:

```rust
// oauth.rs:161, inside McpOAuthConfig::from_env
let issuer = std::env::var("MEMWAL_RELAYER_URL").ok()?;
```

The `?` returns `None`, leaving `config.mcp_oauth` empty, and every OAuth handshake refused with `OauthNotConfigured` (`mcp_proxy.rs:397`). All three environments have `MCP_OAUTH_DELEGATE_ENCRYPTION_KEY` set.

So the **dial address** is what moves, not the public identity:

- `MEMWAL_RELAYER_URL` keeps its meaning exactly — OAuth issuer, and the network `memwal_health` names.
- What the sidecar dials is now **loopback**, unless `MEMWAL_SIDECAR_RELAYER_URL` explicitly says otherwise. Almost nobody should set it: the managed sidecar is a child of the relayer process.

**No deployment has to change an environment variable.** Merging this is the whole fix for dev.

## Make the next one findable

The only reason this was locatable was diffing a Prometheus counter against a wall clock — the relayer's histogram starts when the request *lands*, so a minute spent reaching it reads as a healthy 9.6ms, and the sidecar logged `tool.call` going in and nothing coming out.

| event | level | when |
| --- | --- | --- |
| `tool.slow` (`settled: false`) | warn | **still running** past `MCP_TOOL_SLOW_WARN_MS` (default 5000) |
| `tool.slow` (`settled: true`) | warn | finished over the threshold; `alreadyWarned` pairs them |
| `tool.done` | info | finished under it |
| `tool.failed` | warn | threw; carries `errName`, `errMessage`, `causeCode` |

The in-flight line is load-bearing: a settle-only log cannot see a hang, and a hang is what this was written for.

## Two client fixes in the same package

**A retry that duplicates a paid write.** A `memwal_remember_bulk` whose reply was lost said *"the connection to the relayer dropped… Please retry."* But the relayer returns 202 and finishes in a durable queue, so a client deadline cancels nothing, and `/api/remember/bulk` carries no idempotency key (unlike the single path) with no content dedupe. That advice mints a second paid copy. #900 rewrote this string for the two never-sent cases and copied the already-sent branch byte for byte — and a bulk POSTed on a healthy session lands in that branch. Now a sent **write** says it may have completed and to check with `memwal_recall` first; a sent **read** says plainly that retrying is safe.

**Unpinned launcher.** `.mcp.json`, `.cursor-mcp.json` and `.codex-mcp.json` all shipped `npx -y @mysten-incubation/memwal-mcp` with no version. `npx` then cached `^0.0.5`, and a caret on a `0.0.x` version locks the exact patch — a machine kept launching 0.0.5 for weeks after 0.0.9 through 0.0.12 shipped. Now `@latest`.

## Tests

`cargo test -p memwal-server --bin memwal-server` → **697 passed**, 7 new. The 47 failures are all `PermissionDenied: Operation not permitted` in tests that bind local sockets under this sandbox; none touch `main.rs`.

Sidecar suite → **75 passed** across every file that does not bind a socket, 4 new in `tool-duration-log.test.ts`. `tsc --noEmit` clean in both `services/server/scripts` and `packages/mcp`.

The `packages/mcp` suite could not run here — it binds localhost stubs and the sandbox denies `listen` (EPERM). Checked by hand instead: the only occurrence of the old string in `packages/mcp/test/` is a doc comment in `pending-forward-stalled.test.mjs`, so no test asserts it, and that file covers the never-sent branches this PR leaves alone. A test pinning the new write-case wording still needs adding by whoever can run the suite.

| new test | pins |
| --- | --- |
| `the_deployed_shape_now_dials_loopback_while_naming_the_same_network` | the real prod/staging/dev env shape, fixed with no env change |
| `nothing_set_dials_loopback_and_names_no_network` | the default still reports no network |
| `an_explicit_public_origin_wins_over_the_relayer_url` | precedence |
| `only_the_dedicated_override_can_move_the_dial_off_this_host` | `MEMWAL_RELAYER_URL` can no longer steer the dial |
| `an_explicit_loopback_override_is_not_warned_about` | no false alarm |
| `a_credential_in_the_dial_url_is_never_logged` | userinfo never reaches a log |
| `every_loopback_spelling_is_recognised_as_in_process` | `127.0.0.0/8`, `localhost.`, `[::1]`, `[::ffff:127.0.0.1]`, case |
| `a call still running past the threshold is reported BEFORE it settles` | the watchdog can see a hang |
| `a completed tool call reports how long it took` | duration + dialled address |
| `a call slower than the threshold is warned about, not filed as normal` | reported once, as slow |
| `a failing tool call still reports its duration` | error identity in the structured line |


services/server/scripts/mcp/tools/util.ts
```diff
@@ -58,9 +87,69 @@ export function wrapTool<Args>(
...
+        // How long the call takes is the only number that shows a slow hop to
+        // the relayer. Per-request latency on the relayer side cannot: it
+        // measures the request once it lands, so a minute spent reaching it
+        // reads as a healthy few milliseconds there and as silence here.
+        const startedAt = Date.now();
+        const durationMs = () => Date.now() - startedAt;
+        const outcomeFields = () => ({
+            tool,
+            durationMs: durationMs(),
+            relayerUrl: session.relayerUrl ?? null,
```
ducnmm: **[bug]** The new `outcomeFields()` copies `session.relayerUrl` into every `tool.slow` / `tool.done` / `tool.failed` line with no userinfo strip. That field is the sidecar dial URL (`MEMWAL_SIDECAR_RELAYER_URL` when set). `redact_url_userinfo` in `main.rs` only covers Rust startup warnings and the "dialling relayer at" info line; `a_credential_in_the_dial_url_is_never_logged` never exercises `wrapTool`. `https://ops:hunter2@host` therefore lands in JSON stderr on every slow/done/failed call.

**Suggestion:** Redact userinfo in the sidecar before logging (mirror the Rust helper, or log host+port only). Add a wrapTool test whose session `relayerUrl` contains `user:password@` and assert no secret in captured stderr. Keep passing the real URL to `fetch`.

docs/reference/environment-variables.md
```diff
@@ -177,7 +180,8 @@ These are not all enforced at boot, but most real deployments need them.
 - `MEMWAL_PACKAGE_ID` and `MEMWAL_REGISTRY_ID` are server env vars. Do not replace them with `VITE_*` app env vars.
 - For network-specific `MEMWAL_PACKAGE_ID` and `MEMWAL_REGISTRY_ID` values, see [Contract Overview](/contract/overview).
 - `MEMWAL_RELAYER_URL` is only needed when the sidecar should call a different relayer URL than the Rust server's local port. The Rust server sets it automatically to `http://127.0.0.1:$PORT` for the managed sidecar when it starts.
```
ducnmm: **[suggestion]** The table and the new notes bullet correctly say `MEMWAL_RELAYER_URL` is public identity / OAuth issuer and must stay set. This preceding notes bullet was not updated: "MEMWAL_RELAYER_URL is only needed when the sidecar should call a different relayer URL… The Rust server sets it automatically to `http://127.0.0.1:$PORT`". That is the revision-1 footgun — operators who unset it still take `McpOAuthConfig::from_env` to `None` and every OAuth handshake to `OauthNotConfigured`.

**Suggestion:** Delete or rewrite this bullet so it cannot be read as "unset MEMWAL_RELAYER_URL". Point dial overrides at `MEMWAL_SIDECAR_RELAYER_URL` only.

services/server/scripts/mcp/tools/util.ts
```diff
@@ -58,9 +87,69 @@ export function wrapTool<Args>(
...
+            });
+        }, SLOW_TOOL_WARN_MS);
+        watchdog.unref?.();
+        let warnedInFlight = false;
+        const stopWatchdog = () => {
+            // `hasRef()` is false once the timer has fired, which is how the
+            // settle-time line knows whether the in-flight one already went out
+            // and can avoid reporting the same call twice.
+            warnedInFlight = watchdog.hasRef ? !watchdog.hasRef() : false;
+            clearTimeout(watchdog);
```
ducnmm: **[suggestion]** `alreadyWarned` is computed as `!watchdog.hasRef()` after `watchdog.unref()`. Node's `Timeout.hasRef()` is false immediately after `unref()`, while the timer is still pending. The comment ("hasRef() is false once the timer has fired") is therefore wrong, and `alreadyWarned` is always `true` on settle whenever `hasRef` exists. The in-flight `tool.slow` still fires; only the pairing bit is unreliable.

**Suggestion:** Set a boolean inside the `setTimeout` callback (`warnedInFlight = true` then log). Leave `unref()`. Drop `hasRef`.

docs/reference/environment-variables.md
```diff
@@ -148,10 +148,13 @@ These are not all enforced at boot, but most real deployments need them.
...
+| `MEMWAL_SIDECAR_RELAYER_URL` | `http://127.0.0.1:$PORT` | Address the sidecar **dials** for every MCP tool call. Loopback, because the managed sidecar is a child of the relayer process. Set it only if this sidecar genuinely serves a relayer in another process — a non-loopback value sends each `memwal_*` call out of the container and back through the public edge, and the server warns at startup when it sees one |
+| `MEMWAL_PUBLIC_RELAYER_URL` | falls back to `MEMWAL_RELAYER_URL` | Public origin `memwal_health` reports, when it should differ from `MEMWAL_RELAYER_URL` |
 | `MCP_MAX_TOTAL_SESSIONS` | `1000` | Maximum active MCP sessions across SSE and Streamable HTTP transports |
 | `MCP_MAX_SESSIONS_PER_IP` | `16` | Maximum active MCP sessions from one source IP |
 | `MCP_MAX_NEW_SESSIONS_PER_IP_PER_MIN` | `30` | Maximum new MCP sessions opened by one source IP per minute |
+| `MCP_TOOL_SLOW_WARN_MS` | `5000` | A completed MCP tool call at or above this duration is logged as `tool.slow` at `warn` instead of `tool.done` at `info` |
```
ducnmm: **[nit]** `MCP_TOOL_SLOW_WARN_MS` is documented as changing how a *completed* call is logged. The load-bearing behaviour this PR adds is the in-flight `tool.slow` (`settled: false`) that fires while the call is still running.

**Suggestion:** Mention the in-flight warn in the table cell, not only the settle path.

ducnmm (changes requested):
## Summary

The deployed-shape fix is right: `MEMWAL_RELAYER_URL` stays the public identity / OAuth issuer, the managed sidecar dials `http://127.0.0.1:$PORT` unless `MEMWAL_SIDECAR_RELAYER_URL` is set, and the in-flight `tool.slow` watchdog is the right signal for the 61s hang. One bug: per-call sidecar log lines still emit dial `userinfo`. Two suggestions: leftover notes bullet still describes the old (dangerous) meaning of `MEMWAL_RELAYER_URL`, and `alreadyWarned` is wrong after `unref()`.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 1

services/server/scripts/mcp/tools/util.ts
ducnmm: **[bug]** Still open from the last review. `outcomeFields()` copies `session.relayerUrl` into every `tool.slow` / `tool.done` / `tool.failed` line with no userinfo strip. `redact_url_userinfo` in `main.rs` does not cover wrapTool. `https://ops:hunter2@host` lands in JSON stderr.

**Suggestion:** Redact userinfo in the sidecar before logging. Add a wrapTool test with `user:password@`. Keep the real URL for `fetch`.

packages/mcp/src/bridge.ts
```diff
@@ -2198,23 +2198,69 @@ export async function runBridge(
...
         if (!neverSent) {
+            // The request reached the relayer. What is missing is the reply,
+            // and for a write that distinction is the whole message: the work
+            // may have completed, may still be running, and cannot be assumed
+            // undone. "Please retry" is only safe advice for a read.
+            if (tool !== null && MUTATING_TOOLS.has(tool)) {
+                return {
+                    reason: "no response to a sent write",
+                    opts: {
+                        toolText:
```
ducnmm, line 2240: **[bug]** Sent-write `toolText` correctly says do not repeat / `memwal_recall` first. Existing `packages/mcp/test/orphaned-call.test.mjs` still asserts `/retry/i` on this path ("safe to retry"). CI `MCP / Integration` failed on this head.

**Suggestion:** Pin the new write-case wording and drop "safe to retry". Keep `/retry/i` only for never-sent / sent-read.

services/server/scripts/mcp/__tests__/tool-duration-log.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    const slow = async () => {
+        await new Promise((r) => setTimeout(r, SLOW_THRESHOLD_MS * 2));
+        return ok();
+    };
+
+    const { lines } = await capturingLogs(() =>
+        wrapTool(SESSION, "memwal_health", slow)({})
+    );
+
+    const warned = lines.find((l) => l.event === "tool.slow");
```
ducnmm: **[bug]** `lines.find(event === tool.slow)` hits the in-flight watchdog (`settled: false`), which can fire a millisecond early (`durationMs` 149 vs 150). Flake on Compile & CLI Smoke.

**Suggestion:** Assert on `settled === true`. Do not require in-flight `durationMs >= threshold`.

packages/mcp/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Fixed
 
+- Stop telling the user to retry a write whose reply was lost. A `memwal_remember`, `memwal_remember_bulk` or `memwal_analyze` that was POSTed and then timed out came back as "the connection to the relayer dropped before the result came back. Please retry." — but the relayer answers those with HTTP 202 and finishes the work in a durable queue, so a client-side deadline cancels nothing and the write may already have landed. `/api/remember/bulk` carries no idempotency key, unlike the single path, so following that advice stores a second paid copy that `recall` then hides behind the first. A sent write now says it may have completed, that the timeout did not undo it, and to check with `memwal_recall` before re-saving. A sent read still says plainly that retrying is safe. (WALM-618 follow-up)
```
ducnmm, line 7: **[bug]** These 0.0.13 Fixed bullets are only in the package changelog. `docs/mcp/changelog.mdx` `## 0.0.13` / intro / `answer:` still omit them. No bump (`dev` already 0.0.13).

**Suggestion:** Copy both bullets into the mdx and mention no-retry-on-sent-write in the intro/`answer:`.

docs/reference/environment-variables.md
ducnmm: **[suggestion]** Still unfixed. This notes bullet still reads as "unset MEMWAL_RELAYER_URL". Unsetting it disables OAuth (`McpOAuthConfig::from_env`).

**Suggestion:** Delete or rewrite. Point dial overrides at `MEMWAL_SIDECAR_RELAYER_URL` only.

services/server/scripts/mcp/tools/util.ts
```diff
@@ -58,9 +87,69 @@ export function wrapTool<Args>(
...
+                settled: false,
+            });
+        }, SLOW_TOOL_WARN_MS);
+        watchdog.unref?.();
+        let warnedInFlight = false;
+        const stopWatchdog = () => {
+            // `hasRef()` is false once the timer has fired, which is how the
+            // settle-time line knows whether the in-flight one already went out
+            // and can avoid reporting the same call twice.
+            warnedInFlight = watchdog.hasRef ? !watchdog.hasRef() : false;
```
ducnmm: **[suggestion]** Still unfixed. `hasRef()` is false immediately after `unref()`, so `alreadyWarned` is always true on settle.

**Suggestion:** Set a boolean inside the `setTimeout` callback. Leave `unref()`. Drop `hasRef`.

packages/mcp/plugin/.mcp.json
```diff
@@ -2,7 +2,7 @@
   "mcpServers": {
     "memwal": {
       "command": "npx",
-      "args": ["-y", "@mysten-incubation/memwal-mcp"]
+      "args": ["-y", "@mysten-incubation/memwal-mcp@latest"]
```
ducnmm: **[suggestion]** `@latest` is a moving dist-tag, not a pin, and duplicates open WALM-627 / #913 (which pins `@0.0.13`). Codex installer on this PR is still unpinned.

**Suggestion:** Drop launcher JSON from this PR and leave WALM-627 to #913. If they must ship here, pin the same exact version as `packages/mcp/package.json`.

docs/reference/environment-variables.md
ducnmm: **[nit]** Still unfixed. Table cell describes only the completed-call log. The load-bearing behaviour is in-flight `tool.slow` (`settled: false`).

**Suggestion:** Mention the in-flight warn in the table cell.

ducnmm (changes requested):
## Summary

Re-review of `27e48aa9`. Loopback rewrite is still the right deployed-shape fix, and sent-write copy no longer says retry. Previous REQUEST_CHANGES items are still open (wrapTool userinfo, leftover unset-RELAYER_URL notes, hasRef after unref). New blockers: `orphaned-call.test.mjs` still asserts `/retry/i` on a sent remember (CI MCP Integration failed); slow-call test flakes on the in-flight `tool.slow` line; `docs/mcp/changelog.mdx` was not dumped under unpublished 0.0.13. Launcher `@latest` duplicates WALM-627/#913 and is not a pin.

## Issue counts by severity

- bugs: 4
- suggestions: 3
- nits: 1

## Issues outside the diff

- **[bug]** docs/mcp/changelog.mdx — MCP package changed; `packages/mcp/CHANGELOG.md` has two new 0.0.13 Fixed bullets, but the Mintlify mdx `## 0.0.13` / intro / `answer:` omit them. Copy the bullets. Leave package.json at 0.0.13.

harrymove-ctrl:
Thanks @ducnmm — all four are right, and the bug one was a real leak. Fixed in 195e9b67.

**[bug] dial userinfo in per-call logs.** Correct, and my Rust-side `redact_url_userinfo` gave me false confidence — it only covers the startup lines, and `a_credential_in_the_dial_url_is_never_logged` never goes near `wrapTool`, exactly as you said. Added `redactUrlUserinfo` in the sidecar mirroring the Rust helper, applied it in `outcomeFields()`, and added two tests that go through `wrapTool` with a `user:password@` session URL. One extra decision worth flagging: an unparseable value is now returned as `null` rather than echoed — it cannot be redacted, so it should not be shown. `fetch` still receives the real URL.

**[suggestion] `alreadyWarned` after `unref()`.** You are right and my comment was simply wrong: `hasRef()` is false from the moment `unref()` is called, not once the timer fires, so the bit was always `true` on settle. Now set inside the timer callback, `unref()` kept, `hasRef` dropped.

**[suggestion] stale notes bullet.** Deleted and rewritten. It read as permission to unset `MEMWAL_RELAYER_URL`, which is the revision-1 footgun. Dial overrides now point at `MEMWAL_SIDECAR_RELAYER_URL` only, and the bullet names the OAuth consequence explicitly.

**[nit] `MCP_TOOL_SLOW_WARN_MS` docs.** Agreed — the in-flight warn is the load-bearing half, since a hang never settles. The table cell now leads with it.

---

Both CI failures are also fixed in the same commit, and one of them was mine:

- **Compile & CLI Smoke** — my own test was flaky. It asserted `durationMs >= threshold` on the *first* `tool.slow` line, which is now the in-flight one emitted by the timer at the threshold itself. Its duration sits within a millisecond of the threshold and landed at 149 against 150 on the runner. It now asserts on the settled line.
- **MCP / Integration (login handoff)** — `orphaned-call.test.mjs` asserted that an orphaned `memwal_remember` tells the caller it is "safe to retry". That is the assumption this PR reverses on purpose: the call was POSTed, the relayer answered 202 and finished in a durable queue, and `/api/remember/bulk` carries no idempotency key, so a blind repeat buys a second paid blob. The test now pins the write-case contract instead and is renamed. The negative is scoped to `please retry` rather than `nothing was stored`, because the message deliberately says it *does not mean* nothing was stored.

One limitation I should state rather than paper over: the `packages/mcp` suite cannot run in my sandbox (it binds localhost stubs, `listen` is denied), so `orphaned-call.test.mjs` is verified by CI here, not locally. I checked each new assertion against the emitted string by hand before pushing.

ducnmm (approved):

harrymove-ctrl:
@ducnmm the re-review landed on `27e48aa9`, 36 seconds before I pushed `195e9b67` — so the four "still unfixed" items were already fixed in that commit, which is why CI on it is green (20/20). Two items in your re-review were genuinely new, and both are now addressed in `4f2bad7b`.

**Already fixed in `195e9b67`** (pushed at 04:03:37, your review submitted 04:03:01):

- wrapTool userinfo → `redactUrlUserinfo` in the sidecar, applied in `outcomeFields()`, with two tests going through `wrapTool`. An unparseable value returns `null` rather than being echoed.
- `hasRef()` after `unref()` → boolean set inside the timer callback, `unref()` kept, `hasRef` dropped.
- stale notes bullet → rewritten, points dial overrides at `MEMWAL_SIDECAR_RELAYER_URL` only and names the `McpOAuthConfig::from_env` consequence.
- `MCP_TOOL_SLOW_WARN_MS` table cell → now leads with the in-flight warn.
- `orphaned-call.test.mjs` → asserts the write-case contract, renamed. Scoped the negative to `please retry` rather than `nothing was stored`, because the message deliberately says it *does not mean* nothing was stored.
- slow-call flake → asserts on `settled === true`, exactly as you suggested.

**New, fixed in `4f2bad7b`:**

- **`docs/mcp/changelog.mdx`** — you are right and this one mattered more than it looks. The mdx is the user-facing page and its `answer:` block is what AI search cites, so a fix that lives only in the package CHANGELOG is invisible where people look. Copied the bullet into `### Fixed` and led both the intro and `answer:` with it, since "a lost reply does not mean the write did not land" is the sentence someone needs *before* they retry something that costs money. No version bump.
- **Launcher `@latest`** — agreed, and I have dropped the launcher JSON from this PR entirely. You are right that `@latest` is a moving dist-tag rather than a pin, and #913 does it properly: the exact `@0.0.13` that `package.json` declares, plus the Codex fallback installer, which I had not touched. Two PRs editing the same three files with different values is a conflict for no benefit. Removed the matching CHANGELOG bullet with it.

That puts this PR back to one subject: the sidecar dialling loopback, the in-flight slow-call signal, and the sent-write message.

One thing I still cannot do here and would rather flag than hide: the `packages/mcp` suite does not run in my sandbox (it binds localhost stubs, `listen` is denied), so `orphaned-call.test.mjs` is verified by CI rather than locally. I checked each new assertion against the emitted string by hand before pushing.

PR #914, opened 2026-09-15
fix(mcp): stop advising a retry that duplicates a paid write, and pin the launcher

harrymove-ctrl:
Addresses §1b, §3 and §4 of the 2026-09-14 field report, as far as the npm package can reach them. Companion to #912 (relayer side). Nothing here needs a relayer deploy — it ships with the next `memwal-mcp` release.

## 1. "Please retry" was advice that duplicates a paid write

A `memwal_remember_bulk` whose reply never came back reported:

```
❌ Walrus Memory did not answer this call. The connection to the relayer
   dropped before the result came back. Please retry.
```

Every clause after the first is wrong for a write, and the last one costs money:

- `/api/remember/bulk` returns **HTTP 202** and finishes in a durable Postgres/Apalis queue (`routes/remember.rs:1343` → `jobs.rs:2808`). A client-side deadline cancels nothing — items can still be landing long after the bridge gave up.
- The SDK pre-seeds every unfinished item as `status:"timeout"` and then computes `failed = total - succeeded` (`packages/sdk/src/memwal.ts:554`), so still-running work is reported as failed.
- A genuine `status:"failed"` can also mean the blob was **already minted and paid**: `fail_stale_remember_jobs` rewrites `'uploaded' → 'failed'` while leaving `blob_id` populated (`storage/db.rs:2114`).
- **`/api/remember/bulk` carries no idempotency key** — unlike `RememberRequest`, whose key exists precisely "so an ambiguous timeout can't produce a duplicate paid on-chain blob" (`types.rs:1237`) — and there is no content dedupe on the write path.

So the message told the user to mint a second paid copy of memories that may already be stored, which `recall` then hides behind the first.

#900 rewrote this string for the two cases where the call **never left the bridge**. It copied the already-sent branch across byte for byte, and that is the branch this incident lands in: `postIfCurrent` sets `sent = true` *before* awaiting the POST (`bridge.ts:1100`), deliberately — once the request is issued we can no longer prove it did not run.

This PR splits that branch by what the call does:

| case | message |
| --- | --- |
| sent `memwal_remember` / `_bulk` / `_analyze` | the write may have completed; a timeout does not undo it; check with `memwal_recall` before re-saving |
| sent read | plainly safe to retry — which is what the old text meant |
| never sent (both #900 branches) | unchanged |

## 2. Pin the launcher

All three launcher configs shipped the same unpinned spec:

```json
{ "command": "npx", "args": ["-y", "@mysten-incubation/memwal-mcp"] }
```

Unpinned, `npx` records `^0.0.5` in its cache, and a caret on a `0.0.x` version locks the exact patch — so a reporting machine kept launching 0.0.5 for weeks after 0.0.9 through 0.0.12 shipped, and updating the plugin did not move it. Now `@latest` in `.mcp.json`, `.cursor-mcp.json` and `.codex-mcp.json`.

## What this does NOT fix, and cannot

Worth stating plainly, because the report's top ask is not reachable from this package:

`memwal-mcp` is a pure stdio↔SSE transport. Its dependencies are `@modelcontextprotocol/sdk`, `@noble/ed25519`, `@noble/hashes` and `open` — **it does not contain `@mysten-incubation/memwal` and does not perform the write.** The write runs in the relayer's sidecar against its own npm-pinned SDK (`services/server/scripts/package.json:13` → `"@mysten-incubation/memwal": "0.0.3"`).

So **no client release can make `memwal_remember` faster or stop a bulk from hanging.** That needs, in order: a fix in `packages/sdk` (the bulk fetches pass no `AbortSignal`, and `timeoutMs` is only re-checked between poll attempts so one stalled fetch walks past it to undici's 300s floor), an npm publish of the SDK, a bump of that `0.0.3` pin, and a relayer deploy.

Likewise `ip_active_cap` is untouched here — it is a relayer-side identity problem (`TRUSTED_PROXY_HOPS` is unset on prod, so every user shares one bucket keyed on the Railway edge address).

## Tests

`tsc --noEmit` clean.

The package's own suite **could not be run in this environment**: it binds localhost HTTP stubs and the sandbox denies `listen` with `EPERM`, so `sse-handshake-429.test.mjs`, `pending-forward-stalled.test.mjs` and their siblings fail on the socket before reaching an assertion. This needs a run outside the sandbox before merge. Checked by hand in the meantime: the only occurrence of the old string anywhere in `packages/mcp/test/` is a doc comment in `pending-forward-stalled.test.mjs` describing the behaviour #900 replaced — no test asserts it, and that file covers the never-sent branches this PR leaves alone. A test pinning the new write-case wording still needs to be added by whoever can run the suite.


harrymove-ctrl:
Folded into #912 — keeping this to one PR. The commit here is cherry-picked onto that branch unchanged; nothing is lost.

PR #918, opened 2026-09-15
perf(mcp+relayer): stop memwal_remember blocking on the Walrus write, and stop queueing on a busy wallet

harrymove-ctrl:
## Problem

`memwal_remember` blocked until the background write reached `done`. Measured
against the production relayer that is **20–75s for a single short fact**, with
no error — the write succeeds, it is just slow. Worse than the wait itself: a
call that outlived the client's deadline surfaced as a failure, so the agent
retried, and a retry is a second **paid** Walrus blob for a fact that was
already being written.

Phase breakdown of one job (`rememberAsync` + `getRememberStatus`, 2s polling):

| Phase | Time | % of call |
|---|---|---|
| `accepted` | 1.223s | 3.5% |
| `pending → running` | 2.115s | 6.0% |
| **`running → uploaded`** | **29.640s** | **84.4%** |
| `uploaded → done` | 2.121s | 6.0% |

Status polls answer in 96–190ms, so the API is not the bottleneck. Payload size
is not either: a 1.2 KB fact finished in 30.1s while a 97-character one took
39.4s. Seven production runs: 30.1, 36.0, 37.0, 39.4, 42.6, 75.1s.

Almost none of that time tells the agent anything it can act on — the job is
durably accepted ~1.2s in. Everything after is queue time
(`WALRUS_UPLOAD_PER_WALLET_CONCURRENCY` defaults to 1, so a second write waits
out the first; uploads were observed waiting 3–10s just for a slot).

## Change

`memwal_remember` returns once the relayer has **durably accepted** the job.

`MEMWAL_MCP_REMEMBER_WAIT_MS` still buys a wait, but the default is now **0**,
not 10s. A 10s budget against the measured spread lands in the pending branch
on nearly every call — the caller pays the 10s *and* gets no guarantee. Either
wait long enough to mean it (`90000`, the old always-block behaviour) or do not
wait.

- accepted → returns the `job_id`, states the fact is **not** saved yet, and
  points at `memwal_remember_status`
- finished within a non-zero budget → unchanged result, `blob_id` and explorer
  link
- genuinely failed → still an error envelope, unchanged

A timeout is no longer an error, which is what stops the retry-storm and the
duplicate paid blob.

## Why this does not become silent data loss

Returning at accept is safe from a disconnect: the job is a row in
`remember_jobs` driven by the relayer
(`spawn_persisted_remember_preparation`), not work held in the MCP process.
Closing the client does not cancel it.

It is **not** safe from a job that fails *after* acceptance. That is not
hypothetical — one run during this investigation returned `status: failed`,
`error: "Memory encryption backend is unavailable"` 31.6s after acceptance,
from the SEAL encrypt sidecar being unreachable on `localhost:9000`
(`services/server/src/storage/seal.rs:168`).

So `memwal_remember_status` resolves an in-flight job by id, and the accept
message tells the agent in as many words not to report the fact as saved.
`waitMs=0` reads current state in a single GET — `waitForRememberJob` cannot
express that, because it sleeps before its first poll and a 0ms deadline would
answer "still running" without ever asking the relayer.

`remember-wait.ts` holds the shared budget parsing and the 504-vs-500
discrimination, rather than duplicating it across both tools. That distinction
is load-bearing and easy to get wrong: the SDK ships **no** dedicated error
classes for these, in either the pinned `0.0.x` or the current `0.1.x` line, so
`err.status` is the only stable discriminator — a constructor-name check
silently never fires and every slow write surfaces as an error.

## Dependency bump

`@mysten-incubation/memwal` `0.0.3` → `0.0.4`, the first release exposing
`getRememberStatus`, which the `waitMs=0` path needs. The wider `0.1.x`
upgrade stays with the sidecar SDK work rather than riding along here.

## Streamable HTTP transport

Also adds `MEMWAL_MCP_TRANSPORT` (`sse` by default, `http` opts in). The
Streamable HTTP endpoint answers a call on the same request, so there is no
idle watchdog and no replay-on-reconnect — the machinery behind the multi-minute
reconnect stalls.

## Tests

269 run, 236 pass, **33 fail — the same 33 present on `dev`**, all sandbox port
binding (`Cannot read properties of null (reading 'port')`). `tsc --noEmit`
clean.

## Least-loaded upload wallet

Round-robin `next_index()` handed out the next wallet in sequence whether or
not it was mid-upload. With one upload allowed per wallet, landing on a busy
one costs the full upload ahead of it while other wallets sit idle — observed
as a job waiting 6.2s while the global semaphore reported `available:2,
queued:0`. The contention was never global, only per-wallet:

```
[walrus/upload] limiter_acquired {"keyIndex":6,"waitMs":6221,
  "limits":{"global":{"capacity":8,"available":2,"queued":0},
            "perWalletCapacity":1,"wallet":{"capacity":1,"available":0,"queued":1}}}
```

`least_loaded_index()` picks the wallet with the fewest in-flight attempts,
falling back to round-robin ordering among equals. A `WalletAttemptGuard` marks
it busy for the attempt rather than paired increment/decrement calls — the
upload path has many early returns and every one has to release the slot.

11 new unit tests. Lib suite 466 → 477 passing, same 21 pre-existing failures
(no database in the sandbox).

## Retry spacing

Apalis attaches no retry/backoff layer, so a retriable upload error is
re-polled almost immediately. Production shows the cost — one job, four
wallets, one second:

```
15:10:10  attempt 2/5  key 4  → 503 Too Many Requests
15:10:10  attempt 3/5  key 5  → 503 Too Many Requests
15:10:10  attempt 4/5  key 6  → 503 Too Many Requests
15:10:11  attempt 5/5  key 7  → 503 Too Many Requests   → job dead
```

The upstream limit is time-based, so rotating wallets cannot help; only
waiting can, and nothing was waiting. Upload attempts now reuse the existing
`backoff_duration` schedule (2s, 4s, 8s, 16s) the lock-defer path already
uses. Over 24h, 19 of the sampled upload failures were this one rate-limit and
roughly 23% of jobs reached a second attempt.

## Accepted-then-failed writes are reported, not silent

Returning at accept leaves a window where the write still dies with nobody
listening. `memwal_remember_status` answers for one job, but nothing obliges
an agent to ask, and storing a memory is typically the last thing it does in a
turn — an unasked question is the same as a silent loss.

So `/api/recall` now carries `failed_writes`: this owner's writes that reached
`failed` in the last 24h, capped at 5, and the MCP tool renders them as a
warning naming the job, namespace and the relayer's own error text. Recall is
the call an agent always makes, which is why the report rides there.

Owner-scoped from `AuthInfo`; bounded by window and limit because recall is
the hottest authed route, and covered by the existing
`remember_jobs (owner, status, updated_at DESC)` index. A lookup failure
degrades to "nothing to report" rather than failing the recall. The field is
omitted when empty, so older clients and older relayers see today's response.

## Before deploying

`021_failed_write_report_ack.sql` must be applied before the relayer that
queries `failure_reported_at` starts. The Dockerfile copies migrations "for
reference / manual runs" — nothing applies them at boot — so a deploy without
it leaves the failed-write report dead, and dead *quietly*: the lookup error is
caught so it cannot take the recall down with it, which means the only trace is
one `recall: failed-write report unavailable` line per call.

That is exactly what happened on dev. The column was missing, every recall
logged that line, and the feature whose whole purpose is to stop silent loss
was itself silently off. Applying the migration fixed it, confirmed by a
failure reported on the next recall and absent on the one after (each failure
surfaces once).

## Not addressed here

Still relayer-side, and larger than this PR:

1. `sponsor_execute` at 5.4s average — the largest single remaining write-path
   cost.
2. Sui gRPC `GetObject` saturation — 1865 OK plus 78 `unavailable` in an 8
   minute window, roughly 4 calls/second.
3. Enoki sponsor signer/object mismatch causing retryable `set-metadata`
   failures; `ENOKI_FALLBACK_TO_DIRECT_SIGN` defaults to `"false"`, so
   `scripts/sidecar/enoki.ts:287` throws rather than falling back.


services/server/scripts/mcp/tools/remember-wait.ts
```diff
@@ -0,0 +1,318 @@
...
+ *
+ * What makes returning at accept safe from disconnects: the job is a row in
+ * `remember_jobs` driven by the relayer (`spawn_persisted_remember_preparation`
+ * in services/server/src/routes/remember.rs), not work held in this process.
+ * Closing the client does not cancel it.
+ *
+ * What it is NOT safe from: a job that fails after acceptance. Nothing here
+ * can catch that — only a later `memwal_remember_status` call can.
+ */
+const DEFAULT_REMEMBER_WAIT_MS = 0;
```
ducnmm: **[bug]** `DEFAULT_REMEMBER_WAIT_MS` is `0`, so `memwal_remember` / `memwal_remember_bulk` return as soon as the relayer accepts the job and hand back `job_id` with no `blob_id` (`remember.ts:79` early-returns `pending(0)`). MemWal D1 (2026-09-15) already decided not to change MCP `memwal_remember` to return at 202/accept — keep the wait; agents need `blob_id`. Cutting poll is #902; return-on-accept is a later product ticket. This is a breaking tool-contract change for every client of the hosted sidecar (tools run relayer-side, so it lands on deploy, not on npm upgrade). `docs/mcp/reference.md` still documents that `memwal_remember` waits for the Walrus upload to finish.

**Suggestion:** Default the wait to the old blocking ceiling (`90000`) so the tool still returns `blob_id`. Keep `memwal_remember_status` if useful, and leave return-on-accept behind an explicit operator opt-in (or a later product ticket). Poll-cadence cuts belong with #902, not a contract change.

services/server/scripts/mcp/tools/remember-wait.ts
```diff
@@ -0,0 +1,318 @@
...
+    const shared =
+        `Walrus Memory did not accept the ${what} within ${ACCEPT_DEADLINE_MS / 1000}s — the ` +
+        `relayer is unreachable or not responding. The write may or may not have been queued, ` +
+        `so do NOT tell the user it was saved.`;
+
+    return withDeadline(
+        work,
+        ACCEPT_DEADLINE_MS,
+        opts.idempotent
+            ? `${shared} Retrying in this session is safe: the write carries a content-derived ` +
```
ducnmm: **[bug]** `withAcceptDeadline(..., { idempotent: true })` tells the agent "Retrying in this session is safe: the write carries a content-derived idempotency key". That is only true of the unpublished in-repo SDK (`packages/sdk/src/memwal.ts` `derivedIdempotencyKey`). The sidecar is pinned to npm `@mysten-incubation/memwal@0.1.7` (`services/server/scripts/package.json:13`), whose `rememberAsync` still does `crypto.randomUUID()` and deletes `pendingRememberKeys` once accept returns. `withDeadline` does not cancel the in-flight POST, and the comment at line 282 already says the relayer has usually accepted by the time this fires — so the advice the agent sees is the path that mints a second paid blob. The PR body claimed a `0.0.3` → `0.0.4` bump for `getRememberStatus`; `0.0.4` already has that API (and bulk status / wait). This PR instead jumps to `0.1.7` and still does not ship the derived-key change the new copy depends on.

**Suggestion:** Either pin the sidecar at a published SDK that actually derives the key (after dumping that SDK), or have the sidecar pass its own content-derived `idempotencyKey` into `rememberAsync`. Until then, the single-write timeout copy must not claim retry is safe — it is the same hazard the bulk branch already describes. If `getRememberStatus` is the only need, `0.0.4` is sufficient.

services/server/scripts/mcp/tools/recall.ts
```diff
@@ -55,6 +55,51 @@ function dedupeKey(text: string): string {
...
+            : "";
+        return `  job_id=${jobId}${ns}${why}`;
+    });
+
+    const n = raw.length;
+    return (
+        `\n\n⚠ ${n} earlier ${n === 1 ? "write was" : "writes were"} accepted but then FAILED, ` +
+        `so ${n === 1 ? "that fact is" : "those facts are"} NOT stored:\n` +
+        lines.join("\n") +
+        `\nSend ${n === 1 ? "it" : "them"} again with memwal_remember if still wanted.`
```
ducnmm: **[bug]** Every `memwal_recall` now appends up to 5 owner-scoped jobs that failed in the last 24h and instructs the agent to "Send it/them again with memwal_remember if still wanted." The plaintext is not in the report (the row never stored it), the failed row is not cleared on a successful re-save, and recall is the proactive every-turn tool. On the sidecar's 0.1.7 client (random idempotency keys) a well-behaved agent will mint a new paid write for the same fact on every subsequent recall until the 24h window elapses — including after a retry that already succeeded. That is the retry-storm this PR is trying to stop, moved onto the read path.

**Suggestion:** Do not ask the agent to re-save from recall. Point at `memwal_remember_status` with the `job_id` (terminal failure is already an error there) and omit facts that the caller cannot resubmit without the original text. If a recall warning stays, make it one-shot (mark reported / drop after first fetch) so it cannot re-fire every turn.

packages/mcp/src/auth-required.ts
```diff
@@ -88,6 +88,35 @@ function buildToolDefinitions(proactive: boolean) {
             additionalProperties: false,
         },
     },
+    {
+        name: "memwal_remember_status",
```
ducnmm, line 92: **[bug]** This PR adds `memwal_remember_status`, changes `memwal_remember` / `_bulk` descriptions and server instructions, adds `MEMWAL_MCP_TRANSPORT`, and changes sidecar tool behavior. `@mysten-incubation/memwal-mcp` is `0.0.13` on this branch, on `origin/dev`, on `origin/main`, and on npm. There is no MCP changelog dump (`packages/mcp/CHANGELOG.md` / `docs/mcp/changelog.mdx` including the mdx `answer:`), plugin manifests stay `0.0.13`, and `scripts/verify-manual-sdk-release.mjs` still pins `0.0.13`. After WALM-617 the dump is by hand; shipping this under an already-published version means the next MCP release either republishes 0.0.13 or drops a breaking tool contract with no changelog.

**Suggestion:** Bump `0.0.13` → `0.0.14` in `packages/mcp/package.json`, every plugin manifest, and `scripts/verify-manual-sdk-release.mjs`. Write dual changelogs (`packages/mcp/CHANGELOG.md` + `docs/mcp/changelog.mdx` with an updated `answer:`). Do not use a changeset file.

.changeset/remember-poll-cadence-and-replay-dedupe.md
```diff
@@ -0,0 +1,13 @@
+---
+"@mysten-incubation/memwal": patch
```
ducnmm: **[bug]** `packages/sdk` changes poll cadence, check-first waits, per-request deadlines, and content-derived idempotency keys. `@mysten-incubation/memwal` is `0.1.7` on this branch, `origin/dev`, `origin/main`, and npm (`0.1.7` is already published). Version dump is by hand; a changeset is the wrong mechanism, and there is no `0.1.8` bump, no `packages/sdk/CHANGELOG.md` / `docs/sdk/changelog.mdx` (including `answer:`) update, and `scripts/verify-manual-sdk-release.mjs` still pins `0.1.7`. These SDK changes also do not reach the sidecar, which still depends on the published tarball.

**Suggestion:** Delete the two `.changeset/*.md` files. Bump the TS SDK `0.1.7` → `0.1.8`, dump dual changelogs, and pin `scripts/verify-manual-sdk-release.mjs` at `0.1.8`. Only then can the sidecar take the new idempotency / deadline behaviour (see the remember-wait idempotency comment).

services/server/scripts/mcp/tools/remember-wait.ts
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/** Bound a status wait at its own budget plus the overshoot grace. */
+export function withWaitDeadline<T>(work: Promise<T>, budgetMs: number): Promise<T> {
+    return withDeadline(
+        work,
+        budgetMs + WAIT_OVERSHOOT_GRACE_MS,
+        `Walrus Memory stopped responding while waiting for the write to land. The job is ` +
+            `still queued relayer-side — do NOT tell the user it was saved, and do NOT re-send ` +
+            `the fact. Call memwal_remember_status with the job_id to settle it.`,
```
ducnmm, line 367: **[suggestion]** If an operator restores a non-zero `MEMWAL_MCP_REMEMBER_WAIT_MS`, `withWaitDeadline` expiry throws `MemWalRelayerUnresponsive` with no `job_id` and no `status: 504`. `remember.ts` only maps `isStillRunning` (504) back to `pendingMessage`; this error becomes a tool failure, so the agent loses the only handle that can settle a write that was already accepted. Default wait is 0 so this is off the happy path, but it is the path the env var documents.

**Suggestion:** Catch deadline-exceeded on the remember tools and return `pendingMessage(accepted.job_id, …)` (same as a 504 wait timeout). Put the actual `job_id` in the error text if it must still be an error.

services/server/scripts/mcp/tools/remember-bulk.ts
```diff
@@ -36,33 +52,89 @@ export function registerRememberBulkTool(
...
                 const text = facts[i] ?? "";
                 const blob = r.blob_id ? ` blob_id=${r.blob_id}` : "";
                 const err = r.error ? ` error=${r.error}` : "";
-                return `${i + 1}. [${r.status}]${blob}${err}${text ? ` — ${text}` : ""}`;
+                // `timeout` is not a failure — the write is still running and
+                // its job_id is how the caller settles it later.
+                const state = r.status === "timeout" ? `still uploading, job_id=${r.id}` : r.status;
+                return `${i + 1}. [${state}]${blob}${err}${text ? ` — ${text}` : ""}`;
             });
             const summary = `Saved ${result.succeeded}/${result.total} fact(s) to Walrus Memory (failed=${result.failed}).`;
```
ducnmm: **[suggestion]** On a non-zero wait, `waitForRememberJobs` sets `failed = total - succeeded`, which counts `timeout` (still uploading) as failed. The new mixed-result path then prints `Saved X/Y fact(s) to Walrus Memory (failed=N)` and a tail that says those same jobs are still uploading. Agents that read `failed=` will re-send in-flight writes.

**Suggestion:** Count only terminal `failed` / `not_found` in that summary, or stop using `result.failed` and derive counts from `unfinished` / `done` the way `remember-status.ts` already does.

docs/reference/environment-variables.md
```diff
@@ -70,6 +70,7 @@ The stdio MCP package reads these environment variables directly. A CLI flag tak
 | `MEMWAL_CLIENT_LABEL` | `--label <text>` | `MCP Client` / `Walrus Memory MCP` | Friendly delegate-key label shown in the dashboard |
 | `MEMWAL_MCP_DEBUG` | none | `0` | Set to `1` for verbose stderr logging |
 | `MEMWAL_CREDS_DIR` | none | `~/.memwal` | Directory holding `credentials.json`. Overrides both project-local and `~/.memwal` credentials, re-read on every access so a test can redirect it after import. Mainly for tests, which must not write into the real credential directory |
+| `MEMWAL_MCP_TRANSPORT` | none | `sse` | Which relayer transport the stdio bridge dials. `sse` uses the legacy split (`POST /api/mcp/messages` + `GET /api/mcp/sse`). `http` (aliases `streamable`, `streamable-http`) uses the Streamable HTTP endpoint `/api/mcp`, where a call is answered on the same request — no idle watchdog and no replay-on-reconnect. Unrecognised values fall back to `sse` |
```
ducnmm: **[suggestion]** `MEMWAL_MCP_TRANSPORT=http` is documented as having "no idle watchdog and no replay-on-reconnect". `openRelaySession` still feeds the Streamable session into `runBridge`, which replays the entire `inFlight` map on reconnect (`bridge.ts` around the `bridge.replayed` path) and serializes `send()` on `enqueuePost`. For Streamable HTTP, `transport.send()` waits out the RPC on the same request, so a disconnect mid-send looks like `sent=true` with a non-404 status and the replay can duplicate a write — the failure mode this transport was added to avoid. Default stays SSE, so this is opt-in, but the claim in the env docs is false of the code that landed.

**Suggestion:** Skip inFlight replay (and the SSE-only single-POST chain) when the session is Streamable, or drop the "no replay" wording until that path exists. Don't ship the transport rewrite in the same PR as the remember contract change.

packages/mcp/src/streamable.ts
```diff
@@ -0,0 +1,178 @@
+/**
+ * Streamable HTTP transport for the stdio bridge.
```
ducnmm, line 2: **[suggestion]** One PR changes the MCP remember contract, adds a status tool, rewrites the stdio transport, changes SDK polling/timeouts/idempotency, picks least-loaded wallets, adds upload retry sleep, sweeps orphaned preparations, and attaches `failed_writes` to recall. Several of those are independently mergeable and independently risky; the remember-contract piece is also the one D1 already deferred.

**Suggestion:** Split into (1) relayer upload scheduling + backoff + stale-prep sweep, (2) SDK poll/timeout/idempotency with a real 0.1.8 dump, (3) opt-in Streamable HTTP, (4) a product ticket for return-on-accept. Keep this PR to the non-contract perf work if the goal is "stop queueing on a busy wallet".

services/server/src/storage/db.rs
```diff
@@ -2103,9 +2181,59 @@ impl VectorDb {
         Ok(rows)
     }
 
-    /// Mark worker-claimed remember jobs as failed when no worker has updated
-    /// them within the stale TTL. Pending rows are left alone because they may
-    /// simply be waiting behind legitimate queue backlog.
+    /// Mark remember jobs as failed once nothing can still move them.
```
ducnmm: **[suggestion]** Newly added comments repeatedly narrate the investigation, restate the code, and embed architecture history (this `fail_stale_remember_jobs` block, `remember-wait.ts` / `streamable.ts` file headers, `FailedWrite` / `KeyPool` docs, SDK poll/idempotency essays). Comments here should be short and explain a non-obvious constraint, not the design log.

**Suggestion:** Keep one-liners for the load-bearing predicates (`prepare_claimed_at IS NOT NULL` so bulk/analyze rows are not swept; `status: 504` is retryable). Move the rest to the PR body.

ducnmm (changes requested):
## Summary

This PR mixes a product-level MCP contract change (`memwal_remember` / `_bulk` return at accept with no `blob_id`) with SDK poll/timeout/idempotency work, an opt-in Streamable HTTP transport, wallet least-loaded picking, upload retry backoff, a stale-prep sweeper, and `failed_writes` on recall. The relayer-side wallet picker, backoff, error sanitization, and orphan-prep sweep look carefully tested. Do not merge as-is: default return-on-accept breaks the decided `memwal_remember` tool contract (agents need `blob_id`), the sidecar still uses published SDK 0.1.7 (`crypto.randomUUID()` idempotency) while telling agents retries are content-keyed, and both published packages (MCP 0.0.13, TS SDK 0.1.7) change without a manual version dump.

## Issue counts by severity

- bugs: 5
- suggestions: 5
- nits: 0

ducnmm (changes requested):
## Summary

Re-review of `e903f43c` (unique since 99a9bb3b: honour `retry_after` on 429 / `AUTH_UPSTREAM_UNAVAILABLE`, absorb ≤8s, report longer cooldowns as not-saved). That retry path looks correct and fail-closed for bulk (no idempotency key).

The five blocking findings from 99a9bb3b are still on this head. Do not merge.

## Issue counts by severity

- bugs: 5
- suggestions: 0 (prior suggestions on 99a9bb3b still apply; not restated)
- nits: 0

## Issues still open on this head

- **[bug]** `remember-wait.ts` `DEFAULT_REMEMBER_WAIT_MS = 0` — `memwal_remember` returns at accept with no `blob_id`. D1 already decided keep the wait; poll cut is #902; return-on-accept is a later product ticket.
- **[bug]** `withAcceptDeadline(..., { idempotent: true })` claims a content-derived key. Sidecar is still npm `@mysten-incubation/memwal@0.1.7` (`randomUUID`). Retry copy is the duplicate-blob path.
- **[bug]** `memwal_recall` `failed_writes` tells the agent to `memwal_remember` again every turn for 24h. Retry-storm on the read path; plaintext is not even in the report.
- **[bug]** MCP contract/tool/transport changed; dump still published `0.0.13`. Bump `0.0.14` by hand (no changeset).
- **[bug]** TS SDK poll/timeout/idempotency changed under published `0.1.7` plus `.changeset/*.md`. Delete changesets; bump `0.1.8` dual changelogs + verify-script pin.

ducnmm (commented):
## Summary

Re-review of `da91fed7`. The five blocking findings from `e903f43c` are addressed: default wait is 90s again (accept-and-continue only via `MEMWAL_MCP_REMEMBER_WAIT_MS=0`); sidecar passes a content-derived `idempotencyKey` so the 0.1.7 client does not mint a second UUID; recall `failed_writes` is one-shot and no longer asks the agent to re-save text it does not have; MCP `0.0.14` and SDK `0.1.8` dumped by hand (changesets deleted, verify-script + plugin pins).

Leftover is the mixed-result bulk summary still printing `failed=` from `waitForRememberJobs` (timeouts counted as failed) next to a tail that says those jobs are still uploading.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 0

## Issues outside the unique-commit diff

- **[suggestion]** `services/server/scripts/mcp/tools/remember-bulk.ts:129` — `Saved ${succeeded}/${total} (failed=${result.failed})` still uses `waitForRememberJobs`'s `failed = total - succeeded`, which counts `timeout` as failed. The next paragraph then says those writes are still uploading. Count only terminal `failed` / `not_found`, or derive from `unfinished` the way `remember-status.ts` does.

services/server/scripts/mcp/server.ts
```diff
@@ -50,6 +50,15 @@ const INSTRUCTIONS = [
     "summary. Skip one-off tasks, the current file or bug, and small talk. Use",
     "memwal_remember_bulk when several distinct facts arrived at once.",
     "",
+    "A Walrus write is queued, not instant: memwal_remember and memwal_remember_bulk normally",
```
ducnmm: **[bug]** Product D1 is blob_id on the happy path and no dual-tool; the code default wait is 90s. These initialize instructions still tell the agent that `memwal_remember` / `memwal_remember_bulk` *normally* return job_ids with the facts "ACCEPTED but NOT YET SAVED", that this is "the healthy path, not an error", to tell the user the facts are being saved rather than saved, and to resolve them with `memwal_remember_status`. The stdio copy in `packages/mcp/src/instructions.ts` is the same, and `memwal_remember_bulk`'s tool description still says it "normally returns job_ids with the facts NOT yet saved". Agents follow this even when the tool already returned a blob_id — dual-tool plus extra status polls against the rate-limit budget this PR is trying to protect.

**Suggestion:** Describe the 90s wait as the default: a result with `blob_id` means stored; job_id + `memwal_remember_status` is only the overrun / `MEMWAL_MCP_REMEMBER_WAIT_MS=0` path. Keep the two instruction strings in sync. Soften the bulk tool description from "normally" to "may".

services/server/scripts/mcp/tools/analyze.ts
```diff
@@ -33,29 +54,101 @@ export function registerAnalyzeTool(
...
+                "Extract memorable facts from a longer passage of text (preferences, habits, biographical info, constraints) and save each as a separate Walrus Memory memory. Use this when you want MemWal's LLM to split the facts out of a transcript or notes for you; if you already know the exact facts, use memwal_remember or memwal_remember_bulk instead. The extracted facts come back immediately; if the result says the writes are still in flight it carries job_ids — confirm them with memwal_remember_status rather than telling the user they are saved.",
             inputSchema: ANALYZE_INPUT,
         },
         wrapTool<{ text: string; namespace?: string }>(session, "memwal_analyze", async ({ text, namespace }) => {
-            const result = await session.memwal.analyzeAndWait(text, namespace, {
-                timeoutMs: 180_000,
+            // `analyze` (not `analyzeAndWait`) returns once extraction is done
+            // and every fact has a queued job, which is the point this tool can
+            // usefully answer at.
+            const accepted = await withAcceptDeadline(
```
ducnmm, line 64: **[bug]** `memwal_analyze` wraps `session.memwal.analyze()` — the call that runs the extractor LLM inline — in `withAcceptDeadline`, which is `ACCEPT_DEADLINE_MS` (15s), sized for a remember accept (~1.1s). This PR's own SDK 0.1.8 notes that `/api/analyze` "legitimately outruns" the 30s default and gives it 60s. Sidecar still runs published 0.1.7, so this 15s race is the only bound. A healthy extraction that takes >15s returns `MemWalRelayerUnresponsive` with `idempotent: false` ("Do NOT retry blindly") while `withDeadline` does not cancel the request, so the jobs can still land and the agent is told they did not.

**Suggestion:** Give analyze its own extraction budget (the SDK's 60s, or at least something well above 15s). Keep the 15s cap on remember accept.

services/server/scripts/mcp/tools/remember.ts
```diff
@@ -38,23 +54,64 @@ export function registerRememberTool(
...
+                        {
+                            type: "text" as const,
+                            text: `Saved to Walrus Memory. blob_id=${result.blob_id} namespace=${result.namespace}\nExplorer: ${walruscanBlobUrl(result.blob_id)}`,
+                        },
+                    ],
+                };
+            } catch (err) {
+                // Still running at the deadline is the expected path, not a
+                // failure — the job is durably accepted and keeps going.
+                if (isStillRunning(err)) return pending(Date.now() - startedAt);
```
ducnmm, line 112: **[bug]** After accept, a stalled status poll is the hang this wait bound exists to catch. `waitForRememberJob` on pinned 0.1.7 has no per-request timeout, so `withWaitDeadline` fires at budget+10s with `MemWalRelayerUnresponsive` (no `status: 504`). The catch only treats `isStillRunning` (status 504) as pending and otherwise throws. wrapTool then returns `isError` with "Call memwal_remember_status with the job_id" and never includes `accepted.job_id`. Same pattern on `remember-bulk.ts` / `analyze.ts`: `withWaitDeadline` is uncaught, so a hung batch wait drops every job_id. The pending branch already has the ids; this path throws them away.

**Suggestion:** On `MemWalRelayerUnresponsive` after a successful accept, return `pending(job_id)` / `pendingBulkMessage(entries)` instead of throwing. Include the ids in any error copy that tells the agent to call `memwal_remember_status`.

services/server/scripts/mcp/tools/remember-bulk.ts
```diff
@@ -36,33 +53,95 @@ export function registerRememberBulkTool(
...
```
ducnmm: **[suggestion]** Leftover from the last pass. `waitForRememberJobs` sets `failed = total - succeeded`, so timeouts count as failed. The summary still prints `failed=${result.failed}` while the next paragraph says those writes are still uploading and must not be re-sent. Timeout rows also still interpolate `error=${r.error}` (`"polling timed out after Nms"`); `remember-status.ts` already hides that for non-terminal rows because it read as a failed write. `analyze.ts` has the same `failed=${result.failed}` summary. The tail does say "Do not re-send", so this is confusing rather than an instruction to retry paid writes.

**Suggestion:** Count failed as `status === "failed"` only (or print `timed out=` separately). Skip `error=` on timeout rows, matching the status tool.

docs/reference/environment-variables.md
```diff
@@ -154,6 +155,8 @@ These are not all enforced at boot, but most real deployments need them.
 | `MCP_MAX_TOTAL_SESSIONS` | `1000` | Maximum active MCP sessions across SSE and Streamable HTTP transports |
 | `MCP_MAX_SESSIONS_PER_IP` | `16` | Maximum active MCP sessions from one source IP |
 | `MCP_MAX_NEW_SESSIONS_PER_IP_PER_MIN` | `30` | Maximum new MCP sessions opened by one source IP per minute |
+| `MEMWAL_MCP_REMEMBER_WAIT_MS` | `0` | How long `memwal_remember` / `memwal_remember_bulk` wait for a write to land before returning the job_id instead. `0` returns once the relayer has durably accepted the job (~1s). A value between the two is the worst of both — against a 30-75s completion spread it pays the wait and still returns pending — so either leave it at `0` or set it high enough to mean it. Clamped to `90000`; invalid values fall back to the default |
```
ducnmm, line 158: **[suggestion]** `MEMWAL_MCP_REMEMBER_WAIT_MS` is documented in the relayer table with default `0` and "either leave it at `0` or set it high enough to mean it". Code default is 90_000; `0` is the opt-in D1 deferred. An operator following the table and setting the var to `0` switches the sidecar onto accept-and-continue. `MEMWAL_MCP_ACCEPT_DEADLINE_MS` on the next line is fine; this row's default and "leave it at 0" are the wait=0 leftover.

**Suggestion:** Default `90000`. Say `0` returns at accept and is opt-in. Invalid / unset → 90s, clamped to 90s.

packages/mcp/src/bridge.ts
```diff
@@ -269,6 +270,57 @@ const DEFAULT_CALL_TIMEOUT_MS = SLOWEST_SERVER_TOOL_MS + 60_000;
...
+ *
+ * `DEFAULT_CALL_TIMEOUT_MS` is sized for `memwal_analyze`, the slowest tool
+ * there is. Applying that one number to every call means a request whose reply
+ * is lost — the relayer answered, the stream dropped before it arrived — keeps
+ * the agent blocked for 240s even when the tool could not still be working.
+ * Users read that as a hang and reload the client.
+ *
+ * Each entry is the ceiling the matching tool enforces on itself in
+ * `services/server/scripts/mcp/tools/`: `MAX_REMEMBER_WAIT_MS` for
+ * `memwal_remember` (its default wait is 0 — it returns at accept — but an
```
ducnmm, line 283: **[suggestion]** Comment on `TOOL_DEADLINE_MS` still says `memwal_remember`'s "default wait is 0 — it returns at accept" while the constant it documents is `90_000`. Same leftover in `remember-wait.ts` (`waitedMs === 0` "is the default path, the one an agent reads on nearly every call") and `analyze-fast-return.test.ts` ("The shipped default is 0").

**Suggestion:** Say the default wait is 90s and `0` is opt-in, so the next person does not "fix" the constant back to match the comment.

ducnmm (changes requested):
## Summary

The latency work is in the right shape: default `memwal_remember` wait is still 90s (`parseWaitBudget(undefined) === MAX_REMEMBER_WAIT_MS`), accept-and-continue is `MEMWAL_MCP_REMEMBER_WAIT_MS=0`, the sidecar still mints a content-derived idempotency key (SDK pin stays published 0.1.7), recall failed-writes are one-shot and do not tell the agent to re-save unknown ciphertext, and the dump is MCP 0.0.14 / SDK 0.1.8 via `verify-manual-sdk-release.mjs` with plugin pins and no changesets. Poll cadence is correctly left to #902. The remaining D1 hole is agent-facing copy that still treats return-on-accept + `memwal_remember_status` as the healthy path; there is also a 15s cap around `analyze()` extraction and a hang path that drops the `job_id` the pending result exists to return.

## Issue counts by severity

- bugs: 3
- suggestions: 3
- nits: 0

services/server/scripts/mcp/tools/remember-bulk.ts
```diff
@@ -36,33 +53,95 @@ export function registerRememberBulkTool(
...
+            if (REMEMBER_WAIT_MS === 0) return pending(0);
+
+            const startedAt = Date.now();
+            const namespaces = items.map(
+                (item) => item.namespace ?? session.namespace ?? "default"
+            );
+            // Unlike `waitForRememberJob`, this never throws on expiry — it
+            // reports the stragglers as `timeout` per item, so a batch can come
+            // back part landed and part still in flight.
+            const result = await withWaitDeadline(
```
ducnmm, line 101: **[bug]** The previous hung-wait bug is fixed only on `memwal_remember`. After accept, that tool catches `MemWalRelayerUnresponsive` via `isStillRunning` and returns `pendingMessage` with the `job_id`. `memwal_remember_bulk` (and `memwal_analyze`) still do a bare `await withWaitDeadline(session.memwal.waitForRememberJobs(...))`. Sidecar is pinned to SDK 0.1.7, so a stalled poll does not self-bound; `withWaitDeadline` then rejects, `wrapTool` turns it into `isError: true` with the generic "stopped responding… Call memwal_remember_status with the job_id" text, and the accepted ids never appear. Bulk has no idempotency key, so the agent is invited to re-send a batch that may already be queued. `remember-wait-hang.test.ts` covers only the single-write tool.

**Suggestion:** Use the same try/catch as `remember.ts`: if `isStillRunning(err)`, return `pending(...)` with the already-built `entries` / `job_ids`. Extend the wait-hang test to bulk (and analyze) so a quiet relayer after accept still returns the ids.

services/server/scripts/mcp/tools/analyze.ts
```diff
@@ -33,29 +54,106 @@ export function registerAnalyzeTool(
         {
             ...TOOL_METADATA.memwal_analyze,
             description:
-                "Extract memorable facts from a longer passage of text (preferences, habits, biographical info, constraints) and save each as a separate Walrus Memory memory. Use this when you want MemWal's LLM to split the facts out of a transcript or notes for you; if you already know the exact facts, use memwal_remember or memwal_remember_bulk instead.",
+                "Extract memorable facts from a longer passage of text (preferences, habits, biographical info, constraints) and save each as a separate Walrus Memory memory. Use this when you want MemWal's LLM to split the facts out of a transcript or notes for you; if you already know the exact facts, use memwal_remember or memwal_remember_bulk instead. The extracted facts come back immediately; if the result says the writes are still in flight it carries job_ids — confirm them with memwal_remember_status rather than telling the user they are saved.",
```
ducnmm, line 57: **[bug]** Remember, bulk, sidecar initialize, and stdio instructions were rewritten so the happy path is "this call waits; a `blob_id` means stored; status is the exception." `memwal_analyze`'s advertised description still says the extracted facts come back immediately and to confirm in-flight writes with `memwal_remember_status`. Default `REMEMBER_WAIT_MS` is 90s, so this tool also blocks on the Walrus writes. The description teaches return-on-accept plus a status follow-up as the normal contract — the dual-tool path D1 ruled out.

**Suggestion:** Mirror the remember/bulk wording: extraction still runs inline, then the call waits; success carries `blob_id`s; only a budget miss returns `job_ids` to settle with `memwal_remember_status`.

docs/reference/environment-variables.md
ducnmm, line 158: **[suggestion]** Leftover from the last review. The table still lists `MEMWAL_MCP_REMEMBER_WAIT_MS` default as `0` and tells operators to "leave it at `0` or set it high enough to mean it." `parseWaitBudget(undefined)` returns `MAX_REMEMBER_WAIT_MS` (90000). An operator copying the documented default into env gets accept-and-continue, which D1 made opt-in.

**Suggestion:** Default column `90000`. Describe `0` as the explicit accept-and-continue opt-in, not the shipped default.

packages/mcp/src/bridge.ts
ducnmm, line 283: **[suggestion]** Leftover comments still describe the abandoned default-0 product. `TOOL_DEADLINE_MS` says `memwal_remember` "default wait is 0 — it returns at accept." `pendingMessage` in `remember-wait.ts` still calls the zero-budget wording "the default path, the one an agent reads on nearly every call." `packages/mcp/test/tool-definitions.test.mjs` still says "Both write tools return at accept now."

**Suggestion:** Default wait is 90s; zero is opt-in. Update those comments to match `DEFAULT_REMEMBER_WAIT_MS`.

services/server/scripts/mcp/tools/remember-bulk.ts
```diff
@@ -36,33 +53,95 @@ export function registerRememberBulkTool(
...
                 const text = facts[i] ?? "";
                 const blob = r.blob_id ? ` blob_id=${r.blob_id}` : "";
                 const err = r.error ? ` error=${r.error}` : "";
-                return `${i + 1}. [${r.status}]${blob}${err}${text ? ` — ${text}` : ""}`;
+                // `timeout` is not a failure — the write is still running and
+                // its job_id is how the caller settles it later.
+                const state = r.status === "timeout" ? `still uploading, job_id=${r.id}` : r.status;
+                return `${i + 1}. [${state}]${blob}${err}${text ? ` — ${text}` : ""}`;
             });
             const summary = `Saved ${result.succeeded}/${result.total} fact(s) to Walrus Memory (failed=${result.failed}).`;
```
ducnmm: **[suggestion]** Leftover: a mixed batch still reports timeouts as failures. SDK `waitForRememberJobs` sets `failed: results.length - succeeded`, so a 2/3 land + 1 still-uploading batch becomes `Saved 2/3 fact(s) to Walrus Memory (failed=1)`. Per-line rendering also appends `error=${r.error}` for every row, so a timeout line is `[still uploading, job_id=…] error=polling timed out after 90000ms` — the same `error=`-means-failed misread `remember-status.ts` already special-cased to terminal rows only.

**Suggestion:** Count `failed` from `status === "failed"` (or `result.total - result.succeeded - unfinished.length`), and only print `error=` on terminal rows, matching `settleBatch`.

ducnmm (changes requested):
## Summary

D1 is in place for the surfaces called out last time: default wait is 90s with a real `blob_id` on the happy path, `MEMWAL_MCP_REMEMBER_WAIT_MS=0` is opt-in, initialize/stdio/remember/bulk copy no longer teach dual-tool as the healthy path, and `memwal_analyze` extraction is bounded at 60s rather than the 15s accept ceiling. The hung-wait `job_id` fix is real for `memwal_remember` (the wait-hang test now forces a non-zero budget and asserts the id), but the same `withWaitDeadline` path on bulk and analyze still throws `MemWalRelayerUnresponsive` without returning the accepted ids. Relayer join-shortest-queue plus upload backoff look correct. Leftover docs/comments still describe the abandoned default-0 product.

## Issue counts by severity

- bugs: 2
- suggestions: 3
- nits: 0

ducnmm (approved):

PR #921, opened 2026-09-16
fix(relayer,mcp): return remember at accept and drop the failed-write report

harrymove-ctrl:
## Summary

`memwal_remember` / `_bulk` return at **accept** (`job_id`, ~1s) instead of blocking 30–90s for Walrus. Agents settle with the job-status tool; `blob_id` still means stored. The #918 recall-attached failed-write report (and ADD migration 021) is **reverted**; leftover `failure_reported_at` is dropped `IF EXISTS`. `/health` grows `writes=degraded` when recent durable writes fail and none land, without flipping `write_ready` (CI still gates on accept). Cold start is a floor (#928): do not advertise or name tools the oldest relayer may not serve.

HEAD `7a7eb3d6`. CI green. Merge still **blocked** on ducnmm `CHANGES_REQUESTED` until he re-reviews.

This is the product answer to “saves block the agent for minutes while `/health` still says ok.” It is **not** the `ip_active_cap` / Windows SSE-slot leak.

## Sequence diagrams

GitHub Mermaid treats `;` `()` and the word `and` as syntax. Labels below avoid all three.

### Remember at accept default wait is 0

```mermaid
sequenceDiagram
    autonumber
    participant Agent
    participant MCP as MCP sidecar
    participant Relayer
    participant Jobs as remember_jobs
    participant Walrus
    Agent->>MCP: memwal_remember text
    MCP->>Relayer: POST /api/remember
    Relayer->>Jobs: insert pending then return job_id
    Relayer-->>MCP: 202 accepted with job_id
    MCP-->>Agent: job_id in about 1s - not stored yet
    Note over Relayer,Walrus: background encrypt register upload
    Relayer->>Walrus: durable upload 30 to 60s
    alt upload lands
        Relayer->>Jobs: status done plus blob_id
    else attempts exhausted
        Relayer->>Jobs: status failed immediately
    end
    Agent->>MCP: job-status tool with job_id
    MCP->>Relayer: GET remember job
    Relayer->>Jobs: read status
    Relayer-->>MCP: done or failed or running
    MCP-->>Agent: blob_id saved or error or still uploading
```

### Cold-start tool floor issue 928

```mermaid
sequenceDiagram
    autonumber
    participant Agent
    participant Bridge
    participant Relayer
    Agent->>Bridge: initialize
    Bridge-->>Agent: instructions do not name memwal_remember_status
    Agent->>Bridge: tools/list while SSE is still null
    Bridge-->>Agent: BASELINE_RELAYER_TOOLS only
    alt call a tool the floor does not advertise
        Agent->>Bridge: tools/call memwal_remember_status
        Bridge-->>Agent: local error nothing forwarded
    else call a baseline tool
        Agent->>Bridge: tools/call memwal_remember
        Bridge->>Relayer: forward when session is up
    end
    Relayer-->>Bridge: session live
    Bridge-->>Agent: notifications/tools/list_changed
    Agent->>Bridge: tools/list forwarded
    Bridge->>Relayer: tools/list
    Relayer-->>Bridge: this relayer tools
    Bridge-->>Agent: splice login plus logout
    Note over Bridge: allow-set filled then reconnect clears it
```

### Health writes degraded why 022 exists

```mermaid
sequenceDiagram
    autonumber
    participant LB as Load balancer
    participant Health as health endpoint
    participant Probe as recent_write_outcomes
    participant DB as remember_jobs
    LB->>Health: every tick
    Health-->>LB: HTTP 200 status ok write_ready true
    Note over Health: write_ready stays true so CI wait-for-relayer does not hang
    Health->>Probe: cache miss 30s TTL
    Probe->>DB: count failed vs done in last 15m
    Note over DB: 022 partial index on updated_at with 1s timeout fail open
    alt no successful writes in the window
        Health-->>LB: writes degraded
        Note over LB: status page must not say All Systems Operational
    else some writes landed or probe timed out
        Health-->>LB: writes ok
    end
```

### 022 first-boot race fixed in 7a7eb3d6

```mermaid
sequenceDiagram
    autonumber
    participant A as Replica A
    participant B as Replica B
    participant PG as Postgres
    A->>PG: advisory lock then recover
    A->>PG: CREATE INDEX CONCURRENTLY 022
    B->>PG: waits on the same advisory lock
    A-->>PG: index valid then unlock
    B->>PG: recover sees valid index then skip drop
    B->>PG: CREATE INDEX IF NOT EXISTS is a no-op
    Note over A,B: live build is not treated as a crashed leftover
```

### Open on this HEAD

None from ducnmm's last review. Both 15:14 findings are fixed in `7a7eb3d6`.

### Fixed in `7a7eb3d6` (ducnmm 2026-09-18 15:14)

| Sev | Where | Fix |
|---|---|---|
| **bug** | `recover_invalid_concurrent_indexes` vs live CONCURRENTLY | Skip drop when `pg_stat_progress_create_index` shows a live build. `DROP INDEX CONCURRENTLY IF EXISTS`. Recover + 016/018/022 on one session behind `pg_advisory_lock`, `statement_timeout=0`. |
| **suggestion** | `admin.rs` probe comments | Cache = LB tick rate; 022 = steady state; 1s = backstop. |

### Fixed in this PR (ducnmm threads 1–7 — code done, GitHub threads not all marked resolved)

1. Single-job status probe: `failed` / `not_found` → error, not “still uploading” (`6c4bfa22`).
2. Initialize / cold-start do not name `memwal_remember_status`; gate fails closed until first upstream `tools/list` (`13b6de51`).
3. `reconnect()` clears `upstreamToolNames` + `tools/list_changed` (`ea91d059`).
4. `/health` probe 1s + fail-open + `SET LOCAL` (`9ece083e`); partial index 022 (`e34446a1`); recover INVALID list 016/018/022 (`2a23e871`).
5. Wait-default comments match default=0 (`66f8eca8`, `f25b6862`).
6. Troubleshooting docs: empty recall after accept is expected (`f25b6862`).
7. Status page: `writes=degraded` is **Writes Degraded**, not All Systems Operational (`bbf15165`).

### Intentionally out of this PR

| Item | Why |
|---|---|
| `ip_active_cap` (Theodore: 8/11 sessions lose all memory) | Relayer SSE slot leak / Windows stdio kill. Not a line in #921. Next ticket. |
| Unpin launchers `0.0.14-dev.0` → `@0.0.14` | npm has no `0.0.14` yet. Written into both changelogs; do at publish. |
| Status-page allowlist (`writes === 'ok'`) | Same class as #7, skipped to keep this PR from growing. |
| 013 unique-index recovery | Invalid unique index fails upsert **loudly**. Lives before the recover pass. |

## Confidence score

| | Score | Why |
|---|---|---|
| Remember-at-accept + write-path honesty (the user-visible change) | **90 / 100** | Default wait=0 is wired, tested, and documented. `writes=degraded` + status page + exhausted→failed match the “health said ok, write stored nothing” class. |
| First deploy of 022 / multi-replica boot | **85 / 100** | Recover skips live builds, DROP IF EXISTS, advisory lock + timeout 0. Residual: progress view may be invisible to non-superusers; the lock is the real fence. |
| **Merge as-is** | **88 / 100** | CI on this push still running. Prior review items fixed. Remaining block is ducnmm re-review, not a known open bug. |

**Blocked on:** ducnmm re-review (GitHub still `CHANGES_REQUESTED`). Not blocked on a known open bug.


harrymove-ctrl:
### CI is green — and fixing it took three layers, not one

The migration wiring alone was not enough. Recording the trail because each layer hid the next.

**1. `VectorDb::new()` chain ended at 020.** Wired 021 in. Production is now correct — but the four tests stayed red, identically.

**2. `idem_test_pool` never calls `VectorDb::new()`.** It hand-builds a minimal schema from 005, 012 and 013 by `include_str!`, so nothing added to the production chain reaches it. Added 021 there too. Both edits are needed: the chain is what a deploy runs, the helper is what these tests run.

That fixed the four: `783 passed; 4 failed` → `786 passed; 2 failed`.

**3. The two remaining failures were a different, pre-existing fault** — and one this branch exposed rather than caused:

```
Failed to run migration 011: duplicate key value violates unique constraint
  "pg_type_typname_nsp_index"
DETAIL:  Key (typname, typnamespace)=(mcp_oauth_clients, 2200) already exists.
```

`CREATE TABLE IF NOT EXISTS` is not atomic in Postgres. Two tests calling `VectorDb::new()` at the same moment both pass the existence check and one loses on the `pg_type` insert, so they fail on an unreachable database rather than on anything they assert. `jobs::tests` already serialises its pool behind a `DB_SETUP_LOCK` for exactly this; neither `test_db()` in `storage/db.rs` did. Adding one test to the first module was enough to make two builds overlap. Same guard, declared once so both modules share it, held only across `VectorDb::new()`.

`20 checks | 0 failing`.

### Also added

The first test for either half of the failed-write report. `recent_failed_remember_jobs` and `claim_failed_write_reports` had no coverage at all, which is what let the read/claim split be got wrong silently. It pins the two properties the split depends on: the read is non-consuming, so a recall that dies on an embedding 429 or is abandoned at the SDK's hard abort leaves the failure still reportable; and the claim is exclusive, so two concurrent recalls cannot both tell the agent to re-send the same fact.

### One note that is not a fix

`bc7bf462` records a constraint in the `DEFAULT_REMEMBER_WAIT_MS` comment rather than acting on it. The MCP TypeScript SDK defaults a `tools/call` to `DEFAULT_REQUEST_TIMEOUT_MSEC = 60_000` (`shared/protocol.js:8`, applied at `:712`). Our default wait is `90_000`, so on a host that sets no timeout of its own the pending branch — whose whole purpose is to hand back a `job_id` — arrives 30s after the client gave up. `docs/troubleshooting/overview.md` already documents the symptom, including that retrying duplicates the memory.

Four fixes were designed for it and **all four were refuted**, so the comment records what is verified instead of guessing:

- The legs are sequential, so the ceiling is `ACCEPT_DEADLINE_MS` (15_000) + budget + `WAIT_OVERSHOOT_GRACE_MS` (10_000). Fitting under 60_000 needs a budget of 35_000 or less — against a measured 26.7–47.9s time-to-saved, which is the D1 trade, not a free win.
- `memwal_analyze` cannot be fixed by this constant at any value including 0: its 60_000ms extraction deadline fully precedes the wait.
- Hosts differ. Claude Code sets its own far larger ceiling, so this does not bite there.

Lowering the default is a D1-level decision and belongs with whoever owns it.

harrymove-ctrl:
## Evidence — Theodore ask #1 (Write latency / bulk never returns)

Measured 2026-09-17. Metrics: **accept** (tool returns) vs **time-to-saved** vs **save_rate**.

### Prod — returns **and** saves

Relayer: `https://relayer.memory.walrus.xyz`

| Step | Time | Result |
|---|---:|---|
| `memwal_health` | 1.3s | `write_ready=true` |
| `memwal_remember` | **30.3s** | **returned** + saved |
| `memwal_recall` | 3.1s | **HIT** score=1.000 |

```
REMEMBER_RETURN 30263ms ok
Saved to Walrus Memory.
blob_id=mrRxNzN2ztNlpOYGSWVis9V5nfXyH3K787MrWrZn6VQ
Explorer: https://walruscan.com/mainnet/blob/mrRxNzN2ztNlpOYGSWVis9V5nfXyH3K787MrWrZn6VQ

save_rate=1/1
proves_never_returns_fixed=true
proves_saved=true
```

Second prod run (same day):

```
remember 29344ms ok
blob_id=cSrfZQh0oLeJ4jy_hyeKXUcCdpi77KerIN4-lMbl2EU
Explorer: https://walruscan.com/mainnet/blob/cSrfZQh0oLeJ4jy_hyeKXUcCdpi77KerIN4-lMbl2EU
recall_hit=true
```

**vs Theodore:** his bulk got **no reply for 240s** and **0/5 saved**. Prod now **returns** and **saves** (still ~30s, not “a few seconds”).

---

### Dev — returns fast; durable upload still 503

Relayer: `https://relayer.dev.memwal.ai` (this PR’s runtime)

| Step | Time | Result |
|---|---:|---|
| `memwal_health` | 1.4s | `write_ready=true`, **`writes=degraded`** |
| `memwal_remember` | **1.2s** | **returned** `job_id=17cca09f-9503-4a4f-b571-8e2cd8aabc8e` (NOT yet saved) |
| `memwal_remember_status` | later | **`NOT STORED (failed)`** Walrus 503 timeout |
| `memwal_recall` | — | MISS (correct) |

```
REMEMBER_RETURN 1164ms ok
ACCEPTED, NOT YET SAVED
job_id=17cca09f-9503-4a4f-b571-8e2cd8aabc8e

memwal_remember_status →
0/1 saved, 1 NOT stored.
[NOT STORED (failed)]
error=durable Walrus upload failed (503 Service Unavailable)
code=SHARED_SERVICE_UNAVAILABLE
traceId=41c71595-6a79-404f-986e-9190046cf534

save_rate=0/1
proves_never_returns_fixed=true
proves_saved=false
```

So on dev: **“never returns” is fixed** (1.2s accept). **Durable save is not** (Walrus testnet 503). Health no longer pretends writes are fine (`writes=degraded`).

---

### Ask #1 summary

| Claim | Evidence |
|---|---|
| Tool returns (not 240s hang) | Prod ~30s return; Dev **1.2s** return |
| Fact actually saved | Prod `blob_id` + recall HIT **1/1** |
| Dev Walrus reliability | Still **0/1**, terminal failed via `remember_status` |

Also in this PR: plugin pin fixed to published `@0.0.14-dev.0` (npm has no `0.0.14`). Closed duplicate #929 so this stays the single PR.

harrymove-ctrl:
## Root cause of Dev 503 (not an MCP/client bug)

Ask #1 split into two different problems:

### A) “Never returns” (client/MCP UX) — addressed by #918 / this PR
Dev accept now returns in **~0.4–1.2s** with `job_id`. That is separate from durable save.

### B) Durable save 503 on `relayer.dev` — Walrus **testnet write path**

Live traces (2026-09-17):

1. `encode` / `register_prepared` / `registered` **succeed** (blob on chain).
2. Sliver write then fails:
   - Via public upload-relay: `phase=registered` timeout after **~122s** (`WALRUS_RELAY_TIMEOUT_MS` default).
   - Via `WALRUS_DIRECT_UPLOAD=true`: `Too many failures while writing blob … to nodes` in **~10s**.

Example job `ad7e90f3-c6fd-4e04-9539-ff848e2e8a50`:
- registered blob `7u2hzlkfcXSrreZyZJQCtpWBieDFu89PkBzSI52_HFI`
- then 503 `SHARED_SERVICE_UNAVAILABLE` / too many node write failures
- `save_rate=0/1`, recall MISS

Infra findings on Railway **dev**:
- Service `walrus-upload-relay` had **NO DEPLOYMENT** since May (prod/staging still run `cmdoss/walrus-upload-relay`).
- Relayer was pointed at public `https://upload-relay.testnet.walrus.space` instead of internal relay like prod (`http://walrus-upload-relay.railway.internal:8080`).
- Redeploying the image crashes on start: hardcoded `https://fullnode.testnet.sui.io:443` JSON-RPC is **deprecated** (`Method not found`).

**Prod (mainnet Walrus) saves OK** (~30s + blob_id). Dev failure is testnet storage/relay availability + missing healthy private relay, not the MCP accept-path.

Ops follow-ups (outside this PR’s code):
1. Bring up `walrus-upload-relay` on dev with a working Sui RPC/gRPC (not public JSON-RPC fullnode).
2. Point `WALRUS_UPLOAD_RELAY_URL` at `http://walrus-upload-relay.railway.internal:8080` once healthy.
3. Until then, durable remember on dev will keep failing after accept.

harrymove-ctrl (commented):
## Delivery review (standalone, **full** depth — operator asked “do all”)

**Occasion:** boundary + pre-ship. Trigger: #918 follow-ups join MCP pin, persisted schema (`021`), `/health` wire, remember job terminal status, and a deployable relayer.  
**Range:** `origin/dev`…`b4601350` (22 commits, +2573/−346, 36 files). Base `dev`.  
**CI:** all required checks green, including `Server / Clippy + Unit tests`.  
**Verdict: approve merge into `dev`.** Not a prod-promotion review by itself.

Teo’s four asks are the contract this branch is being judged against (plus the PR’s own “#918 follow-ups / 021 / claim-at-delivery”).

---

### What I read / ran / checked

- PR body, file list, `gh pr checks 921`.
- `services/server/src/jobs.rs` (`update_remember_job_after_wallet_error`, `exhausted_by`, `classify_sidecar_error` catch-all Transient, `MAX_ATTEMPTS=5`, test `exhausted_transient_upload_marks_the_row_failed_immediately`).
- Prod `0ffbacbb` same function: 5/5 transient stays `running` (no `attempt_info`).
- `admin.rs` `durable_writes_degraded` + test `durable_writes_degraded_only_when_nothing_lands`; `recent_write_outcomes` in `db.rs`.
- `health.ts`, `remember-status.ts`, `remember-wait.ts` (`isStillRunning` / `isRateLimited`).
- `021_failed_write_report_ack.sql`, `VectorDb` `MIGRATIONS_AFTER_INDEX_RECOVERY` includes 021.
- `recall.rs` `claim_for_delivery` (stamp at response, not in the prefetch spawn).
- `concurrency.ts` one-shot release + `sidecar-slot-lifecycle.test.ts`.
- Plugin `.mcp.json` / `.codex-mcp.json` / `.cursor-mcp.json` pin `@0.0.14-dev.0`; `plugin.json` still `0.0.14`; `install_codex_hooks.mjs` reads `.mcp.json`.
- CHANGELOG 0.0.14: default remember **still waits for `done`**.
- Live earlier this session: exhausted job `163e99b9-…` reached `failed` ~510s on the deployed 921 tip.

Lenses 1–7 all run on this pass (inline; fan-out agent hit 401 on grok-4.6).

---

### Lens findings (verified)

#### 1. Correctness

| ID | Sev | Finding | Disposition |
|---|---|---|---|
| C1 | — | Exhausted transient 503 marks `remember_jobs.status=failed` on attempt 5/5 (`jobs.rs` ~222–226, test ~4232). Prod still leaves `running`. Live job confirmed. | **Discard as defect** — this is the intended fix. |
| C2 | — | Slot release is idempotent; double-release test exists. | **Discard as defect.** |
| C3 | — | Failed-write ack is claimed at recall **delivery** (`recall.rs` `claim_for_delivery`), not in the prefetch spawn. | **Discard as defect.** |
| C4 | Med | `/health` `writes=degraded` is **false if any write succeeded in 15 minutes**, even with ≥3 failures (`durable_writes_degraded`: `if succeeded > 0 { false }`, test at `admin.rs` ~1373–1384). Teo’s sequence (bulk 0/5 then 3 singles OK) still shows `write_ready=true` and `writes=ok`. | **Rescope** — this is a total-outage detector, not “this bulk died.” Do not claim it closes Teo’s health bullet. |

#### 2. Spec / AC (Teo + this PR)

| Ask | This PR vs `dev` | Closed for **Teo on prod**? |
|---|---|---|
| 1a Remember/bulk must not hang 240s | Wait budget + `job_id` (from #918, still default-wait-for-`done` per CHANGELOG) | **No** — #918+#921 are not on prod/staging |
| 1b Fail fast + real cause | 5/5 → `failed` now | **No** until promoted |
| 1c Health must not look OK while writes die | `writes=degraded` only when **zero** successes | **Partial**, and not on prod |
| 2 `ip_active_cap` | Not this PR (`TRUSTED_PROXY_HOPS` is ops) | Client 0.0.13; hops on prod |
| 3 Pin npx | `@0.0.14-dev.0` (npm has it; no `0.0.14`) | **Better than `dev`’s broken `@0.0.14`**. Teo-stable pin remains **0.0.13** until 0.0.14 is published |
| 4 0.0.13 latest + WALM-618 | Not this PR | Already true on npm |

PR-local AC (021 applied, claim-at-delivery, 918 follow-ups): **met**.  
“Single save in a few seconds”: **not** this PR’s default (`CHANGELOG` 0.0.14 Changed: still blocks until `blob_id`).

#### 3. Code quality

| ID | Sev | Finding | Disposition |
|---|---|---|---|
| Q1 | Low | `db.rs` ~108–111 comments that `recent_failed_remember_jobs` **writes** on every recall. The write is `claim_failed_write_reports` at delivery. | **Fix now** if you touch the file; else **defer** — stale comment only. |

#### 4. Cross-surface

| ID | Sev | Finding | Disposition |
|---|---|---|---|
| X1 | Med | `plugin.json` / `package.json` = `0.0.14`; launchers = `@0.0.14-dev.0`. Codex installer now follows `.mcp.json`. Anything still reading `plugin.json` version will `npx @0.0.14` → 404. | **Fix now** before a marketplace ship: keep `plugin.json` in lockstep with the launcher pin, **or** publish `0.0.14`. |
| X2 | Low | PR title is “021 + four #918 follow-ups”; the tip also has pin, exhausted-fail, slot logs, timeout knob. Body is still accurate enough. | **Discard** as blocker. |

#### 5. Edge cases

| ID | Sev | Finding | Disposition |
|---|---|---|---|
| E1 | — | Defer/congestion callers pass `attempt_info=None`, so a lock loser cannot mark failed on 5/5 while a winner runs. | **Discard as defect.** |
| E2 | Low | `recent_write_outcomes` only counts `failed`/`done`/`uploaded`. Jobs still `running` after spent retries **do not** trip degraded — which is why C1 matters. | **Discard** once C1 ships. |

#### 6. Security

No new authz surface. Failed-write report runs `sanitize_job_error_for_client` (`recall.rs` ~47–57) so pool-wallet hex / “insufficient balance” does not leak to tenants. Slot logs include `jobId` only. Pin uses public npm. **No security findings.**

#### 7. Production readiness

| ID | Sev | Finding | Disposition |
|---|---|---|---|
| P1 | — | `021` is `ADD COLUMN IF NOT EXISTS` + partial index; hooked in `VectorDb::new()`. Fresh DBs (CI) get it; long-lived DBs no-op. | **Discard as defect.** |
| P2 | Med | Merge to `dev` then **marketplace** would launch **`0.0.14-dev.0`**, not `latest`. Fine for internal `dev`. Wrong for Teo if that marketplace is what he updates. | **Fix now** on the ship path: Teo stays on **`@0.0.13`** until a real `0.0.14` publish. |
| P3 | — | Promoting Teo-ask-1 to **prod** is `dev` (with this PR) → staging → prod, **and** a matching MCP publish. This PR targeting `dev` is the first hop, not the last. | Informational. |

---

### Verdict

**Approve merge to `dev`.** CI is green; 021 unblocks fresh schemas; exhausted-fail, status 429 handling, claim-at-delivery, and the unpublished-`0.0.14` pin break are real follow-ups to #918.

**Do not tell Teo Ask 1 is closed** until that `dev` train is on **prod** and the plugin pin he installs is a **published** npm version.

**Do not wait on testnet 503** to merge this.

Suggested next hop (ops, not this review): merge #921 → `dev`, then staging → prod, keep marketplace users on `@0.0.13` until `0.0.14` exists on npm.

harrymove-ctrl (commented):
## Delivery review #2 (standalone, **full** depth — docs + “can we delete 021?”)

Previous review: approve merge to `dev`. This pass is extra depth on **docs** and a full consumer trace of **migration 021**. Operator asked whether 021 can be deleted because it “doesn’t seem needed.”

**Verdict unchanged: approve merge to `dev`.**  
**Do not delete 021.** Deleting it without reverting #918’s queries keeps CI red and breaks fresh schemas.

---

### 021 — can you delete it?

**No.** The SQL file already exists on **`origin/dev`** (shipped with #918). The **queries already use the column**. What #921 adds is only **applying** that file in `VectorDb::new()`.

`origin/dev` `VectorDb` chain ends at **020**. Production/CI schema is that `include_str!` list, not `sqlx::migrate!`. So:

| Piece | On `origin/dev` today | If you delete 021 from this PR |
|---|---|---|
| `migrations/021_failed_write_report_ack.sql` | **Already there** | File remains on `dev` unless you revert #918 |
| `failure_reported_at` in SQL | `remember.rs` reclaim/INSERT/SELECT; `db.rs` UPDATE/filter | **Still referenced** → `42703 column does not exist` on fresh DB |
| `VectorDb::new()` applies 021 | **Missing** (CI red: 4 tests) | Stays missing → CI stays red |
| Recall `failed_writes` + MCP `formatFailedWrites` | Already on `dev` | Still runs, still needs the column |

Call sites that **require the column** (verified `rg`):

- `services/server/src/routes/remember.rs` ~1070, ~1080, ~1649, ~1663 (reclaim clears it; tests insert/select it)
- `services/server/src/storage/db.rs` `recent_failed_remember_jobs` (`AND failure_reported_at IS NULL`), `claim_failed_write_reports` (`SET failure_reported_at = NOW()`)
- `services/server/src/routes/recall.rs` `failed_writes_for` / `claim_for_delivery`
- MCP `tools/recall.ts` `formatFailedWrites`

To drop 021 you would have to **revert the whole failed-write report feature from #918** (recall payload, reclaim SQL, tests, MCP formatter). That is not “delete one migration.” It is a product revert of “tell the agent a write died after accept.”

`ADD COLUMN IF NOT EXISTS` matches `migrations/README.md`: every listed file re-runs on every boot and must be idempotent.

**Disposition:** **Discard “delete 021.”** Keep the VectorDb hook. That *is* the CI fix this PR exists for.

---

### Docs lens (this was thin last time)

Checked: `docs/reference/environment-variables.md` (only docs file in the diff), `docs/relayer/api-reference.md`, `docs/mcp/changelog.mdx`, `packages/mcp/CHANGELOG.md`, `docs/troubleshooting/overview.md`, `services/server/migrations/README.md`.

| ID | Sev | Finding | Disposition |
|---|---|---|---|
| D1 | **High** | `docs/relayer/api-reference.md` § `GET /health` still says `writes` is **`"ok"` or `"paused"`**. Code now emits **`"degraded"`** (`types.rs` `writes_health_status`, `admin.rs` probe, MCP `health.ts`). Integrators following the published API doc will never look for the Teo signal. | **Fix now** — document `"degraded"`: recent durable writes failing **and** none landing in 15 min; `write_ready` stays true; HTTP 200. Also state the mixed-success case (any success → still `"ok"`). |
| D2 | Med | `docs/mcp/changelog.mdx` 0.0.14 is missing the **cold-start `BASELINE_RELAYER_TOOLS` / tool-not-served** bullet that `packages/mcp/CHANGELOG.md` already has (#928). Living docs lag the package changelog. | **Fix now** if 0.0.14 is what this branch ships. |
| D3 | Med | `docs/mcp/changelog.mdx` still says plugin pins **`@0.0.13`**. Launchers in this PR pin **`@0.0.14-dev.0`**. FAQ frontmatter still: “latest MCP package release is 0.0.13.” | **Fix now** on the marketplace path, or keep Teo on 0.0.13 and **don’t** merge the plugin pin to the published marketplace until `0.0.14` exists on npm. |
| D4 | Low | Env docs correctly flipped `MEMWAL_MCP_REMEMBER_WAIT_MS` default **0 → 90000** (matches CHANGELOG “still wait for `done`”). Walrus timeout/concurrency knobs added. Good. | — |
| D5 | Low | No doc for exhausted-retry → immediate `failed` (status/MCP). Operators will still read Slack 5/5 as “stuck uploading.” | **Defer** — optional troubleshooting note. |

---

### Other lenses (re-verified, not repeated in full)

- **Correctness:** 5/5 → `failed` still correct; slot release still idempotent; claim-at-delivery still correct. `writes=degraded` still **all-or-nothing** in the 15 min window (`admin.rs` test `durable_writes_degraded_only_when_nothing_lands`).
- **AC:** Teo Ask 1 not closed on prod. This PR’s own AC (apply 021, stop eating the failure report) **requires keeping 021**.
- **Security:** no change. Failed-write path still sanitizes `error_msg`.
- **Prod readiness:** 021 is the safe/required hop. Deleting it is **unsafe**. Promote `dev`→staging→prod after merge; pin Teo to published npm.

---

### What you should do

1. **Do not delete 021.** It is not optional once #918’s SQL is on `dev`.
2. **Docs fix-now:** `docs/relayer/api-reference.md` `writes: degraded`; align `docs/mcp/changelog.mdx` 0.0.14 + pin story.
3. **Then merge #921 → `dev`.**

harrymove-ctrl (commented):
## Docs hard-review (the three items + what they hide)

Checked at `2adfb261` against code. The two “stale docs” findings from the last review are **closed on this tip**. A worse contradiction is **not**.

### 1. `docs/relayer/api-reference.md` — `writes` only ok/paused

**Status: fixed** at `2adfb261`. Lines 85–89 now list `"ok"` / `"degraded"` / `"paused"` and match `writes_health_status` + `DURABLE_WRITE_WINDOW` (15 min) + `DURABLE_WRITE_FAILURE_THRESHOLD` (3) + “any success ⇒ not degraded”.

Nits (not the original bug):

- Response JSON example still only shows `"writes": "ok"`. Fine as the happy-path sample; a one-line note that `"degraded"` / `"paused"` also appear would help.
- `docs/reference/environment-variables.md` `WRITES_PAUSED` row still only mentions `writes: "paused"`. Cross-link, not a lie.

### 2. `docs/mcp/changelog.mdx` — 0.0.14 cold-start + pin

**Status: fixed** at `2adfb261`.

- Cold-start `BASELINE_RELAYER_TOOLS` / #928 bullet is present (line 40), same as `packages/mcp/CHANGELOG.md`.
- In-tree pin is documented as `@0.0.14-dev.0`; npm `latest` stays 0.0.13.
- The **0.0.13** section still saying pin `@0.0.13` is **correct history** (WALM-627), not drift.

Remaining: `plugin.json` `"version": "0.0.14"` vs launchers `@0.0.14-dev.0` (not a docs file; still a 404 if anything npxes the plugin.json version).

### 3. Env docs — `MEMWAL_MCP_REMEMBER_WAIT_MS` default 90000 + Walrus knobs

**Walrus knobs: match code.**

| Doc default | Code |
|---|---|
| `WALRUS_DIRECT_UPLOAD` `false` | `=== "true"` else false |
| `WALRUS_RELAY_TIMEOUT_MS` `120000` | `clients.ts` `\|\| 120_000` (no clamp) |
| `WALRUS_STORAGE_NODE_TIMEOUT_MS` `30000` range 1000–180000 | `parsePositiveIntEnv(..., 30_000, 1_000, 180_000)` |
| `WALRUS_UPLOAD_MAX_CONCURRENCY` size of key pool | `Math.max(1, SERVER_SUI_PRIVATE_KEYS.length \|\| 1)` |
| `WALRUS_UPLOAD_PER_WALLET_CONCURRENCY` `1` | default 1 |
| `WALRUS_UPLOAD_ACQUIRE_TIMEOUT_MS` `120000` | default 120_000, clamp 1_000–180_000 |

“Keep storage-node timeout under acquire timeout” is **advice**, not enforced.

**WAIT default: docs and changelog LIE about the code. This is the hard finding.**

```
remember-wait.ts
  DEFAULT_REMEMBER_WAIT_MS = 0          // return at accept
  comment: “do not wait at all”; operator sets 90000 to restore block-until-done

docs/reference/environment-variables.md
  default column: 90000
  text: “The default is the full ceiling, so a successful call carries a real blob_id”

packages/mcp/CHANGELOG.md + docs/mcp/changelog.mdx ### Changed
  “memwal_remember keeps blocking until the write reaches done”
  “Returning at accept is available behind MEMWAL_MCP_REMEMBER_WAIT_MS=0”

docs/mcp/changelog.mdx frontmatter (one sentence, two stories)
  “keeps memwal_remember waiting for done by default
   (MEMWAL_MCP_REMEMBER_WAIT_MS=0 returns at accept)”
```

Unset env ⇒ **0 ms wait** (accept-and-continue). Env table and the 0.0.14 “Changed” bullet describe the **opposite**. The 90000 table default was aligned to the changelog, not to `DEFAULT_REMEMBER_WAIT_MS`.

**Fix now:** pick one contract and make all four surfaces say it.

- If the product is accept-by-default (what the sidecar actually does): env default `0`; delete/rewrite the CHANGELOG “keeps blocking until done” bullet; frontmatter should not say “waiting for done by default”.
- If the product is block-until-done: change `DEFAULT_REMEMBER_WAIT_MS` to `90_000` and keep the current env/changelog. That is a **behavior** change, not a docs tweak.

Until that is one story, do not tell Teo or Henry “default remember waits 90s for blob_id” — **it does not**.

harrymove-ctrl:
## Delivery review #3 (standalone, **full** depth — binding “max”)

Could not use GitHub “request changes” (own PR). **Verdict is still: do not merge until C5 + X3 are fixed.**

**HEAD:** `d03ac413` vs `origin/dev`.

### What I read / ran
- Range `origin/dev..HEAD`, `gh pr checks 921`.
- Wait: `DEFAULT_REMEMBER_WAIT_MS = 0`; env default `0`; changelogs “return at accept”.
- Still lying: `packages/mcp/src/instructions.ts:42-49`, `services/server/scripts/mcp/server.ts:53-60`.
- CI MCP 105289217298: `not ok 130` `tool-definitions.test.mjs:124` regex `/NOT yet saved|NOT saved/i` vs copy “NOT **stored** yet”.
- CI Clippy: `orphaned_preparation_is_failed` — deadlock on **020**, not 021.
- 021 file gone; no `failure_reported_at` left.
- API docs list `writes=degraded`. Plugin pin `@0.0.14-dev.0`; `plugin.json` still `0.0.14`.

### Prior findings
| ID | Now |
|---|---|
| API `writes` only ok/paused | **Closed** (`2adfb261`) |
| Changelog 0.0.14 cold-start + pin | **Closed** (`2adfb261`) |
| Env wait default 90000 vs code 0 | **Docs closed** (`d03ac413`); instructions **not** (X3) |
| Delete 021 | **Done** (`da73a486`) |
| `writes=degraded` needs 0 successes | Still by design |

### Open — fix now
1. **C5 (CI red):** `cold-start write tools warn that a result may not be saved yet` — description says “NOT stored yet”, test wants `NOT saved`. Widen regex or say “NOT saved”.
2. **X3:** MCP initialize instructions still say remember **waits** and a success is **`blob_id` / stored**; `job_id` is the “exception”. Default wait is **0**; `job_id` is the **normal** result. Same text in `instructions.ts` and `scripts/mcp/server.ts`.

### Discard
Clippy deadlock on migration 020 = flake (seen before). Do not restore 021.

### Do not merge to `dev` while MCP login-handoff is red.

harrymove-ctrl:
## Delivery review (HEAD `812cfe5b`)

**CI: all required checks green**, including MCP login-handoff and Clippy + unit tests.

**Verdict: approve merge to `dev`.**

Closed since last request-changes:
- Wait default is `0` (accept / `job_id`); env + changelog + tool copy + initialize instructions agree.
- Cold-start pending-warning test matches `NOT stored` / `Settle`.
- Migration 021 and the recall `failed_writes` report are gone (intentional).
- `GET /health` `writes=degraded` is documented.

Nits (do not block `dev`):
- PR **title still says “apply migration 021”** — that work was reverted. Retitle before merge.
- `plugin.json` version `0.0.14` vs launcher `@0.0.14-dev.0`. Keep marketplace users on published `@0.0.13` until `0.0.14` exists on npm.
- Teo Ask 1 is **not closed on prod** until this train is staging → prod.

Promote path: merge this PR → `dev` → staging → prod. Do not wait on testnet 503.

services/server/scripts/mcp/tools/remember-status.ts
```diff
@@ -149,7 +151,29 @@ export function registerRememberStatusTool(
...
+                            }
+                        } catch (probe) {
+                            if (isRateLimited(probe)) {
+                                throw rateLimitedError(
+                                    probe,
+                                    "check whether the write landed"
+                                );
+                            }
+                        }
+                        return stillRunning(job_id);
```
ducnmm: **[bug]** The new confirming probe on the single-job wait path treats a terminal job as still in flight. After `waitForRememberJob` times out (504 / `isStillRunning`), the probe reads `getRememberStatus`. `done` returns `saved`, and pending/running return `stillRunning` with the state — but `failed` and `not_found` fall through the empty `if` and hit `return stillRunning(job_id)`. That is the case this probe exists for: polls that were all 429s (or otherwise swallowed) so the SDK stamped timeout even though the job already died. The agent is then told not to re-send, which is silent loss. The budget-0 path already throws `nameJobError` for these two statuses, and the batch probe replaces rows so `failed`/`not_found` render as `NOT stored`. Only this branch launders them into progress. The GET is also unbounded (`withAcceptDeadline` is used on the batch probe and on the budget-0 read, not here), so a stalled confirm can hang the tool after the wait already expired.

**Suggestion:** Mirror the budget-0 single-job handling inside the probe: `failed` → `nameJobError` with status 500, `not_found` → 404, and wrap `getRememberStatus` in `withAcceptDeadline(..., { idempotent: true })`. Add a test next to the batch rate-limit cases: wait times out, probe returns `failed`, result is an error and must not match `/still uploading/i`.

packages/mcp/src/instructions.ts
```diff
@@ -39,14 +39,12 @@ export const PROACTIVE_INSTRUCTIONS = [
...
-    "",
-    "If a write outruns that budget the call returns job_ids instead, saying the facts are",
-    "ACCEPTED but NOT YET SAVED. That is the exception, not the normal result. When it",
-    "happens, say the facts are being saved rather than that they are saved, do NOT re-send",
-    "them — that queues duplicates — and resolve the ids with memwal_remember_status (it takes",
-    "job_id, or job_ids for a whole batch). Only a blob_id means a fact is stored.",
+    "By default memwal_remember and memwal_remember_bulk return in ~1s once the relayer has",
+    "accepted the job (job_id / job_ids). The Walrus write continues in the background (~30-60s)",
+    "and the fact is NOT stored yet. That is the normal result. Do not claim it is saved.",
+    "Do NOT re-send the same text — that queues duplicates. Resolve with memwal_remember_status",
```
ducnmm: **[bug]** Cold-start tool descriptions were rewritten so they no longer name `memwal_remember_status`, and a test now fails if a cold-start description points at an unadvertised tool — but `PROACTIVE_INSTRUCTIONS` (injected at `initialize`, before any `tools/list`) still says “Resolve with memwal_remember_status”. That is the same #928 defect the new test describes: a description is an instruction the agent follows, so naming an unadvertised tool is the same defect as listing it. The local refusal cannot save a call that follows this sentence during the cold-start window: `upstreamToolNames` is only filled from a forwarded `tools/list`, the cold-start list is answered locally and does not populate it, and the gate is skipped while the set is empty. `tool-not-served.test.mjs` waits for `list_changed` plus a re-list before calling, so it never covers this path. A cached client tool list or this initialize text will still forward `memwal_remember_status` into a 0.0.13 relayer and sit in `inFlight` until the orphan deadline.

**Suggestion:** Stop naming `memwal_remember_status` in `PROACTIVE_INSTRUCTIONS`; use the same “job-status tool this server advertises / re-list tools” wording as the cold-start remember descriptions. Until an upstream `tools/list` has been seen, refuse `tools/call` names that are not in the cold-start list + `LOCAL_TOOL_NAMES` instead of forwarding them. Extend the tool-not-served test to call the missing tool after initialize / cold `tools/list` and before the first forwarded list, and assert a fast local error. The sidecar copy in `server.ts` can keep the explicit name — that process actually registers the tool.

packages/mcp/src/bridge.ts
```diff
@@ -1348,6 +1360,12 @@ export async function runBridge(
      * client surfaces them in its tool palette. */
     const pendingListIds = new Set<string | number>();
 
+    /** Tool names the CONNECTED relayer advertised on its last `tools/list`,
+     * minus the ones we serve locally. Empty until the client has listed tools
+     * at least once over a live session — until then we know nothing about the
+     * relayer's capabilities and gate nothing. */
+    const upstreamToolNames = new Set<string>();
```
ducnmm, line 1374: **[suggestion]** `upstreamToolNames` is never cleared on reconnect. `reconnect()` publishes a new SSE session (and can swap `relayerUrl` on `login-credentials-updated`) but leaves the last advertised set in place until the client happens to re-list. A stale allow re-forwards a tool the new relayer does not serve (the #928 hang). A stale deny refuses a tool the new relayer does serve until the next `tools/list`.

**Suggestion:** `upstreamToolNames.clear()` at the start of a successful reconnect so the gate fails open until the next upstream list, matching the documented “empty means we know nothing” rule.

services/server/src/storage/db.rs
```diff
@@ -1837,106 +1847,57 @@ impl VectorDb {
...
-             RETURNING id, namespace, error_msg, updated_at",
-        )
-        .bind(owner)
-        .bind(chrono::Utc::now() - chrono::Duration::from_std(window).unwrap_or_default())
-        .bind(limit)
-        .fetch_all(&self.pool)
+        let since =
+            chrono::Utc::now() - chrono::Duration::from_std(window).unwrap_or_default();
+        let outcome = sqlx::query_as::<_, (i64, i64)>(
+            "SELECT
```
ducnmm, line 2205: **[suggestion]** `/health` now runs `recent_write_outcomes` on every cache miss. The predicate is `status IN ('failed','done','uploaded') AND updated_at >= $1` against `remember_jobs`, which the comment correctly notes is not indexed on `updated_at` alone and is never pruned. Unlike `write_ready`, this probe has no statement timeout. A sequential scan on a large table can stall the liveness handler for well beyond the 30s cache TTL; a hung `/health` is worse than failing open.

**Suggestion:** Bound the query the same way as the Postgres size probe (1s, fail open). If the scan is expected to matter in production, add a matching index or restrict the scan to a cheap partial index on recent terminal rows.

services/server/scripts/mcp/tools/remember-wait.ts
```diff
@@ -61,26 +61,49 @@ export const MAX_REMEMBER_WAIT_MS = 90_000;
...
+ * It is NOT safe from a job that fails after acceptance — that is real, and
+ * dev proved it on 2026-09-17 when every register transaction was rejected
+ * post-accept. Only a later `memwal_remember_status` call surfaces that, which
+ * is why the pending message points at it and refuses to claim the fact is
+ * stored.
+ *
+ * An operator who wants the old always-block behaviour sets
+ * `MEMWAL_MCP_REMEMBER_WAIT_MS=90000` knowingly.
+ *
+ * This constant does not fix `memwal_analyze`: its extraction leg is a
```
ducnmm: **[suggestion]** The new default-wait comment is both history-narration (“This REVERSES D1…”) and factually stale: it says `memwal_analyze`’s extraction leg is still a 60_000ms deadline that exceeds the client ceiling for any wait value, but this same PR sets `ANALYZE_EXTRACTION_DEADLINE_MS` to 45_000 and tests that it sits under the 60s client timeout.

**Suggestion:** Drop the design-history and the outdated analyze claim. One or two sentences on why the default is 0 (client 60s ceiling; in-between budgets still return pending) is enough.

ducnmm (changes requested):
## Summary

This branch started as four #918 follow-ups and then absorbed a much larger set of live-incident fixes (cold-start tool floor, wait-at-accept default, rate-limit wording, register-transaction expiration, writes=degraded, dropping the failed-write report and migration 021). The analyze/bulk job_id retention, the `timeout`≠failed rendering, and the migration-pipeline completeness check look correct. The two real defects are on the new status-probe path (a terminal failure can still be reported as “still uploading”) and on the #928 wording/gate (initialize instructions still name `memwal_remember_status`, and the local refusal does not run until after the first upstream `tools/list`).

## Issue counts by severity

- bugs: 2
- suggestions: 3
- nits: 0

harrymove-ctrl:
Addressed the two review bugs, plus the leftover-column cleanup that the 021 revert left open on live `relayer.dev`.

- **remember-status probe** (`6c4bfa22`): a wait timeout whose confirming GET is `failed` / `not_found` is now an error, not “still uploading”. GET is bounded with `withAcceptDeadline`.
- **#928 cold-start window** (`13b6de51`): initialize instructions no longer name `memwal_remember_status`. Until the first upstream `tools/list`, only the cold-start floor is callable.
- **021 leftover on live dev** (`c007c94f`): `021_drop_failed_write_report_ack.sql` is `DROP INDEX/COLUMN IF EXISTS`. Railway statuses show PR builds with the ADD wired (`5377867c` and others) deployed to `relayer.dev.memwal.ai` before the revert; `IF EXISTS` keeps CI / origin/dev-only DBs a no-op.

The three suggestions (clear `upstreamToolNames` on reconnect, statement timeout on `recent_write_outcomes`, trim the wait-default comment) are not in this push.

services/server/scripts/mcp/tools/remember-status.ts
harrymove-ctrl: Fixed in 6c4bfa22.

The probe now interprets the GET *outside* the catch, so `failed` / `not_found` cannot fall through to `stillRunning`. Shared `settleFromStatus` with the budget-0 path (`failed` → `MemWalRememberJobFailed` 500, `not_found` → 404). The GET is wrapped in `withAcceptDeadline(..., { idempotent: true })`. A failed probe that is not a 429 still returns still-uploading rather than inventing a terminal outcome.

Tests in `status-rate-limit.test.ts`: wait times out, probe returns `failed` / `not_found` → `isError`, must not match `/still uploading/i`.

packages/mcp/src/instructions.ts
harrymove-ctrl: Fixed in 13b6de51.

`PROACTIVE_INSTRUCTIONS` now uses the same “job-status tool this server advertises (re-list tools if you do not see one)” wording as the cold-start remember descriptions — it no longer names `memwal_remember_status`. The sidecar copy in `server.ts` still does; that process actually registers the tool.

Until an upstream `tools/list` has been seen, a `tools/call` for a name outside the cold-start floor + locally-served tools is refused locally (same `unknownToolText` as the post-list gate). `tool-not-served.test.mjs` now calls the missing tool after initialize / cold `tools/list` and before the first forwarded list, and asserts a fast local error that never reached the relayer.

packages/mcp/src/bridge.ts
harrymove-ctrl: Fixed in ea91d059.

`reconnect()` now `upstreamToolNames.clear()`s when the new session is published, and emits `tools/list_changed` so the client re-lists. Empty still means the cold-start floor (fail-closed), not fail-open — a stale allow after swapping onto a 0.0.13 relayer is the #928 hang.

Test: first session advertises `memwal_remember_status`, second session is 0.0.13 and stays silent on that name. After reconnect, a call without re-listing is refused locally and never reaches session-2.

services/server/src/storage/db.rs
harrymove-ctrl: Fixed in 9ece083e.

The probe is wrapped in the same 1s `tokio::time::timeout` as the Postgres size check and fails open on timeout (or any query error). The statement itself also `SET LOCAL statement_timeout = 1000ms` inside a transaction, because the pool GUC is the startup bound (up to 300s) and a tokio timeout alone would leave the scan occupying a backend.

services/server/scripts/mcp/tools/remember-wait.ts
harrymove-ctrl: Fixed in 66f8eca8.

The comment is now two sentences: default is 0 because the MCP client's 60s ceiling cannot fit a real Walrus wait, and any in-between budget still returns pending. Restore block-until-done with `MEMWAL_MCP_REMEMBER_WAIT_MS=90000`. D1 history and the stale 60s analyze claim are gone.

harrymove-ctrl:
The three review suggestions are in as well:

- `ea91d059` — clear `upstreamToolNames` on reconnect (fail-closed to the cold-start floor until the next list; `tools/list_changed` so the client refreshes).
- `9ece083e` — `/health` write-outcome probe bounded at 1s, fail open; Postgres `SET LOCAL statement_timeout=1000ms` so the scan is actually cancelled.
- `66f8eca8` — wait-default comment trimmed.

HEAD is `66f8eca8`.

services/server/scripts/mcp/__tests__/remember-fast-return.test.ts
```diff
@@ -104,23 +104,26 @@ function textOf(result: unknown): string {
...
-test("the default is the full wait, and zero is opt-in", () => {
-    // D1: a result means the fact landed, so an unset budget blocks to
-    // terminal. The in-between is the setting to avoid — a budget under the
-    // real completion time pays the wait AND still returns pending.
-    assert.equal(parseWaitBudget(undefined), MAX_REMEMBER_WAIT_MS);
-    assert.equal(parseWaitBudget(""), MAX_REMEMBER_WAIT_MS);
-    // ...and this file asked for the opt-in path at the top.
+test("the default returns at accept, and blocking is opt-in", () => {
+    // Reverses D1. A full-ceiling call runs ACCEPT_DEADLINE_MS + 90_000 +
+    // WAIT_OVERSHOOT_GRACE_MS = 115s against the MCP SDK's 60s default
```
ducnmm: **[suggestion]** The same file now asserts `parseWaitBudget(undefined) === 0` (“the default returns at accept”), but this new comment still narrates D1 history (“Reverses D1… honour a result means it landed”). The file header (lines 1–3, not in this hunk) still says accept-and-continue “is NOT the default”. Those comments are now false in a file this PR edited.

**Suggestion:** Drop the D1 / “not the default” wording. The env assignment is only so the module under test reads 0 even if the process already had another value; the product default is 0.

ducnmm (changes requested):
## Summary

All five prior findings are fixed on `66f8eca8`. The status probe now shares `settleFromStatus` so `failed` / `not_found` cannot fall through to still-uploading; initialize instructions no longer name `memwal_remember_status` and a cold-start `tools/call` is fail-closed on the baseline floor; reconnect clears `upstreamToolNames` and emits `tools/list_changed`; `/health` bounds `recent_write_outcomes` with a 1s tokio timeout, fail-open, and `SET LOCAL statement_timeout`; the wait-default comment is now “return at accept / restore with 90000”, with D1 history and the stale 60s analyze claim gone. Plugin launchers pin `@mysten-incubation/memwal-mcp@0.0.14-dev.0` (that version exists on npm; `0.0.14` does not), and CI is green. One leftover docs lie about the new default remains.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

## Issues outside the diff

- **[bug]** docs/mcp/reference.md:489 -- The published MCP troubleshooting guide still says `memwal_remember` waits for the Walrus upload to finish before returning. This PR’s default is `MEMWAL_MCP_REMEMBER_WAIT_MS=0` (return at accept with `job_id`; the fact is not stored yet). Changelog and env-var docs were updated; this sentence was not. An operator following “Recall returns nothing right after a remember” will treat a ~1s accept as a completed write and retry recall too early — or assume a missing recall is only an index lag.
  - **Suggestion:** Rewrite that section for return-at-accept: a remember reply is not stored until `blob_id` (or `memwal_remember_status` reports saved); empty recall immediately after accept is expected. Point at `MEMWAL_MCP_REMEMBER_WAIT_MS=90000` only as the opt-in wait-for-`blob_id` path.

services/server/scripts/mcp/__tests__/remember-fast-return.test.ts
harrymove-ctrl: Fixed in f25b6862.

File header and the `parseWaitBudget(undefined) === 0` comment no longer say accept-and-continue is not the default / Reverses D1. The env assignment is only so the module under test reads 0 even if the process already had another value; the product default is 0.

harrymove-ctrl:
Second review leftovers on `66f8eca8`, fixed in `f25b6862`:

- **docs/mcp/reference.md** (and the matching section in `docs/troubleshooting/overview.md`): no longer says `memwal_remember` waits for the Walrus upload. Empty recall right after accept is expected; settle with the job-status tool / `blob_id`. `MEMWAL_MCP_REMEMBER_WAIT_MS=90000` is the opt-in wait.
- **remember-fast-return.test.ts**: comments no longer call accept-and-continue “NOT the default”.

HEAD `f25b6862`. CI on this push still running.

harrymove-ctrl:
Follow-up from the final review of `f25b6862`:

- `bbf15165` — status page: `writes=degraded` is no longer rendered as All Systems Operational. Banner/pill say **Writes Degraded**.
- `e34446a1` — `022_remember_jobs_recent_outcomes.sql`: partial index on `updated_at DESC` for terminal rows, so the `/health` probe does not fail-open forever on a large `remember_jobs`.
- `e508190b` — storage-node timeout comment no longer confuses `uploadRelay.timeout` with a per-node timeout; `writes_health_status` doc names degraded; cold-start remember text says `blob_id` in the same reply means stored; changelog notes unpinning `0.0.14-dev.0`.

HEAD `e508190b`.

services/server/src/storage/db.rs
```diff
@@ -1195,58 +1496,50 @@ async fn backfill_updated_at(pool: &PgPool) -> Result<(), AppError> {
...
+             WHERE pg_class.relname = $1",
+        )
+        .bind(index)
+        .fetch_optional(pool)
+        .await
+        .map_err(|e| AppError::Internal(format!("Failed to check validity of {}: {}", index, e)))?;
+
+        // None means no index by that name -- nothing built it yet, which
+        // its own migration handles. Some(true) is a healthy index.
+        if index_is_invalid != Some(false) {
```
ducnmm, line 1589: **[bug]** `recover_invalid_concurrent_indexes` treats `pg_index.indisvalid = false` as "interrupted leftover, drop it." That flag is also false for the entire duration of a live `CREATE INDEX CONCURRENTLY`. 022 is a new CONCURRENTLY index on never-pruned `remember_jobs` (the table the probe already could not scan in 1s), built inside `VectorDb::new()` with no advisory lock and no `pg_stat_progress_create_index` check. Two replicas booting together — first deploy, restart-all, scale-up — both run recover-then-create: the follower sees the leader's in-progress (or just-finished) catalog row as INVALID, then `DROP INDEX CONCURRENTLY` without `IF EXISTS` (line 1540). `DROP INDEX CONCURRENTLY` conflicts with the in-progress build, waits it out, and then drops the now-valid index so this replica can rebuild it; if the sibling already dropped the leftover, the unqualified DROP errors and `VectorDb::new()` fails (`main.rs` `.expect`s that). 016/018 are already present on prod so this hole was latent; 022 is the first boot that actually builds under this loop.

**Suggestion:** Skip the drop when `pg_stat_progress_create_index` shows a live build for that `indexrelid`. Use `DROP INDEX CONCURRENTLY IF EXISTS`. Serialize recover+022 with a session advisory lock so only one replica builds. Optionally `SET statement_timeout = 0` for that build so a large `remember_jobs` cannot leave INVALID by hitting the server default, then fail the boot.

services/server/src/routes/admin.rs
```diff
@@ -150,11 +150,25 @@ pub async fn health(State(state): State<Arc<AppState>>) -> Json<HealthResponse>
...
 const WRITE_READY_CACHE_TTL: std::time::Duration = std::time::Duration::from_secs(2);
+/// How far back `/health` looks when judging whether durable writes land.
+/// Long enough to span a few upload attempts with their backoff, short
+/// enough that recovery shows up without an operator waiting.
+const DURABLE_WRITE_WINDOW: std::time::Duration = std::time::Duration::from_secs(15 * 60);
+/// Failures inside the window before the write path is called degraded.
+/// Writes fail individually all the time; one or two is not an outage.
+const DURABLE_WRITE_FAILURE_THRESHOLD: i64 = 3;
+/// Cached far longer than `write_ready`: this one aggregates over
+/// `remember_jobs`, which has no index on `updated_at` alone, so it must
```
ducnmm: **[suggestion]** After 022, `/health`'s probe is served by `remember_jobs_recent_outcomes_idx` on `updated_at`. This comment still says the query "has no index on `updated_at` alone," and the probe doc at line 356 still says the 1s bound is to stop "a sequential scan of `remember_jobs`." Both were true when the probe landed; they are now the opposite of why 022 exists, and they will send the next reader hunting for a missing index that is already there.

**Suggestion:** Reword to: cached so `/health` does not hit `remember_jobs` on every load-balancer tick; 022's partial `updated_at` index is what keeps the 1s bound a backstop rather than the steady state.

ducnmm (changes requested):
## Summary

The post-66f8eca8 delta is mostly docs honesty plus two real follow-ups: the status page now treats `writes=degraded` as degraded rather than operational, and 022 adds a partial `remember_jobs(updated_at)` index so `/health`'s write-outcome probe can stop failing open on a seq scan. Prior findings 1–7 are actually fixed on this HEAD (`settleFromStatus` reports `failed`/`not_found`; initialize/cold-start copy no longer names `memwal_remember_status`; `reconnect()` clears `upstreamToolNames`; the probe is still 1s/fail-open; wait-default comments no longer narrate D1; `docs/mcp/reference.md:489` and the fast-return test comments match return-at-accept). Leftover wait-for-`blob_id` claims in reference, troubleshooting, changelogs, env-var docs, initialize instructions, tool descriptions, and cold-start copy are gone. The remaining defect is the new invalid-index recovery: it cannot tell a live `CREATE INDEX CONCURRENTLY` from a crashed one, which is the first-deploy path for 022.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

services/server/src/storage/db.rs
harrymove-ctrl: Fixed in 7a7eb3d6.

- Skip the drop when `pg_stat_progress_create_index` shows a live build for that `indexrelid` (`indisvalid=false` is also the in-progress CONCURRENTLY state).
- `DROP INDEX CONCURRENTLY IF EXISTS` so a sibling that already cleaned up does not fail `VectorDb::new`.
- Recover + 016/018/022 run on **one session** behind `pg_advisory_lock(872122, 22)` so two replicas cannot interleave recover and build.
- That session sets `statement_timeout = 0` (restored before the connection returns to the pool) so a large `remember_jobs` cannot leave 022 INVALID by hitting the server GUC.

Tests: drop SQL contains `IF EXISTS`; live-build → do not drop.

services/server/src/routes/admin.rs
harrymove-ctrl: Fixed in 7a7eb3d6.

Comments now say the 30s cache is so `/health` does not hit `remember_jobs` on every load-balancer tick, 022's partial `updated_at` index is the steady-state path, and the 1s bound is a backstop rather than the expected scan.

services/server/src/storage/db.rs
```diff
@@ -1195,62 +1542,178 @@ async fn backfill_updated_at(pool: &PgPool) -> Result<(), AppError> {
...
+/// `remember_jobs` cannot leave 022 INVALID by hitting the server GUC.
+async fn recover_and_rebuild_concurrent_indexes(pool: &PgPool) -> Result<(), AppError> {
+    let mut conn = pool.acquire().await.map_err(|e| {
+        AppError::Internal(format!(
+            "Failed to acquire a connection for concurrent-index rebuild: {}",
+            e
+        ))
+    })?;
+
+    sqlx::query("SELECT pg_advisory_lock($1, $2)")
```
ducnmm: **[bug]** `recover_and_rebuild_concurrent_indexes` takes a session-level `pg_advisory_lock` and then runs recover + `CREATE INDEX CONCURRENTLY` on that same backend. That lock choice is right (CIC cannot sit in a transaction, so `pg_advisory_xact_lock` cannot cover the build). The wait is not. Replica B blocks inside `SELECT pg_advisory_lock($1, $2)`, which is an open statement with a snapshot. Replica A, holding the lock, is in 022's `CREATE INDEX CONCURRENTLY`. CIC's locker waits ("waiting for writers before build", "waiting for old snapshots") wait for transactions that can see `remember_jobs` to finish — including B. B is waiting for A's advisory lock. Postgres aborts one side with `deadlock detected`. If CIC is the aborted party, 022 is left INVALID and this `VectorDb::new` fails; a follower retrying immediately can re-enter the same wait during the next snapshot barrier. `DROP INDEX CONCURRENTLY` in recover has the same waiter-snapshot deadlock. Unlock-on-error and `IF EXISTS` do run after a failed rebuild (1677–1685), and `statement_timeout` is restored even on error (1711–1714). The progress skip at 1591–1600 does not close this: `pg_stat_progress_create_index` only shows other backends to superusers / `pg_read_all_stats`, recover runs before this session's own CIC, and an unreadable view is treated as "not live" (`Ok(false)` at 1649). Same-version replicas therefore never DROP a live sibling build — they deadlock instead of interleaving — but the first 022 deploy on a large `remember_jobs` is still a boot failure, which is the scenario the lock was added to make safe.

**Suggestion:** Wait with `pg_try_advisory_lock` plus a Rust sleep so waiters hold no Postgres snapshot, then keep recover + 016/018/022 on that locked session with `statement_timeout = 0`. Do not `continue` past a live build and then unlock; either the try-lock is the fence, or wait until `pg_stat_progress_create_index` is empty *while still holding the lock*. If `pg_advisory_unlock` fails, close/detach the connection instead of returning it to the sqlx pool still holding a session lock (the leak this file already warns about at `admit_storage_reservations`).

ducnmm (changes requested):
## Summary

Prior finding 2 is fixed on 7a7eb3d6: the `/health` comments now describe 022 as the steady-state path and the 1s bound as a backstop. Prior finding 1 is only partly fixed. Recover + 016/018/022 do run on one session behind `pg_advisory_lock(872122, 22)` with `DROP INDEX CONCURRENTLY IF EXISTS`, a progress-table skip, and `statement_timeout = 0` restored after the build. Unlock runs after rebuild errors. The remaining hole is the fence they added for the two-replica boot: a blocking session `pg_advisory_lock` waiter holds a snapshot for the whole wait, which deadlocks with `CREATE INDEX CONCURRENTLY` (and with `DROP INDEX CONCURRENTLY`) — the exact first-022 / restart-all case the lock was meant to serialize.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

ducnmm (approved):
## Summary

Re-reviewed `09db12d7`. The advisory-lock deadlock is fixed: `recover_and_rebuild_concurrent_indexes` polls `pg_try_advisory_lock` and sleeps in Rust between probes, so a waiter holds no snapshot; the wait is capped at 20 minutes; a failed unlock or failed `statement_timeout` restore closes the connection instead of returning it to the pool. Prior findings (status probe, wait-default docs, reconnect tool set, `writes=degraded`) stay fixed. No remaining issues.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #931, opened 2026-09-18
fix(mcp,server,sdk): say why a recall timed out (WALM-396)

nikola0x0:
## Ticket

WALM-396 — https://linear.app/mysten-labs/issue/WALM-396/mcp-bridge-timeout-error-doesnt-indicate-root-cause-relayer-down-vs

## Why is this needed?

A recall that runs out of time tells the agent nothing it can act on:

- sidecar (SDK 0.1.7): `Tool error: This operation was aborted`
- bridge, reply lost: `did not answer this call … safe to retry`, after **240s**

The agent can't tell a dead relayer from a wrong URL from one slow step, so it can't choose between retrying, waiting and reconfiguring.

## What changed?

Every timeout message now has the same three lines:

```
❌ Walrus Memory memwal_recall timed out.
Cause: No answer within the SDK's time limit. The relayer answered its health check, so the time went on this call: inside the relayer, or on a service it waits on first (such as the Sui fullnode).
Relayer health: ok (126ms, v0.1.0).
Next step: Repeating this call cannot store a duplicate, so it is safe to retry once. If it keeps happening, report it with the time of the call.
```

| Part | Change |
|---|---|
| **SDK** (0.1.8, unreleased) | `recall()` sends `deadline_ms: 15000` in the signed body. A body field, not a header: older relayers ignore it, and a header would need the CORS allow-list deployed first. |
| **Relayer** | With `deadline_ms`, the recall stops ~1s before it (counted from request arrival, 2s floor) and returns `504 {code: "RECALL_TIMEOUT", stage}`. Stage is `embed`, `vector_search`, `walrus_download` or `seal_decrypt`, tracked in a task-local, so `Engine::fetch_batch` keeps its signature. **Without `deadline_ms` (Python, older SDKs), nothing changes.** A recall the caller hangs up on logs the stage it was in. |
| **Sidecar** | `wrapTool` names the stage for `RECALL_TIMEOUT`. For an SDK timeout or a failed connect it checks the relayer's `/health` (2s) first, and only blames the relayer if that fails too. Writes are never told a retry is safe. Other errors keep their old wording. |
| **Bridge** | A sent call whose reply never arrives is answered after a `/health` check (3s, `MEMWAL_MCP_HEALTH_PROBE_MS`). A late reply, logout or shutdown during the check still wins, so the call never gets two answers. `memwal_recall` gets a 90s ceiling: a lost reply is answered at 2 min instead of 4. |

## Testing

- Unit: SDK 130/130, bridge 159/159, sidecar 327/327. Relayer: all pass against a local pgvector DB, except 4 remember-job tests that fail on `dev` too (migration `021` isn't applied; #921 fixes it).
- Live, against `relayer.dev.memwal.ai` (reads only):
  - branch SDK recall works; `deadline_ms=1` is ignored by the current relayer
  - branch bridge: normal calls unchanged; with a 1s call timeout, the lost-reply message carries the real `/health` result, and the late real reply is dropped
  - branch sidecar with the real SDK 0.1.7 against a fake relayer that stalls or drops `/api/recall`: all three messages come out right (above is the stall case)
- **Not tested live:** the relayer's 504 against real Walrus/SEAL, since it needs this relayer deployed. Worth a check on staging, including that the edge proxy passes the 504 body through.

## Notes for reviewers

- Recall timeouts will now show as 504s (`recall_timeout`) in relayer metrics. Today they're invisible: the caller just disconnects.
- The sidecar only gets the stage name once its SDK pin moves from 0.1.7 to 0.1.8 (follow-up after 0.1.8 is published). Until then it uses the health-check path.
- Overlaps #883 (WALM-598) in SDK `recall()`; whichever merges second rebases.
- Known gaps: the courtesy `failed_writes` lookup is awaited under the last stage, so a stall there would be reported as that stage. No test drives the full `recall` handler; the deadline and stage logic is tested directly in `engine/stage.rs`.


services/server/src/engine/stage.rs
```diff
@@ -0,0 +1,249 @@
...
+/// How long a recall may run for a caller that said it waits `deadline_ms`,
+/// given `already` spent since the request arrived (auth, rate limiting).
+/// `None` when it said nothing: those callers keep today's behaviour.
+pub fn budget_for(deadline_ms: Option<u64>, already: Duration) -> Option<Duration> {
+    let caller = Duration::from_millis(deadline_ms?.min(MAX_DEADLINE_MS));
+    Some(
+        caller
+            .saturating_sub(DEADLINE_MARGIN)
+            .saturating_sub(already)
+            .max(MIN_BUDGET),
```
ducnmm: **[suggestion]** `budget_for` subtracts `already` so the 504 can land before the caller's abort, then `.max(MIN_BUDGET)` undoes that once remaining time after the 1s margin is under 2s. The unit test locks this in: a 15s deadline with 20s already spent still runs 2s. In that case (cold on-chain auth) the SDK has already aborted, the 504 is wasted, and the caller still sees a blind timeout — the failure WALM-396 is meant to fix. A short explicit `deadline_ms` is documented as getting a 2s floor; overrunning a deadline that is already exhausted is not.

**Suggestion:** If `caller - DEADLINE_MARGIN - already` is zero, return `RecallTimeout` immediately (stage `embed`, or whatever the marker already holds) instead of starting work. Keep the 2s floor only when the caller sent a short deadline and time still remains.

services/server/src/routes/recall.rs
```diff
@@ -279,14 +331,14 @@ pub async fn recall(
...
             owner
         );
-        return Ok(Json(RecallResponse {
+        return Ok(RecallResponse {
             results: vec![],
             total: 0,
             dropped_count: 0,
             // Reported even with no hits: an empty recall is exactly when a
             // caller is most likely to be looking for the fact that failed.
             failed_writes: failed_writes.await.unwrap_or_default(),
```
ducnmm: **[suggestion]** `failed_writes.await` sits inside `recall_pipeline`, so it is on the caller deadline. That join is a courtesy report spawned before the deadline starts; hydration and ranking may already be done when this await is the thing that expires. A 504 then discards a finished recall and mis-labels the stage as `vector_search` (empty hits) or `seal_decrypt` (hydrated path). The same await is at recall.rs:430.

**Suggestion:** Join `failed_writes` outside `run_with_deadline`, or bound that join with a tiny timeout and omit the report rather than failing the recall.

packages/mcp/src/bridge.ts
```diff
@@ -2439,6 +2482,48 @@ export async function runBridge(
...
+                // up with this one call stuck (WALM-396). Only the answer
+                // waits on the probe; the bookkeeping stays synchronous.
+                if (entry.probing) continue;
+                entry.probing = true;
+                const relayerUrl = creds?.relayerUrl ?? config.relayerUrl;
+                // `probeRelayerHealth` does not reject, but a call left with
+                // `probing` set and no answer is the one outcome this path
+                // must never have, so a rejection still answers it.
+                sweepProbe ??= probeRelayerHealth(relayerUrl, healthProbeMs).catch(() => null);
+                void sweepProbe.then((probe) => {
```
ducnmm: **[suggestion]** `probeRelayerHealth(...).catch(() => null)` covers a rejecting probe, and the comment says a call left with `probing` set and no answer must never happen. The `then` callback that actually answers the call has no `.catch`. If `log.warn` or `failRequest` throws, that is an unhandled rejection, `entry.probing` stays true, later sweeps skip the id, and the client is never answered.

**Suggestion:** Chain `.catch` on the `then` (or wrap the callback in try/finally) so a throw still `failRequest`s and never leaves `probing` stuck.

packages/mcp/src/bridge.ts
```diff
@@ -2348,23 +2373,39 @@ export async function runBridge(
...
                     },
                 };
             }
+            // Not "this call only reads": `memwal_restore` re-indexes. What
+            // makes a retry safe is that none of these can store a duplicate.
+            const nextStep =
+                described === null || described.reachable
+                    ? "Repeating this call cannot store a duplicate, so it is safe to retry " +
+                      "once. If it keeps happening, report it with the time of the call."
+                    : "Repeating this call cannot store a duplicate, so it is safe to retry " +
```
ducnmm, line 2388: **[suggestion]** Unreachable-relayer copy reads "it is safe to retry once that is resolved". An agent can parse that as "retry once" rather than "retry after the outage is resolved".

**Suggestion:** Something like: "Safe to retry once the relayer is reachable again — wait a minute if the problem is on the relayer's side."

services/server/scripts/mcp/tools/failure.ts
```diff
@@ -0,0 +1,227 @@
+/**
```
ducnmm, line 1: **[suggestion]** Several new comments narrate WALM-396 / the old "This operation was aborted" wording instead of a short why. Same pattern at packages/mcp/src/health-probe.ts:1, packages/sdk/src/memwal.ts:811, services/server/src/engine/stage.rs:1, and packages/mcp/src/bridge.ts:296.

**Suggestion:** Keep a one-line why (e.g. "Relayer cuts the recall just before the caller's abort so the 504 can name the stage"). Drop ticket ids, previous wording, and the 2-vs-4-minute history from code comments; that belongs in the changelog.

ducnmm (commented):
## Summary

The change is coherent: the TypeScript SDK puts `deadline_ms` in the signed `/api/recall` body, the relayer stops about a second early and returns `504 RECALL_TIMEOUT` with the in-flight stage, and both MCP surfaces (sidecar `wrapTool` and the stdio bridge) turn a timeout or a lost reply into Cause / Relayer health / Next step. Stage tracking via task-local `enter`, the late-SSE-vs-probe identity check, and the SDK `serverCode`/`cause` shape all look correct, and the new tests cover the important races. The 15s/fast-auth path should beat the SDK abort; the 2s budget floor can still lose that race after slow auth.

## Issue counts by severity

- bugs: 0
- suggestions: 5
- nits: 0

services/server/src/engine/stage.rs
nikola0x0: Fixed in 7983781. `budget_for` now returns `Unbounded | Run | Exhausted`. The 2s floor applies only to a short caller deadline, never after subtracting time already spent. If nothing is left when the handler starts, the relayer answers the 504 at once with stage `auth` and does not start work. Tests: `a_deadline_spent_before_the_recall_starts_is_exhausted`, `time_spent_before_the_handler_comes_out_of_the_budget`.

services/server/src/routes/recall.rs
nikola0x0: Fixed in 7983781. The pipeline no longer awaits the report: `recall` joins it after `run_with_deadline`, bounded by the same deadline (`join_failed_writes`), and drops it with a warn rather than hold a finished recall past the deadline. Callers without a deadline still wait for it as before. Tests in `routes::recall::tests`.

packages/mcp/src/bridge.ts
nikola0x0: Fixed in 7983781. The answering callback now has its own `.catch`, which clears `probing` and logs `bridge.call_orphaned_answer_failed`, so the next sweep answers the call.

packages/mcp/src/bridge.ts
nikola0x0: Fixed in 7983781: "safe to retry after the relayer is reachable again — wait a minute if the problem is on the relayer's side."

services/server/scripts/mcp/tools/failure.ts
nikola0x0: Trimmed in 7983781. Code comments keep a one-line why; ticket ids, the old wording and the 2-vs-4-minute history stay in the changelog.

nikola0x0:
The failing check, `Server / Clippy + Unit tests`, is inherited from `dev`: `dev` has failed the same job since `de98e6cf` (Sep 16) on the same 4 `routes::remember::tests`, because migration `021_failed_write_report_ack.sql` is not applied. #921 fixes it. This branch runs 792 passed / 4 failed against dev's 783 / 4; the extra 9 are this PR's tests.

packages/mcp/src/bridge.ts
```diff
@@ -2439,6 +2479,58 @@ export async function runBridge(
...
             if (elapsedMs <= deadlineMs) continue;
+            // `initialize` is answered locally and gets no reply here, so
+            // there is nothing to explain and nothing to probe for.
+            if (!neverSent && entry.msg.method !== "initialize") {
+                // A sent call: ask the relayer's `/health` before answering,
+                // so the message can say whether it is down, unreachable, or
+                // up with this one call stuck. Only the answer waits on the
+                // probe; the bookkeeping stays synchronous.
+                if (entry.probing) continue;
+                entry.probing = true;
```
ducnmm, line 2617: **[bug]** When a sent call expires, the sweeper sets `entry.probing = true` and `continue`s without removing it from `inFlight` or `closedOutIds`. That is required so a late SSE reply can still win, but `reconnect()` still replays the entire `inFlight` map. A handshake that lands during the probe (default 3s, up to 60s) will therefore POST the call again after the sweeper has already decided the reply is lost. For `memwal_remember_bulk` / `memwal_analyze` that is a duplicate write: `/api/remember/bulk` has no idempotency key, and the error the probe then writes tells the agent not to retry. Previously expiry called `failRequest` synchronously, so a reconnect that completed after the deadline would not replay the call. The new test covers a late reply on the same stream, not this path.

**Suggestion:** Do not replay entries with `probing === true` (the old stream is already dead, so late-reply-wins cannot come from it). Alternatively, move the entry out of `inFlight` when the probe starts and keep a side table the SSE pump consults so a genuine reply still wins without reconnect treating the call as outstanding.

packages/sdk/src/memwal.ts
```diff
@@ -808,6 +808,9 @@ export class MemWal {
                 // request byte-identical and the relayer applies its own
                 // "relevance" default.
                 sort: options.sort,
+                // How long this call waits, so the relayer can stop just
+                // short of it and name the step it was stuck in.
+                deadline_ms: RECALL_REQUEST_TIMEOUT_MS,
```
ducnmm: **[suggestion]** `deadline_ms` is the same constant as the `fetch` abort (`RECALL_REQUEST_TIMEOUT_MS` = 15000). The relayer subtracts a 1s margin from request *arrival*, while the SDK timer covers the whole `fetch`, including DNS/TCP/TLS. Connect time therefore eats the margin that is supposed to let the 504 beat `MemWalRequestTimeout`. On a cold connection that takes more than ~1s, the caller still sees a generic SDK timeout and the sidecar falls back to the `/health` path, which is the failure mode this change is meant to avoid. Subsequent keep-alive recalls are fine; the first request to a distant or slow-TLS relayer is not.

**Suggestion:** Either send a deadline shorter than the abort (e.g. abort at 15s, `deadline_ms: 14000`, relayer margin only for the response), or measure remaining time after `fetch` has connected if the runtime exposes it. Keep the two values derived from one source so they cannot drift.

ducnmm (changes requested):
## Summary

This is a well-aimed change: the SDK now puts `deadline_ms` in the signed recall body, the relayer cuts the pipeline ~1s early and names the stage (`embed` / `vector_search` / `walrus_download` / `seal_decrypt` / `auth`), and both MCP surfaces turn opaque aborts into Cause / Relayer health / Next step text, with writes still told not to retry. Stage tracking via a task-local marker (so `fetch_batch` keeps its signature), the exhausted-auth fast path, and moving `failed_writes` off the timed future all look correct, and the new unit tests cover the important cases (late SSE reply wins, 504 body shape, SDK `deadline_ms`, sidecar wording). The main residual risk is the bridge leaving an expired call in `inFlight` for the whole health probe, which reconnect replay still treats as live work.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

packages/mcp/src/bridge.ts
nikola0x0: Fixed in 878c09a5. The sweeper now sets `orphaned` on the entry alongside `probing` when it takes the call, and the replay loop skips those (`bridge.replay_skipped_orphaned`). The entry still stays in `inFlight`, so a late reply on a live stream can win as before — it just can never be posted again. `orphaned` is never cleared, unlike `probing`, so the failed-answer path can't let a replay back in on the next tick either.

Test: the mock ends the SSE stream the moment the probe starts, the bridge reconnects inside the probe window, and `memwal_recall` is asserted to have been posted exactly once. Without the gate it posts twice (checked).

packages/sdk/src/memwal.ts
nikola0x0: Fixed in 878c09a5. `RECALL_DEADLINE_MS = RECALL_REQUEST_TIMEOUT_MS - 1_000`, so the body carries 14000 against the 15s abort and the two are derived from one source. The relayer's margin now only has to pay for the reply's trip back, not for the connect the SDK timer has already been counting.

I didn't take the measure-after-connect option: the SDK has to run on Node, browsers and Workers, and there's no portable hook for when `fetch` finished connecting — it's the same reason `deadlineSignal` is hand-rolled instead of using `AbortSignal.any`. Test and both changelogs updated, plus `docs/relayer/api-reference.md`, which quoted `15000`.

ducnmm (approved):
LGTM, the orphaned call flag in bridge replay and the recall deadline offset fully address the requested changes, and all CI checks pass.

